PR #3, opened 2022-11-29
Migrate delegated routing

ajnavarro:
- Run go mod tidy
- Rewrite import paths when needed
- Change package names when needed


routing/http/README.md
```diff
@@ -0,0 +1,19 @@
+Repository Name
```
willscott: can you provide a real readme here?

guseggert (approved):

PR #9, opened 2022-12-05
routing/http: feat: limit the resp body payload

guseggert:


routing/http/client/transport.go
```diff
@@ -0,0 +1,38 @@
...
+type limitReadCloser struct {
+	limit     int64
+	bytesRead int64
+	io.ReadCloser
+}
+
+func (l *limitReadCloser) Read(p []byte) (int, error) {
+	n, err := l.ReadCloser.Read(p)
+	l.bytesRead += int64(n)
+	if l.bytesRead > l.limit {
```
willscott, line 34: the https://pkg.go.dev/io#LimitReader has slightly more complex logic here - i don't know there's worry about determinism here in terms of having a successful read up to the limit, even if your reads on the underlying reader don't match up with the boundary

willscott (approved):

ajnavarro:
We could/should use the standard go API to do this. We have a wrapping handler https://pkg.go.dev/net/http#MaxBytesHandler or a reader https://pkg.go.dev/net/http#MaxBytesReader

ajnavarro (changes requested):
We could/should use the standard go API to do this. We have a wrapping handler https://pkg.go.dev/net/http#MaxBytesHandler or a reader https://pkg.go.dev/net/http#MaxBytesReader

routing/http/client/transport.go
guseggert: Yeah I didn't use LimitReader because you can't tell if it stopped reading because it hit the limit, since it returns EOF in that case. So the error would be some JSON parsing error instead of "payload limit exceeded". 

The logic completely changes if you want to detect if you go *over* the limit, and it's non-trivial to do it precisely and I can't think of a practical reason that it needs to be precise. I'm guessing that, since you approved this, that you can't think of one that's worth pursuing either?

guseggert:
> We could/should use the standard go API to do this. We have a wrapping handler https://pkg.go.dev/net/http#MaxBytesHandler or a reader https://pkg.go.dev/net/http#MaxBytesReader

Not sure those apply? They are for servers handling reqs, not clients handling responses.

ajnavarro:
@guseggert shall we limit at the server level too?

ajnavarro (approved):

ajnavarro:
> Not sure those apply? They are for servers handling reqs, not clients handling responses.

Yeah, you are right, for me it is strange that there is no standard way to do this on the client side, having helpers on the server side.

guseggert:
> @guseggert shall we limit at the server level too?

Yeah we could, I was focused purely on client here since that is what we are imminently shipping, mind if I follow-up with this in a separate commit?

guseggert:
part of https://github.com/ipfs/go-delegated-routing/issues/63

PR #13, opened 2022-12-06
Migrate go-namesys

guseggert:
Kubo integration branch: https://github.com/ipfs/kubo/tree/chore/migrate-go-namesys


hacdias:
@guseggert any plans to update this? Or can we close it (namesys is out of sync now)?

PR #14, opened 2022-12-07
routing/http: feat: add client latency measure

guseggert:
Note that this measure includes some labels that also record:

- HTTP status codes
- HTTP method and path
- Errors returned by the HTTP client
- Call counts (included as part of histogram measure)


masih (approved):

guseggert:
part of https://github.com/ipfs/go-delegated-routing/issues/63

PR #18, opened 2022-12-21
routing/http: feat: add streaming support

guseggert:
This adds streaming support to the routing/v1 client and server by changing the interfaces to use iterators instead of slices, and adding content type negotation to the client and server with `application/x-ndjson` support.


routing/http/client/client.go
```diff
@@ -69,6 +82,12 @@ func WithProviderInfo(peerID peer.ID, addrs []multiaddr.Multiaddr) option {
 	}
 }
 
+func WithStreamResultsRequired() option {
+	return func(c *client) {
+		c.accepts = mediaTypeJSON
```
ajnavarro: shall this be `mediaTypeNDJSON`?

routing/http/types/iter/iter.go
```diff
@@ -0,0 +1,28 @@
+package iter
+
+// Iter is an iterator of aribtrary values.
+// Iterators are generally not goroutine-safe, to make them safe just read from them into a channel.
+// For our use cases, these usually have a single reader. This motivates iterators instead of channels,
+// since the overhead of goroutines+channels has a significant performance cost.
+// Using an iterator, you can read results directly without necessarily involving the Go scheduler.
+type Iter[T any] interface {
+	// Next returns the next element, true if an attempt was made to get the next element, and any error that occurred.
+	Next() (T, bool, error)
```
ajnavarro: We should add a Close() method to avoid the need of exhausting the iterator to free resources.

routing/http/client/client.go
```diff
@@ -96,10 +116,13 @@ func New(baseURL string, opts ...option) (*client, error) {
...
 
-func (c *client) FindProviders(ctx context.Context, key cid.Cid) (provs []types.ProviderResponse, err error) {
+func (c *client) FindProviders(ctx context.Context, key cid.Cid) (provs iter.Iter[types.ProviderResponse], err error) {
 	measurement := newMeasurement("FindProviders")
 	defer func() {
-		measurement.length = len(provs)
+		if sliceIter, ok := provs.(*iter.SliceIter[types.ProviderResponse]); ok {
+			length := len(sliceIter.Slice)
+			measurement.length = &length
+		}
```
ajnavarro: We could send a measurement to the iterator and add a counter to measure received elements on close or on finish.

routing/http/server/server.go
```diff
@@ -141,16 +157,100 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
...
+			if err != nil {
+				writeErr(w, "FindProviders", http.StatusBadRequest, fmt.Errorf("unable to parse Accept header: %w", err))
+				return
+			}
+
+			switch mediaType {
+			case "application/json":
+				supportsJSON = true
+			case "application/x-ndjson":
+				supportsNDJSON = true
```
ajnavarro: shall we use here the previously defined constants?

routing/http/server/server.go
```diff
@@ -141,16 +157,100 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
...
+		}
+		providers = append(providers, v)
+		i++
+	}
+	response := jsontypes.ReadProvidersResponse{Providers: providers}
+	writeJSONResult(w, "FindProviders", response)
+}
+
+func (s *server) findProvidersNDJSON(w http.ResponseWriter, provIter iter.Iter[types.ProviderResponse]) {
+	w.Header().Set("Content-Type", "application/x-ndjson")
```
ajnavarro: the same as the previous comment: Shall we use here the previously defined constant?

routing/http/server/server.go
```diff
@@ -141,16 +157,100 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
...
+	w.WriteHeader(http.StatusOK)
+	for {
+		v, ok, err := provIter.Next()
+		if !ok {
+			break
+		}
+		if err != nil {
+			logger.Errorw("FindProviders ndjson iterator error", "Error", err)
```
ajnavarro: shall we stop iteration when error here?

routing/http/types/iter/iter.go
ajnavarro: WDYT about:

```suggestion
	Next(ctx context.Context) (T, error)
	io.Closer
```
Where instead of using a boolean to know if we still have new elements, we use `io.EOF`? (this just simplify the method definition and also a little bit the code and the implementation IMHO)

ajnavarro (dismissed):

routing/http/types/iter/iter.go
guseggert: I don't feel strongly about either. I always found `io.EOF` to be clunky since it's not really an error so it's easy to miss or handle incorrectly due to being mislabeled, but OTOH it's still idiomatic and having three return values can also be confusing.  I'll try the EOF and see how it looks.

guseggert: I also don't think a ctx is necessary (including in JSONIter where I included one), this is controlled by the underlying reader being closed (e.g. when reading client-side and the req context is canceled, the body reader is closed and the context err is returned).

guseggert: I tried it with `io.EOF` and it was pretty annoying to have to do `if errors.Is(err, io.EOF){...}` instead of `if ok {...}`.  Definitely adding `Close()` though.

ajnavarro: > I tried it with io.EOF and it was pretty annoying to have to do if errors.Is(err, io.EOF){...} instead of if ok {...}. 

io.EOF is intended to be used like:

```go
if err == io.EOF {
    break
}
if err != nil {
    return err
}

```

I think is quite similar to `if ok {...}` and 100% idiomatic. In any case, the final decision is up to you, I don't want to bikeshed.

routing/http/types/ndjson/provider.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (p *readProvidersResponseIter) Next() (types.ProviderResponse, bool, error) {
+	v, ok, err := p.Iter.Next()
+	if err != nil {
+		return nil, false, err
+	}
+	if !ok {
+		return nil, false, nil
+	}
```
ajnavarro: Iterator chaining is a good example to see how this can be simplified using only two return params:

```go
v, err := p.iter.Next()
if err != nil {
   return nil, err
}
...
```

routing/http/types/ndjson/provider.go
```diff
@@ -0,0 +1,96 @@
...
+	switch v.Schema {
+	case types.SchemaBitswap:
+		var prov types.ReadBitswapProviderRecord
+		err := json.Unmarshal(v.Bytes, &prov)
+		if err != nil {
+			return nil, false, err
+		}
+		return &prov, true, nil
```
ajnavarro: Another example how having only two params simplifies iterator implementations and it is less error prone (there is no chance to flip an ok boolean when is not the case):

```go
err := json.Unmarshal(...)
return &prov, err
```

routing/http/types/iter/iter.go
ajnavarro: Sorry, after reviewing the code, I changed my mind :rofl: . Implementation-wise, using 2 return parameters instead of 3 has several advantages:
- Less error-prone. The developer doesn't have to decide when the returned boolean has to be true or false, it depends only on the error output. 
- Easy to chain iterators, less code doing the same. Check: https://github.com/ipfs/go-libipfs/pull/18#discussion_r1072012310
- Easy to integrate with existing functions. AFAIK there is no example on the Go SDK doing what we are doing with the Iterator, returning an error and an ok flag value simultaneously. That makes integration with existing APIS more verbose. See example here: https://github.com/ipfs/go-libipfs/pull/18#discussion_r1072015158

guseggert: > Less error-prone. The developer doesn't have to decide when the returned boolean has to be true or false, it depends only on the error output.

I disagree with this characterization, because you have to remember to check io.EOF. Typically I associate that with io.Reader, not other random non-io fns.

The other points are ultimately about ease of use, which is a secondary concern to me. Iterating over a list of stuff that does preemptible I/O, and could error out at any point, is not easy, and the complex interface reflects that instead of hiding it behind a sentinel error.

routing/http/types/ndjson/provider.go
guseggert: btw this is an example of why you should use `errors.Is(err, io.EOF)`, I almost always wrap these kinds of errors with some context which would break `err == io.EOF` checks

ajnavarro: `io.EOF` is intended to be used like `err == io.EOF` and must not be wrapped because it is a sentinel error: https://github.com/golang/go/issues/39155

routing/http/types/iter/iter.go
ajnavarro: > I disagree with this characterization, because you have to remember to check io.EOF

This is the same as saying that you have to remember to check the flag... you have to check something to know when to stop iterating. Using a sentinel error makes to only have one place to check. Errors that need to be returned by an interface must be described on the Godocs methods, that's how people will know how to check when the iterator is finished.

My point is that this Iterator API is like IO API, if it returns an error, then you must stop reading, so having a boolean flag duplicates that information. Every time an error is returned on an iterator implementation, the flag will be false.

That's why I think a sentinel error like io.EOF is a good idea here: we simplify the implementations, we simplify the interaction with other standard libraries, and we avoid duplicated method outputs.

routing/http/server/server.go
```diff
@@ -141,17 +162,106 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
...
+		handlerFunc = s.findProvidersJSON
+	} else {
+		for _, accept := range accepts {
+			mediaType, _, err := mime.ParseMediaType(accept)
+			if err != nil {
+				writeErr(w, "FindProviders", http.StatusBadRequest, fmt.Errorf("unable to parse Accept header: %w", err))
+				return
+			}
+
+			switch mediaType {
```
masih: Check for `*/*`?

routing/http/types/iter/iter.go
ajnavarro: @guseggert WDYT about https://pkg.go.dev/database/sql#Rows approach?

Interface:
```go
type Iter[T any] interface {
    Next() bool
    Err() error
    Value() T
    Close() error
}
```

Example:
```go
if iter.Next() {
    val := iter.Value()
    ....
}

if err := iter.Err(); err != nil {
    panic(err)
}

if err := iter.Close(); err != nil {
    panic(err)
}
```

No io.EOF check, no 3 return values, we are following the standard API approach

guseggert: sgtm thanks for finding this

routing/http/types/ndjson/provider.go
guseggert: even worse then that you cant use standard idioms with it

Jorropo: > btw this is an example of why you should use `errors.Is(err, io.EOF)`, I almost always wrap these kinds of errors with some context which would break `err == io.EOF` checks

It is widely accepted that `io.EOF` must not be wrapped:
> EOF is the error returned by Read when no more input is available. (Read must return EOF itself, not an error wrapping EOF, because callers will test for EOF using ==.) Functions should return EOF only to signal a graceful end of input. If the EOF occurs unexpectedly in a structured data stream, the appropriate error is either ErrUnexpectedEOF or some other error giving more detail.

guseggert: > EOF is the error returned by Read when no more input is available

This is not Read()...

BigLep:
@guseggert : what's remaining so this can move out of draft?

routing/http/client/client.go
```diff
@@ -118,43 +140,94 @@ func New(baseURL string, opts ...option) (*client, error) {
...
+}
+
+func (c *measuringIter[T]) Close() error {
+	c.m.record(c.ctx)
+	return c.Iter.Close()
+}
+
+func (c *client) FindProviders(ctx context.Context, key cid.Cid) (provs iter.ResultIter[types.ProviderResponse], err error) {
+	// TODO test measurements
+	m := newMeasurement("FindProviders")
```
aschmahmann, line 167: Is this still a todo?

routing/http/client/client.go
```diff
@@ -118,43 +140,94 @@ func New(baseURL string, opts ...option) (*client, error) {
...
-	measurement.err = err
-	measurement.latency = c.clock.Since(start)
+	m.err = err
+	m.latency = c.clock.Since(start)
 
 	if err != nil {
+		m.record(ctx)
 		return nil, err
 	}
-	defer resp.Body.Close()
```
aschmahmann, line 144: switching to calling `Body.Close()` everywhere is fine, but feels a bit easy to miss. Wondering if something like this might be better:
```
defer func() { 
  if err != nil || isSyncCall {
      resp.Body.Close()
  }
}
```

routing/http/types/iter/map.go
```diff
@@ -0,0 +1,39 @@
+package iter
+
+// Map invokes f on each element of iter.
+func Map[T any, U any](iter Iter[T], f func(t T) U) *MapIter[T, U] {
```
aschmahmann, line 4: Unused code?

routing/http/client/client.go
guseggert: yes, nothing has changed here I'm just making a note that the measurements aren't tested

BigLep:
2023-05-09 maintainer conversation: 
- We want to land this before @guseggert departs
- [x] @guseggert rebase
- @hacdias will review
- [x] @guseggert confirm that this can be bubbled up to Kubo and the cid.contact calls still work (verify it works in production)
- Leave some notes in https://github.com/ipfs/specs/issues/344.  Notes around error handling for example.

guseggert:
Re: testing with cid.contact, I talked with @masih and I was wrong--this is not implemented in cid.contact yet. It is for IPNI APIs but not for routing/v1. Confirmed by wiring into Kubo, req Accept=application/x-ndjson,application/json but resp Content-Type=application/json.  Tracking issue for this on their side: https://github.com/ipni/indexstar/issues/92

I think the best option is to just go ahead and ship this in Kubo and let cid.contact own the integration testing when they are ready to deploy.

hacdias (approved):
LGTM, thanks! I also went over previous comments to check if everything was addressed. Let's finally merge this!

Issue #17, opened 2022-12-09
routing/http/client: allow custom User-Agent


lidel:
Currently, `routing/http/client` sends requests with generic  `User-Agent: Go-http-client/1.1` from [golang's `http` package](https://go.dev/src/net/http/request.go#L508).

Before we ship with Kubo 0.18, this library should:
- [x] Change implicit default to own `User-Agent: go-libipfs/routing/http/client[/version]` 
  - Ideally, the value should conform to RFC 7231 (see http://tools.ietf.org/html/rfc7231#section-5.5.3), so maybe `go-libipfs/routing/http/client` or `go-libipfs/routing/http/client/{version or commit}` (having version is nice but may be hard to pull off – [go-libp2p's identify protocol uses just golang package name for the similar purpose](https://github.com/libp2p/go-libp2p/blob/2c82176c35be49a887482382c2056a27a727c7d3/p2p/protocol/identify/id.go#L49) and I think its fine here too. 
- [x] Allow Kubo for passing own version string via `func WithUserAgent(ua string) option`
  - We want IPNI at cid.contact to know which versions of Kubo are sending requests, to make informed decision about supporting legacy endpoints, making breaking changes etc.  (cc @willscott)

PR #27, opened 2023-01-13
chore: release 0.2.0

hacdias:


lidel (changes requested):
I'd like to include  https://github.com/ipfs/go-libipfs/pull/31 before we make this release, so we can bubble up both to Kubo.

lidel:
Let's continue in https://github.com/ipfs/go-libipfs/pull/32 which includes #17

PR #31, opened 2023-01-14
feat(routing/http/client): allow custom User-Agent

### LINKED ISSUE #17: routing/http/client: allow custom User-Agent

lidel:
This PR Closes #17 

- Adds `WithUserAgent` which allows Kubo (or any other user of `routing/http/client`) to set own version
  - Kubo PR:  https://github.com/ipfs/kubo/pull/9550
- Uses module version if reported via `ReadBuildInfo()`  when no override is present and module is imported from `github.com/ipfs/go-libipfs/routing/http/client` – allows us to quickly tell if a request was made by stable official release or some hacky fork. 
- Does nothing  otherwise. This sends `User-Agent: Go-http-client/1.1` (current state)

### DIFF
--- a/routing/http/client/transport.go
+++ b/routing/http/client/transport.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"reflect"
 	"runtime/debug"
+	"strings"
 )
 
 type ResponseBodyLimitedTransport struct {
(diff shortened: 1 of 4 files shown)


routing/http/client/transport.go
```diff
@@ -36,3 +41,35 @@ func (l *limitReadCloser) Read(p []byte) (int, error) {
 	}
 	return n, err
 }
+
+// ImportPath is the canonical import path that allows us to identify
+// official client builds vs modified forks, and use that info in User-Agent header.
+const ImportPath = "github.com/ipfs/go-libipfs/routing/http/client"
+
+// moduleVersion returns a useful user agent version string allowing us to
+// identify requests coming from officialxl releases of this module.
```
guseggert:
```suggestion
// identify requests coming from official releases of this module.
```

routing/http/client/transport.go
```diff
@@ -36,3 +41,35 @@ func (l *limitReadCloser) Read(p []byte) (int, error) {
...
+func moduleVersion() string {
```
guseggert: Can we add a test for this?  Really I'm interested in making sure a test fails if this code is moved and `ImportPath` is not updated (or use reflection or something to get the import path dynamically).

guseggert (dismissed):

routing/http/client/transport.go
lidel: @guseggert simplified and added test in https://github.com/ipfs/go-libipfs/pull/31/commits/4e11e8b46555c5f623eec39fb25f5d6c56f4ea08

routing/http/client/transport.go
```diff
@@ -36,3 +43,37 @@ func (l *limitReadCloser) Read(p []byte) (int, error) {
...
+			}
+		}
+		if module != nil {
+			ua += "@" + module.Version
+			return
+		}
+		ua += "@unknown"
+	}
+	return
+}
```
ajnavarro, line 79: This might be a bit complex. Can we just have a default string header as a simpler default? something like `router/http/go`

Edit: Actually, I don't think the actual default header `User-Agent: Go-http-client/1.1`  is bad, isn't it?

lidel: Defaulting to `User-Agent: Go-http-client/1.1` is not bad, but does not tell us when someone used our library, and when they wrote their own client.  

Having dynamic default set to package name will tell us what version of lib is in the use, how fast people upgrade, and if they fork, we will see the repo name and be able to inspect what changes were made.

Overall, less guesswork for IPFS stewards, more visibility into ecosystem.

lidel:
As per verbal from today's standup, merging to include this in Kubo 0.18, continued in https://github.com/ipfs/kubo/pull/9550

PR #34, opened 2023-01-21
feat: support UnixFS 1.5 file mode and modification times

kstuart:
Adds support for storing and retrieving file mode and last modification time.

Support added to:
- [X] Files
- [X] LinkFiles
- [X] Webfiles
- [X] Directories
- [X] Tar Archives

When the TAR archive (headers) include a file mode or modification time, the extractor will restore that metadata when supported for the underlying filesystem.

The Golang runtime currently does not support changing the times on a symlink, for Linux and some BSDs a custom solution has been implemented, for Darwin this is not the case so when copying a symlink to the filesystem the last modification time is not updated. Since for concrete files and directories stored modes and modification times are faithfully restored to the filesystem this should not be a breaking issue, a similar solution to that implemented for Linux/BSDs is likely implementable by a developer with access to a Darwin platform.

Replaces PRs:
- https://github.com/ipfs/go-ipfs-files/pull/31
- https://github.com/ipfs/tar-utils/pull/11

Relates to https://github.com/ipfs/kubo/issues/6920

### DIFF
--- a/files/file_test.go
+++ b/files/file_test.go
@@ -70,7 +70,7 @@ func TestMultipartFiles(t *testing.T) {
 	data := `
 --Boundary!
 Content-Type: text/plain
-Content-Disposition: form-data; name="file-0?mode=0754&mtime=1604320500&mtime-nsecs=55555"; filename="file1"
+Content-Disposition: form-data; name="file-0?mode=0754&mtime=1604320500&mtime-nsecs=55555"; filename="%C2%A3%E1%BA%9E%C7%91%C7%93%C3%86+%C3%A6+%E2%99%AB%E2%99%AC"
 Some-Header: beep
 
 beep
@@ -122,7 +122,7 @@ implicit file2
 	CheckDir(t, dir, []Event{
 		{
 			kind:  TFile,
-			name:  "file1",
+			name:  "£ẞǑǓÆ æ ♫♬",
 			value: "beep",
 			mode:  0754,
 			mtime: time.Unix(1604320500, 55555),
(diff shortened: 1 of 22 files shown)


zacharywhitley:
Any status on where this is? It seems to be stalled.

kstuart:
@zacharywhitley, unfortunately I don't think there's any real interest in this feature; This has been awaiting review for some time, and the completed https://github.com/ipfs/kubo/issues/6920 PRs prior to the refactoring met the same fate.

files/file_test.go
```diff
@@ -48,36 +50,62 @@ func TestReaderFiles(t *testing.T) {
...
+	rf := NewReaderStatFile(reader, stat)
+	if rf.Mode() != mode {
+		t.Fatalf("Expected file mode to be [%v] but got [%v]", mode, rf.Mode())
+	}
+	if rf.ModTime() != mtime {
+		t.Fatalf("Expected file modified time to be [%v] but got [%v]", mtime, rf.ModTime())
+	}
+}
+
 func TestMultipartFiles(t *testing.T) {
```
lidel, line 69: :thought_balloon:  are these required, and we need to effectively break the wire format? Could we avoid it and be backward compatible?

If so, we want to keep original tests without `mode` and `mtime` being sent over the wire. 
Perhaps refactor into separate `TestMultipartFiles` and `TestMultipartFilesWithMtime`?

lidel (commented):
See https://github.com/ipfs/kubo/pull/7754#pullrequestreview-2063687831

Something we should add in this PR is to update `boxo/gateway` as well, and expose `mtime` as `Last-Modified` on HTTP Gateway responses (if present)

gammazero:
Replaced by PR #653

PR #35, opened 2023-01-23
Use `PUT` as method to insert provider records

masih:
Change the implementation of HTTP delegated routing provider records to match the IPIP-337 specification. The specification was changed to use `PUT`.

Close request body after decoding when handling provide.

### DIFF
--- a/routing/http/client/client.go
+++ b/routing/http/client/client.go
@@ -211,7 +211,7 @@ func (c *client) provideSignedBitswapRecord(ctx context.Context, bswp *types.Wri
 		return 0, err
 	}
 
-	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(b))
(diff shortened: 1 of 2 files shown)


Jorropo:
AFAIT you removed the `PUT` operations from the spec https://github.com/ipfs/specs/pull/370.

Is it because you plan to open a new IPIP ? What is the source of truth here ?

willscott:
> Is it because you plan to open a new IPIP ? What is the source of truth here ?

correct. the put hasn't been deferred from 337, but that doesn't mean we aren't working towards it and I believe the convention is that code comes before spec.

masih:
As soon as IPIP-337 is merged PR to add `PUT` back will be opened. For now, this PR fixes what's definitely not going to be, which is `POST` requests.

Jorropo (approved):
Sounds good to me then. :slightly_smiling_face:

PR #58, opened 2023-01-26
Migrate `go-block-format`

Jorropo:
Needed for ipfs/go-block-format#45 

Todo:
- [X] Make this PR ready
- [x] Create stub types in [`go-block-format`](https://github.com/ipfs/go-block-format)
- [x] Move issues
- [x] Close initial PRs (Use Gus's script)
- [x] Merge and release https://github.com/ipfs/go-block-format/pull/37
- [ ] ~~Archive repo https://github.com/ipfs/go-block-format~~ (we shouldn't do this now, I'll need to update the stubs swith new stubs if we proceed with ipfs/go-block-format#45)

PR #59, opened 2023-01-26
blocks: change Block to a value struct

### LINKED ISSUE #45: valuify `blocks.Block`
`blocks.Block` being an interface force at least two allocations per blocks (`[]byte` + the `block.Block` implementation).
If this was a value struct or even an type Block []byte (we already know the CID in most cases), we would only allocate the []byte and embed or pass by registers (decomposed stack) the rest.

This is gonna break a lot of stuff depending of how brutal we are.

### LINKED ISSUE #57: `block.Block` being an interface force at least two allocations per blocks (`[]byte` + the `block.Block` implementation).

Jorropo:
The goal is to stop doing two allocations for each block (now it will only allocate the `[]byte` buffer, and pass the `cid.Cid, []byte` pair by decomposed registers or stack).
This way of changing does not change the syntax for trivial uses of block.Block, so in theory we will have to update only producers of block.Block, not consumers.

Depends on (include) ipfs/boxo#58
Fixes ipfs/go-block-format#45

Todo:
- [ ] Test bubbling up this PR in Kubo and sees how much break.
- [ ] Test bubbling up this PR in Lotus and sees how much break.

### DIFF
--- a/blocks/blocks.go
+++ b/blocks/blocks.go
@@ -16,66 +16,57 @@ import (
 // according to the contents. It is currently used only when debugging.
 var ErrWrongHash = errors.New("data did not match given hash")
 
-// Block provides abstraction for blocks implementations.
-type Block interface {
-	RawData() []byte
-	Cid() cid.Cid
-	String() string
-	Loggable() map[string]interface{}
-}
-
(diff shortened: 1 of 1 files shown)


Jorropo:
Replaced by ipfs/boxo#192

Issue #68, opened 2022-10-23
[ipfs/go-bitswap] Bitswap client usage unintuitive


aschmahmann:
Since ipfs/go-bitswap#570 has made it possible to use the client and server pieces of go-bitswap independently it would seem reasonable that someone could replace:

```
bsnet := bsnet.NewFromIpfsHost(h, &rhelp.Null{})
bswap := bitswap.New(ctx, bsnet, bstore)
block, err := bswap.GetBlock(ctx, someCID)
```

with

```
bsnet := bsnet.NewFromIpfsHost(h, &rhelp.Null{})
bswap := bsclient.New(ctx, bsnet, bstore)
block, err := bswap.GetBlock(ctx, someCID)
```

especially since there are no compile-time errors or comments on the client that would indicate otherwise. However, unless `bsnet.Start(bswap)` is called on the client (not required when using `bitswap.New`) the code will silently hang as the response messages are not being processed.

IMO this footgun should be removed so that `bsclient.New` can drop-in for `bitswap.New` when no server functionality is required. An easy, if inelegant, option might be to call `bsnet.Start(bswap)` unless an option is passed in telling it not to and then passing that option internally in the `bitswap.New` constructor that creates the client. If not resolved at least there should be some comments on the client (and I suspect server) constructor letting people know they need to register their client/server with the BitswapNetwork in order to function.

wdyt @Jorropo @guseggert ?

Jorropo:
Yeah should be fixed.

PR #143, opened 2023-01-31
docs(examples): add UnixFS file download over Bitswap

aschmahmann:
This is a starting point to discuss how we want to manage our examples for go-libipfs. It also IMO gives some illustration on some of the UX pains of working with the existing libraries, places where we can add nicer Go helper utilities, etc.


.github/workflows/test-examples-against-branch.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: echo "C:/msys64/usr/bin" >> $GITHUB_PATH
+      - name: Use go mod from current branch
+        with:
+          run:
+            # in addition to the standard tests we want to test the examples against the current branch
+            # however, that version might be in a fork so we need to replace the dependency
+
+            # make sure the examples run against the current version of kubo
+            go mod edit -replace github.com/ipfs/go-libipfs=./..
+            go mod tidy
```
aschmahmann: @galargh do you have any better suggestions on how to do this, or thoughts on if this is a bad idea? My thinking here is that there are two issues we'd like to avoid:
1. The examples being out of date for a long time because nobody updated them and CI didn't fail/notify us that anything bad was happening
2. The examples not being usable with `main` due to some breaking change, so when someone tries to use the examples they don't work.
   - As we add more examples it'll also give us a bit of a heads up as to how basic use cases will be affected by a given breaking change along with any announcements we may want to call out in release notes.
   
Note: I mostly copied the Unified CI Go Tests but removed the codecov checks

examples/unixfs-file-cid/README.md
```diff
@@ -0,0 +1,58 @@
+# Downloading a UnixFS file
```
aschmahmann, line 1: This whole thing is very rough and a lot of the style was copied from the go-libp2p examples. Some feedback from some of the folks working on/near docs (where we should be pointing people to understand general concepts/protocols) would be great. (e.g. @2color)

examples/unixfs-file-cid/main.go
```diff
@@ -0,0 +1,231 @@
...
+	addr := h.Addrs()[0]
+	return addr.Encapsulate(hostAddr).String()
+}
+
+// The CID of the file with the number 0 to 100k, built with the parameters:
+// CIDv1 links, a 256bit sha2-256 hash function, raw-leaves, a balanced layout, 256kiB chunks, and 174 max links per block
+const fileCid = "bafybeiecq2irw4fl5vunnxo6cegoutv4de63h7n27tekkjtak3jrvrzzhe"
+
+// createFile0to100k creates a file with the number 0 to 100k
+func createFile0to100k() ([]byte, error) {
```
aschmahmann, line 132: Note: I chose a larger file rather than just "Hello World!" to try not to confuse people with https://docs.ipfs.tech/concepts/hashing/#content-identifiers-are-not-file-hashes

examples/unixfs-file-cid/main.go
```diff
@@ -0,0 +1,231 @@
...
+	nd, err := balanced.Layout(ufsBuilder) // Arrange the graph with a balanced layout
+	if err != nil {
+		return cid.Undef, nil, err
+	}
+
+	// Start listening on the Bitswap protocol
+	// For this example we're not leveraging any content routing (DHT, IPNI, delegated routing requests, etc.) as we know the peer we are fetching from
+	n := bsnet.NewFromIpfsHost(h, routinghelpers.Null{})
+	bswap := bsserver.New(ctx, n, bs)
+	n.Start(bswap)
```
aschmahmann, line 183: The need for this unintuitive dance with `Start` is something we should fix #68

examples/unixfs-file-cid/main.go
```diff
@@ -0,0 +1,231 @@
...
+		NoCopy:  false,
+	}
+	ufsBuilder, err := ufsImportParams.New(chunker.NewSizeSplitter(fileReader, chunker.DefaultBlockSize)) // Split the file up into fixed sized 256KiB chunks
+	if err != nil {
+		return cid.Undef, nil, err
+	}
+	nd, err := balanced.Layout(ufsBuilder) // Arrange the graph with a balanced layout
+	if err != nil {
+		return cid.Undef, nil, err
+	}
```
aschmahmann, line 177: Wow... this is just a lot. I could simplify down by using some of the helpers that just use defaults. However:
1. I'd like the examples to use things closer to recommended usage (e.g. definitely raw leaves, and also CIDv1 links)
2. All this complexity seems like roughly the minimum to do non-default things and finding out how to change things to match the kubo CLI shouldn't be miserable for people

Hopefully some moving of packages into this repo + adding some nicer usability wrappers will help here. Although maybe just adding some docs and having a struct that makes more sense (e.g. `NoCopy` is really tied to how https://github.com/ipfs/go-filestore works rather than being user configurable)

examples/unixfs-file-cid/main.go
```diff
@@ -0,0 +1,231 @@
...
+	// ask peers it is connected to for content so this will work
+	if err := h.Connect(ctx, *info); err != nil {
+		return nil, err
+	}
+
+	dserv := merkledag.NewReadOnlyDagService(merkledag.NewSession(ctx, merkledag.NewDAGService(blockservice.New(blockstore.NewBlockstore(datastore.NewNullDatastore()), bswap))))
+	nd, err := dserv.Get(ctx, c)
+	if err != nil {
+		return nil, err
+	}
```
aschmahmann, line 216: Honestly not so bad other than the fact that:
1. This whole thing is kind of a mouthful
2. We need to load a block before doing the UnixFS things which will in turn load even more blocks... why can't it just do it for me?

examples/unixfs-file-cid/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Directly connect to the peer that we know has the content
+	// Generally this peer will come from whatever content routing system is provided, however go-bitswap will also
+	// ask peers it is connected to for content so this will work
+	if err := h.Connect(ctx, *info); err != nil {
+		return nil, err
+	}
```
aschmahmann, line 210: Probably the "safer" way to do this if you were running this for real would be to create a content router that always returned the the target peer had the data. This is because if the remote peer disconnects from you (or you from them) you have no way to re-establish the connection. Not sure if it's worth adding more complexity to the code here though.

Ideally any API changes around sessions should make doing this less painful, although if we needed a special content router that wouldn't be the end of the world (some food for thought @Jorropo).

examples/unixfs-file-cid/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"bytes"
+	"context"
+	"github.com/ipfs/go-cid"
+	"testing"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+func TestBitswapFetch(t *testing.T) {
```
aschmahmann, line 12: @guseggert any thoughts on how you'd like the testing to work here? Would you prefer something that runs the binary itself, reads logs as in https://github.com/libp2p/go-libp2p/blob/34dc11547a10e7abe0a3fc615dbf839818554cab/examples/echo/main_test.go#L11, or something else?

examples/go.mod
```diff
@@ -0,0 +1,139 @@
...
+	github.com/ipfs/go-ipfs-blockstore v1.2.0
+	github.com/ipfs/go-ipfs-chunker v0.0.5
+	github.com/ipfs/go-ipfs-exchange-offline v0.3.0
+	github.com/ipfs/go-libipfs v0.4.0
+	github.com/ipfs/go-merkledag v0.9.0
+	github.com/ipfs/go-unixfs v0.4.2
+	github.com/libp2p/go-libp2p v0.24.2
+	github.com/libp2p/go-libp2p-routing-helpers v0.6.0
+	github.com/multiformats/go-multiaddr v0.8.0
+	github.com/multiformats/go-multicodec v0.7.0
```
Jorropo: I think the only dep that isn't in `github.com/ipfs/go-libipfs`'s `go.mod` is `github.com/libp2p/go-libp2p-routing-helpers v0.6.0`.

I think you should just keep one `go.mod` file, it gets annoying to update multiple `go.mod` files, either you forget and get reminded by CI, or you don't and they get out of sync.

.github/workflows/test-examples-against-branch.yml
galargh: I think it makes sense. I know in kubo it's put together the other way round, as in examples reference relative parent - https://github.com/ipfs/kubo/blob/master/docs/examples/kubo-as-a-library/go.mod#L5-L7. One nice feature that approach comes with is that `go test ./...` performs the same checks the CI does.

examples/go.mod
hacdias: My question is whether that will scale to other examples we may do after?

aschmahmann: yeah, I think having one go.mod for all the examples is fine but I would keep separate from the main. We can use `../go-libipfs` as well, but I'd like examples to also always be usable from only released repos if possible. Ideally this would happen without requiring people to rewind to the release commit of go-libipfs in GitHub, but if that's too annoying we can skip it

hacdias: @aschmahmann since we've merged two examples in the meanwhile, would you mind rebasing this on the current code so I can have an idea of the changes you're suggesting?

> I'd like examples to also always be usable from only released repos if possible.

I think this would require us to only update examples after releasing, which can lead us to have out of date examples that no longer work.

aschmahmann: Yep, will rebase.

> I think this would require us to only update examples after releasing, which can lead us to have out of date examples that no longer work.

Well they'd always work, they just wouldn't work with the newest library. Ideally when there are breaking changes there's some deprecation period where we encourage people to switch over. To some extent watching all of our examples break/require hard-coding to a go-libipfs commit (or to `..`) allows us to feel some of that pain and consider what this will mean for our downstream users.

However, the key there was "if possible", sometimes its not or the tradeoff isn't worthwhile. That being said if people prefer this route that's fine as long as we add CI checks to make sure the dependencies get updated (i.e. if we're going the `..` method I don't want go-libipfs depending on go-multihash v0.3.0 while the example depends on v0.2.0 which is implicitly being upgraded.

hacdias: > as we add CI checks to make sure the dependencies get updated

The CI currently already checks if [all `go.mod`s](https://github.com/ipfs/go-libipfs/blob/main/.github/workflows/go-check.yml#L31-L40) are updated, so I think we're good on that.

examples/README.md
```diff
@@ -6,5 +6,6 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 
 ## Examples and Tutorials
 
+- [Fetching a UnixFS file by CID](./unixfs-file-cid)
```
aschmahmann, line 9: Not sure what order we want for these. Just did alphabetical by title, feel free to change though.

examples/unixfs-file-cid/main.go
Jorropo: you could mock a content routing that always return you `info` but that would be very confusing to new comers, not the goal of an example, that looks fine to me. I guess the only better way would be to actually use the dht.

examples/unixfs-file-cid/main.go
Jorropo: :+1: should fix (not in this PR)

hacdias (approved):
I'm happy to merge this PR once the remanning dependency conflicts are solved. I agree that there is a lot of complexity and I think that is something we should work on by creating helper functions for common use-cases.

PR #145, opened 2023-01-31
feat(gateway): improve GO API interface, remove Writable API

### LINKED ISSUE #61: gateway: post-extraction GO API cleanup
We're currently working on extracting the Gateway code from `kubo` to `go-libipfs` (https://github.com/ipfs/kubo/issues/8524). After the extraction, we should improve the code, such that it is more useable. There are currently a few points that, in my opinion, make the Gateway handler harder to use than it should:

1. It requires more APIs than necessary.
	- From `NodeAPI.Block()`, we only use `Get` and `Stat`, but we are requiring the full interface (`Get`, `Put`, `Rm`, `Stat`).
	- From `NodeAPI.Dag()`, we only use `Get`, but we are requiring the full `APIDagService` interface, which contains a lot more functions.
2. The handler requires both an online and offline APIs. The offline API is only used once with the `Dag().Get` function. Perhaps we could add an option to that method `Offline: true` or `LocalOnly: true` or `NoFetch: true` instead. This would allow us to only require the consumers of the Gateway code to provide one Node API.

Things to do:

- [x] Define a minimal set of APIs required by the Gateway. https://github.com/ipfs/go-libipfs/issues/61
- [x] Remove Writable API. https://github.com/ipfs/go-libipfs/issues/61 & https://github.com/ipfs/kubo/pull/9616
	- The Writable API is undocumented and experimental feature of Kubo. We want to support a POST/PUT API in the future that will be incompatible. So it is a good idea to remove it right now and do not carry this bagage.

hacdias:
Closes #61

This PR does the following:

1. Simplifies `NodeAPI` (renamed `API`), such that we only require the bare minimum functions we need on the gateway. These have each the minimum required amount of functions required. This will allow other developers to use our gateway code much more easily.
2. Consolidates the `offlineAPI` as a single function on `API` (`IsCached`).
3. Removes the Writable Gateway. The code for the Writable Gateway is moved to Kubo codebase and deprecated https://github.com/ipfs/kubo/issues/9622. See https://github.com/ipfs/kubo/pull/9616.

Please **note**: gateway sharness tests are expected to fail since they are run against Kubo's `master` version. The PR in Kubo with this changes can be viewed at: https://github.com/ipfs/kubo/pull/9616

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -5,33 +5,38 @@ import (
 	"net/http"
 	"sort"
 
-	coreiface "github.com/ipfs/interface-go-ipfs-core"
-	path "github.com/ipfs/interface-go-ipfs-core/path"
+	cid "github.com/ipfs/go-cid"
+	"github.com/ipfs/go-libipfs/blocks"
+	"github.com/ipfs/go-libipfs/files"
+	iface "github.com/ipfs/interface-go-ipfs-core"
+	options "github.com/ipfs/interface-go-ipfs-core/options"
+	"github.com/ipfs/interface-go-ipfs-core/path"
 )
 
-// Config is the configuration that will be applied when creating a new gateway
-// handler.
+// Config is the configuration used when creating a new gateway handler.
 type Config struct {
-	Headers  map[string][]string
-	Writable bool
+	Headers map[string][]string
 }
 
-// NodeAPI defines the minimal set of API services required by a gateway handler
-type NodeAPI interface {
-	// Unixfs returns an implementation of Unixfs API
-	Unixfs() coreiface.UnixfsAPI
+// API defines the minimal set of API services required for a gateway handler.
+type API interface {
+	// GetUnixFsNode returns a read-only handle to a file tree referenced by a path.
+	GetUnixFsNode(context.Context, path.Path) (files.Node, error)
 
-	// Block returns an implementation of Block API
-	Block() coreiface.BlockAPI
+	// LsUnixFsDir returns the list of links in a directory.
+	LsUnixFsDir(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
 
-	// Dag returns an implementation of Dag API
-	Dag() coreiface.APIDagService
+	// GetBlock return a block from a certain CID.
+	GetBlock(context.Context, cid.Cid) (blocks.Block, error)
 
-	// Routing returns an implementation of Routing API.
-	// Used for returning signed IPNS records, see IPIP-0328
-	Routing() coreiface.RoutingAPI
+	// GetIPNSRecord retrieves the best IPNS record for a given CID (libp2p-key)
+	// from the routing system.
+	GetIPNSRecord(context.Context, cid.Cid) ([]byte, error)
 
-	// ResolvePath resolves the path using Unixfs resolver
+	// IsCached returns whether or not the path exists locally.
+	IsCached(context.Context, path.Path) bool
+
+	// ResolvePath resolves the path using UnixFS resolver
 	ResolvePath(context.Context, path.Path) (path.Resolved, error)
 }
 
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -9,7 +9,6 @@ import (
 	"github.com/ipfs/go-libipfs/blocks"
 	"github.com/ipfs/go-libipfs/files"
 	iface "github.com/ipfs/interface-go-ipfs-core"
-	options "github.com/ipfs/interface-go-ipfs-core/options"
 	"github.com/ipfs/interface-go-ipfs-core/path"
 )
 
@@ -21,10 +20,10 @@ type Config struct {
 // API defines the minimal set of API services required for a gateway handler.
 type API interface {
 	// GetUnixFsNode returns a read-only handle to a file tree referenced by a path.
-	GetUnixFsNode(context.Context, path.Path) (files.Node, error)
+	GetUnixFsNode(context.Context, path.Resolved) (files.Node, error)
 
 	// LsUnixFsDir returns the list of links in a directory.
-	LsUnixFsDir(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
+	LsUnixFsDir(context.Context, path.Resolved) (<-chan iface.DirEntry, error)
 
 	// GetBlock return a block from a certain CID.
 	GetBlock(context.Context, cid.Cid) (blocks.Block, error)
@@ -36,7 +35,9 @@ type API interface {
 	// IsCached returns whether or not the path exists locally.
 	IsCached(context.Context, path.Path) bool
 
-	// ResolvePath resolves the path using UnixFS resolver
+	// ResolvePath resolves the path using UnixFS resolver. If the path does not
+	// exist due to a missing link, it should return an error of type:
+	// https://pkg.go.dev/github.com/ipfs/go-path@v0.3.0/resolver#ErrNoLink
 	ResolvePath(context.Context, path.Path) (path.Resolved, error)
 }
 
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"net/textproto"
 	"net/url"
-	"os"
 	gopath "path"
 	"regexp"
 	"runtime/debug"
@@ -18,11 +17,7 @@ import (
 
 	cid "github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
-	"github.com/ipfs/go-libipfs/files"
 	logging "github.com/ipfs/go-log"
-	dag "github.com/ipfs/go-merkledag"
-	mfs "github.com/ipfs/go-mfs"
-	path "github.com/ipfs/go-path"
 	"github.com/ipfs/go-path/resolver"
 	coreiface "github.com/ipfs/interface-go-ipfs-core"
 	ipath "github.com/ipfs/interface-go-ipfs-core/path"
@@ -71,9 +66,8 @@ type redirectTemplateData struct {
 // handler is a HTTP handler that serves IPFS objects (accessible by default at /ipfs/<path>)
 // (it serves requests like GET /ipfs/QmVRzPKPzNtSrEzBFm2UZfxmPAgnaLke4DMcerbsGGSaFe/link)
 type handler struct {
-	config     Config
-	api        NodeAPI
-	offlineAPI NodeAPI
+	config Config
+	api    API
 
 	// generic metrics
 	firstContentBlockGetMetric *prometheus.HistogramVec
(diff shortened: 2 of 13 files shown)


hacdias:
**Note**: I would also prefer to completely remove the Offline API and instead adding an option to Dag().Get() to do LocalOnly.

hacdias:
I converted this to draft because I will be doing to more things: removing the Writable gateway and updating the interface to make it more simple.

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
-	// Unixfs returns an implementation of Unixfs API
-	Unixfs() coreiface.UnixfsAPI
+	// GetUnixFs returns a read-only handle to a file tree referenced by a path.
+	GetUnixFs(context.Context, path.Path) (files.Node, error)
+
+	// LsUnixFs returns the list of links in a directory.
+	LsUnixFs(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
+
+	// GetBlock attempts to resolve the path and return a reader for data in the block.
+	GetBlock(context.Context, path.Path) (io.Reader, error)
```
lidel: nits:
- Since we have a separate `ResolvePath`, does it make sense to change this to require  [`Resolved`](https://github.com/ipfs/interface-go-ipfs-core/blob/master/path/path.go#L44) path instead?
- It returns a `BlockReader`, not a block bytes

```suggestion
	BlockReader(context.Context, path.Resolved) (io.Reader, error)
```

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
-	// Block returns an implementation of Block API
-	Block() coreiface.BlockAPI
+	// StatBlockOffline returns information on a block without making any network
+	// connection. In case the block does not exist locally, an error should be returned.
+	StatBlockOffline(context.Context, path.Path) (iface.BlockStat, error)
 
-	// Dag returns an implementation of Dag API
-	Dag() coreiface.APIDagService
+	// GetNode returns the IPLD node for the path.
+	GetNode(context.Context, cid.Cid) (ipld.Node, error)
```
lidel: Confusing name, `GetUnixFs` returns `Node` too. Make this explicit:

```suggestion
	GetIPLDNode(context.Context, cid.Cid) (ipld.Node, error)
```

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
+	LsUnixFs(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
+
+	// GetBlock attempts to resolve the path and return a reader for data in the block.
+	GetBlock(context.Context, path.Path) (io.Reader, error)
 
-	// Block returns an implementation of Block API
-	Block() coreiface.BlockAPI
+	// StatBlockOffline returns information on a block without making any network
+	// connection. In case the block does not exist locally, an error should be returned.
+	StatBlockOffline(context.Context, path.Path) (iface.BlockStat, error)
```
lidel: Gateway ignores `BlockStat` – we only care if it errors (block missing from local blockstore) or not, so it works as a simple boolean. Make it explicit and clear:

```suggestion
	IsCached(context.Context, path.Path) (bool)
```

gateway/handler_car.go
```diff
@@ -89,10 +88,10 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
 // FIXME(@Jorropo): https://github.com/ipld/go-car/issues/315
 type dagStore struct {
-	dag coreiface.APIDagService
+	api NodeAPI
 	ctx context.Context
 }
 
 func (ds dagStore) Get(_ context.Context, c cid.Cid) (blocks.Block, error) {
-	return ds.dag.Get(ds.ctx, c)
+	return ds.api.GetNode(ds.ctx, c)
```
lidel: :thought_balloon: Interesting, we don't need  `ipld.Node`  returned by `Get[IPLD]Node` here, [`readStore` only uses `blocks.Block`](https://github.com/ipld/go-car/blob/268b65bab2b7f45aa05494b943992f5b9bbab75e/car.go#L26).


:point_right:   Are we able to simplify and switch to `blocks.Block` created with `BlockReader`,
removing dependency on IPLD if someone wants to only use this library for hosting CARs and Blocks?



(:thought_balloon:  This is short term cleanup. Long term, CAR response handler will be refactored  again, and in the case of bifrost-gateway most likely we will be doing pass-through of a CAR generated  on Saturn nodes.)

gateway/handler_codec.go
```diff
@@ -182,7 +182,7 @@ func (i *handler) serveCodecRaw(ctx context.Context, w http.ResponseWriter, r *h
 
 // serveCodecConverted returns payload converted to codec specified in toCodec
 func (i *handler) serveCodecConverted(ctx context.Context, w http.ResponseWriter, r *http.Request, resolvedPath ipath.Resolved, contentPath ipath.Path, toCodec mc.Code, modtime time.Time) {
-	obj, err := i.api.Dag().Get(ctx, resolvedPath.Cid())
+	obj, err := i.api.GetNode(ctx, resolvedPath.Cid())
```
lidel: :thought_balloon:  This seems to be the only place we use `Get[IPLD]Node`.

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
-	// Dag returns an implementation of Dag API
-	Dag() coreiface.APIDagService
+	// GetNode returns the IPLD node for the path.
+	GetNode(context.Context, cid.Cid) (ipld.Node, error)
 
-	// Routing returns an implementation of Routing API.
-	// Used for returning signed IPNS records, see IPIP-0328
-	Routing() coreiface.RoutingAPI
+	// GetRouting retrieves the best value for a given key from the routing system.
+	GetRouting(context.Context, string) ([]byte, error)
```
lidel: We only use this for IPNS, make it explicit, and require a valid Cid instead of a string:

```suggestion
	// GetIPNSRecord retrieves the best IPNS record for a given CID (libp2p-key) from the routing system.
	GetIPNSRecord(context.Context, cid.Cid) ([]byte, error)
```

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
 
-// NodeAPI defines the minimal set of API services required by a gateway handler
+// NodeAPI defines the minimal set of API services required for a read-only
+// gateway handler. These are, for the most part, subsets of the core interface.
+
 type NodeAPI interface {
-	// Unixfs returns an implementation of Unixfs API
-	Unixfs() coreiface.UnixfsAPI
+	// GetUnixFs returns a read-only handle to a file tree referenced by a path.
+	GetUnixFs(context.Context, path.Path) (files.Node, error)
```
lidel:
```suggestion
	GetUnixFsNode(context.Context, path.Path) (files.Node, error)
```

gateway/gateway.go
```diff
@@ -2,36 +2,48 @@ package gateway
...
+
+	// LsUnixFs returns the list of links in a directory.
+	LsUnixFs(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
```
lidel:
```suggestion
	LsUnixFsDir(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
```

gateway/handler.go
```diff
@@ -305,20 +278,6 @@ func (i *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			i.postHandler(w, r)
-			return
-		case http.MethodPut:
-			i.putHandler(w, r)
-			return
-		case http.MethodDelete:
-			i.deleteHandler(w, r)
-			return
-		}
-	}
```
lidel, line 320: :information_source:  I've filled https://github.com/ipfs/kubo/issues/9622 to decide if we move this code back to kubo and do soft deprecation, or break existing users.

lidel (changes requested):
Thank you for identifying these @hacdias. 

I've dropped some ideas to simplify it further + avoid confusing names + follow up for cleaning up writable mode in Kubo.

gateway/handler_codec.go
hacdias: Yeah, I don't think we can get "rid of" IPLD since we have the conversions. Perhaps we can do something along the lines of: GetBlock, convert it to a node ourselves, etc. Let me try.

hacdias: Actually managed to remove it :D

gateway/handler_unixfs_dir.go
```diff
@@ -104,14 +109,7 @@ func (i *handler) serveDirectory(ctx context.Context, w http.ResponseWriter, r *
...
 	}
 
-	// Optimization: use Unixfs.Ls without resolving children, but using the
-	// cumulative DAG size as the file size. This allows for a fast listing
-	// while keeping a good enough Size field.
-	results, err := i.api.LsUnixFsDir(ctx,
-		resolvedPath,
-		options.Unixfs.ResolveChildren(false),
-		options.Unixfs.UseCumulativeSize(true),
-	)
```
hacdias, line 114: Moved this optimization to Kubo's codebase. Each gateway implementation should use whatever they want, some might be faster, some slower. It's the implementers choice. This also allowed me to simplify the API signature to remove the options. 

```diff
- LsUnixFsDir(context.Context, path.Path, ...options.UnixfsLsOption) (<-chan iface.DirEntry, error)
+ LsUnixFsDir(context.Context, path.Resolved) (<-chan iface.DirEntry, error)
```

gateway/handler_unixfs_dir.go
```diff
@@ -61,9 +60,15 @@ func (i *handler) serveDirectory(ctx context.Context, w http.ResponseWriter, r *
...
-	idx, err := i.api.GetUnixFsNode(ctx, idxPath)
+	idxResolvedPath, err := i.api.ResolvePath(ctx, idxPath)
 	switch err.(type) {
 	case nil:
+		idx, err := i.api.GetUnixFsNode(ctx, idxResolvedPath)
+		if err != nil {
+			internalWebError(w, err)
+			return
+		}
+
```
hacdias, line 71: I updated the signature of the getter for UnixFS files:

```diff
- GetUnixFsNode(context.Context, path.Path) (files.Node, error)
+ GetUnixFsNode(context.Context, path.Resolved) (files.Node, error)
```

We already have `ResolvePath` at our disposal. Let's simplify. The change in this code is to reflect that.

gateway/gateway.go
```diff
@@ -5,33 +5,39 @@ import (
...
-	Writable bool
+	Headers map[string][]string
 }
 
-// NodeAPI defines the minimal set of API services required by a gateway handler
-type NodeAPI interface {
-	// Unixfs returns an implementation of Unixfs API
-	Unixfs() coreiface.UnixfsAPI
+// API defines the minimal set of API services required for a gateway handler.
+type API interface {
```
lidel, line 21: nit:  something like `IPFSBackend` would be more descriptive, but can be changed later, won't block on this one thing  (this interface will most likely get revamp in near future anyway)

lidel (approved):
Thank you @hacdias, moving `Gateway.Writable` to https://github.com/ipfs/kubo/pull/9616 is :+1:  
Merging to unblock other PRs based on this.

Issue #142, opened 2023-01-31
gateway: create example that proxies to requests to ?format=raw


lidel:
Make a binary that runs the libipfs gateway code using a virtual blockstore that sends [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) (`?format=raw`) requests to some gateway.

This will be a good warmup before we get custom blockstore from Saturn,
will also allow us to quickly see if the performance is as bad as we expect.

- for now in the same repo, `examples/gateway/block-proxy-gateway` (its late, needs better name)
- CLI tool with two  parameters: port to listen and URL of gateway that supports `?format=raw` responses

This may depend on, or help with  cleaning up API in #61

Jorropo:
Related: https://github.com/ipfs/go-libipfs/issues/139 (tangently)

PR #147, opened 2023-01-31
docs: add example of gateway backed by CAR file

### LINKED ISSUE #141: gateway: create example that hosts a CAR file
Make a binary that runs the libipfs gateway code against a car as a blockstore.

This will be useful as warmup for our gateway work, and also for running compliance tests.

- for now in the same repo, `examples/gateway/car-file-gateway`
- CLI tool with two  parameters: port to listen and locaiton of CAR file

This may depend on, or help with  cleaning up API in #61

hacdias:
This PR adds:

- Example on how to build a Gateway backed by a CAR file and a few smoke tests that check directory traversing, as well as a DAG-JSON response. Closes #141.
- CI that runs the tests from the examples based on #143.
- A readme for the examples based on #143.

Depends on #145.


hacdias:
I'm almost sure that some of the issues on this example will be fixed on top of https://github.com/ipfs/go-libipfs/pull/36.

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,221 @@
...
+func (api *carGateway) lsFromLinks(ctx context.Context, ndlinks []*ipld.Link, settings *options.UnixfsLsSettings) (<-chan iface.DirEntry, error) {
+	links := make(chan iface.DirEntry, len(ndlinks))
+	for _, l := range ndlinks {
+		lr := ft.LinkResult{Link: &ipld.Link{Name: l.Name, Size: l.Size, Cid: l.Cid}}
+
+		links <- api.processLink(ctx, lr, settings) // TODO: can be parallel if settings.Async
+	}
+	close(links)
+	return links, nil
+}
```
hacdias: Code borrowed from https://github.com/iand/meridian. I wonder if there's an easier way to do this. This is massive and I don't want people to have to implement this overly complicated thing.

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,221 @@
...
+
+	ipath := ipfspath.Path(pth.String())
+
+	ipldFetcher := bsfetcher.NewFetcherConfig(api.bsrv)
+	fetcher := ipldFetcher.WithReifier(unixfsnode.Reify)
+	resolver := ipfspathresolver.NewBasicResolver(fetcher)
+
+	node, rest, err := resolver.ResolveToLastNode(ctx, ipath)
+	if err != nil {
+		return nil, fmt.Errorf("resolve to last node: %w", err)
```
hacdias, line 74: This is failing to resolve links. Didn't figure out why yet.

hacdias:
@lidel @aschmahmann I'm happy to receive reviews here. The code I mentioned in the comment above for `LsUnixFsDir` is gigantic and I feel like this is something that I don't want to ask users to have to write. Perhaps wrapping it around a helper could be beneficial?

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,234 @@
...
+func (api *carGateway) GetUnixFsNode(ctx context.Context, p path.Resolved) (files.Node, error) {
+	nd, err := api.resolveNode(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+
+	return unixfile.NewUnixfsFile(ctx, api.dsrv, nd)
+}
+
+func (api *carGateway) LsUnixFsDir(ctx context.Context, p path.Resolved) (<-chan iface.DirEntry, error) {
```
aschmahmann: @hacdias you mentioned that this function was too complicated and you wanted a helper. This is the similar, but (incorrect) code that's copy-pasted from an older version of kubo, you want the latest https://github.com/ipfs/kubo/blob/4e8271b1021f7d31e7b201640e6214a521f328e7/core/coreapi/unixfs.go#L221 which makes adding a helper seem reasonable.

However, it looks like you can toss most of the helper function code given that you know the various parameters (e.g. no resolving children and using cumulative (DAG) size). Will comment more below

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,234 @@
...
+	settings, err := options.UnixfsLsOptions(
+		options.Unixfs.ResolveChildren(false),
+		options.Unixfs.UseCumulativeSize(true),
+	)
```
aschmahmann: These are passed into a function you control you don't need this

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,234 @@
...
+
+	dagnode, err := api.resolveNode(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+
+	dir, err := uio.NewDirectoryFromNode(api.dsrv, dagnode)
+	if err == uio.ErrNotADir {
+		return api.lsFromLinks(ctx, dagnode.Links(), settings)
+	}
```
aschmahmann: This is `LsUnixFsDir` we can't get a `uio.ErrNotADir`, right?

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,234 @@
...
+
+	switch lnk.Cid.Type() {
+	case cid.Raw:
+		// No need to check with raw leaves
+		lnk.Type = iface.TFile
+		lnk.Size = linkres.Link.Size
+	case cid.DagProtobuf:
+		if !settings.ResolveChildren {
+			break
+		}
```
aschmahmann: We know this is true which means basically all of the complexity around types can be removed.

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,234 @@
...
+			switch d.Type() {
+			case ft.TFile, ft.TRaw:
+				lnk.Type = iface.TFile
+			case ft.THAMTShard, ft.TDirectory, ft.TMetadata:
+				lnk.Type = iface.TDirectory
+			case ft.TSymlink:
+				lnk.Type = iface.TSymlink
+				lnk.Target = string(d.Data())
+			}
+			lnk.Size = d.FileSize()
```
aschmahmann: See comment linking kubo code. We're looking at the cheaper cumulative size function

aschmahmann (changes requested):
Left some comments on the complexity with Ls. At a high level though, this PR is a little difficult to merge with no tests since we have no idea if it's correct or not 🙃 (e.g. issues with go-ipfs-files, or directory listing).

You could add a couple smoke tests in a CAR for now and then move onto #142 which IIUC should be roughly testable using sharness tests. Hopefully the code from #142 and here should be almost entirely shared and could make its way into a helper function to make this easier and better tested.

Later as we do more refactoring we can attempt to make most of the complexity here go away with better types that are specific to what the gateway code currently needs.

examples/gateway/car-file-gateway/api.go
hacdias: @aschmahmann yup, I just noticed this and I'm gonna cleanup. It's funny because I was the one who updated the original version last 😅 .

examples/gateway/car-file-gateway/api.go
hacdias: @aschmahmann that's a good question. I traced it back to this commit from 5 years ago: https://github.com/ipfs/kubo/commit/a22cae1bce7dfa9b67ed5c969c469e259eac0fd9 - unfortunately it is not connected to any PR so I can get some more context. I removed it locally and it works well with the directory I am testing with. I wonder if there's some old format that led to this decision with two different types. Do you know anything about it?

hacdias: As per our other convo, this seems to be related to `ipfs ls` also supporting listing the links of files: https://github.com/ipfs/kubo/issues/9541

hacdias:
@aschmahmann thanks for the review. I have tackled the comments and massively simplified the example. I will work on adding some smoke tests.

> (...) helper function to make this easier and better tested. (...)

I'm almost entirely sure that the only difference between this PR and #142 will be the block service. It is quite trivial to have a Gateway API implementation given a block service that conforms to the interface.

examples/gateway/car-file-gateway/api.go
```diff
@@ -0,0 +1,185 @@
...
+	return nil, errors.New("not implemented")
+}
+
+func (api *carGateway) IsCached(ctx context.Context, p ifacepath.Path) bool {
+	rp, err := api.ResolvePath(ctx, p)
+	if err != nil {
+		return false
+	}
+
+	_, err = api.blockService.GetBlock(ctx, rp.Cid())
```
aschmahmann: Why not use blockstore.Has here?

hacdias: Because I didn't know that. Still a noob around these interfaces. Thanks for pointing out!

aschmahmann: Lol no problem, the abstraction layers here could be fewer and cleaner. But hey, we're getting done opportunities to make things easier now 🙃

aschmahmann (approved):
Generally LGTM. We've got some basic tests and hopefully we'll get the main machinery better tested in #142. Approving to remove my prior "Request changes" since it looks good and because I might not be available later today to approve.

hacdias:
@aschmahmann I'm just adding CI for the tests still :)

examples/gateway-car/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Then, you can start the gateway with:
+
+
+```
+./gateway -c data.car -p 8040
+```
+
+Now you can access the gateway in [localhost:8040](http://localhost:8040). It will
```
lidel: We either need to change this to `127.0.0.1` or add subdomain support on `localhost` (https://github.com/ipfs/go-libipfs/issues/149).

Website hosting will usually require subdomain support (otherwise static website generators have invalid relative paths, and no Origin isolation), so I'd include `localhost` subdomain gw in this example.

examples/gateway-car/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	gateway, err := newBlocksGateway(blockService)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	handler := newHandler(gateway, *portPtr)
+
+	address := ":" + strconv.Itoa(*portPtr)
+	log.Printf("Listening on %s", address)
```
lidel: nit: can we make it listen on localhost by defaut and print a clickable URL?

Currently, it's listening on all IPs but prints only port:
```
./gateway -c docs.car -p 8040
2023/02/02 03:30:50 Listening on :8040
```

:thought_balloon: A neat example would be to print not only the gateway listener, but also direct links to roots from CAR header (roots are optional, but `dag export` always has one):

```
./gateway -c docs.car -p 8040
2023/02/02 03:30:50 Listening on https://localhost:8040
2023/02/02 03:30:50 Hosting CAR root at https://localhost:8040/ipfs/QmR2rs1JDUf9UBpLu5ibsT18ACVqRUBLFnP5q8nqoPR6m8
```

lidel (approved):
lgtm, if you address small UX suggestions below, fine to merge

It may be useful to update this PR after  https://github.com/ipfs/go-libipfs/issues/149 is done, and then expose subdomain on `cid.ipfs.localhost:port` but also ok to merge as-is for now

hacdias:
@lidel I made the updates you requested and added a note to #149 so I don't forget to update this example once I move the DNSLink and subdomain code here. Regarding roots, I print all roots, whether there's 0 or 10. I will merge once CI is green.

Issue #149, opened 2023-02-01
gateway: migrate subdomain and dnslink code


lidel:
Logic that parses `Host` header and translates it to content path is still in Kubo.

We need to migrate it as well, because the gateway binary for Saturn  MUST also support subdomain (for `dweb.link`) and DNSLink gateways (for things like `https://en.wikipedia-on-ipfs.org`).

Ref. 
- kubo
  - https://github.com/ipfs/kubo/blob/f54b2bcf6061219f7f481e8660e6707ae3bc3c38/cmd/ipfs/daemon.go#L841-L842
  - https://github.com/ipfs/kubo/blob/4e8271b1021f7d31e7b201640e6214a521f328e7/core/corehttp/hostname.go#L42-L43
- https://github.com/ipfs/specs/blob/main/http-gateways/SUBDOMAIN_GATEWAY.md
- https://github.com/ipfs/specs/blob/main/http-gateways/DNSLINK_GATEWAY.md

Other ToDos:

- [ ] Update CAR Gateway example to support subdomains on localhost

lidel:
@hacdias  this is lower priority than PoC examples (#141  and #142), but something we will need for the production binary for `dweb.link`.


Not sure how tricky this is, but these additional types of gateways are essentially [wrappers executed before passing request to the path one](https://github.com/ipfs/kubo/blob/f54b2bcf6061219f7f481e8660e6707ae3bc3c38/cmd/ipfs/daemon.go#L841-L842).

I think this is a good patter/model to follow in our library.
[See `mux` and `childMux` in  this file](https://github.com/ipfs/kubo/blob/4e8271b1021f7d31e7b201640e6214a521f328e7/core/corehttp/hostname.go#L42-L43).

hacdias:
(Oops, moving on...)

> I think this is a good patter/model to follow in our library. [See mux and childMux in this file](https://github.com/ipfs/kubo/blob/4e8271b1021f7d31e7b201640e6214a521f328e7/core/corehttp/hostname.go#L42-L43).

@lidel Any reason to do so instead of doing the common middleware pattern we see in Go?

```go
func Middleware(http.Handler) http.Handler
```

`*http.Mux` is a `http.Handler` but the opposite is not necessarily true. That is how I implemented it right now, but I can change if there's any reason to use a different pattern that what is common in the Go-land.

lidel:
[...]

PR #151, opened 2023-02-01
docs: add example of gateway that proxies to ?format=raw

### LINKED ISSUE #142: gateway: create example that proxies to requests to ?format=raw

hacdias:
Closes #142. Depends on #147.

Since this is largely based on #147, I'm basing this on top of that. I reorganized the functions so I could re-use them. Part of the fetcher code is mostly based on https://github.com/filecoin-saturn/caboose.

I just have to see how I can pass-through DNSLink and IPNS requests for resolution.


hacdias:
Feedback welcome @lidel @aschmahmann. I'm not sure to which point IPNS is within the scope of this PR. We could use the raw keys, but we would still have to deal with DNSLink (#149).

lidel:
While it is fine to have this example without IPNS/DNSLink,  the bifrost production binary will need it – I've wrote some thoughts https://github.com/protocol/bifrost-gateway/issues/5.

@hacdias perhaps you could implement a PoC here that uses HTTP RPC at `https://node[0-2].delegate.ipfs.io/api/v0/resolve` for `/ipns/` resolution and `/api/v0/routing/get` for IPNS raw records?

hacdias:
@lidel I reworked and added IPNS resolution: DNSLink goes through normal DNS requests implemented in the `namesys` package, and then I added a proxy routing that does IPNS routing requests via `?format=ipns-record`. Haven't added any tests yet, but it would be great to get some feedback.

examples/gateway/blocks.go
```diff
@@ -121,11 +162,16 @@ func (api *blocksGateway) ResolvePath(ctx context.Context, p ifacepath.Path) (if
 		return nil, err
 	}
 
-	if p.Namespace() != "ipfs" {
+	ipath := ipfspath.Path(p.String())
+	ipath, err := resolve.ResolveIPNS(ctx, api.namesys, ipath)
+	if err != nil {
+		return nil, err
+	}
```
lidel: nit:  wrap this in `if ipath.Segments()[0] == "ipns" {`  to skip ResolveIPNS for  `/ipfs` ?

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,30 @@
+# Proxy Gateway
+
+This is an example that shows how to build a Gateway backed by the contents
+of another gateway. In this example, we implement two major structures:
```
lidel:
```suggestion
This is an example of building a Gateway that uses  `application/vnd.ipld.raw`
responses from another gateway acting as a remote blockstore and IPNS resolver.

In this example, we implement two major structures:
```

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,30 @@
...
+## Build
+
+```bash
+> go build -o proxy
+```
+
+## Usage
+
+First, you need a compliant gateway that supports both RAW and IPNS Record response
+types. Once you have it, run the proxy gateway with its address as the host parameter:
```
lidel:
```suggestion
First, you need a compliant gateway that supports both [RAW Block](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) and IPNS Record response
types. Once you have it, run the proxy gateway with its address as the host parameter:
```

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+```
+./proxy -h ipfs.io -p 8040
+```
+
+Now you can access the gateway in [127.0.0.1:8040](http://127.0.0.1:8040). It will
+behave like a regular IPFS Gateway, except for the fact that all contents are provided
+from the CAR file. Therefore, things such as IPNS resolution and fetching contents
+from nodes in the IPFS network won't work.
```
lidel:
```suggestion
behave like a regular IPFS Gateway, except for the fact that it runs no libp2p, and has no local blockstore.
All contents are provided by a remote  gateway and fetched as RAW Blocks and Records, and verified locally.
```

examples/gateway/proxy/blockstore.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return &proxyStore{
+		host:       host,
+		httpClient: client,
+	}
+}
+
+func (ps *proxyStore) fetch(ctx context.Context, c cid.Cid) (blocks.Block, error) {
+	u, err := url.Parse(fmt.Sprintf("http://%s/ipfs/%s?format=raw", ps.host, c))
```
lidel: nit: `http://%s` adds unnecessary roundtrip for every block request (ipfs.io redirects to https://).

Host passed via `-h` is not enough, less headache if you make it `-g` and take full gateway URL, that way it works fine for:
- `-g http://127.0.0.1:8080`
- `-g https://ipfs.io`

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,30 @@
+# Proxy Gateway
```
lidel: More clear what it does.

```suggestion
# Gateway as a Proxy for Remote Blockstore
```

examples/README.md
```diff
@@ -6,4 +6,5 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 
 ## Examples and Tutorials
 
-- [Gateway backed by a CAR file](./gateway-car)
+- [Gateway backed by a CAR file](./gateway/car)
+- [Gateway that proxies block and IPNS requests](./gateway/proxy)
```
lidel:
```suggestion
- [Gateway backed by a remote blockstore and IPNS resolver](./gateway/proxy)
```

examples/gateway/proxy/blockstore.go
```diff
@@ -0,0 +1,111 @@
...
+	resp, err := ps.httpClient.Do(&http.Request{
+		Method: http.MethodGet,
+		URL:    u,
+		Header: http.Header{
+			"Accept": []string{"application/vnd.ipld.raw"},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
```
lidel, line 54: Need to check Status code of the response and return remote error if was anything other than 200.

For example, ipfs.io returns 500 and that breaks page:

> 2023-02-03-021516_677x81_scrot

```console
$  curl 'https://ipfs.io/ipfs/filecoin.io?format=raw' | xxd 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   194  100   194    0     0    710      0 --:--:-- --:--:-- --:--:--   710
00000000: 3c68 746d 6c3e 0d0a 3c68 6561 643e 3c74  <html>..<head><t
00000010: 6974 6c65 3e35 3033 2053 6572 7669 6365  itle>503 Service
00000020: 2054 656d 706f 7261 7269 6c79 2055 6e61   Temporarily Una
00000030: 7661 696c 6162 6c65 3c2f 7469 746c 653e  vailable</title>
00000040: 3c2f 6865 6164 3e0d 0a3c 626f 6479 3e0d  </head>..<body>.
00000050: 0a3c 6365 6e74 6572 3e3c 6831 3e35 3033  .<center><h1>503
00000060: 2053 6572 7669 6365 2054 656d 706f 7261   Service Tempora
00000070: 7269 6c79 2055 6e61 7661 696c 6162 6c65  rily Unavailable
00000080: 3c2f 6831 3e3c 2f63 656e 7465 723e 0d0a  </h1></center>..
00000090: 3c68 723e 3c63 656e 7465 723e 6f70 656e  <hr><center>open
000000a0: 7265 7374 793c 2f63 656e 7465 723e 0d0a  resty</center>..
000000b0: 3c2f 626f 6479 3e0d 0a3c 2f68 746d 6c3e  </body>..</html>
000000c0: 0d0a
```

examples/gateway/proxy/routing.go
```diff
@@ -0,0 +1,126 @@
...
+		Method: http.MethodGet,
+		URL:    u,
+		Header: http.Header{
+			"Accept": []string{"application/vnd.ipfs.ipns-record"},
+		},
+	})
+	if err != nil {
+		fmt.Println(err)
+		return nil, err
+	}
```
lidel, line 94: Same here, need to check Status code of the response and return remote error if was anything other than 200.

lidel (changes requested):
Namesys with custom routing lgtm, 
but for this to be an useful example it needs better  error handling (details inline)

examples/gateway/proxy/blockstore.go
hacdias: I made some updates, but I don't think we're gonna get exactly what you're expecting. Don't forget that this is just a block store and a routing mechanism and we have to conform to the interfaces. I can't just return the error so I build an error with that info.

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This is an example of building a Gateway that uses  `application/vnd.ipld.raw`
+responses from another gateway acting as a remote blockstore and IPNS resolver.
+
+In this example, we implement two major structures:
+
+- [Block Store](./blockstore.go), which forwards the block requests to the backend
+gateway using `?format=raw`, and
+- [Routing System](./routing.go), which forwards the IPNS requests to the backend
+gateway using `?format=ipns-record`. In addition, DNSLink lookups are done locally.
```
lidel, line 15: We did not ship this yet, let's add a note:

```suggestion
- [Routing System](./routing.go), which forwards the IPNS requests to the backend
gateway using `?format=ipns-record`. In addition, DNSLink lookups are done locally.
  - Note: `ipns-record` was introduced just recently in [IPIP-351](https://github.com/ipfs/specs/pull/351) and reference support for it will ship in Kubo 0.19. Until that happens, it may not be supported by public gateways yet. 
```

examples/gateway/blocks.go
```diff
@@ -1,9 +1,9 @@
-package main
+package gateway
```
lidel: nit: having `blocks.go`  it at the top level  looks awkward.
if we have code common to multiple examples, it will be friendlier to move it to `/examples/gateway/common`

examples/gateway/proxy/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Usage
+
+First, you need a compliant gateway that supports both [RAW Block](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) and IPNS Record response
+types. Once you have it, run the proxy gateway with its address as the host parameter:
+
+
+```
+./proxy -h https://ipfs.io -p 8040
+```
```
lidel: :thought_balloon: it may not be clear why someone would want to run a "proxy".  Added some context at the top of the page, but here, renaming it to something like  "verifying-proxy"  will remove guesswork, and communicate a clear value:

````suggestion
```bash
> go build -o verifying-proxy
```

## Usage

First, you need a compliant gateway that supports both [RAW Block](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) and IPNS Record response
types. Once you have it, run the proxy gateway with its address as the host parameter:


```
./verifying-proxy -h https://ipfs.io -p 8040
```
````

File: examples/gateway/proxy/README.md [hunk 13]
```diff
@@ -0,0 +1,31 @@
+# Gateway as a Proxy for Remote Blockstore
+
+This is an example of building a Gateway that uses  `application/vnd.ipld.raw`
+responses from another gateway acting as a remote blockstore and IPNS resolver.
```
lidel:
```suggestion
# Gateway as a Verifying Proxy for Untrusted Remote Blockstore

This is an example of building a Gateway that uses  `application/vnd.ipld.raw`
responses from another gateway acting as a remote blockstore and IPNS resolver.

Key benefits:
1. Verifies raw blocks and IPNS records fetched from untrusted third-party gateways.
2. The proxy provides web gateway functionalities: returns deserialized files and websites, including index.html support, while the remote gateway only needs to support block responses.
```

lidel (approved):
@hacdias lgtm! small asks below, then feel free to merge (ok to add subdomains later) :+1:

hacdias:
@lidel I updated the blockstore to add validation by default, which makes sense for this example, since we are proxying requests to a remote _untrusted_ blockstore. I had missed that. I also added two smoke tests to ensure we get an HTTP error when the remote gives invalid content, and gives 200 and the right content when the content is indeed valid. I will merge once CI is green.

PR #153, opened 2023-02-02
feat: migrate subdomain and dnslink code

### LINKED ISSUE #149: gateway: migrate subdomain and dnslink code

hacdias:
Closes #149.

Some things I want to improve:

- [x] Do not require a name system. Instead, require some function or simple interface dedicated only to solve DNSLink domains. We don't need the full thing.
- [x] Simplify the configuration and leave part of the configuration magic for Kubo.
- [x] Remove `/api` and `/p2p` from hostname and keep in Kubo (related to point before)
- [x] Working example
- [x] Fix the cause of failing Kubo Go tests (https://github.com/ipfs/kubo/pull/9624)
	- ~I'm think it might be caused by us now calling `ResolvePath` instead of `namesys.Resolve`. This is because the namesys.Resolve is part of ResolvePath and ResolvePath does more than just namesys.Resolve. So we catch some weird errors here.~ Unlikely. ResolvePath returns the error of namesys.Resolve.
	- See https://app.circleci.com/pipelines/github/ipfs/kubo/9010/workflows/15e0dfe2-0ee6-4e86-8056-c9cb7d65521c/jobs/97462
	- See https://github.com/ipfs/kubo/pull/9624#discussion_r1095494351
- [x] Restore commented test (will require some work but will also give opportunity to move other some Go gateway tests); or move back to Kubo for now (may involve copying a few helper functions).
- [x] PR in Kubo (https://github.com/ipfs/kubo/pull/9624)

**Note**: sharness expected to fail as we changed the Gateway API interface.


gateway/hostname.go
```diff
@@ -0,0 +1,623 @@
...
+	dns "github.com/miekg/dns"
+
+	mbase "github.com/multiformats/go-multibase"
+
+	nsopts "github.com/ipfs/interface-go-ipfs-core/options/namesys"
+)
+
+type HostnameConfig struct {
+	// NameSystem is the name system used to resolve TXT records for DNSLink hosts.
+	NameSystem namesys.NameSystem
```
lidel: In practice, subdomains use it for IPNS too (used inside ResolvePath, but here we require it explicitly).

We either need to clarify comment, or remove dependency on Namesys and have something like `GetDNSLinkRecord` to match `GetIPNSRecord`

Note:  :point_right:  either way we need to use namesys (even if it is internal to `ResolvePath`). We have special casing for things like `.eth` that won't work with DNS resolver alone. 

```suggestion
	// NameSystem is the name system used to resolve content paths starting with /ipns/id.
	// It finds  TXT records for DNSLink hosts and IPNS records for CID with libp2p-key in subdomains.
	NameSystem namesys.NameSystem
```

gateway/hostname.go
```diff
@@ -0,0 +1,623 @@
...
+
+	// PublicGateways configures behavior of known public gateways.
+	// Each key is a fully qualified domain name (FQDN).
+	PublicGateways map[string]*Specification
+}
+
+// Specification is the specification of an IPFS Public Gateway.
+type Specification struct {
+	// Paths is explicit list of path prefixes that should be handled by
+	// this gateway. Example: `["/ipfs", "/ipns", "/api"]`
```
lidel: _:thought_balloon:just a suggestion,  if this is a pain to clean up, fine to keep it around for now._

```suggestion
	// this gateway. Example: `["/ipfs", "/ipns"]`
```

gateway/hostname.go
```diff
@@ -0,0 +1,623 @@
...
+	NoDNSLink bool
+
+	// InlineDNSLink configures this gateway to always inline DNSLink names
+	// (FQDN) into a single DNS label in order to interop with wildcard TLS certs
+	// and Origin per CID isolation provided by rules like https://publicsuffix.org
+	InlineDNSLink bool
+}
+
+// DefaultPaths are the default path prefixes that are handled by default.
+var DefaultPaths = []string{"/ipfs/", "/ipns/", "/api/", "/p2p/"}
```
lidel: _:thought_balloon:just a suggestion,  if this is a pain to clean up, fine to keep it around for now._

This may require refactor in Kubo, but we dont want these Kubo-specific thingss in go-libipfs.

```suggestion
// DefaultPaths are the default path prefixes that are handled by default.
var DefaultPaths = []string{"/ipfs/", "/ipns/"}
```

gateway/hostname.go
hacdias: I'm confused. This is just the hostname part, AFAIK we only need the NameSys to handle DNSLink. We don't call `ResolvePath` in the hostname code. ResolvePath is provided by the implementer.

lidel: Yes, here it only uses it for checking if `id` in  `/ipns/id`  is a DNSLink name (if it is not, then we assume IPNS).
Introducing entire   `NameSystem`  just for that is overkill.

Thoughts on removing `NameSystem` and using `ResolvePath` instead?

This way it is up to `ResolvePath` implementation to decide if they support IPNS, DNSLink, and how they resolve them.

Removes a lot of API surface, makes it more flexible.

hacdias: @lidel I will try doing so.

gateway/hostname_test.go
```diff
@@ -0,0 +1,316 @@
...
+// 	ns := mockNamesys{}
+// 	n, err := newNodeWithMockNamesys(ns)
+// 	if err != nil {
+// 		t.Fatal(err)
+// 	}
+// 	coreAPI, err := coreapi.NewCoreAPI(n)
+// 	if err != nil {
+// 		t.Fatal(err)
+// 	}
+// 	testCID, err := coreAPI.Unixfs().Add(n.Context(), files.NewBytesFile([]byte("fnord")))
```
lidel: :thought_balloon:  try replacing `testCID` made by `Unixfs().Add` with  inline identity cid like `bafkqaglimvwgy3zakrsxg5cun5jxkyten5wwc2lokvjeycq`

It will behave as a raw block, but wont require any IO.

You can create them like this:
```
$ echo 'hello TestToSubdomainURL' | ipfs add --raw-leaves --inline -Q
```

And then can use in tests without having to do `Unixfs().Add`

hacdias:
@lidel I updated the PR and the Kubo PR: https://github.com/ipfs/kubo/pull/9624. Every test is passing now. I'm not very happy how I ended up handling it: I added a new function to the interface. Read more about the issue here: https://github.com/ipfs/kubo/pull/9624#discussion_r1095494351 - I feel like the alternative would require me to do deeper changes to the Hostname code.

gateway/gateway.go
```diff
@@ -39,6 +39,11 @@ type API interface {
 	// exist due to a missing link, it should return an error of type:
 	// https://pkg.go.dev/github.com/ipfs/go-path@v0.3.0/resolver#ErrNoLink
 	ResolvePath(context.Context, path.Path) (path.Resolved, error)
+
+	// HasDNSLinkRecord returns if the provided path has a DNSLink TXT record.
+	// It does not perform any validation, only checks for the existence of
+	// a DNSLink TXT record.
+	HasDNSLinkRecord(context.Context, string) bool
```
lidel: (as noted in https://github.com/ipfs/kubo/pull/9624#discussion_r1096844466)

I feel this will not only have more utility (allow us passing value to tracing/debug added in the future), but will match existing `GetIPNSRecord` more closely.

```suggestion
	// GetDNSLinkRecord returns if the provided FQDN has a DNSLink TXT record.
	// It does not perform recursive resolution like ResolvePath, only checks for the existence of
	// DNSLink TXT record with path starting with /ipfs/ or /ipns/ and returns the path as-is.
	GetDNSLinkRecord(context.Context, string) (path.Path, error)
```

lidel (changes requested):

lidel (approved):
Thank you @hacdias!

Pushed small doc changes because extracted things lacked context outside Kubo. 

Good enough for now, but we will refine the way configuration happens in the future.
(after `bifrost-gateway`  / Kubo 0.19)

Issue #146, opened 2023-01-31
gateway: migrate sharness tests from Kubo


hacdias:
This is simply a tracking issue to migrate the gateway tests here. This will involve migrating the **many** sharness tests we currently have. And then we can stop running Kubo's tests in go-libipfs CI.

Related work:
- https://github.com/ipfs/gateway-conformance
- https://github.com/ipfs/kubo/pull/9657

BigLep:
2023-03-30 maintainer converation: not going to do this as instead going to rely on conformance tests for coverage: https://github.com/ipfs/boxo/pull/237

PR #156, opened 2023-02-07
test(gateway): migrate Go tests from Kubo

hacdias:
Part of #146. Migrates remaining gateway Go tests from Kubo (see removal in Kubo: https://github.com/ipfs/kubo/pull/9627). Since we don't have a UnixFS API here, I created a  `fixtures.car` file. This file has a single root. Under that root, there is a directory for each test (that requires data) with the name of the test.


gateway/gateway_test.go
```diff
@@ -60,44 +86,681 @@ func (m mockNamesys) GetResolver(subs string) (namesys.Resolver, bool) {
...
+	dagService   format.DAGService
+	resolver     resolver.Resolver
+	namesys      mockNamesys
 }
 
-func newMockApi() *mockApi {
-	return &mockApi{
-		ns: mockNamesys{},
+func newMockAPI(t *testing.T) (*mockAPI, cid.Cid) {
+	r, err := os.Open("./testdata/fixtures.car")
```
lidel, line 98: @hacdias fysa I've moved fixtures to `testdata/` subdir, it seems to be a good convention to follow in future: 

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests.
> – `go help test`

lidel (approved):
Thank you for cleaning this up :heart:

PR #159, opened 2023-02-08
sync: update CI config files

galargh:


galargh:
`peerSeq` is only ever incremented in `GeneratePeers`. Was the plan to use it as peer id (instead of `i`)? 

https://github.com/ipfs/go-libipfs/blob/df98127afa4cbac74eedde9b405439f149abd11b/bitswap/internal/testutil/testutil.go#L56-L67

galargh:
I replaced `i` with `peerSeq`. Might be worth checking if that was the reason why all the tests that use `GeneratePeers` were marked as flaky. I'm going to merge this at EOD if there are no objections.

PR #161, opened 2023-02-10
chore: make gocritic happier

Jorropo:
Solve most results of:
```console
$ gocritic check ./... 2>&1 | grep -v "_test\.go:"
```


hacdias (changes requested):
LGTM, but please rebase.

PR #155, opened 2023-02-06
feat(gateway): add TAR, IPNS Record, DAG-* histograms and spans

hacdias:
Closes #154.

- Kubo PR: https://github.com/ipfs/kubo/pull/9626

PR #167, opened 2023-02-13
fix(gateway): ensure ipfs_http_gw_get_duration_seconds gets updated

### LINKED ISSUE #162: gateway: ipfs_http_gw_get_duration_seconds metric is not updated
https://github.com/ipfs/go-libipfs/blob/1c6ed2e96f7753a3452607e2a429311a31099cf4/gateway/handler.go#L240-L242

It should be updated  for every request type (global histogram across all response types), does not seem to be happening atm

hacdias:
In #155 I overlooked where I was returning the success status and ended up returning `false` for all UnixFS files. I added more nested success status returning so the metric is more accurate now. Closes #162.

### DIFF
--- a/gateway/handler_unixfs.go
+++ b/gateway/handler_unixfs.go
@@ -29,8 +29,7 @@ func (i *handler) serveUnixFS(ctx context.Context, w http.ResponseWriter, r *htt
 	// Handling Unixfs file
 	if f, ok := dr.(files.File); ok {
 		logger.Debugw("serving unixfs file", "path", contentPath)
-		i.serveFile(ctx, w, r, resolvedPath, contentPath, f, begin)
-		return false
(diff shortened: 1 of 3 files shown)


hacdias:
CI failures seem to be related to https://github.com/ipfs/go-libipfs/commit/8c0677966f9d01d77b7cbedf88b8abfacae22422. it's being worked in Kubo.

lidel (changes requested):
ack.  we need to make decision what to do with failing sharness, not having it is risk, not having correct metrics is also bad.

Let's wait for answer on https://github.com/ipfs/kubo/pull/9643#issuecomment-1428987793

lidel (approved):
Thanks, re-run sharness,green,  lgtm

PR #168, opened 2023-02-13
feat(gateway): metric for implicit index.html in dirs

### LINKED ISSUE #166: gateway: add metric for implicit `index.html`
Ref. https://github.com/ipfs/go-libipfs/blob/1c6ed2e96f7753a3452607e2a429311a31099cf4/gateway/handler.go#L244-L253

We track (histograms):
-  `gw_unixfs_file_get_duration_seconds`  (files) and
-   `gw_unixfs_gen_dir_listing_get_duration_seconds` (generated html dir listings). 

We are missing a metric that tracks the case where request was for a directory with index.html. It is necessary to see how popular this use case is, and if we should introduce optimizations to avoid round trips for `index.html`. 

## Proposed change

Add new histogram ([here?](https://github.com/ipfs/go-libipfs/blob/1c6ed2e96f7753a3452607e2a429311a31099cf4/gateway/handler_unixfs_dir.go#L78-L80) for tracking how long it took to find and load `index.html` when a dir path was requested:

-   `gw_unixfs_dir_index-html_get_duration_seconds`

hacdias:
Fixes #166, based on #162.

This is an interesting one. Please note that serving `index.html` produces a metric! Should we add a new metric and ensure only one is emitted, or should we emit both?

https://github.com/ipfs/go-libipfs/blob/1c6ed2e96f7753a3452607e2a429311a31099cf4/gateway/handler_unixfs_dir.go#L78-L80

https://github.com/ipfs/go-libipfs/blob/1c6ed2e96f7753a3452607e2a429311a31099cf4/gateway/handler_unixfs_file.go#L99-L102

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -74,14 +74,15 @@ type handler struct {
 	unixfsGetMetric            *prometheus.SummaryVec // deprecated, use firstContentBlockGetMetric
 
 	// response type metrics
-	getMetric                 *prometheus.HistogramVec
-	unixfsFileGetMetric       *prometheus.HistogramVec
-	unixfsGenDirGetMetric     *prometheus.HistogramVec
-	carStreamGetMetric        *prometheus.HistogramVec
-	rawBlockGetMetric         *prometheus.HistogramVec
-	tarStreamGetMetric        *prometheus.HistogramVec
-	jsoncborDocumentGetMetric *prometheus.HistogramVec
-	ipnsRecordGetMetric       *prometheus.HistogramVec
+	getMetric                    *prometheus.HistogramVec
+	unixfsFileGetMetric          *prometheus.HistogramVec
+	unixfsDirIndexGetMetric      *prometheus.HistogramVec
+	unixfsGenDirListingGetMetric *prometheus.HistogramVec
+	carStreamGetMetric           *prometheus.HistogramVec
+	rawBlockGetMetric            *prometheus.HistogramVec
+	tarStreamGetMetric           *prometheus.HistogramVec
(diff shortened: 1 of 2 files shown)


gateway/handler.go
```diff
@@ -246,6 +247,11 @@ func newHandler(c Config, api API) *handler {
 			"gw_unixfs_file_get_duration_seconds",
 			"The time to serve an entire UnixFS file from the gateway.",
 		),
+		// UnixFS: time it takes to find and serve an index.html file with directory listing.
+		unixfsDirIndexGetMetric: newHistogramMetric(
+			"gw_unixfs_dir_index-html_get_duration_seconds",
+			"The time to find and serve an index.html file for a directory listing from the gateway.",
```
```suggestion
		// UnixFS: time it takes to find and serve an index.html file on behalf of a directory.
		unixfsDirIndexGetMetric: newHistogramMetric(
			"gw_unixfs_dir_index-html_get_duration_seconds",
			"The time to serve an index.html file on behalf of a directory from the gateway. This is a subset of gw_unixfs_file_get_duration_seconds.",
```

lidel (changes requested):
Thanks! 

> Please note that serving index.html produces a metric! Should we add a new metric and ensure only one is emitted, or should we emit both?

Yes, we want to emit both. This one is a subset of `serveFile`, a special case which we want to track separately to see how often it is done. 

Lgtm, but this is blocked by https://github.com/ipfs/go-libipfs/pull/167#pullrequestreview-1296716489 too, if it does not get resolved, we will discuss our options during 1:1.

lidel (approved):
rebased, re-run sharness, all green now, thanks!

PR #177, opened 2023-02-17
fix: do not send back content length for car files

### LINKED ISSUE #9651: sharness - t0118 is broken - response for application/vnd.ipld.car SHOULD NOT include Content-Length
### Checklist

- [X] This is a bug report, not a question. Ask questions on [discuss.ipfs.io](https://discuss.ipfs.io).
- [X] I have searched on the [issue tracker](https://github.com/ipfs/kubo/issues?q=is%3Aissue) for my bug.
- [X] I am running the latest [kubo version](https://dist.ipfs.tech/#kubo) or have an issue updating.

### Installation method

ipfs-update or dist.ipfs.tech

### Version

```Text
Kubo version: 0.18.1
Repo version: 13
System version: arm64/darwin
Golang version: go1.19.1
```

### Config

_No response_

### Description

In sharness t0118 it seems that the test is misusing the invert option in grep:

https://github.com/ipfs/kubo/blob/ed4d6b7d4187512ede4aeb19188a383d9aa55d45/test/sharness/t0118-gateway-car.sh#L93-L96

The comment says that we should NOT return a `Content-Length`, but we do:

```
› cat curl_output | grep Content-Length                                                                                                                        ☹
< Access-Control-Expose-Headers: Content-Length
< Content-Length: 127
```

(tested with Kubo and a car gateway from [go-libipfs](https://github.com/ipfs/go-libipfs/tree/main/examples/gateway/car))

Changing the test to : `! grep "< Content-Length:" curl_output` reveals the error.

I found a few other cases where we use `grep -v` and that might have a similar issue:

```
> ack 'grep.*-v' ./ - | pbcopy
t0119-prometheus.sh
t0250-files-api.sh
t0140-swarm.sh
t0060-daemon.sh
t0080-repo.sh
t0230-channel-streaming-http-content-type.sh
t0141-addfilter.sh
t0046-id-hash.sh
```

laurentsenta:
> we should not have this header because we don't know how long CAR stream will be

Fixes https://github.com/ipfs/kubo/issues/9651

⚠️ This relies on a patch to golang: https://github.com/golang/go/pull/58578


laurentsenta:
Team, I think I found the "correct" fix for this issue, but it requires a change in golang's `net/http` library. I created the PR there, we might have to keep this PR on hold for some time (marking as draft in the meantime).

Let me know if we need this patched faster, even if we have to implement a workaround.

galargh:
> Let me know if we need this patched faster, even if we have to implement a workaround.

Is the workaround to remove `Content-Length` header from the response? That sounds simple enough and we wouldn't have to wait the entire Go release cycle to fix the behaviour here.

BigLep:
@laurentsenta : what are the next steps here.  Are the changes making it into the next go release?  If not, I think we should do workaround for now to not keep this open for 6+ months.

laurentsenta:
Thanks for the bump, so the patches have been merged and should get in for the next release (1.21):

https://go-review.googlesource.com/c/net/+/471535
https://go-review.googlesource.com/c/go/+/495115

(cc @Jorropo)

@galargh sorry for missing your message

> Is the workaround to remove Content-Length header from the response? That sounds simple enough and we wouldn't have to wait the entire Go release cycle to fix the behaviour here.

From my understanding, the header is written [directly to the stream](https://go-review.googlesource.com/c/go/+/495115/7/src/net/http/server.go#1231); I don't think you can "remove" it like you'd remove an object from a map.

This means: we either copy and paste and maintain the whole chunkWriter to use our patch, or we wrap the regular writer with a custom writer that buffers the stream and filter the header we want to drop. That wrapper would have to be HTTP-aware (you want to drop "Content-Length: ####" when it's a header, not in the payload).

The golang's patch seemed like a better option here.

laurentsenta:
Discussion also happening in https://github.com/ipfs/gateway-conformance/pull/64

lidel:
I believe we no longer need this PR for reasons listed in https://github.com/ipfs/kubo/issues/9651#issuecomment-1654381331, namely:
> - Content-Length can be returned when gateway knows the total size, and in golang (boxo/kubo) it may happen when a small DAG is flushed by the response writer

Thank you @laurentsenta  for pointing this out, forced us to remove ambiguity around mandatory/optional behavior here.

PR #183, opened 2023-02-23
refactor: use assert in remaining gateway tests

hacdias:
Just use the `assert` library more thoroughly.


hacdias:
I'm not exactly sure how the coverage decreased, and there is no coverage diff. It doesn't make sense to me since I did not change the tests themselves.

lidel (approved):

PR #187, opened 2023-02-27
fix(gateway): return 500 for all /ip[nf]s/id failures

### LINKED ISSUE #185: gateway: unify root identifier resolution errors
> Extracted from https://github.com/ipfs/kubo/pull/9660#pullrequestreview-1310527962 so we dont forget

I was thinking about this again, both from semantic perspective, but also given your feedback on how hairy this gets.
I am leaning towards returning 500s for all invalid root identifiers, just for the sake of consistency and simplicity

@hacdias  would like to hear your thoughts if below makes sense (if you dont have any concerns, feel free to execute on this):

- `/ipfs/this-is-not-a-cid` →  we were thinking about 400 Bad Request, but it may be better/easier to return 500 Server Error, because maybe today we know this CID is not valid, but in 10 years the same software could be wrong? Maybe it is a valid CIDv8 that this old software can't recognize? 400 could be wrong, 500 will always kinda fit.
- `/ipns/this-domain-does-not-exist` → 500 (we dont know if domain does not exist, or was resolution error)
- `/ipns/{cid-libp2p-key-that-cant-be-resolved}` → 500  (if it is a valid key, resolution error, if not, we will fallbck to dnslink, and that also cant be proven negative)

## TLDR

Return HTTP 500 Server Error for all  `/ip[nf]s/id`  failures. 

## TODO

- [x] PR here: https://github.com/ipfs/go-libipfs/pull/187
  - [x] Side-quest:  [`IsErrNotFound` func](https://github.com/ipfs/go-libipfs/blob/00e02499517393b13a22ae293091e861e6f796ca/gateway/handler.go#L581-L586) probably should not have [`routing.ErrNotFound`](https://github.com/libp2p/go-libp2p/blob/1ad0a50be62b1e00388b9d85fc997d66e8910c9d/core/routing/routing.go#L14-L15), because these errors are for /ipns/id, which is we always want to be 500.
- [x] PR in Kubo and tests passing: https://github.com/ipfs/kubo/pull/9660
- [x] PR against ipfs/specs, documenting this behavior https://github.com/ipfs/specs/pull/380

hacdias:
Closes #185.

- [ ] Kubo PR: https://github.com/ipfs/kubo/pull/9660
- [ ] Specs PR https://github.com/ipfs/specs/pull/380

### DIFF
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -352,9 +352,11 @@ func TestGatewayGet(t *testing.T) {
 	}{
 		{"127.0.0.1:8080", "/", http.StatusNotFound, "404 page not found\n"},
 		{"127.0.0.1:8080", "/" + k.Cid().String(), http.StatusNotFound, "404 page not found\n"},
+		{"127.0.0.1:8080", "/ipfs/this-is-not-a-cid", http.StatusInternalServerError, "failed to resolve /ipfs/this-is-not-a-cid: invalid path \"/ipfs/this-is-not-a-cid\": invalid CID: illegal base32 data at input byte 3\n"},
 		{"127.0.0.1:8080", k.String(), http.StatusOK, "fnord"},
 		{"127.0.0.1:8080", "/ipns/nxdomain.example.com", http.StatusInternalServerError, "failed to resolve /ipns/nxdomain.example.com: " + namesys.ErrResolveFailed.Error() + "\n"},
 		{"127.0.0.1:8080", "/ipns/%0D%0A%0D%0Ahello", http.StatusInternalServerError, "failed to resolve /ipns/\\r\\n\\r\\nhello: " + namesys.ErrResolveFailed.Error() + "\n"},
+		{"127.0.0.1:8080", "/ipns/k51qzi5uqu5djucgtwlxrbfiyfez1nb0ct58q5s4owg6se02evza05dfgi6tw5", http.StatusInternalServerError, "failed to resolve /ipns/k51qzi5uqu5djucgtwlxrbfiyfez1nb0ct58q5s4owg6se02evza05dfgi6tw5: " + namesys.ErrResolveFailed.Error() + "\n"},
 		{"127.0.0.1:8080", "/ipns/example.com", http.StatusOK, "fnord"},
 		{"example.com", "/", http.StatusOK, "fnord"},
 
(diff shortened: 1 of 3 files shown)


gateway/gateway_test.go
```diff
@@ -352,9 +352,11 @@ func TestGatewayGet(t *testing.T) {
 	}{
 		{"127.0.0.1:8080", "/", http.StatusNotFound, "404 page not found\n"},
 		{"127.0.0.1:8080", "/" + k.Cid().String(), http.StatusNotFound, "404 page not found\n"},
+		{"127.0.0.1:8080", "/ipfs/this-is-not-a-cid", http.StatusInternalServerError, "failed to resolve /ipfs/this-is-not-a-cid: invalid path \"/ipfs/this-is-not-a-cid\": invalid CID: illegal base32 data at input byte 3\n"},
```
hacdias, line 355: I really don't like this tests where we compare the body for errors. Error information may change slightly depending on the backing implementation. I think rewriting these tests can be part of #156 or even #183.

Yes, we should compare the body for files we are expecting I think, and stuff like custom 404s, but not necessarily for errors. As long as the status is correct, the test should pass. Similarly to `TestGatewayInternalServerErrorInvalidPath`.

lidel: @hacdias these don't do much for UX, but also we dont expect these messages to change often.

Once set in stone, make good canaries to hint someone in transitive dependencies messed up refactor :) 
I would keep this, at least until go-libipfs/gateway and rhea work is in progress – better to have more tests than less.

I remember in the past an external library (like go-cid or go-multihash/base, don't remember which one) changed something, and we detected it in Kubo, 2 layers later, because a specific error message changed to a generic one.

ps. At some point we will add more friendly HTML error responses if a request was made with `Accept` that includes `text/html` (web browser), like we did for dag-cbor.

gateway/handler.go
```diff
@@ -580,16 +575,13 @@ func webError(w http.ResponseWriter, err error, defaultCode int) {
 }
 
 func isErrNotFound(err error) bool {
-	return isErrNoLink(err) ||
-		errors.Is(err, routing.ErrNotFound) ||
-		err == routing.ErrNotFound ||
```
lidel, line 585: :thought_balloon:  :+1:  for removing these, it looked like a dead code that could cause hard-to-debug bugs in the futture.

lidel (approved):
Thank you.

Merging this to fix Kubo, unify behavior , and unblock future spec work around 502 and 504.

PR #194, opened 2023-03-03
feat: support HTTP 429 with Retry-After

### LINKED ISSUE #188: gateway: add 429 handling and ErrGatewayRetryAfter
We need a mechanism for signaling to the client when gateway has a cached failure, and when it is ok to retry.
The initial use will be https://github.com/filecoin-saturn/caboose/issues/44 but this will be useful beyond that – in the future, gateway operators will be able to write custom throttling logic, and this error will take care of HTTP semantics.

I propose we:
-  reuse [HTTP 429 Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) error code with  [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) HTTP header
- [ ] in go-libipfs/gateway: add special handling of  `gateway.ErrGatewayRetryAfter` in the [handler.go#L565-L578](https://github.com/ipfs/go-libipfs/blob/0d5d1ca6e251214a99604288379f73fbe873bd95/gateway/handler.go#L565-L578), so it produces HTTP 429 response  and if `ErrGatewayRetryAfter.RetryAfter` attribute is set to value `>0` set the header on the response.
- [x] specs are already in place https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#429-too-many-requests

hacdias:
This closes #188.

- Moved all - I think - logic related to errors to `errors.go`. I also added some simply sanity tests for this new error as it is its own custom type to support the `RetryAfter` field.
- Added `gateway.ErrTooManyRequests`. Returning an error of this type will make the gateway return 429. If the field `RetryAfter` above 0, it will set the `Retry-After` header. Also added a more "e2e" test for this, albeit arguable unnecessary since I added tests for `webError`.

### DIFF
--- a/gateway/errors.go
+++ b/gateway/errors.go
@@ -3,8 +3,10 @@ package gateway
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 
 	ipld "github.com/ipfs/go-ipld-format"
 	"github.com/ipfs/go-path/resolver"
@@ -16,11 +18,15 @@ var (
 )
 
 type ErrTooManyRequests struct {
-	RetryAfter uint64
+	RetryAfter time.Duration
 }
 
 func (e *ErrTooManyRequests) Error() string {
-	return http.StatusText(http.StatusTooManyRequests)
+	text := http.StatusText(http.StatusTooManyRequests)
+	if e.RetryAfter != 0 {
+		text += fmt.Sprintf(", retry after %s", e.RetryAfterHuman())
+	}
+	return text
 }
 
 func (e *ErrTooManyRequests) Is(err error) bool {
@@ -30,6 +36,18 @@ func (e *ErrTooManyRequests) Is(err error) bool {
 	default:
 		return false
 	}
+}
+
+func (e *ErrTooManyRequests) RetryAfterRoundSeconds() time.Duration {
+	return e.RetryAfter.Round(time.Second)
+}
+
+func (e *ErrTooManyRequests) RetryAfterHuman() string {
+	return e.RetryAfterRoundSeconds().String()
+}
+
+func (e *ErrTooManyRequests) RetryAfterHeader() string {
+	return strconv.Itoa(int(e.RetryAfterRoundSeconds().Seconds()))
 }
 
 func webError(w http.ResponseWriter, err error, defaultCode int) {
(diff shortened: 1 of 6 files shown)


gateway/errors.go
```diff
@@ -0,0 +1,98 @@
...
+			w.Header().Set("Retry-After", strconv.FormatUint(tooManyRequests.RetryAfter, 10))
+		}
+
+		code = http.StatusTooManyRequests
+	}
+
+	http.Error(w, err.Error(), code)
+	if code >= 500 {
+		log.Warnf("server error: %s", err)
+	}
```
lidel: I assume this was for debug? I'd remove it – this will be spamming logs too much.

gateway/errors.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/ipfs/go-path/resolver"
+)
+
+var (
+	ErrGatewayTimeout = errors.New(http.StatusText(http.StatusGatewayTimeout))
+	ErrBadGateway     = errors.New(http.StatusText(http.StatusBadGateway))
+)
+
+type ErrTooManyRequests struct {
+	RetryAfter uint64
```
lidel: nit:  switch to time.Duration like we do [here](https://github.com/filecoin-saturn/caboose/blob/98dc6074d2940302f61c23a870adc86d681cbbee/caboose.go#L106).
consistency + makes things easier to work with if we need to round things to full seconds

We also want helpersm this will help us avoid writing the same boilerplate everywhere (would refactor caboose to use this):


- `func (e *ErrToomanyRequests) RetryAfterRoundSeconds() time.Duration` doing `RetryAfter.Round(time.Second)`
- `func (e *ErrToomanyRequests) RetryAfterHuman() string` which does `RetryAfterRoundSeconds.String()` (human-readable string rounded to full seconds like `5m` `1h30s`)
- `func (e *ErrToomanyRequests) RetryAfterHeader() string` which does `strconv.Itoa(int(RetryAfterRoundSeconds.Seconds()))` (the value for `Retry-After` HTTP header)

lidel (changes requested):

lidel (approved):
:thought_balloon: something I've realized during this review is that `Retry-After` is not limited to 429 Too Many Requests, which is used when user is sending too many requests. There are cases where  `Retry-After` is returned with  503 Service Unavailable (when it is not specific to user or their IP, could be global outage). 

Pushed two commits which replace all standalone errors with generic wrappers: `ErrorResponse` and `ErrorRetryAfter`, and attempted to wire that up in https://github.com/ipfs/bifrost-gateway/pull/56  – existing tests pass and does the trick, but I feel we could simplify things somehow.

@hacdias lmk if you have any idea how to make this better (if not, we can merge as-is, to unblock `Retry-After`, and refine later)

hacdias:
@lidel I think the way you rewrote it is a good idea. I'm not sure how we could simplify this now, but maybe later we will figure something out. Merged for now.

PR #195, opened 2023-03-06
bitswap/client: add option to disable duplicated block stats

hsanjuan:
Keeping a counter should not require query-ing the blockstore repeteadly for every single block received, so I have added an option to disable it.

Currently there is the assumption that doing Has() calls is cheap. That can be correct for most cases but also badly incorrect for others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The following emojis are used to highlight certain changes:
 
 * `boxo/gateway`:
   * A new `WithResolver(...)` option can be used with `NewBlocksBackend(...)` allowing the user to pass their custom `Resolver` implementation.
+* `boxo/bitswap/client`:
+  * A new `WithoutDuplicatedBlockStats()` option can be used with `bitswap.New` and `bsclient.New`. This disable accounting for duplicated blocks, which requires a `blockstore.Has()` lookup for every received block and thus, can impact performance.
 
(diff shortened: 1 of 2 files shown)


Jorropo:
I don't think anyone is using this to compare against the actual blockstore, I think it's more meant when you download the same block from multiple nodes.
There an in memory LRU or refactoring this to compare against our wantlist (if we get a block that isn't in our wantlist anymore we can assume it may be a duplicate). Would solve the same thing.

hsanjuan:
I am not sure what this stat is being used for or why it is implemented like this, I just want the ability to disable it.

aschmahmann:
@Jorropo any objection to merging this?

hacdias:
Ping @Jorropo

Jorropo (approved):
That fine, it solves the problem and allows to optin.

Jorropo:
@hsanjuan sorry for the delay, this code is good and doesn't make anything worst even if it could be optimized.
Could you please add a changelog entry ?

hsanjuan:
@Jorropo now without allocating an empty slice

Jorropo (approved):
LGTM

Jorropo:
A test would be good to catch regressions but I think this is fine.

hsanjuan:
Merge and will do a follow up with a test some day. :eyes:

PR #199, opened 2023-03-08
fix(gateway): 500 on panic, recover on WithHostname

mathew-cf:
By default, go will write a 200 status code if none is written. During a panic, a 500 is more appropriate. This will be a no-op if the status code has already been written.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -283,6 +283,8 @@ func (i *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			log.Error("A panic occurred in the gateway handler!")
 			log.Error(r)
 			debug.PrintStack()
+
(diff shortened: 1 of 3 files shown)


hacdias (approved):
This will also likely emit a `http: superfluous response.WriteHeader` in case something has already been written to the client. I think that is a fine trade-off to ensure that the client receives a 500 when there is a panic though. Thoughts @lidel?

guseggert:
Can we add a test for this?

guseggert (changes requested):
Can we add a test for this?

hacdias:
@guseggert I stepped on here and added a test myself. Also, added the panic handler to the `WithHostname` middleware. As of right now, if `GetDNSLinkRecord` panics, there's no error recovery.

guseggert (approved):

PR #176, opened 2023-02-16
feat: refactor gateway api to operate on higher level semantics

aschmahmann:
> Kubo PR that runs sharness tests: https://github.com/ipfs/kubo/pull/9681 (sharness in this PR will fail, but MUST be green in the linked PR)

This is a start at a gateway refactor so that we operate on higher level semantics rather than lower level ones ipfs/kubo#173.

This enables us to have all the gateway HTTP code separate from whether the data was retrieved over Bitswap, via requesting the entire result as a CAR file, etc.

This probably could've been done while still keeping a bunch of the traversal code in the gateway (and collecting metrics like first block times, etc.). However, IMO this seemed unnecessary given that it should be the job of the API implementations to verify the data is correct and the gateway code's job to "HTTP-ify" it and record some higher level metrics.

The result is the a lot of the code has been inverted, since instead of the order being "get a bit -> verify -> get more -> verify..." we have "resolve mutable -> do top level request and get async response -> process async response and emit the correct headers".

Given that the implementations (e.g. of the BlockGateway or the interface-go-ipfs-core one) aren't built yet this is all obviously still in flux, but wanted to put something up in case people think this approach is nuts 😅. For this to all be usable I suspect we'll probably need to have some shared utilities so we don't replicate some of the annoying gateway traversal code multiple times (e.g. handling _redirects).

Also addresses:

- https://github.com/ipfs/bifrost-gateway/issues/60

Requirements:

~- [ ] https://github.com/ipfs/go-unixfs/pull/142~ No longer needed.

Issue #198, opened 2023-03-08
Consolidate IPFS Path libraries under `boxo/path`


lidel:
## How many <del>clowns</del> path deps can you fit in <del>a clown car</del> one line of code?

During review of Gateway API refactor from https://github.com/ipfs/go-libipfs/pull/176,  I've realized there are cases, when we use three "path types" from three different libraries. This culminated in this fun one-liner:

```go
imPath, err := NewImmutablePath(ifacepath.New(ipath.String()))
```

- `ipath` is https://github.com/ipfs/go-path (includes resolvers, see https://github.com/ipfs/kubo/pull/9750)
   - bonus: old vendored version in https://github.com/filecoin-project/lotus/tree/master/lib/oldpath
- `ipfspath` is https://github.com/ipfs/interface-go-ipfs-core/tree/master/path
- `ImmutablePath` is recent invention in `go-libipfs/gateway` because above did not provide a type for representing  this state

## Proposed path forward

Thoughts on deduplicating/consolidating under `boxo/path`?
It honestly feels like if go-libipfs has a purpose, it is to provide a solution for reusable things like this.

I feel we will make better Gateway  API in #176 if we don't use `ifacepath` but  use a single `go-libipfs/path` library that covers `Path`, `Immutable` and `Resolved` states with distinct types/interfaces. 

I assume we want to follow "copy" strategy from https://github.com/ipfs/go-libipfs/issues/191 here, to minimize blast radius?

We could make this consolidation in steps, first switch gateway, so the public interface from https://github.com/ipfs/go-libipfs/pull/176 only uses types from `boxo/path`, and in the future switch namesys/kubo to that as well.

## Things to look out for

During the consolidation, the following issues and Pull Requests from the relevant repositories should be taken into consideration.

- https://github.com/ipfs/go-path/issues/54
- https://github.com/ipfs/go-path/pull/29
- https://github.com/ipfs/go-path/pull/61
- https://github.com/ipfs/interface-go-ipfs-core/issues/15
- https://github.com/ipfs/interface-go-ipfs-core/issues/33

cc  @ipfs/kubo-maintainers  for feedback (paths are everywhere, and as much as I'd like to clean this up, I don't want to add unnecessary work for anyone)

hacdias:
[...]

PR #204, opened 2023-03-15
refactor(gateway): use a single path struct internally

aschmahmann:
Builds off of #176 for separate review and merge.

Things that feel good about the PR:
- We are mostly passing a single type around instead of many
- Less reliance on variable naming conventions since we can rely on field names instead

Some things that don't feel great about this PR:
- Which fields in the struct are filled out at any given point in the code is implied rather than type checked
   - Could potentially be fixed by having more types as we progress from `maybeMutable -> immutable -> resolved to last CID + path`
- We use `path.Resolved` which isn't type checked as immutable (if we wanted we could switch to a `finalCID` and `finalPathRemainder` which would probably do the job) 
- There's a bit of a disconnect around _redirects 200s where the finalResolvedPath includes following the redirect but the rest of the contentPathRequest struct is pre-redirect (not sure if this matters much, or if there's an obvious clean way to handle this other than introducing another type or passing more variables around)

Note: aside from this straight refactor there are some issues likely uncovered here (e.g. changing handleNonUnixFSRequestErrors to take a contentPath and choosing a consistent path type to report with).


hacdias (commented):
The changes look fine overall, but I share some of your sentiments. I would first like to ask why we do have 4 paths now. Before we had only 2, then on #176 we have 3. Why 4 now? Is it possible to remove any? 

For example, do we need `immutablePartiallyResolvedPath`, or could we potentially just jump straight to `finalResolvedPath`?

aschmahmann:
> Before we had only 2

We had only two types before `ipath.Path` and `ipath.Resolved`.

> then on https://github.com/ipfs/go-libipfs/pull/176 we have 3

In that PR there are 3 Go types, but 4 different kinds of paths being moved around:
- the requested (maybe mutable) path (`ipath.Path`)
- the immutable version of the requested path (`ImmutablePath`), which was added so we could use some type safety to make sure the data was actually immutable and not have to do type checks or wasted resolutions in `IPFSBackend` implementations
- `maybeResolvedImPath`, not a great name but takes into account that we have done some resolution of the path
- the final resolved path

> Why 4 now? Is it possible to remove any?
> For example, do we need immutablePartiallyResolvedPath, or could we potentially just jump straight to finalResolvedPath?

Nothing new here, but yes I think we could probably remove `immutablePartiallyResolvedPath`. You might have to do a bit of type massaging since `ImmutablePath` is not a `ipath.Resolved` and some functions need to take an `ImmutablePath`.

If you have a way you'd like to change `ImmutablePath` or the function signatures to make things better that's fine. Alternatively, you can probably just `immutablePartiallyResolvedPath` and use type conversions. I'd probably change `finalResolvedPath` to be two fields or a struct with CID + Remainder components though so you don't accidentally "unresolve" part of the path by passing `resolvedPath.String()`.

hacdias:
@lidel @aschmahmann is this something we still want to do, or can this be done in #198?

lidel:
The amount of force pushes we did here makes it impossible to review :see_no_evil: 
@aschmahmann  do you remember which commits are relevant here, we can do interactiver rebase and drop everything else?


I think  we probably want to park this until we have time  for https://github.com/ipfs/boxo/issues/198.

aschmahmann:
closing this until we take a stab at #198

PR #205, opened 2023-03-15
feat: return 400 on /ipfs/invalid-cid 

### LINKED ISSUE #200: gateway: return 400 on /ipfs/invalid-cid
On #185 we unified the behaviour of all resolution errors to return 500. However, this raised the concern that requests to `/ipfs/invalid-cid` should still return 400. Initially, we decided not to do so since we could not find an easy way to return it and with a unified behaviour, it would be easier for implementations.

## TODO

- [ ] Create PR here to ensure `/ipfs/invalid-cid` returns 400 Bad Request. Likely, we have to detect the invalid CID error type somehow. We may want to implement the `.Is` interface for simplicity. https://github.com/ipfs/kubo/pull/9726
- [ ] Update ipfs/specs to reflect this change: https://github.com/ipfs/specs/pull/384

hacdias:
This closes #200. Before merging:

- [x] https://github.com/ipfs/go-cid/pull/150
	- [x] Merged
	- [x] Tagged
	- [x] Updated here

After merging:

- [ ] https://github.com/ipfs/kubo/pull/9726
- [ ] https://github.com/ipfs/specs/pull/384

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/ipfs/go-blockservice v0.5.0
-	github.com/ipfs/go-cid v0.3.2
+	github.com/ipfs/go-cid v0.4.0
 	github.com/ipfs/go-datastore v0.6.0
 	github.com/ipfs/go-fetcher v1.6.1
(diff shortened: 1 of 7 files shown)


guseggert (approved):

hacdias:
Sharness and go tests passing in https://github.com/ipfs/kubo/pull/9726. Merging.

PR #212, opened 2023-03-16
feat: bump bitswap message package to v1

guseggert:
This resolves issues registering protobufs from bitswap in go-libipfs when protobufs from go-bitswap are also registered in the same process.

We should really get off of gogoproto, the whole project is deprecated, but that is a larger effort.


guseggert:
I tested this as follows:

```
import (
	"fmt"
	"testing"

	"github.com/ipfs/go-bitswap/message"
	libipfsmsg "github.com/ipfs/go-libipfs/bitswap/message"
)

func TestThing(t *testing.T) {
	msg1 := message.New(true)
	msg2 := libipfsmsg.New(true)
	fmt.Printf("%s\n%s\n", msg1, msg2)
}
```


prior to this change, that panics like we have seen

```
go test -v -count=1 -run='TestThing' .
panic: proto: duplicate enum registered: bitswap.message.pb.Message_BlockPresenceType

goroutine 1 [running]:
github.com/gogo/protobuf/proto.RegisterEnum(...)
	/home/gus/.asdf/installs/golang/1.19.3/packages/pkg/mod/github.com/gogo/protobuf@v1.3.2/proto/properties.go:520
github.com/ipfs/go-libipfs/bitswap/message/pb.init.0()
	/home/gus/.asdf/installs/golang/1.19.3/packages/pkg/mod/github.com/ipfs/go-libipfs@v0.6.2/bitswap/message/pb/message.pb.go:371 +0x35a
FAIL
```

After the change it succeeds:

```
go test -v -count=1 -run='TestThing' .
=== RUN   TestThing
&{%!s(bool=true) map[] map[] map[] %!s(int32=0)}
&{%!s(bool=true) map[] map[] map[] %!s(int32=0)}
--- PASS: TestThing (0.00s)
PASS
```

Jorropo (approved):
This is stupid to begin with, I don't like it, but it works and I don't think it's gonna break anyone.

PR #217, opened 2023-03-20
Boxo

### LINKED ISSUE #215: Rename go-libipfs to boxo
## Done Criteria
1. go-libipfs the repo is renamed to boxo in such a way that existing consumers aren't broken when they dependended on an explicit version of go-libipfs 
2. key consumers of go-libipfs use boxo instead (e.g., Kubo)
3. public docs in this repo that use the term "go-libipfs" have been updated to say "boxo"

## Why Important
A project's name can set expectations on what a project provides to consumers.  We don't want someone to think that they need to use go-libipfs if writing an IPFS implementation in Go.  It may be helpful, but it isn't required.  Worked started in lead up to IPFS Thing 2022 and is ongoing (https://github.com/ipfs/specs/issues/381 ) to drive the message home that IPFS compatibility is very broad.  For example, they must support addressability using CIDs and expose operations (e.g., provision, retrieval) on resources using CIDs.  IPFS implementations don't need to support all the functionality of Kubo or all the functionality that is possible with go-libipfs.  

## Why are we doing this now?
"boxo" has had this history about its name:
* 2021Q4: project pitched and experimented with in "ipfs/go-libipfs": https://github.com/ipfs/kubo/issues/8543
* 202211: started making active effort on it in an incremental fashion. In that process, we renamed "go-libipfs" to "libkubo", but reverted back to go-libipfs because we found it "to be even more confusing since its intention is to have code that is reusable and not Kubo-specific" (see https://github.com/ipfs/kubo/issues/8543#issuecomment-1331161667)
* 202303: @ipfs/kubo-maintainers got feedback from @jbenet on how the "go-libipfs" name can cause confusion about what constitutes an IPFS implementation or set expectations about this repo that we don't want (i.e., that one needs to use this repo for one's IPFS implementation written in go).  

Renames aren't free, but it's better to do it now before even more start depending on it as part of https://github.com/ipfs/boxo/issues/196.

## Why was this name chosen? 
1. ["boxo" is physical toolbox vendor](https://boxousa.com/).  "boxo" the project is intended to be a **toolbox** of modules, tools, and examples that can aid one in their go-based IPFS implementation.
2. Similarity between the "box" and IPFS's "cube" shape.
3. Audible resonance with "kubo", which is the archeologically-rich soil from which "boxo" has sprouted from and "kubo" is at least currently one of the significant consumers of "boxo".
4. "boxo" is short.

## Why wasn't the community given input into the name?
The request for a name change came late while we were part way through https://github.com/ipfs/boxo/issues/196.  That work was already drawing on and impacting other work.  Team resolve to be in limbo was low 1) given current situation and 2) having done a lot of renaming last year with go-ipfs and Kubo, and the name was compelling enough for the reasons listed.

## Tasks
- [x] Validate that github proxying works as expected
- [x] Rename this repo
- [x] Update dependencies
   - [x] kad-dht: https://github.com/libp2p/go-libp2p-kad-dht/pull/832
   - [x] Lotus: https://github.com/filecoin-project/lotus/pull/10562
   - [x] Kubo: https://github.com/ipfs/kubo/pull/9746
   - [x] go-ds-crdt https://github.com/ipfs/go-ds-crdt/pull/200
   - [x] ipfs-lite https://github.com/hsanjuan/ipfs-lite/pull/292
   - [x] ipfs-cluster https://github.com/ipfs-cluster/ipfs-cluster/pull/1901
- [x] Find/replace go-libipfs in this repo

Jorropo:
Depends on:
- https://github.com/ipfs/interface-go-ipfs-core/pull/111
- https://github.com/ipfs/go-block-format/pull/40
- https://github.com/ipfs/go-ipfs-files/pull/60
- https://github.com/ipld/go-car/pull/402
- https://github.com/ipfs/go-unixfs/pull/145

Fixes #215


version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "v0.7.0"
+  "version": "v0.7.1"
```
guseggert: should probably be v0.8.0

README.md
```diff
@@ -31,75 +31,75 @@ go-libipfs 🍌
...
 
 ### Is everything related to IPFS in the Go ecosystem in this repo?
 
-No.  Not everything related to IPFS is intended to be in go-libipfs. View it as a starter toolbox (potentially among multiple).  If you’d like to build an IPFS implementation with Go, here are some tools you might want that are maintained by a group that has long term commitments to the IPFS project.  There are certainly repos that others maintainer that aren't included here (e.g., ipfs/go-car) which are still useful to IPFS implementations. It's expected and fine for new IPFS functionality to be developed that won't be part of go-libipfs.  
+No.  Not everything related to IPFS is intended to be in Boxo. View it as a starter toolbox (potentially among multiple).  If you’d like to build an IPFS implementation with Go, here are some tools you might want that are maintained by a group that has long term commitments to the IPFS project.  There are certainly repos that others maintainer that aren't included here (e.g., ipfs/go-car) which are still useful to IPFS implementations. It's expected and fine for new IPFS functionality to be developed that won't be part of Boxo.
```
guseggert, line 72: we don't need this section anymore (nor the "Does Boxo == IPFS?" section)

BigLep: I don't think it hurts though to have this though to be clear.

Jorropo: I would like to keep this too.

Jorropo:
Closed in favor of https://github.com/ipfs/boxo/pull/220

PR #223, opened 2023-03-24
fix: ensure (dag)-cbor/json codecs are registered

hacdias:
Moving https://github.com/ipfs/bifrost-gateway/pull/64 here, since that's where we convert stuff between codecs.


lidel (approved):
Thank you (tests for these response types will live in https://github.com/ipfs/gateway-conformance, will be added to this repo after we have some breathing room and do it in https://github.com/ipfs/bifrost-gateway/issues/63 first )

PR #229, opened 2023-03-28
refactor(examples): proxy gateway to use exchange

aschmahmann:
Using a blockstore isn't really the right interface here (e.g. `Has` doesn't really make any sense). The `exchange.Interface` or better the `exchange.Fetcher` interfaces are more appropriate and less confusing.

This also removes all the `notImplemented` errors. The functions not implemented are essentially optional, so descriptions as to what they are for have been left there but they otherwise should not cause issues.

Note: there seems to be some good cause here to simplify the examples (and make them more production-like) by adding:
- An implementation of the exchange interface that takes an `exchange.Fetcher`.
- An implementation of an LRU blockstore

Both of these are effectively implemented in bifrost-gateway (https://github.com/ipfs/bifrost-gateway/pull/61). So once we're happy with how things pan out there we should consider porting implementations back to boxo.

Explicitly looking for feedback such as:
- Things to fix
- Comments to add/remove
- If we should hold off on merging this until some of the bifrost-gateway pieces have moved back here so the example is more production-like

Other feedback welcome too 😄

### DIFF
--- a/examples/gateway/proxy/blockstore.go
+++ b/examples/gateway/proxy/blockstore.go
@@ -2,47 +2,38 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 
-	blockstore "github.com/ipfs/boxo/blockstore"
-	"github.com/ipfs/go-block-format"
+	"github.com/ipfs/boxo/exchange"
+	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 )
 
-var (
-	errNotImplemented = errors.New("not implemented")
-)
-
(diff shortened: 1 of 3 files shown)


hacdias (approved):
LGTM. Tbf, when I wrote this for the first time I wasn't really aware of the other abstractions (like the exchange). It does indeed make more sense to use it.

hacdias:
I rebased on top of #246 since I made quite a few changes to the examples there and we enabled more conformance tests, just to run more tests here.

lidel (approved):
Rebased since bunch of refactor had overlap with https://github.com/ipfs/boxo/pull/246, now this PR is much smaller. lgtm

PR #230, opened 2023-03-28
feat(libp2p): add defaults for libp2p host construction

### LINKED ISSUE #208: Provide Stock Libp2p Config For IPFS
Yesterday I learned that if I initialize a Libp2p node with default options, I won't be able to connect to some large IPFS providers (like web3.storage). 

Most prominently, this is cause JS based IPFS implementations are usually still using Mplex as opposed to Yamux. That said, digging through Kubo source code, there are SO many customizations to Libp2p parameters it's hard to know how much might be different, and how much this will affect connectivity.

To make things easier for folks to people to spin up an IPFS stack using this library, it would be nice, either as a module or in example code, to provide an ideal stock configuration for maximizing connectivity on the network.

Currently, figuring out the right Libp2p configs requires scanning a ton of Kubo startup code, which, I can do, having poured through it several times over the years, but is a HUGE lift for anyone new to the ecosytem.

For reference, where we ended up in Lassie after the exercise of copying Kubo configs is here: https://github.com/filecoin-project/lassie/pull/157

aschmahmann:
fixes #208 

Overall idea: Sometimes go-libp2p's defaults are going to make sense for the wider base of users than just the IPFS ecosystem. It'd be nice, as part of providing a toolbox for helping people build IPFS implementations, if we helped them take care of some of the defaults that we've found useful.

The main examples of where this seems particularly useful is if go-libp2p wants to discourage the use of a feature for future implementations that isn't yet practical for IPFS applications.
- Keeping mplex support is the current example
- I don't recall the history here, but it's something that could've happen around the SECIO deprecation as well

In theory this is also true about new features go-libp2p doesn't yet want to enable by default. However, this has not shown to be a strong case so far.

Looking for feedback here for:
1. Are there more/different defaults we should use here?
2. Is this the right API to expose?
3. Is this the right package to put this in?
   - Consider that we probably also want to put in some defaults for DHT construction in anticipation of finally doing https://github.com/libp2p/go-libp2p-kad-dht/issues/597. Where would we want those helpers to land?

As always, other feedback welcome too 😄

### DIFF
--- a/libp2p/default_host.go
+++ b/libp2p/default_host.go
@@ -1,0 +1,61 @@
+package libp2p
+
+import (
+	"os"
+
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/p2p/muxer/mplex"
+	"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
+)
+
+// Complete list of default options and when to fallback on them.
+//
+// Please *DON'T* specify default options any other way. Putting this all here
+// makes tracking defaults *much* easier.
+var defaults = []struct {
+	fallback func(cfg *libp2p.Config) bool
+	opt      libp2p.Option
+}{
+	{
+		fallback: func(cfg *libp2p.Config) bool { return cfg.Muxers == nil },
+		opt:      libp2p.ChainOptions(libp2p.Muxer(yamux.ID, yamuxTransport()), libp2p.Muxer(mplex.ID, mplex.DefaultTransport)),
+	},
+}
+
+func yamuxTransport() network.Multiplexer {
+	tpt := *yamux.DefaultTransport
+	tpt.AcceptBacklog = 512
+	if os.Getenv("YAMUX_DEBUG") != "" {
+		tpt.LogOutput = os.Stderr
+	}
+	return &tpt
+}
+
+// fallbackDefaults applies default options to the libp2p node if and only if no
+// other relevant options have been applied. will be appended to the options
+// passed into NewHost.
+var fallbackDefaults libp2p.Option = func(cfg *libp2p.Config) error {
+	for _, def := range defaults {
+		if !def.fallback(cfg) {
+			continue
+		}
+		if err := cfg.Apply(def.opt); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// NewHost creates a new libp2p host with default options found helpful by some existing IPFS implementations.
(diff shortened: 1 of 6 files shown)


libp2p/default_host.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func yamuxTransport() network.Multiplexer {
+	tpt := *yamux.DefaultTransport
+	tpt.AcceptBacklog = 512
+	if os.Getenv("YAMUX_DEBUG") != "" {
+		tpt.LogOutput = os.Stderr
+	}
+	return &tpt
+}
```
aschmahmann, line 34: @marten-seemann any thoughts on the `AcceptBacklog` change here. This seems to have been in kubo https://github.com/ipfs/kubo/blob/405b1d2dcd976b9917f3b82036ce5266352bdca9/core/node/libp2p/smux.go#L18 for years and is in lotus as well https://github.com/filecoin-project/lotus/blob/687f0c070fbe59a18f391a7d8cbc9a8a8838df18/node/modules/lp2p/smux.go#L14.

Should this become a default in go-libp2p, should we consider changing this here to something else, etc.?

libp2p/default_host.go
```diff
@@ -0,0 +1,61 @@
...
+// NewHost creates a new libp2p host with default options found helpful by some existing IPFS implementations.
+// These defaults are not guaranteed to be stable over time and any new options provided will replace ones
+// of the existing type.
+//
+// For example, adding a new transport will remove all the existing transports unless they are also added back in the
+// passed options.
+//
+// See the libp2p options for more information
+func NewHost(opts ...libp2p.Option) (host.Host, error) {
+	return libp2p.New(append(opts, fallbackDefaults)...)
```
Jorropo, line 60: I would use `NewWithoutDefaults` and exhaustively list all things this prevents againt a consumer updating libp2p which could create unexpected changes.

Also I think we should include a helper function like `NewHostAndRouting` which also setups a DHT (or ipni, ...) for you, because this is the most complex here in setting up `go-libp2p`. But this could be made as an other PR.

libp2p/default_host.go
hacdias: Ping @marten-seemann

libp2p/default_host.go
hacdias: I do agree with @Jorropo that this would be beneficial. Any thoughts on this @aschmahmann?

libp2p/default_host.go
marten-seemann: The default value is 256. I don't think it should be necessary to increase it at all, 256 streams is massive already.

lidel:
Closing for now  – see  triage notes from https://github.com/ipfs/boxo/issues/208#issuecomment-1658437071

PR #232, opened 2023-03-28
blockservice: remove busyloop in getBlocks by removing batching

Jorropo:
It makes the code quite more complex and there is no evidance it helps for anything.

Capture d’écran du 2023-03-29 01-26-46
This reduce the CPU usage of `ipfs get` (and other code that use the blockservice `getBlocks`) by ~7x (from 200% CPU usage to 30%).

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -296,6 +296,7 @@ func getBlocks(ctx context.Context, ks []cid.Cid, bs blockstore.Blockstore, fget
 		}
 
 		if !allValid {
+			// can't shift in place because we don't want to clobber callers.
 			ks2 := make([]cid.Cid, 0, len(ks))
 			for _, c := range ks {
 				// hash security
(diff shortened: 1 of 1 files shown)


blockservice/blockservice.go
```diff
@@ -333,52 +334,39 @@ func getBlocks(ctx context.Context, ks []cid.Cid, bs blockstore.Blockstore, fget
...
-				case <-ctx.Done():
+			var b blocks.Block
+			select {
+			case v, ok := <-rblocks:
+				if !ok {
 					return
-				default:
-					break batchLoop
 				}
+				b = v
```
MichaelMure, line 345:
```suggestion
				cache[0] = v
```
maybe?

MichaelMure (approved):

guseggert (approved):

blockservice/blockservice.go
Jorropo: But then we would need to write `bs.Put(ctx, cache[0])` which isn't pretty either.

I think the real fix here is to add a new non variadic `NotifyNewBlock`.

PR #234, opened 2023-03-29
chore: release v0.8.0-rc3

aschmahmann:


version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "v0.8.0-rc2"
```
aschmahmann, line 2: I haven't updated the example project's dependency in this PR. Not sure if that's something to do now or post a release.

guseggert: Are there reasons not to?  I can think of reasons to update, such as running tests, but not for not updating.

Jorropo (commented):
We were reserving `v0.8.0` for the no breakage release, the goal is that for whatever else you are using before hand the only migrations needed is updating import paths and updating the `go.mod` file, I see you updated the gateway interface to the new one, so I think this deserve a `v0.9.0-rc1`.

Gus wants to write more documentation that why `v0.8.0` is not released yet, we can work on a `release/v0.8.0` branch for this it's not an issue.

version.json
aschmahmann: Wasn't sure what to update to. WDYT commit or the tag number? Either way probably let's merge this now and we'll document this for the next one (trying to get kubo and other PRs merged and don't want to wait for more CI runs at the moment)

guseggert: sgtm

guseggert:
> We were reserving `v0.8.0` for the no breakage release, the goal is that for whatever else you are using before hand the only migrations needed is updating import paths and updating the `go.mod` file, I see you updated the gateway interface to the new one, so I think this deserve a `v0.9.0-rc1`.
> 
> Gus wants to write more documentation that why `v0.8.0` is not released yet, we can work on a `release/v0.8.0` branch for this it's not an issue.

I think that since this is code not being migrated from existing libraries that it should be okay for v0.8.0.  I just want to make sure there's a release where people can upgrade to boxo without having to change anything but import paths, so as long as this is not being migrated from another repo that people consume (which I believe is the case), it should be okay to include in v0.8.0. (I think?)

guseggert (approved):

PR #236, opened 2023-03-30
feat(gateway): redirect ipns b58mh to cid

### LINKED ISSUE #9749: Gateway support for signed IPNS records doesn't work with raw base58 PeerIDs
### Checklist

- [X] This is a bug report, not a question. Ask questions on [discuss.ipfs.io](https://discuss.ipfs.io).
- [X] I have searched on the [issue tracker](https://github.com/ipfs/kubo/issues?q=is%3Aissue) for my bug.
- [X] I am running the latest [kubo version](https://dist.ipfs.tech/#kubo) or have an issue updating.

### Installation method

ipfs-desktop

### Version

```Text
Kubo version: 0.19.0-196321958
Repo version: 13
System version: arm64/darwin
Golang version: go1.20.2
```

### Description

[Kubo 0.19](https://github.com/ipfs/kubo/releases/tag/v0.19.0) add Gateway support for returning signed (verifiable) IPNS records. 

However, from initial testing, this only works for CID encodings of IPNS names (base32 and base36) but not with the raw base58 multihash.

- ✅ base32 CID: https://ipfs.io/ipns/bafzaajaiaejcbzdibmxyzdjbbehgvizh6g5tikvy47mshdy6gwbruvgwvd24seje?format=ipns-record 
- ✅ base36 CID: https://ipfs.io/ipns/k51qzi5uqu5dlvj2baxnqndepeb86cbk3ng7n3i46uzyxzyqj2xjonzllnv0v8?format=ipns-record
- ❌ raw base58 multihash (b58mh) https://ipfs.io/ipns/12D3KooWRBy97UB99e3J6hiPesre1MZeuNQvfan4gBziswrRJsNK This one fails with the error: "selected encoding not supported"

This is strange since all those URLs work without the `?format=ipns-record`.

## Suggestions

- Consider adding support for raw base58 multihash IPNS names. Given that this is always scoped to `/ipns/[ipns_name], there shouldn't be too much ambiguity
- Improve the error to be more specific, e.g. `selected encoding of the IPNS name not supported. Try passing an IPNS name encoded as a base32 or base36 CID.`
- Update the [IPFS HTTP Gateway spec](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#get-ipnsnamepathparams) to include the supported encodings for IPNS names.

hacdias:
Closes https://github.com/ipfs/kubo/issues/9749. Redirects base58 encoded peer IDs to CID under the `/ipns` namespace.

Sharness will fail here. Kubo PR: https://github.com/ipfs/kubo/pull/9785

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -19,6 +19,8 @@ import (
 	ipath "github.com/ipfs/boxo/coreiface/path"
 	cid "github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/multiformats/go-multibase"
 	prometheus "github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -201,6 +203,10 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 
 	if err := handleServiceWorkerRegistration(r); err != nil {
 		webRequestError(w, err)
+		return
+	}
+
+	if handleIpnsB58mhToCidRedirection(w, r) {
 		return
 	}
 
@@ -726,6 +732,57 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
 	}
 
 	return nil
+}
+
+// handleIpnsB58mhToCidRedirection redirects from /ipns/b58mh to /ipns/cid in
+// the most cost-effective way.
+func handleIpnsB58mhToCidRedirection(w http.ResponseWriter, r *http.Request) bool {
+	if _, dnslink := r.Context().Value(DNSLinkHostnameKey).(string); dnslink {
+		// For DNSLink hostnames, do not perform redirection in order to not break
+		// website. For example, if `example.net` is backed by `/ipns/base58`, we
+		// must NOT redirect to `example.net/ipns/base36-id`.
+		return false
+	}
+
+	if w.Header().Get("Location") != "" {
+		// Ignore this if there is already a redirection in place. This happens
+		// if there is a subdomain redirection. In that case, the path is already
+		// converted to CIDv1.
+		return false
+	}
+
+	pathParts := strings.Split(r.URL.Path, "/")
+	if len(pathParts) < 3 {
+		return false
+	}
+
+	if pathParts[1] != "ipns" {
+		return false
+	}
+
+	id, err := peer.Decode(pathParts[2])
+	if err != nil {
+		return false
+	}
+
+	// Convert the peer ID to a CIDv1.
+	cid := peer.ToCid(id)
+
+	// Encode CID in base36 to match the subdomain URLs.
+	encodedCID, err := cid.StringOfBase(multibase.Base36)
+	if err != nil {
+		return false
+	}
+
+	// If the CID was already encoded, do not redirect.
+	if encodedCID == pathParts[2] {
+		return false
+	}
+
+	pathParts[2] = encodedCID
+	r.URL.Path = strings.Join(pathParts, "/")
+	http.Redirect(w, r, r.URL.String(), http.StatusFound)
+	return true
 }
 
 // Attempt to fix redundant /ipfs/ namespace as long as resulting
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -546,3 +546,30 @@ func TestGoGetSupport(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, http.StatusOK, res.StatusCode)
 }
+
+func TestIpnsBase58MultihashRedirect(t *testing.T) {
+	ts, _, _ := newTestServerAndNode(t, nil)
+	t.Logf("test server url: %s", ts.URL)
+
+	t.Run("ED25519 Base58-encoded key", func(t *testing.T) {
+		t.Parallel()
+
+		req, err := http.NewRequest(http.MethodGet, ts.URL+"/ipns/12D3KooWRBy97UB99e3J6hiPesre1MZeuNQvfan4gBziswrRJsNK?keep=query", nil)
+		assert.Nil(t, err)
+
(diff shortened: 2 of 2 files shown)


.github/workflows/gateway-sharness.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           repository: ipfs/kubo
           path: kubo
-          ref: 503edee648e29c62888f05fa146ab13d9c65077d
+          ref: master
```
hacdias: This was supposed to have been reverted after #220!

hacdias:
@lidel afaik, the sharness tests that are failing are subdomain related. So before they would go `/ipns/b58mh` --> `{cidv1}.ipns...`, now there's two steps. `/ipns/b58mh` --> `/ipns/cid` --> `{cidv1}.ipns...`. Can we just remove those tests then? The redirection  `/ipns/b58mh` --> `/ipns/cid` is tested on this PR and `/ipns/cid` --> `{cidv1}.ipns...` is already tested by other tests.

.github/workflows/gateway-sharness.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         with:
           repository: ipfs/kubo
           path: kubo
+          ref: 6ae10703f1d25c8c3f1c5526a803f4b95af13525
```
lidel: :thought_balloon: i think its better to have failing sharness on a PR here as long Kubo is green  than to forget we did this and merge 

```suggestion
```

gateway/handler.go
```diff
@@ -727,6 +733,45 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
...
+	// expensive. Base58 encoded peer IDs will start with a '1'.
+	if !strings.HasPrefix(s, "1") {
+		return false
+	}
+
+	// Decode the base58 encoded sha256 or identity multihash.
+	m, err := multihash.FromB58String(s)
+	if err != nil {
+		return false
+	}
```
lidel: Avoid reimplementing `peer.Decode`.

iirc only ED25519 start with `1`,  legacy RSA keys start with `Qm` 
just use `peer.Decode` here, it already does the right thing first, won't be computationally expensive:
https://github.com/libp2p/go-libp2p/blob/88132e976f8e711facab6e5c548236378e351115/core/peer/peer.go#L132

gateway/handler.go
```diff
@@ -727,6 +733,45 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
...
+
+	pathParts[2] = peer.ToCid(peer.ID(m)).String()
```
lidel: change this from default base32 to base36 CID, so if someone inspects redirect, it will have the same ID as the one on subdomain (better DX if someone unfamiliar with IPFS stack tries to debug things)

lidel (changes requested):

gateway/handler.go
hacdias: @lidel `peer.Decode` will also work with b58mh, leading to infinite redirects. That's what I was trying to avoid here. How do you recommend going forward? Decoding it __always__ and comparing it with the string we already have?

gateway/handler.go
hacdias: If I use `peer.Decode` as you asked, we will also trigger this for CIDv0-encoded peer IDs, and that will fail.

hacdias:
@lidel I updated to use `peer.Decode` as you asked, but please note that we will be executing this for more cases than before, that is, we will also be executing `peer.Decode` for Peer IDs that are already CIDv1s. Then, we compare with what we already have and only redirect if they don't match. I also moved the testing to the Kubo PR.

lidel (approved):
Thanks, I think normalization to base36 reduces confusion (IPNS Names always looking the same).

Good for now, I think we will improve it further when we have time for productizing `boxo/ipns` as part of https://github.com/ipfs/specs/issues/376.

(not merging yet, as I want to  check other PRs and merge in least painful order)

gateway/handler.go
```diff
@@ -728,6 +734,52 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
 	return nil
 }
 
+// handleIpnsB58mhToCidRedirection redirects from /ipns/b58mh to /ipns/cid in
+// the most cost-effective way.
+func handleIpnsB58mhToCidRedirection(w http.ResponseWriter, r *http.Request) bool {
+	fmt.Println(w.Header().Get("Location"))
```
```suggestion
```

gateway/handler.go
```diff
@@ -728,6 +734,52 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
...
+	if w.Header().Get("Location") != "" {
+		// Ignore this if there is already a redirection in place. This happens
+		// if there is a subdomain redirection. In that case, the path is already
+		// converted to CIDv1.
+		return false
+	}
+
+	pathParts := strings.Split(r.URL.Path, "/")
+	fmt.Println(pathParts)
```
```suggestion
```

lidel (approved):

gateway/handler.go
```diff
@@ -728,6 +734,50 @@ func handleServiceWorkerRegistration(r *http.Request) (err *ErrorResponse) {
...
+	}
+
+	// If the CID was already encoded, do not redirect.
+	if encodedCID == pathParts[2] {
+		return false
+	}
+
+	pathParts[2] = encodedCID
+	r.URL.Path = strings.Join(pathParts, "/")
+	http.Redirect(w, r, r.URL.String(), http.StatusFound)
```
lidel, line 784: @hacdias  I just realized this might  break  DNSLink websites that have DNS TXT record with legacy value `dnslink=/ipns/Qm..`   or `dnslink=/ipns/1..`.

In such situations, `example.net` being backed by `/ipns/legacy-id` would redirect to `example.net/ipns/base36-id` which will return 404.

Mind adding a check that redirects ONLY when  `DNSLinkHostnameKey` is missing?

lidel (changes requested):

gateway/handler.go
hacdias: Done!

lidel (approved):

PR #237, opened 2023-03-30
ci: add gateway-conformance tests for car-gateway

galargh:
This PR adds https://github.com/ipfs/gateway-conformance tests for CAR gateway example.

Note that `subdomain-gateway` spec is currently disabled in the `gateway-conformance` suite.

### DIFF
--- a/.github/workflows/gateway-conformance.yml
+++ b/.github/workflows/gateway-conformance.yml
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
 
+concurrency:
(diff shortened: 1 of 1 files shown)


hacdias (approved):
LGTM. I will be merging this. I'm investigating the subdomain in https://github.com/ipfs/boxo/pull/246. It's probably similar to the issues we had in Bifrost (https://github.com/ipfs/bifrost-gateway/pull/66), but I think there's something else too. Again, thanks @galargh @laurentsenta for your work on this!

PR #238, opened 2023-03-30
ci: reduce the number of test-examples workflow runs

galargh:
###### Triggers

This PR will reduce the number of workflow runs for `test-examples` workflow. After this change, we will run `test-examples` on every commit in a PR and on every push to the `main` branch. We will stop running `test-examples` on pushes to branches other than `main`.

A similar change is coming soon to the workflows distributed through Unified CI - https://github.com/protocol/.github/pull/489. Once it lands there, I'm going to update `boxo` immediately.

###### Concurrency

This PR also makes sure that we're running `test-example` only for the HEAD of the PR.

Corresponding uCI change - https://github.com/protocol/.github/pull/505.

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -1,5 +1,10 @@
-on: [push, pull_request]
 name: Go Test Examples
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   unit:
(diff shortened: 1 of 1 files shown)


galargh:
Actually, let me close this. We don't need the `test-examples` workflow at all because `test` already runs the tests for `examples`.

PR #240, opened 2023-03-30
feat: switch gateway GetRange API to match Get

aschmahmann:
Also fixes serving range requests for /ipfs/bafydir when we know we'll get the index.html or _redirect.

cc https://github.com/ipfs/bifrost-gateway/pull/61

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -91,10 +91,11 @@ type IPFSBackend interface {
 	// Size, and Cid.
 	Get(context.Context, ImmutablePath) (ContentPathMetadata, *GetResponse, error)
 
-	// GetRange returns a full UnixFS file object. Ranges passed in are advisory for pre-fetching data, however
-	// consumers of this function may require extra data beyond the passed ranges (e.g. the initial bit of the file
-	// might be used for content type sniffing even if only the end of the file is requested).
-	GetRange(context.Context, ImmutablePath, ...GetRange) (ContentPathMetadata, files.File, error)
+	// GetRange returns a full UnixFS file object, or a UnixFS directory. Ranges passed in are advisory for pre-fetching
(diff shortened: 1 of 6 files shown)


gateway/gateway.go
```diff
@@ -91,10 +91,11 @@ type IPFSBackend interface {
...
 
-	// GetRange returns a full UnixFS file object. Ranges passed in are advisory for pre-fetching data, however
-	// consumers of this function may require extra data beyond the passed ranges (e.g. the initial bit of the file
-	// might be used for content type sniffing even if only the end of the file is requested).
-	GetRange(context.Context, ImmutablePath, ...GetRange) (ContentPathMetadata, files.File, error)
+	// GetRange returns a full UnixFS file object, or a UnixFS directory. Ranges passed in are advisory for pre-fetching
+	// data, however consumers of this function may require extra data beyond the passed ranges (e.g. the initial bit of
+	// the file might be used for content type sniffing even if only the end of the file is requested).
+	// Note: there is currently no semantic meaning attached to a range request for a directory
+	GetRange(context.Context, ImmutablePath, ...GetRange) (ContentPathMetadata, *GetResponse, error)
```
aschmahmann, line 98: Looking at the code now, should we just have a single function `Get(context.Context, ImmutablePath, ...GetRange) (ContentPathMetadata, *GetResponse, error)` and drop `GetRange` entirely (since passing no ranges is equivalent to asking for the entire object)?

hacdias: I think that's a good idea.

lidel:
I've cherry-picked this into https://github.com/ipfs/boxo/pull/245, will continue there, Incl. `Get` and `GetRange` consolidation.

gateway/gateway.go
lidel: I will do it in https://github.com/ipfs/boxo/pull/245

PR #241, opened 2023-03-30
feat: expose closer for GetResponse

aschmahmann:
This is more meant as a discussion then necessarily for merging as is.

## High level problem

Given that we return interfaces that dynamically pull data for some of the `gateway.IPFSBackend` requests (e.g. `Get`) the implementation servicing these requests might want to know when a given interface is done being used so it can deal with any wrap up related to the request. This is doable for most requests by overriding the `Close () error` function, however with `Get` this can't be done because there is no closer associated with the directoryMetadata.

### Bonus problem
Additionally, even if `directoryMetadata` had a close function it wouldn't be possible for middleware to add in an additional close function to the loop (e.g. an implementation that wraps another implementations `Get` function and wants to add a close function).

### Specific issue

Listing so as to allow addressing of https://en.wikipedia.org/wiki/XY_problem concerns, and encourage alternative solutions if anyone has a better idea.

In implementing a CAR (and blocks) retrieval based implementation of `gateway.IPFSBackend` I ended up (at least for now) doing this by:
1. Asynchronously starting a fetch for the relevant blocks + CARs
2. Using a BlocksGateway internally which is back by a blockservice
3. Feeding the blocks received over the network into the blockservice

In order for the passed in blockservice to be able to get blocks from lots of different inbound graph requests it needs to communicate with those other requests. However, we don't want that communication to continue once the HTTP request is done. While I was able to overload `Close() error` for most functions I couldn't do anything for `Get` so I used `runtime.SetFinalizer` for now which isn't great https://github.com/ipfs/bifrost-gateway/pull/61#discussion_r1153489302.

## Solution

The draft solution here is one of a family of related ones, but basically the idea is being able to set a close function that will get called when the response is done being used.

## Alternatives

Assume the channel will either be closed due to finishing, or will have a context cancellation. Wait on a goroutine for directory listing closures. This would work, but not using a goroutine would be nice.

Also, either way we need to solve the bonus problem of allowing users access to the internals (e.g. exposing the fields, adding getters, adding modifer functions, etc.)

@lidel @hacdias thoughts, suggestions?


lidel (commented):
Late on my end, but quick thought: stumbled upon this problem in https://github.com/ipfs/boxo/pull/241 and  `defer getResp.Close()` proposed here feels like the lesser evil, or maybe even a way of simplifying code (instead of closing [file.File here](https://github.com/ipfs/boxo/pull/245/files#diff-5b9dda4053dba85f902d2464389f79cdfca0976873fd206acaefcb09784d6e98R88) we would be doing close on parent struct etc), a bit easier to reason about than goroutine conventions/assumptions.

PR #242, opened 2023-03-30
blockservice & exchange & bitswap: add non variadic NotifyNewBlock

Jorropo:
Variadicts in go are just syntactic sugar around passing a slice, that  means all go memory reachability rules apply, this force the compiler to  heap allocate the variadic slice for virtual call, because the  implementation is allowed to leak the slice (and go's interprocedural optimisations do not cover virtuals).

Passing a block without variadic will pass the itab either on the stack  or decomposed through registers. Skipping having to allocate a slice.


MichaelMure (approved):

exchange/interface.go
```diff
@@ -13,6 +13,8 @@ import (
 type Interface interface { // type Exchanger interface
 	Fetcher
 
+	// NotifyNewBlock tells the exchange that a new block is available and can be served.
+	NotifyNewBlock(ctx context.Context, blocks blocks.Block) error
```
aschmahmann, line 17: Any idea how widespread the dependency on this interface (and therefore the result of the breakage) is and the benefits of implementing this? For example, does this show up at all in profiles as a meaningful thing?

Jorropo: I think everyone relies either on the offline exchange (in order to wrap a blockstore) or bitswap. I don't belive many implementations exists, even in outside codebases.

Even if someone else implement this, just forwarding the call to `NotifyNewBlocks` with a slice of 1, this is legal and easy to write.
This just aims to get over golang's lacking interprocedural virtual optimisations, it does not change any feature.

Most of our other APIs have both single and multiple versions for this reason (`Put` vs `PutMany`, `GetBlock` vs `GetBlocks`, ...).

BigLep:
@Jorropo : I don't see an issue linked to this PR.  In the absence of this, can you share more on what's motivating this PR?

Jorropo:
@BigLep the PR and Commit bodies:

> Variadicts in go are just syntactic sugar around passing a slice, that means all go memory reachability rules apply, this force the compiler to heap allocate the variadic slice for virtual call, because the implementation is allowed to leak the slice (and go's interprocedural optimisations do not cover virtuals).
> Passing a block without variadic will pass the itab either on the stack or decomposed through registers. Skipping having to allocate a slice.

I can add the explanation about keeping it in-line with our other exchange APIs.

---

This seems like a good usecase to try the idea @guseggert is cooking (some automated code migrator), that would add the `NotifyNewBlock` method (which redirects to `NotifyNewBlocks`) automatically if you run it.

hacdias (approved):
I don't have strong opinions on this. It seems a relatively small change that can bring memory benefits - although arguably low. If others agree, I'm fine to have it merged.

FYSA: I rebased the PR and added the updated exchange for the gateway examples.

BigLep:
2023-04-06 conversation: will defer for now.  Want to have clarity on how we handle breaking changes in Boxo.

gammazero:
This is ready to merge. Let's decide if we want this or not.

lidel (commented):
Lgtm? As noted in https://github.com/ipfs/boxo/pull/242/files#r1153577309 this follows existing convention from other places (`Put` vs `PutMany`, `GetBlock` vs `GetBlocks` etc), and simplifies code slightly. If someone has custom implementation, should be trivial to migrate.

@aschmahmann any concerns?

gammazero:
CI fails when calling `s.exchange.NotifyNewBlock(ctx, o)` in `blockservice.go` [here](https://github.com/ipfs/boxo/pull/242/files#diff-c3167a605aeb6a2f9430f5730ee6508f88e15022a746998ff6ec8ba013d0d6dcR179), but succeeds when calling `s.exchange.NotifyNewBlocks(ctx, o)`. Ther should be no difference.

bitswap/bitswap.go
```diff
@@ -95,6 +95,13 @@ func New(ctx context.Context, net network.BitSwapNetwork, providerFinder routing
 	return bs
 }
 
+func (bs *Bitswap) NotifyNewBlock(ctx context.Context, blk blocks.Block) error {
+	return multierr.Combine(
+		bs.Client.NotifyNewBlock(ctx, blk),
+		bs.Server.NotifyNewBlock(ctx, blk),
+	)
```
gammazero:
```suggestion
	if bs.Server != nil {
		return multierr.Combine(
			bs.Client.NotifyNewBlock(ctx, blk),
			bs.Server.NotifyNewBlock(ctx, blk),
		}
	}
	return bs.Client.NotifyNewBlock(ctx, blk)
```

bitswap/server/server.go
```diff
@@ -377,6 +377,17 @@ func (bs *Server) Stat() (Stat, error) {
 	return s, nil
 }
 
+// NotifyNewBlock announces the existence of block to this bitswap service. The
```
gammazero:
```suggestion
// NotifyNewBlock announces the existence of a block to this bitswap service. The
```

bitswap/server/server.go
```diff
@@ -377,6 +377,17 @@ func (bs *Server) Stat() (Stat, error) {
...
+// service will potentially notify its peers.
+// Bitswap itself doesn't store new blocks. It's the caller responsibility to ensure
+// that those blocks are available in the blockstore before calling this function.
+func (bs *Server) NotifyNewBlock(ctx context.Context, blk blocks.Block) error {
+	// Call to the variadic to avoid code duplication.
+	// This is actually fine to do because no calls is virtual the compiler is able
```
gammazero:
```suggestion
	// This is fine to do because if no calls are virtual then the compiler is able
```

bitswap/server/server.go
```diff
@@ -377,6 +377,17 @@ func (bs *Server) Stat() (Stat, error) {
...
 }
 
+// NotifyNewBlock announces the existence of a block to this bitswap service. The
+// service will potentially notify its peers.
+// Bitswap itself doesn't store new blocks. It's the caller responsibility to ensure
+// that those blocks are available in the blockstore before calling this function.
+func (bs *Server) NotifyNewBlock(ctx context.Context, blk blocks.Block) error {
+	// Call to the variadic to avoid code duplication.
+	// This is fine to do because if no calls are virtual then the compiler is able
+	// to see that the slice does not leak and the slice is stack allocated.
```
gammazero, line 387: It is not clear whether this is true or not in the golang documentation.

bitswap/client/client.go
```diff
@@ -426,6 +426,16 @@ func (bs *Client) GetBlocks(ctx context.Context, keys []cid.Cid) (<-chan blocks.
 	return session.GetBlocks(ctx, keys)
 }
 
+// NotifyNewBlock announces the existence of blocks to this bitswap service.
+// Bitswap itself doesn't store new blocks. It's the caller responsibility to ensure
+// that those blocks are available in the blockstore before calling this function.
+func (bs *Client) NotifyNewBlock(ctx context.Context, blk blocks.Block) error {
+	// Call to the variadic to avoid code duplication.
+	// This is actually fine to do because no calls is virtual the compiler is able
+	// to see that the slice does not leak and the slice is stack allocated.
```
gammazero: Escape analysis results seem to contradict this:

>
```
./client.go:432:7: leaking param content: bs
./client.go:432:34: leaking param: ctx
./client.go:432:55: leaking param: blk
```

gammazero:
I do not think it is worth the extra complexity to have separate functions, one for a single block and one for multiple blocks.

Issue #218, opened 2023-03-21
Handle relationship between boxo and go-car and go-merkledag


BigLep:
# Current
## Done criteria
boxo/car is removed.  boxo will use ipld/go-car.

# vvv OLD vvv
## Done Criteria
Decision is made on the relation to be had between ipfs/boxo and ipld/go-car and ipfs/go-merkledag.  This could mean having a dependency on ipfs/boxo (e.g., https://github.com/ipfs/go-merkledag/pull/100 ) or boxo taking a fork of the relevant functionality it needs and maintaining it.

## Why Important
The https://github.com/ipfs/boxo/issues/196 effort (and particularly https://github.com/ipfs/boxo/pull/206)  can't be completed without decisions here, and that work is blocking other refactoring work like https://github.com/ipfs/boxo/pull/176.

## Resulting Actions
Below are the actions that are being taken as a result of the 2023-03-21 meeting:

✅ **Item 1: Get go-car decoupled:**
- [x] @Jorropo (with @rvagg review) revert deprecation to ipfs/go-block-format that was done in https://github.com/ipfs/go-block-format/pull/37
  - https://github.com/ipfs/go-block-format/pull/41 
- [x] @Jorropo release ipfs/go-block-format
- [x] @Jorropo (with @rvagg review) replace go-libipfs/blocks with ipfs/go-block-format in ipld/go-car.  (Includes removing dependency in ipld/go-car on ipfs/go-libipfs)
- [x] @Jorropo replace go-libipfs/blocks with ipfs/go-block-format in ipfs/go-unixfs
- [x] @Jorropo remove dependency in ipfs/go-unixfs on ipfs/go-libipfs: https://github.com/ipfs/go-unixfs/pull/146

✅ **Item 2: copy in go-merkledag to boxo**
This will get handled as part of https://github.com/ipfs/boxo/issues/202
A row will be added for go-merkledag.
No deprecated types will be added.  
A CODOWNERS will be added.
https://github.com/ipfs/go-merkledag/pull/101

Additional items that have come up since>
1. ✅ Surface this with maintainers (folks in PL EngRes #bedrock): https://filecoinproject.slack.com/archives/C03FK0JJ10E/p1680108077842429
2. Clarify that boxo/ipld/car is not intended as a replacement for ipld/go-car: https://github.com/ipfs/boxo/pull/244

BigLep:
# 2023-03-21 Sync Meeting

## Attendees
1. @BigLep 
[...]

PR #244, opened 2023-03-30
Clarify ipld/car relatioship with ipld/go-car

BigLep:
See conversation in https://github.com/ipfs/boxo/issues/218#issuecomment-1490916807

### DIFF
--- a/cmd/boxo-migrate/internal/config.go
+++ b/cmd/boxo-migrate/internal/config.go
@@ -39,7 +39,10 @@ var DefaultConfig = Config{
 		"github.com/ipfs/go-ipfs-routing":                "github.com/ipfs/boxo/routing",
 		"github.com/ipfs/go-ipfs-exchange-interface":     "github.com/ipfs/boxo/exchange",
 		"github.com/ipfs/go-merkledag":                   "github.com/ipfs/boxo/ipld/merkledag",
-		"github.com/ipld/go-car":                         "github.com/ipfs/boxo/ipld/car",
+		// By default, users can/should keep using ipld/go-car.
+		// Updating ipld/go-car was useful for other repos that are also maintained by Boxo maintainers like Kubo
+		// More info is in https://github.com/ipfs/boxo/issues/218
+		// "github.com/ipld/go-car":                         "github.com/ipfs/boxo/ipld/car",
 
 		// Pre Boxo rename
 		"github.com/ipfs/go-libipfs/gateway":               "github.com/ipfs/boxo/gateway",
--- a/ipld/car/README.md
+++ b/ipld/car/README.md
@@ -1,9 +1,18 @@
-go-car (go!)
-==================
+# ❗ About CAR libraries written in go
+As of 202303, this is a copy of [`ipld/go-car`](github.com/ipld/go-car).
+`ipld/go-car` is a full-featured car library that is actively maintained.
+Boxo has smaller car needs and will be will be refactored and thinned out soon.
+Boxo can't depend on `ipld/go-car` because it would cause `ipld/go-car` to have dependency on Boxo, 
+and `ipld/go-car` maintainers want to be able to develop indepdently.
 
-[![](https://img.shields.io/badge/made%20by-Protocol%20Labs-blue.svg?style=flat-square)](https://protocol.ai)
-[![Go Reference](https://pkg.go.dev/badge/github.com/boxo/ipld/car.svg)](https://pkg.go.dev/github.com/boxo/ipld/car)
+***A user can always use `ipld/go-car` in conjuection with `boxo/ipld/car`.***
+Please use `ipld/go-car` as needed!
 
+More details and discussion are [here](https://github.com/ipfs/boxo/issues/218).
+
+---
+
+# About this module
 > Work with car (Content addressed ARchive) files!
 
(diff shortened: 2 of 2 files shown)


willscott:
The dependency issue here seems to be on the `blocks` library and whether to use the boxo instance versus the older one in `go-block-format`.

There are only a couple calls / references in there, and the definition of what a block is will be really quite stable at this point.

go-car could remove this dependency with a bit of refactoring by defining the interface locally, so that it can be compatible with either of these implementations.

If there wasn't a dependency re-factor on `ipld/go-car` would boxo be philosophically okay with using that as a canonical version, or is the hope to pursue divergence per the reference to having this implementation "thinned out"?
Is there an issue or reference to what complexity in the current car library we hope can be removed?

BigLep:
> The dependency issue here seems to be on the `blocks` library and whether to use the boxo instance versus the older one in `go-block-format`.

I believe tha's right.

> go-car could remove this dependency with a bit of refactoring by defining the interface locally, so that it can be compatible with either of these implementations.

Cool!
 
> If there wasn't a dependency re-factor on `ipld/go-car` would boxo be philosophically okay with using that as a canonical version, or is the hope to pursue divergence per the reference to having this implementation "thinned out"? 

Per https://github.com/ipfs/boxo/issues/218#issuecomment-1490916807 , go car libraries isn't where we were wanting to apply maintenance or innovation energy.  I know the team is comfortable building and working with car tooling if necessary (e.g., @Jorropo's different car related libraries in https://github.com/Jorropo/linux2ipfs and https://github.com/Jorropo/go-featheripfs).

I expect the team will want a break for a month or two before considering getting rid of boxo/ipld/car given the fatigue on all the renaming and path updating the last month.

I know I'm being non-committal here, not because have another secret plan, but because don't know at this point.

> Is there an issue or reference to what complexity in the current car library we hope can be removed?

Good point.  I don't think so: https://github.com/ipld/go-car/issues?q=is%3Aissue+jorropo+
I'll ask @Jorropo to do this now...

willscott:
> Per https://github.com/ipfs/boxo/issues/218#issuecomment-1490916807 , go car libraries isn't where we were wanting to apply maintenance or innovation energy.

The angle I'm taking here is that there have been a number of PRs over the last year to fix bugs or otherwise improve the car-v1 codebase that is in the go-car repo (example [1](https://github.com/ipld/go-car/pull/357) [2](https://github.com/ipld/go-car/pull/331) [3](https://github.com/ipld/go-car/pull/312)). By having a separate instance in boxo, there's a duplicate maintenance burden we're incurring.

In the meantime we probably want to have some sense of the expected process: - is it on go-car to notify boxo that there's a PR there that probably should get pulled into the boxo implementation and vice versa? or should the boxo maintainers be watching go-car for such relevant PRs and vice versa?

willscott:
>>    The dependency issue here seems to be on the blocks library and whether to use the boxo instance versus the older one in go-block-format.

> I believe tha's right.

I looked into potentially providing an interface, but this seems to not be the case.

The boxo fork of car is using the same external `go-block-format` interface as ipld/go-car [1](https://github.com/ipfs/boxo/blob/main/ipld/car/car.go#L10).

As such, I don't think there's a refactor-needed / incompatibility in linking against the external ipld/go-car as it is. If that isn't true we should track down what it is we are proposing needs to be done.

BigLep:
> we probably want to have some sense of the expected process: - is it on go-car to notify boxo that there's a PR there that probably should get pulled into the boxo implementation and vice versa? or should the boxo maintainers be watching go-car for such relevant PRs and vice versa?

Good point @willscott - thanks for raising.

I think Boxo in general needs a mechanism here to know if all the copied repos (including ipld/go-car) have received "important updates".  I created https://github.com/ipfs/boxo/issues/270 to track this work (and included an for how dependabot could be leveraged).  

For ipld/go-car specifically, I don't want you and the current maintainers to have to do anything extra than you already do for notifying consumers of releases/changes.  I assume that happens through the normal github release and dependabot flow, but if there is anywhere else we should be watching let us know.  Certainly if there are important bug fixes or performance improvements, we're not going to say "no" to extra communication in #boxo-maintainers FIL slack or ipfs/boxo issue, but that is not expected.  

Thanks!

willscott:
That makes sense - my angle here has been to make sure that if nothing else, the pull from the ipld instance can remain as trivial as possible - so making sure we understand what would prevent the code from being identical between the two.

BigLep:
@willscott : concerning `go-block-format` I'll need @Jorropo to chime in.  I believe the issue manifests due to nested types, but I'll let others speak with authority here.

BigLep:
Remaining work here @Jorropo :
1. Give a shipit / merge on the PR itself
2. Make it clear what ipld/go-car would need to change dependency-wise so Boxo could depend on it again.  I think it would be nice if we were in a position of not needing to copy that code because of dependencies, but I can understand if we keep the copy because of different needs.
3. :white_check_mark: Specify what you'd like ipld/go-car to ideally do (you've done this per https://github.com/ipld/go-car/issues/416 )

BigLep:
Per IPFS Thing 2023 conversations, we are going to be able to remove Boxo's ipld/go-car copy from Boxo.  When that happens, this PR can just be closed.

BigLep:
I'm closing this PR since we'll be removing boxo/car per https://github.com/ipfs/boxo/issues/218#issuecomment-1590156907

PR #245, opened 2023-03-31
feat(gateway): IPFSBackend  metrics and HTTP range support

lidel:
This PR is part of https://github.com/ipfs/bifrost-gateway/pull/61

- Moves metric-related code to `gateway/metrics.go`
- Adds opentelemetry span for every IPFSBackend API call 
- Adds success/failure histogram for each API call ipfs_gw_backend_api_call_duration_seconds 

## TODO

- [x] `Get` and `GetRange` are now the same, as suggested in https://github.com/ipfs/boxo/pull/240#pullrequestreview-1365554977
- [x] fix tests: check why Kubo sharness fails
  - [x] Kubo sharness / CI  in https://github.com/ipfs/kubo/pull/9786 must be green

### DIFF
--- a/blockservice/blockservice.go
+++ b/blockservice/blockservice.go
@@ -244,7 +244,7 @@ func getBlock(ctx context.Context, c cid.Cid, bs blockstore.Blockstore, fget fun
 
 		// TODO be careful checking ErrNotFound. If the underlying
 		// implementation changes, this will break.
-		logger.Debug("Blockservice: Searching bitswap")
+		logger.Debug("BlockService: Searching")
 		blk, err := f.GetBlock(ctx, c)
 		if err != nil {
 			return nil, err
@@ -262,7 +262,7 @@ func getBlock(ctx context.Context, c cid.Cid, bs blockstore.Blockstore, fget fun
 		return blk, nil
 	}
 
-	logger.Debug("Blockservice GetBlock: Not found")
+	logger.Debug("BlockService GetBlock: Not found")
 	return nil, err
 }
 
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -54,11 +54,13 @@ type ContentPathMetadata struct {
 	ContentType      string // Only used for UnixFS requests
 }
 
-// GetRange describes a range request within a UnixFS file. From and To mostly follow HTTP Range Request semantics.
+// ByteRange describes a range request within a UnixFS file. From and To mostly follow [HTTP Byte Range] Request semantics.
 // From >= 0 and To = nil: Get the file (From, Length)
 // From >= 0 and To >= 0: Get the range (From, To)
 // From >= 0 and To <0: Get the range (From, Length - To)
-type GetRange struct {
+//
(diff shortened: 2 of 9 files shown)


gateway/metrics.go
```diff
@@ -0,0 +1,312 @@
...
+	begin := time.Now()
+	name := "IPFSBackend.GetCAR"
+	ctx, span := spanTrace(ctx, name, trace.WithAttributes(attribute.String("path", path.String())))
+	defer span.End()
+
+	md, rc, errCh, err := b.api.GetCAR(ctx, path)
+
+	// TODO: handle errCh
+	b.updateApiCallMetric(name, err, begin)
+	return md, rc, errCh, err
```
lidel, line 140: - [ ] TODO

lidel, line 128: :thought_balloon: (late friday wrap-up)  @aschmahmann  `updateApiCallMetric` will increase counter for success or failure based on `err`. Are we missing anything critical if we only check `err` here and ignore `errCh`? 
(This should not block this PR, we can fix metric later)

blockservice/blockservice.go
```diff
@@ -244,7 +244,7 @@ func getBlock(ctx context.Context, c cid.Cid, bs blockstore.Blockstore, fget fun
 
 		// TODO be careful checking ErrNotFound. If the underlying
 		// implementation changes, this will break.
-		logger.Debug("Blockservice: Searching bitswap")
+		logger.Debug("BlockService: Searching")
```
lidel, line 247: :information_source: there is no bitswap in bifrost-gateway, yet we got super confusing "Searching bitswap" message here ;)

gateway/metrics.go
lidel, line 1: :information_source:  moved metrics code here

gateway/gateway.go
```diff
@@ -86,15 +88,24 @@ func NewGetResponseFromDirectoryListing(dagSize uint64, entries <-chan unixfs.Li
...
+	//
+	// Range notes:
+	//   - Generating response to a range request may require additional data
+	//     beyond the passed ranges (e.g. a single byte range from the middle of a
+	//     file will still need magic bytes from the very beginning for content
+	//     type sniffing).
+	//   - A range request for a directory currently holds no semantic meaning.
+	//
+	// [HTTP Byte Ranges]: https://httpwg.org/specs/rfc9110.html#rfc.section.14.1.2
+	Get(context.Context, ImmutablePath, ...ByteRange) (ContentPathMetadata, *GetResponse, error)
```
lidel, line 108: :information_source: removed `GetRange`, we now have `Get` with optional range requests

lidel (commented):
@aschmahmann @hacdias I think this is ready for review.

TLDR: adds metrics, fixes range requests, merges `Get` and `GetRange` into one func.

Sharness in Kubo PR is green: https://github.com/ipfs/kubo/pull/9786

hacdias (approved):
Thanks for this. I tested it locally and it indeed solves the range requests. It LGTM. Nice call to move the metrics stuff to a separate file. It was quite a clutter!

PR #246, opened 2023-03-31
ci(gateway): enable subdomain conformance

hacdias:
This PR enables the subdomain conformance gateway tests (added in #237). Since we are using the CAR example to run the Gateway tests, we need to make it more conformant, of course, similarly to what happened in https://github.com/ipfs/bifrost-gateway/pull/66.

- Enables subdomain conformance tests on `example.net` (domain used in examples)
- Adds `CONNECT` support to the gateway examples.
- Extract the handler creation login into the `common` subpackage since it was the same for both examples.


examples/gateway/common/handler.go
```diff
@@ -0,0 +1,81 @@
...
+func withConnect(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodConnect {
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
```
hacdias, line 81: ℹ️ In the future, we may want to provide this as part of the gateway module itself as a helper, since it's a very common thing to need. All of our implementations (Kubo, Bifrost, Examples) need this since all use the `http.ServeMux` at some point.

.github/workflows/gateway-conformance.yml
```diff
@@ -47,7 +47,6 @@ jobs:
           xml: output.xml
           html: output.html
           markdown: output.md
```
galargh, line 49: We can stick with example.net :) 
```suggestion
          markdown: output.md
          subdomain-url: http://example.net
```

galargh (approved):

lidel (approved):
Thank you! :ship: 

(while looking at this, found unrelated nit: https://github.com/ipfs/boxo/pull/176#pullrequestreview-1366830427)

PR #249, opened 2023-03-31
fix: remove binary that was accidentally checked in

guseggert:


Jorropo (approved):
Whoops.

Jorropo (changes requested):
`.gitignore` ?

Jorropo (approved):

guseggert:
to remove the bin from the tree i'm going to force amend and force push this

PR #252, opened 2023-04-03
feat(gateway)!: deserialised responses turned off by default

### LINKED ISSUE #225: gateway: add trustless-only mode
## What

There should be a way to only expose [response types required by trustless mode](https://docs.ipfs.tech/reference/http/gateway/#trustless-verifiable-retrieval).

The trustless-only mode must have two key features:
- client is provided with ability to fetch all information necessary for verifying and deserializing data (Block, CAR, and `ipfs-record` from [IPIP-351](https://github.com/ipfs/specs/pull/351)) end-to-end.
-   it is impossible to make a mistake and send request that delegated trust to gateway
  - when enabled, trusted responses are disabled
  - for example, if someone sends request without explicit `Accept` or `?format`, gateway returns HTTP error 501 Not Implemented stating only verifiable response types are supported

## How 

TBD, we need some sane defaults that also account for users of library not shooting themselves in the foot if they do nothing.

- Only trustless responses by default 
  - add implicit exception for localhost / 127.0.0.1 / ::1) 
  - enabling trusted responses require explicit opt-in per hostname

## Why 

Hard lessons from project Rhea / Saturn about the tyranny of the default. Exposing deserialized responses in cases where a project only needs a subset of the entire gateway spec creates a surface for abuse.

It is way, way less work for everyone if boxo/gateway library provides a single configuration option to allow deserialized responses on non-localhost hostnames.

hacdias:
Closes #225. Adds support for [trustless and trusted](https://specs.ipfs.tech/http-gateways/trustless-gateway/) gateway modes.

- Default is Trustless Mode as per #225. In addition,
	- Can be globally overwritten by setting `Config.TrustedMode` to `true`
	- Can be overwritten per known gateway by setting `gateway.TrustedMode` to `true/false`
	- ~`localhost`, `127.0.0.1` and `::1` are implicit trusted mode.~
- Cleans up the following:
	- Moves all gateway configuration to `Config`, including `NoDNSLink` and `PublicGateways`. This is motivated by the fact that we now need to access this information from the main handler in order to determine if it's a trustless or trusted gateway.
	- Reworked the context hostname keys. See code comments for explanation.
- Examples are set in trusted mode because that's what our tests expect.
	- Perhaps in a separate PR we could rework the tests for the proxy version to instead use the trustless gateway as it is a perfect example.
- In trustless mode, requests for trusted resources return <del>501</del> 406.
- Includes many tests, except for `/ipns/{cid}?format=ipns-key` because we still don't have a good way of doing that.

Sharness will fail here because structures and defaults have changed. Kubo PR: https://github.com/ipfs/kubo/pull/9789

@lidel @aschmahmann I may have missed some very specific case. Please take a look at let me know what you think.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -15,7 +15,69 @@ import (
 
 // Config is the configuration used when creating a new gateway handler.
 type Config struct {
+	// Headers is a map containing all the headers that should be sent by default
+	// in all requests. You can define custom headers, as well as add the recommended
+	// headers via AddAccessControlHeaders.
 	Headers map[string][]string
+
+	// DeserializedResponses configures this gateway to support returning data
+	// in deserialized format. By default, the gateway will only support
+	// trustless, verifiable [application/vnd.ipld.raw] and
+	// [application/vnd.ipld.car] responses, operating as a [Trustless Gateway].
+	//
+	// This global flag can be overridden per FQDN in PublicGateways map.
+	//
+	// [application/vnd.ipld.raw]: https://www.iana.org/assignments/media-types/application/vnd.ipld.raw
+	// [application/vnd.ipld.car]: https://www.iana.org/assignments/media-types/application/vnd.ipld.car
+	// [Trustless Gateway]: https://specs.ipfs.tech/http-gateways/trustless-gateway/
+	DeserializedResponses bool
+
+	// NoDNSLink configures the gateway to _not_ perform DNS TXT record lookups in
+	// response to requests with values in `Host` HTTP header. This flag can be
+	// overridden per FQDN in PublicGateways. To be used with WithHostname.
+	NoDNSLink bool
+
+	// PublicGateways configures the behavior of known public gateways. Each key is
+	// a fully qualified domain name (FQDN). To be used with WithHostname.
+	PublicGateways map[string]*Specification
+}
+
+// Specification is the specification of an IPFS Public Gateway.
+type Specification struct {
+	// Paths is explicit list of path prefixes that should be handled by
+	// this gateway. Example: `["/ipfs", "/ipns"]`
+	// Useful if you only want to support immutable `/ipfs`.
+	Paths []string
+
+	// UseSubdomains indicates whether or not this is a [Subdomain Gateway].
+	//
+	// If this flag is set, any `/ipns/$id` and/or `/ipfs/$id` paths in Paths
+	// will be permanently redirected to `http(s)://$id.[ipns|ipfs].$gateway/`.
+	//
+	// We do not support using both paths and subdomains for a single domain
+	// for security reasons ([Origin isolation]).
+	//
+	// [Subdomain Gateway]: https://specs.ipfs.tech/http-gateways/subdomain-gateway/
+	// [Origin isolation]: https://en.wikipedia.org/wiki/Same-origin_policy
+	UseSubdomains bool
+
+	// NoDNSLink configures this gateway to _not_ resolve DNSLink for the
+	// specific FQDN provided in `Host` HTTP header. Useful when you want to
+	// explicitly allow or refuse hosting a single hostname. To refuse all
+	// DNSLinks in `Host` processing, set NoDNSLink in Config instead. This setting
+	// overrides the global setting.
+	NoDNSLink bool
+
+	// InlineDNSLink configures this gateway to always inline DNSLink names
+	// (FQDN) into a single DNS label in order to interop with wildcard TLS certs
+	// and Origin per CID isolation provided by rules like https://publicsuffix.org
+	//
+	// This should be set to true if you use HTTPS.
+	InlineDNSLink bool
+
+	// DeserializedResponses configures this gateway to support returning data
+	// in deserialized format. This setting overrides the global setting.
+	DeserializedResponses bool
 }
 
 // TODO: Is this what we want for ImmutablePath?
@@ -221,7 +283,17 @@ func AddAccessControlHeaders(headers map[string][]string) {
 type RequestContextKey string
 
 const (
-	DNSLinkHostnameKey RequestContextKey = "dnslink-hostname"
+	// GatewayHostnameKey is the key for the hostname at which the gateway is
+	// operating. It may be a DNSLink, Subdomain or Regular gateway.
 	GatewayHostnameKey RequestContextKey = "gw-hostname"
-	ContentPathKey     RequestContextKey = "content-path"
+
+	// DNSLinkHostnameKey is the key for the hostname of a DNSLink Gateway:
+	// https://specs.ipfs.tech/http-gateways/dnslink-gateway/
+	DNSLinkHostnameKey RequestContextKey = "dnslink-hostname"
+
+	// SubdomainHostnameKey is the key for the hostname of a Subdomain Gateway:
+	// https://specs.ipfs.tech/http-gateways/subdomain-gateway/
+	SubdomainHostnameKey RequestContextKey = "subdomain-hostname"
+
+	ContentPathKey RequestContextKey = "content-path"
 )
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -198,14 +198,20 @@ func newTestServerAndNode(t *testing.T, ns mockNamesys) (*httptest.Server, *mock
 }
 
 func newTestServer(t *testing.T, api IPFSBackend) *httptest.Server {
-	config := Config{Headers: map[string][]string{}}
+	return newTestServerWithConfig(t, api, Config{
+		Headers:               map[string][]string{},
+		DeserializedResponses: true,
+	})
+}
+
+func newTestServerWithConfig(t *testing.T, api IPFSBackend, config Config) *httptest.Server {
 	AddAccessControlHeaders(config.Headers)
 
 	handler := NewHandler(config, api)
 	mux := http.NewServeMux()
 	mux.Handle("/ipfs/", handler)
 	mux.Handle("/ipns/", handler)
-	handler = WithHostname(mux, api, map[string]*Specification{}, false)
+	handler = WithHostname(config, api, mux)
 
 	ts := httptest.NewServer(handler)
 	t.Cleanup(func() { ts.Close() })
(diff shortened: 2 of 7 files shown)


gateway/gateway_test.go
```diff
@@ -544,3 +547,40 @@ func TestGoGetSupport(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, http.StatusOK, res.StatusCode)
 }
+
+func TestTrustlessMode(t *testing.T) {
```
hacdias: Only tested for `/ipfs`. Not sure how we can simulate `/ipns` here with keys. We can only simulate DNSLink the way we did in the other tests. This is something we should figure out at some point.

gateway/handler.go
```diff
@@ -277,6 +277,19 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if !i.isTrustedMode(r) {
+		switch responseFormat {
+		case "application/vnd.ipld.raw",
```
aschmahmann: Flagging that IMO for trustless mode we should error if the request is not for `/ipfs/cid` (i.e. any IPNS or path elements require trust)

gateway/handler.go
```diff
@@ -277,6 +277,19 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
...
+			"application/vnd.ipld.car",
```
aschmahmann: Flagging that IMO for trustless mode we should error if the request is not for `/ipfs/cid` (i.e. any IPNS or path elements require trust).

We can loosen this up to also allow for paths once we have CARs returning the blocks in the path as well.

aschmahmann (dismissed):

gateway/handler.go
hacdias: Your definitely right, I will fix this (same for comment below).

gateway/handler.go
hacdias: Just to add here that `/ipns/{cid}?format=ipns-record` is also trustless. Just waiting for https://github.com/ipfs/specs/pull/351/ to be merged to show up in the specs website, as its been implemented for quite some time already. /cc @lidel

gateway/handler.go
```diff
@@ -314,6 +322,68 @@ func (i *handler) addUserHeaders(w http.ResponseWriter) {
...
+func (i *handler) isValidTrustlessRequest(contentPath ipath.Path, responseFormat string) bool {
+	// Only allow "/{#1}/{#2}"-like paths.
+	trimmedPath := strings.Trim(contentPath.String(), "/")
+	pathComponents := strings.Split(trimmedPath, "/")
+	if len(pathComponents) != 2 {
+		return false
+	}
+
+	if contentPath.Namespace() == "ipns" {
+		// Only ipns records allowed.
```
```suggestion
		// Only ipns records allowed until https://github.com/ipfs/specs/issues/369 is resolved
```

gateway/handler.go
```diff
@@ -314,6 +322,68 @@ func (i *handler) addUserHeaders(w http.ResponseWriter) {
...
+		if responseFormat != "application/vnd.ipfs.ipns-record" {
+			return false
+		}
+
+		// Only valid peers, no DNSLink.
+		if _, err := peer.Decode(pathComponents[1]); err != nil {
```
lidel: :thought_balloon: now that we've moved go-ipns into `boxo/ipns`, it may be a good idea to add `IpnsName.Decode()` and `IpnsName.String()` there so we don't use PeerIDs in code related to IPNS.

(Internally it would still use `peer.Decode`, but allows us to control things like base and supported legacy encodings independently of go-libp2p/peer defaults.

gateway/hostname.go
```diff
@@ -121,7 +83,7 @@ func WithHostname(next http.Handler, api IPFSBackend, publicGateways map[string]
 			if !gw.NoDNSLink && hasDNSLinkRecord(r.Context(), api, host) {
 				// rewrite path and handle as DNSLink
 				r.URL.Path = "/ipns/" + stripPort(host) + r.URL.Path
-				next.ServeHTTP(w, withHostnameContext(r, host))
+				next.ServeHTTP(w, withSubdomainContext(r, host))
```
lidel: :thought_balloon:  why are we using `withSubdomainContext`  ad not `withDNSLinkContext` here?

iiuc this is  a code path that  handles DNSLink in `Host` header (not `dweb.link/ipns/example.com` but `example.com` case)

gateway/gateway.go
```diff
@@ -15,7 +15,62 @@ import (
...
+	// in all requests. You can define custom headers, as well as add the recommended
+	// headers via AddAccessControlHeaders.
 	Headers map[string][]string
+
+	// TrustedMode configures this gateway to allow trusted requests. By default,
+	// the gateway will operate in trustless mode, as defined in the specification:
+	// https://specs.ipfs.tech/http-gateways/trustless-gateway/.
+	//
+	// This only applies to hostnames not defined under PublicGateways. In addition,
+	// the hostnames localhost, 127.0.0.1 and ::1 are considered trusted by default.
```
lidel: :warning:  :thought_balloon: I wonder if this is the right place to make this call.
Implicit defaults like that feel like a surface for security problems, when someone just uses library, then exposes it over 127.0.0.1 and then exposes that via Nginx to the public internet.

User did nothing wrong, just used default state of library, and exposed themselves to hosting phishing / bad content in deserialized form.

I think we should change this, so the default state for `boxo/gateway`  library is trustless, even on localhost.

@hacdias  mind refactoring things, so by default library is trustless-only, and we explicitly enable trusted mode on localhost names in examples? :pray:

lidel (changes requested):

gateway/hostname.go
hacdias: We never used the equivalent of `withDNSLinkContext` here. I don't know. Probably because the code is too flaky and if we did set it, then the cross-origin check would return `true` and links would be incorrect. I never wrote this code, but I think that's the reason.

hacdias: We never used the equivalent of `withDNSLinkContext` here. I'm not really sure why. I will add it and check if sharness passes. It looks like a bug but I don't really see why there aren't complaints. It's possible it was something that was missed on https://github.com/ipfs/kubo/pull/8890

gateway/handler.go
hacdias: Added, and can also be used for #236. We can rebase and update in whichever is the last PR to be merged.

ipns/name.go
lidel, line 1: :thought_balloon: Is `Name` from `ipns/name.go`  used anywhere? Looks like these changes are  left-over from something else? Can we remove them from this PR?

lidel (changes requested):
I took this for a spin with fresh eyes and functionally looks ok, but  the UX felt off. 

My bad, I know I suggested going with "TrustedMode" :see_no_evil:  The problem is deeper than just having something that work fine as boolean.

I feel we need to make another pass at the name (i know.. but its important to get this one right)  -- @hacdias lmk thoughts on the below direction:

### Problem

We've created a single flag, and this in turn builds a false dichotomy around "trust".

People familiar with content-addressing understand what "trustless" mean and interpret "trusted" deserialized responses as a "step back", paradoxically, something "less useful, because it requires trust". 

However everyone outside our bubble will see "trust" and "no trust" and always pick "trust". 


### Proposed change

If we replace "trusted" with "deserialized"  it would convey a slightly different meaning:

* **Deserialized Mode**:  (disabled by default, user opt-in) This configuration mode indicates that the gateways supports returning data in a deserialized format, allowing for direct data manipulation and interpretation without the need for further processing, at the cost of the client being unable to do the verification (unless Trustless Mode is enabled as well).
  - We want this to be disabled by default, implementers having to explicitly opt-in in their configs.

* **Trustless Mode**: (implicitly enabled) This mode indicates if gateway supports returning content-addressed data  in its original state, without any deserialization or interpretation, allowing for external verification and processing.
  - We want this to be ALWAYS enabled. The ability to fetch a raw block should not be negotiable, this is the core function of HTTP Gateway

### Rationale

My main point is to avoid the semantic trap around "trust", and convey functional meaning instead.

By using "deserialized [responses]" instead of "trusted [mode]," the focus shifts to the nature of the responses being returned by the gateway, highlighting whether they are in a processed (deserialized) or raw (unprocessed) form. It is what matters to the end user, and the developer that integrates  `boxo/gateway` in their own app -- we can't expect people to understand "trusted" vs "trustless".

By having two separate terms, we avoid false dichotomy: "deserialized" and  "trustless" are siblings, not the opposites (and we always support trustless, while deserialized is opt-in).

### TLDR

:point_right:  I propose we rename `TrustedMode` to `DeserializedResponses`.
We still can keep it as bool (disabled by default, require opt-in) but the configuration flag becomes self-explanatory and removes ambiguity/confusion around the semantic meaning of "trust".

ipns/name.go
hacdias: Yes, it is used in <del>`isSerializedRequest`</del> `isTrustlessRequest`

ipns/name.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"strings"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/multiformats/go-multibase"
+)
+
+// Name is an IPNS name
+type Name string
```
lidel: Hm.. if we make this public, we should design this for the future instead of `peer.ID`

- Including `/ipns/` prefix makes it tedious to remove it every time one interacts with name -- all we care is the multihash inside of CID, and that should be stored
- lets model this more closely after  https://github.com/libp2p/go-libp2p/blob/master/core/peer/peer.go
    - the CIDv1 Base36 normalization  should happen inside `String()` 
- use it instead of `peer.ID` in `boxo/ipns` library

    
But this is bigger chunk of design work, I feel we would want to do this properly when we tackle IPNS cleanup from something like https://github.com/ipfs/specs/issues/376 (doing breaking API changes in one release)

@hacdias mind removing `ipns.Name` from this PR for now? (ok to open separate PR wit it)
We don't really use it yet, all we care is confirming the component parses as a valid peer.ID, so maybe:

```diff
- if _, err := ipns.Decode(pathComponents[1]); err != nil {
+ // TODO: replace with ipns.Name as part of https://github.com/ipfs/specs/issues/376
+ if _, err := peer.Decode(pathComponents[1]); err != nil {
```

lidel (dismissed):
@hacdias I've pushed some tweaks in godocs and  tiny HTTP fixes/renames for readability (details in the commits).

The `DeserializedResponse` handling inside of  `boxo/gateway looks ok, but we should move `ipns.Name` to separate PR (details below).

lidel (approved):
Thank you @hacdias :heart: 

Hopefully, this will make it easier for people to leverage https://specs.ipfs.tech/http-gateways/trustless-gateway/ and expose deserialized responses only when they really need to.

PR #255, opened 2023-04-03
fix(gw): panic in handler_car.go

lidel:
Potential fix for https://github.com/ipfs/bifrost-gateway/issues/72


willscott (approved):
this looks like it should fix setting the header to nil

lidel:
Confirmed it works https://github.com/ipfs/bifrost-gateway/issues/72#issuecomment-1494955417

PR #256, opened 2023-04-04
feat(gateway): trace context header support

hacdias:
Ref.
- https://www.w3.org/TR/trace-context/
- https://github.com/ipfs/bifrost-gateway/issues/68

This PR:

- Adds `tracing` sub-package based on Kubo's in order to support `OTEL_TRACES_EXPORTER` easily across the stack. This may be removed in the future, once `opentelemetry-go` supports it by default.
- Add documentation in `docs/tracing.md` on how to use tracing including a script on how to generate a `traceparent` header.
	- In this document, I show how to generate a traceparent header that can be passed onto a request. Then, the trace id from this header can be used to find the request in Jaeger. We may want to open an issue to investigate a way of "back-propagating" a tracing id to the user, such that the user can get the header without having to generate it. Perhaps `trace-id`. I would avoid using the exact same headers in Trace Context spec as per https://www.w3.org/TR/trace-context/#other-risks
- Updates the gateway examples to support tracing.
- **Kubo PR**: https://github.com/ipfs/kubo/pull/9811
	- Changes Kubo to use the helper provided here, instead of duplicating code.
- **Bifrost Gateway**: https://github.com/ipfs/bifrost-gateway/pull/74
	- Adds tracing for the first time in Bifrost Gateway, using the functions defined here.


gateway/metrics.go
```diff
@@ -294,7 +295,20 @@ func newHistogramMetric(name string, help string) *prometheus.HistogramVec {
 	return histogramMetric
 }
 
+var tracer = otel.Tracer("boxo/gateway")
+
 // spanTrace starts a new span using the standard IPFS tracing conventions.
-func spanTrace(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-	return otel.Tracer("boxo").Start(ctx, fmt.Sprintf("%s.%s", " Gateway", spanName), opts...)
+func spanTrace(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	return tracer.Start(ctx, name, opts...)
```
hacdias: I simplified this here. If we need to use the previous nomenclature I'm happy to change it.

examples/gateway/common/tracing.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func NewTracerProvider() *trace.TracerProvider {
+	tp := trace.NewTracerProvider()
+
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
+
+	return tp
+}
```
Jorropo:
```suggestion
```

examples/gateway/proxy/main.go
```diff
@@ -37,6 +38,12 @@ func main() {
 		log.Fatal(err)
 	}
 
+	ctx := context.Background()
+	tp := common.NewTracerProvider()
+	defer func() {
+		_ = tp.Shutdown(ctx)
+	}()
+
```
Jorropo:
```suggestion
```

gateway/metrics.go
```diff
@@ -294,7 +295,20 @@ func newHistogramMetric(name string, help string) *prometheus.HistogramVec {
...
+// withTraceContext sets the Trace Context (https://github.com/w3c/trace-context)
+// headers on the request based on the context and the propagator.
+func withTraceContext(next http.Handler) http.Handler {
+	props := otel.GetTextMapPropagator()
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		props.Inject(r.Context(), propagation.HeaderCarrier(w.Header()))
+		next.ServeHTTP(w, r)
+	})
 }
```
Jorropo:
```suggestion
```

examples/gateway/common/handler.go
```diff
@@ -58,10 +59,15 @@ func NewHandler(gwAPI gateway.IPFSBackend) http.Handler {
...
 
-	// Finally, wrap with the withConnect middleware. This is required since we use
+	// Then, wrap with the withConnect middleware. This is required since we use
 	// http.ServeMux which does not support CONNECT by default.
 	handler = withConnect(handler)
 
+	// Finally, wrap with the otelhttp handler. This will allow the tracing system
+	// to work and for correct propagation of tracing headers. This step is optional
+	// and only required if you want to use tracing.
+	handler = otelhttp.NewHandler(handler, "Gateway.Request")
```
Jorropo, line 70:
```suggestion
	handler = otelhttp.NewHandler(handler, "Gateway.Request", otelhttp.WithPropagators(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{})))
```

examples/gateway/common/handler.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/ipfs/boxo/gateway"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
```
Jorropo, line 9:
```suggestion
	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
	"go.opentelemetry.io/otel/propagation"
```

Jorropo (dismissed):
Simpler and clearer.

Jorropo:
I think we are getting otel wrong, we set globals and force tracing onto everyone.

I belive we should have options where consumers must pass a tracer or propagator or ... in order to enable tracing, else it does nothing. Doing dependency injection instead of setting globals.

hacdias:
@Jorropo did you run the example tests with your suggestions? The header is then not present. If you manage to make it work, I will apply your suggestions. 

I tried exactly what you did first, and it didn't work. That's why I did it the way I did.

Jorropo:
@hacdias no but we dont expect to see a header.

We want the gateway to be a children of Saturn, Nginx front end, ... this needs to accept not emit traceparent headers.

fwiw this already work for cmds in Kubo.

hacdias:
@Jorropo hmm... how can we then test if it's working properly? Afaik, that's the reason it wasn't added in https://github.com/ipfs/kubo/pull/9169, but then the tests from https://github.com/ipfs/kubo/pull/9180 are also invalid.

Jorropo:
@hacdias you can test it yourself and send a screenshot i dont think this require ci.
Ive left it out previously because i wanted to known if that is compatible with the Nginx otel tracing.

hacdias:
Okay, I had to read more about this. Indeed, my impression is that we **should not emit** headers, only accept them and pass them down.

> When vendors include traceparent and tracestate headers in responses, these values may inadvertently be passed to cross-origin callers. Vendors should ensure that they include only these response headers when responding to systems that participated in the trace. - https://www.w3.org/TR/trace-context/#other-risks

(This is also the only section where responses are mentioned, hence my conclusion.)

But also, "Vendors should ensure that they include only these response headers when responding to systems that participated in the trace." is not very clear to me. If I send a `traceparent` to this PR with your suggestions, I don't get the header back. I think that we need a mix of your changes and my changes. I am going to try something.

--

Then, I don't understand why the initial issue for this subject was to add `traceparent` in the response (https://github.com/ipfs/kubo/issues/9168 and https://github.com/ipfs/kubo/pull/9180). CC @iand

hacdias:
After reading the specification, I think @Jorropo is right. The specification clearly states that the `traceparent` header should be sent top-down on the HTTP **requests,** that is, the gateway should accept it and pass it down to further requests. However, the gateway is not supposed to send it back in the HTTP **response**.

> When vendors include traceparent and tracestate headers in responses, these values may inadvertently be passed to cross-origin callers. Vendors should ensure that they include only these response headers when responding to systems that participated in the trace. - https://www.w3.org/TR/trace-context/#other-risks

Therefore, there isn't much to do. This PR adds OTel tracing to the example to show how it should be added, as an example. It also changes the default HTTP Client for the examples such that we use the `otelhttp` transport to ensure the tracing headers are passed down. I am going to update the relevant Bifrost Gateway and Kubo PRs. However, those do **not** depend on this.

I want to add that we already have tracing spans in all handlers and in the IPFSBackend. I made the naming consistent in #261. The data is there. It is now the responsibility of the user of `boxo/gateway` whether or not they want to enable tracing or not.

lidel (changes requested):
We need to add a test for this, to protect from refactor-driven regressions in the future.

Perhaps  `examples/gateway/proxy`? It provides good setup for this.
Enable a tracing reporter and test with and without `traceparent` in request, and then make assert if it was also passed to the proxy backend.

hacdias:
@lidel I added two tests:

1. If the parent request includes `traceparent`, the down the line requests include `{version}-{trace id}-`. Please note that according to the spec the remaining parts can change.
2. If there is no `traceheader`, one is created for requests down the line.

An important thing to note is that for all of this to work, we have to use `http.NewRequestWithContext` to ensure that the context with the tracing information is preserved.

hacdias:
I updated the main comment https://github.com/ipfs/boxo/pull/256#issue-1653578967 with the current information.

docs/tracing.md
```diff
@@ -0,0 +1,143 @@
...
+    -p 9411:9411 \
+    jaegertracing/all-in-one
+```
+
+Then, in other terminal, start the CAR Gateway example (or Kubo), with Jaeger
+tracing enabled:
+
+```
+$ OTEL_TRACES_EXPORTER=jaeger ./car-gateway -c data.car -p 8040
+```
```
guseggert: a copy-paste example for Kubo would be nice, as that's a common use case here (probably more common that CAR gateway?)

docs/tracing.md
```diff
@@ -0,0 +1,143 @@
...
+  - [`Zipkin Exporter`](#zipkin-exporter)
+  - [`File Exporter`](#file-exporter)
+  - [`OTEL_PROPAGATORS`](#otel_propagators)
+- [Using Jaeger UI](#using-jaeger-ui)
+- [Generate `traceparent` Header](#generate-traceparent-header)
+
+## Environment Variables
+
+For advanced configurations, such as ratio-based sampling, please see also the
+[OpenTelemetry Environment Variable Specification].
```
guseggert, line 28: missing URL?  (also a few cases above)

docs/tracing.md
```diff
@@ -0,0 +1,143 @@
...
+```
+$ OTEL_TRACES_EXPORTER=jaeger ./car-gateway -c data.car -p 8040
+```
+
+Finally, the [Jaeger UI] is available at http://localhost:16686.
+
+## Generate `traceparent` Header
+
+If you want to trace a specific request and want to have its tracing ID, you can
+generate a `Traceparent` header. According to the [Trace Context] specification,
```
guseggert, line 109: missing URL

docs/tracing.md
```diff
@@ -0,0 +1,143 @@
...
+> trace-flags      = 2HEXDIGLC   ; 8 bit flags. Currently, only one bit is used. See below for details
+> ```
+
+To generate a valid `Traceparent` header value, the following script can be used:
+
+```bash
+version="00" # fixed in spec at 00
+trace_id="$(cat /dev/urandom | tr -dc 'a-f0-9' | fold -w 32 | head -n 1)"
+parent_id="00$(cat /dev/urandom | tr -dc 'a-f0-9' | fold -w 14 | head -n 1)"
+tace_flag="01"   # sampled
```
guseggert: trace_flag ?

docs/tracing.md
hacdias: Not missing, they're on the bottom. It works with Markdown.

docs/tracing.md
hacdias: Not missing. URLs defined at the end according to MD spec.

docs/tracing.md
hacdias: Yes: https://w3c.github.io/trace-context/#trace-flags

docs/tracing.md
hacdias: Updated to use Kubo.

docs/tracing.md
Winterhuman: I think they mean "tace_flag" is wrong

hacdias: Typo fixed

docs/tracing.md
```diff
@@ -0,0 +1,142 @@
...
+    -p 5778:5778 \
+    -p 16686:16686 \
+    -p 14268:14268 \
+    -p 14269:14269 \
+    -p 14250:14250 \
+    -p 9411:9411 \
+    jaegertracing/all-in-one
+```
+
+Then, in other terminal, start the Kubo daemon, with Jaeger tracing enabled:
```
lidel:
```suggestion
Then, in other terminal, start the app that uses `boxo/tracing` internally (e.g., a Kubo daemon), with Jaeger exporter enabled:
```

lidel (approved):
I took this for a spin with Jaeger UI in both Kubo (https://github.com/ipfs/kubo/pull/9811) and bifrost-gateway (https://github.com/ipfs/bifrost-gateway/pull/74) and as far i could tell works as expected.

| Kubo | bifrost-gw  with GRAPH_BACKEND=true |
| ---- | ---- |
|  Screenshot 2023-04-12 at 00-52-42 Jaeger UI | Screenshot 2023-04-12 at 01-02-57 Jaeger UI |



Merging so we can leverage `traceparent` in bifrost-gateway for Rhea, but also can  start leveraging it in places like IPNI HTTP client at https://github.com/ipfs/boxo/tree/main/routing/http/client

docs/tracing.md
guseggert: slick, i didn't know MD did this

Issue #250, opened 2023-03-31
Release v0.8.0


guseggert:
- [x] Verify your [GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification) is configured in local git and GitHub
- [x] Ensure Boxo and Kubo are checked out on your system
- [x] Create a release branch called `release-vX.Y.Z` and push it
  * Use `master` as base if `Z == 0`
  * Use `release` as base if `Z > 0`
- [x] Tidy the changelog and push any changelog changes to `release-vX.Y.Z`
- [x] Create a draft PR merging `release-vX.Y.Z` into `release`
- [x] If `Z > 0`, then cherry-pick the necessary commits from `master` into `release-vX.Y.Z` using `git cherry-pick -x <commit>`
- [x] Ensure Boxo tests are passing
- [x] Ensure Kubo tests are passing
  - [x] Go to Kubo dir and run `go get github.com/ipfs/boxo@<commit_hash>` using the commit hash of the `release-vX.Y.Z` branch
  - [x] Run `go mod tidy` in repo root and in `docs/examples/kubo-as-a-library`
  - [x] Commit the changes and open a draft PR in Kubo
  - [x] Paste a link to the Kubo PR in the Boxo PR, so reviewers can verify the Kubo test run
  - [x] Verify the CI passes
- [x] Add a commit in `release-vX.Y.Z` bumping the version in `version.json` to `vX.Y.Z`
- [x] Add a "release" label to the Boxo PR
- [x] After the release checker creates a draft release, copy-paste the changelog into the draft
- [x] Wait for approval from another Boxo maintainer
- [x] Merge the PR into `release`, _using "Create a Merge Commit"_, and do not delete the `release-vX.Y.X` branch
  - [x] Verify the tag is created automatically by the releaser action
- [x] Create a PR merging `release` into `main`
- [x] Create the GitHub release

One todo we don't want to forget: this needs to change to v0.8.0: https://github.com/ipfs/boxo/blob/5dee8e4a31c8a12defd050adeaacf03288a463c8/cmd/boxo-migrate/boxomigrate.go#L71

guseggert:
Release PR: https://github.com/ipfs/boxo/pull/257

PR #257, opened 2023-04-04
Release v0.8.0

guseggert:
Note: when this PR is merged, the release tag will be created

Kubo PR: https://github.com/ipfs/kubo/pull/9793

Tracking issue with the release checklist: https://github.com/ipfs/boxo/issues/250

Reviewers, if you can, please sanity check that release checklist, since it is new. After this release, I'll make an issue template out of it.


hacdias (approved):

CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
Jorropo, line 5: I would like to consider using a file per version, like for Kubo, aside of rendering issues (which should be fine for a while). I just like having a per version file because this is easy to link to and always have the ToC at the right place even if newer versions are released.

Feel free to resolve this if you want to keep this like this.

guseggert:
Awesome thanks.  I also want to wait on @Jorropo to look at this to sanity-check the repo consolidation stuff before pulling the trigger.

CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
```
Jorropo, line 8: I'm instantly confused, I had to open https://keepachangelog.com/en/1.0.0/ just to understand the changelog, one benefit of individual files is that we can open new ones without clobbering old files and having to move stuff into an `## [Unreleased]` section.

If anything I would name this `not yet released` but again that fine.

CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+### Changed
+
+- Removed a mention of "bitswap" in blockservice debug logs
+- Changed the Bitswap message package from "bitswap.message.pb" to "bitswap.message.v1.pb" to avoid protobuf panics due to duplicate registration with [go-bitswap](https://github.com/ipfs/go-bitswap)
+- Remove a busyloop in blockservice getBlocks by removing batching when caching
+
+### Deprecated
+
```
Jorropo, line 50: I would add `None` so this look less like a bug.

CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
...
+### Removed
+
```
Jorropo, line 54: Same `None` here.

galargh (approved):
All looks good on the release automation front :)

Jorropo (changes requested):
I want to see https://github.com/ipfs/boxo/pull/263 merged first, after that and looking at my comments you can implicitly view this review as an approval.

CHANGELOG.md
guseggert: The intention here is to use https://github.com/pl-strflt/changelog-driven-release so that the changelogs drive the release version and the GitHub release notes. The benefit is that there isn't a scramble at release time to assemble the changelogs, they get accumulated in [Unreleased] and then you change [Unrelease] to [vX.Y.Z] and the release happens. The goal is to reduce the cost of cutting a release so that we cut them more often.

CHANGELOG.md
guseggert: > I had to open https://keepachangelog.com/en/1.0.0/ just to understand the changelog

This is a feature not a bug? Following a standard format is not only easier on humans but enables automation like https://github.com/pl-strflt/changelog-driven-release

Jorropo (approved):

CHANGELOG.md
BigLep: I think @Jorropo has a good point about the durability of the links when we do one file per version.  Related, it makes the URLs easily hackable.  For example I know I can always do https://github.com/ipfs/kubo/blob/master/docs/changelogs/v0.Y.md#-highlights to get the highlights section.

I don't think this is a blocker, but maybe a feature request to add to https://github.com/pl-strflt/changelog-driven-release about supporting multiple files.

Issue #190, opened 2023-03-03
[WIP] Frequently Asked Questions about this project


BigLep:
## Table of Contents
- [Purpose](#purpose)
- [Rational FAQs](#rational-faqs)
  - [What are the reasons/goals for doing repo consolidation?](#what-are-the-reasonsgoals-for-doing-repo-consolidation)
  - [Why is the Boxo project consolidating repos and using example-driven development despite ecosystem friction?](#why-is-the-boxo-project-consolidating-repos-and-using-example-driven-development-despite-ecosystem-friction)
  - [Why do maintainers care about reducing the number of repos they maintain?](#why-do-maintainers-care-about-reducing-the-number-of-repos-they-maintain)
  - [What alternatives were considered and why were they ruled against?](#what-alternatives-were-considered-and-why-were-they-ruled-against)
    - [Just focus on docs/examples in a repo with a known set of versions that play nice together](#just-focus-on-docsexamples-in-a-repo-with-a-known-set-of-versions-that-play-nice-together)
  - [Given this isn't free for maintainers or those affected, what has been the cost/benefit analysis?](#given-this-isnt-free-for-maintainers-or-those-affected-what-has-been-the-costbenefit-analysis)
  - [What are some concrete usecases where repo consolidation would have been beneficial?](#what-are-some-concrete-usecases-where-repo-consolidation-would-have-been-beneficial)
- [Copied or Migrated Repos FAQs](#copied-or-migrated-repos-faqs)
  - [What repos have been copied into Boxo?](#what-repos-have-been-copied-into-boxo)
  - [What will happen if there is a security issue in one of the legacy ipfs/\* repos that has been copied into boxo?](#what-will-happen-if-there-is-a-security-issue-in-one-of-the-legacy-ipfs-repos-that-has-been-copied-into-boxo)
  - [I don't like the "deprecated" warnings in "not maintained" repos that have been moved to boxo.  What can be done about this?](#i-dont-like-the-deprecated-warnings-in-not-maintained-repos-that-have-been-moved-to-boxo--what-can-be-done-about-this)
  - [How does one claim ownership of a "not maintained" repo?](#how-does-one-claim-ownership-of-a-not-maintained-repo)
  - [Will the "not maintained" ipfs/\* repos be left around to rot?](#will-the-not-maintained-ipfs-repos-be-left-around-to-rot)
- [Logistics FAQs](#logistics-faqs)
[...]

PR #260, opened 2023-04-05
Added FAQ and README updates

BigLep:
Getting the FAQ into a more durable location: https://github.com/ipfs/boxo/issues/190

This also involved pulling in various parts from https://github.com/ipfs/kubo/issues/8543 so it can get resolved.  

Various slack conversations like #boxo-maintainers thread https://filecoinproject.slack.com/archives/C04M8232QRW/p1677103751480109 were also reviewed.

Some new concepts were added in here like our deprecation policy.


BigLep (commented):
@guseggert : feel free to take over and edit at will.  I wanted to get us over the hump of not having some of this info in a more durable location that we couldn't PR.

docs/FAQ.md
```diff
@@ -0,0 +1,177 @@
...
+- [Copied or Migrated Repos FAQ](#copied-or-migrated-repos-faq)
+  - [What repos have been copied into Boxo?](#what-repos-have-been-copied-into-boxo)
+  - [What will happen if there is a security issue in one of the legacy ipfs/\* repos that has been copied into boxo?](#what-will-happen-if-there-is-a-security-issue-in-one-of-the-legacy-ipfs-repos-that-has-been-copied-into-boxo)
+  - [How does Boxo make sure it doesn't miss any important changes that are in the original/source/copied repos?](#how-does-boxo-make-sure-it-doesnt-miss-any-important-changes-that-are-in-the-originalsourcecopied-repos)
+  - [I don't like the "deprecated" warnings in "not maintained" repos that have been moved to boxo.  What can be done about this?](#i-dont-like-the-deprecated-warnings-in-not-maintained-repos-that-have-been-moved-to-boxo--what-can-be-done-about-this)
+  - [How does one claim ownership of a "not maintained" repo?](#how-does-one-claim-ownership-of-a-not-maintained-repo)
+  - [Will the "not maintained" ipfs/\* repos be left around to rot?](#will-the-not-maintained-ipfs-repos-be-left-around-to-rot)
+
+## Rationale FAQ
```
guseggert: This seems weird to call this an "FAQ", I don't think we're really going to maintain this as an FAQ based on what users frequently ask? We don't have any evidence that anyone outside of PL cares about these questions, and also it seems more like just some historical context around why Boxo exists and why a bunch of repos got moved into it. Instinctively this stuff feels more appropriate for a GH issue that we link to in the FAQ/README if someone wants to know the decision making that led to Boxo.

I guess from another angle I'm concerned about over-indexing on PL as Boxo's users, they are definitely users but I think we want to really grow our userbase outside of PL, and an FAQ like this might drive those kinds of users away. If it were me I'd be like "Gosh this looks like it's a huge ordeal to use Boxo", which isn't the case and isn't the impression we want to give.

BigLep: That's good feedback Gus.  Stepping back it makes me think we should lean into Github wikis more for this kind of long form documentation.  It has better discoverability and navigation properties than issues and doesn't clog up the issue tracker.  (@Jorropo has floated this idea recently.). I'll do this now.

BigLep:
@guseggert : ready for re-review.  I moved the content into wiki pages: https://github.com/ipfs/boxo/wiki

The README refavtor is still there.  

I think it likely makes sense to move more of the README to wiki as well so we have a central place for our docs.

docs/FAQ.md
guseggert: That's amazing I love it :bowing_man:

guseggert (approved):

Issue #233, opened 2023-03-29
Gateway: too many CoreAPI.ResolvePath calls


Jorropo:
Here is a trace I captured from loading `dist.ipfs.io`'s favicon (on `v0.8.0-rc2`):
![Capture d’écran du 2023-03-29 05-52-07](https://user-images.githubusercontent.com/24391983/228422547-5ebcbd06-1c8a-4ee9-a5d7-c7a33c1c9f11.png)

You can see that it loads the same ressource over and over, there is no point to restarting from `/ipns/dist.ipfs.io` all the time.
Ideally once we resolve a path component we followup all future resolution from this path component or better.

They all endup calling into `namesys`.
![Capture d’écran du 2023-03-29 05-58-33](https://user-images.githubusercontent.com/24391983/228423166-56fd80f5-cc31-4426-9067-67682cac5b30.png)

It doesn't look like the latency impact is very bad because all but the first call to ResolvePath and first call to `BlockAPI.Get` are very fast, so they are probably hitting a cache, it looks like a low hanging fruit that would reduce allocations and CPU usage a bit tho.

Jorropo:
This looks like it could be the cullprit (if anything it at least expose the same bug):

https://github.com/ipfs/boxo/blob/5bbb21bbd458a833f32e25e87c89b7d2c7857400/gateway/handler_unixfs__redirects.go#L261

This code should shift the last match into a variable while iteratively traversing the folders.

I'll wait on https://github.com/ipfs/boxo/pull/176 before doing anything.

hacdias:
@Jorropo I think this has been solved in the meanwhile through the successive refactors. It seems like we only hit Namesys once now. I tried with a few different pages and only see a single call. 

![image](https://github.com/ipfs/boxo/assets/5447088/532e2be9-9350-48bc-be9e-48995aa5ae38)

Feel free to re-open if you still see many calls.

PR #261, opened 2023-04-05
refactor(gateway): consistent tracing names

hacdias:
In the context of https://github.com/ipfs/bifrost-gateway/issues/68, I decided to make the tracing names more consistent. When we moved the gateway to boxo, and then on #176, we added and changed some of the tracings, but they were not consistent.

What I did here:

- Prepended `Handler.` to all handler tracings, similar to what we had on the `IPFSBackend` with `IPFSBackend.`.
- Changed the tracer name to `boxo/gateway` to clearly indicate the sub-package ~~and stopped pre-pending `Gateway.` in the tracing name, since we already have the gateway part in the tracer name.~~.

### DIFF
--- a/gateway/handler_block.go
+++ b/gateway/handler_block.go
@@ -12,7 +12,7 @@ import (
 
 // serveRawBlock returns bytes behind a raw block
 func (i *handler) serveRawBlock(ctx context.Context, w http.ResponseWriter, r *http.Request, imPath ImmutablePath, contentPath ipath.Path, begin time.Time) bool {
-	ctx, span := spanTrace(ctx, "ServeRawBlock", trace.WithAttributes(attribute.String("path", imPath.String())))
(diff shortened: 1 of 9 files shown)


lidel (commented):
@hacdias do you have a screenshot how it looks like in Jaeger UI when traced in Kubo or bifrost-gateway (whichever is easier for you)? 

See https://github.com/ipfs/kubo/blob/master/docs/environment-variables.md#how-to-use-jaeger-ui
(examples in https://github.com/ipfs/boxo/issues/233, just want to eyeball it so we are happy with how it looks)

hacdias:
image

@lidel here it is. I re-added the `Gateway.` prefix. I thought the tracer name would be more prominent but it wasn't! It's better now! (also see https://github.com/ipfs/kubo/pull/9797).

lidel (approved):
lgtm

> Screenshot 2023-04-12 at 02-49-44 Jaeger UI

PR #263, opened 2023-04-05
cmd/boxo-migrate: add an error message if we do not find a .git folder

Jorropo:


cmd/boxo-migrate/boxomigrate.go
```diff
@@ -67,6 +73,33 @@ func main() {
...
+						}
+
+						for {
+							g := filepath.Join(p, ".git")
+							_, err := os.Stat(g)
+							if err == nil {
+								break
+							}
+							if p == "/" {
+								return fmt.Errorf(`
```
guseggert, line 90: this is not portable, should check if `p == previousP`

Jorropo: Thx fixed

guseggert (approved):

Issue #196, opened 2023-03-07
[Milestone] Getting boxo "over the hump"


BigLep:
## Background
boxo has been in limbo state for a couple of months which doesn't maximize its utility for maintainers and potential consumers and it doesn't instill confidence in existing consumers who are impacted.

This is a tracking issue for the work for getting boxo to a place of providing value for its intended audience and being clear for those involved and affected on what they can expect going forward.

## Done Criteria
The `Tasks` below are completed.  This includes both the various communication pieces and the resulting release in https://github.com/ipfs/boxo/issues/202. 

## Phases
- [x] ⬇️⬇️⬇️ Planning phase ⬇️⬇️⬇️
   - [x] https://github.com/ipfs/boxo/issues/191
   - [x] https://github.com/ipfs/boxo/issues/174
- [x] ⬇️⬇️⬇️ Staging/validation/RC phase ⬇️⬇️⬇️
   - [x] https://github.com/ipfs/boxo/issues/218
   - [x] https://github.com/ipfs/boxo/issues/215
   - [x] Copy in repos listed in https://github.com/ipfs/boxo/issues/202
   - [x] Cut an RC: https://github.com/ipfs/boxo/releases/tag/v0.8.0-rc1
   - [x] Validate consolidation RC with Kubo: https://github.com/ipfs/kubo/pull/9746
   - [x] Validate consolidation RC with Lotus: https://github.com/filecoin-project/lotus/pull/10562
   - [x] https://github.com/ipfs/boxo/issues/291
   - [x] Lotus cleanup: https://github.com/filecoin-project/lotus/pull/10921
   - [x] Validate consolidation RC with ipfs-cluster
- [ ] ⬇️⬇️⬇️ Documentation/community-focus phase ⬇️⬇️⬇️
   - [x] https://github.com/ipfs/boxo/issues/189
   - [x] https://github.com/ipfs/boxo/issues/181
   - [x] Populate the FAQs in https://github.com/ipfs/boxo/issues/190
   - [x] Close out https://github.com/ipfs/kubo/issues/8543 by pointing to this issue
   - [x] https://github.com/ipfs/boxo/issues/170
   - [ ] https://github.com/ipfs/boxo/issues/203
- [x] ⬇️⬇️⬇️ Disruption phase ⬇️⬇️⬇️
   - [x] Make the boxo release with the copy consolidated repos: https://github.com/ipfs/boxo/pull/257
[...]

PR #264, opened 2023-04-05
Release v0.8.0

guseggert:
Note the releaser and release checker will fail here since the release has already occurred, this is just syncing up `main` with the `release` branch.


aschmahmann (approved):
LGTM (although I haven't read the changelog too closely)

BigLep (commented):
Great!

My only takeaway from the changelog in this format is that it doesn't necessarily leave me with "why should I care" and "how can I learn more".  

Why should I care? - Maybe adding a highlights or tldr section as a place forus to put a bit of commentary.  In this particularly release I would want to say things like (quick - off the cuff):

"Library code for HTTP gateways keeps getting better.  A major refactor was done to support make requests for blocks rather than individual blocks.  This was desired functionality for Kubo and bifrost-gateway."

"We made big steps in Boxo being a capable toolbox for IPFS implementations with getting many repos under one roof so we can avoid version issues between them and apply improvement refactors that were proactively cost-prohibitive before.  You can track this effort in https://github.com/ipfs/boxo/issues/196"

Learn more? 
Maybe we just need to add more issue links in the changelog comments.  I realize someone can go through gitblame to figure this out if they really want to, but I think it's nice to make it easy.  For example, linking to the blockservice/busyloop issue would have helped make it clearer the significance.

BigLep:
Also, in future, maybe we should have some linking between this PR and https://github.com/ipfs/boxo/pull/257 and/or some naming conventions to easily differentiate

guseggert:
Yeah I'm not too fond of the "Keep a Changelog" format TBH, but I can see the value in just using some consistent format that people (and automation) know how to deal with, even if it's not perfect. I don't have a strong opinion either way here.

I would prefer to not have a bunch of prose that ties everything together, just b/c it's a pain to write at release time, I'm really trying to minimize the amount of work it takes to do a release. Maybe we could come up with a format that keeps things discrete but still clearly communicates "motivation" to users?

Issue/PR links sound good.

PR #273, opened 2023-04-06
provider: refactor to only maintain one batched implementation and add throughput callback

Jorropo:
This code does a few things:
1. Remove the simple provider to avoid duplicating features.
2. Add the support for single providing on the batched provider.
3. Fix a bugs in the batched provider.
4. Add support for a throughputCallback in the batched provider.
6. Add support for an offline mode of the batched provider (stuff is exclusively pushed onto the queue).
5. Move the batched provider to be the only provider and make the queue implementation private.


hacdias:
@Jorropo can you update this once you have time so that the tests are passing? I can take a look after (also add some motivation on why this is beneficial for more context, please).

BigLep:
Yes, agreed that the context for where this is coming from and why we're prioritizing it now would be great.  It's not currently on the 0.21 iteration.  (I'll add it now.)

hacdias:
Is this ready to be reviewed @Jorropo?

BigLep:
Related to https://github.com/ipfs/kubo/issues/9703

BigLep:
2023-05-23 maintainer conversation:

Context: we know this will get rewritten again when it moves into the DHT.  As a result, this doesn't have to be perfect.  

@guillaumemichel : can you please look at this (given this is relevant to future code for the DHT)?  We want to make sure you have context here when future DHT rewriting happens.

@aschmahmann will look at the test code.

@Jorropo : add a changelog.md entry

guillaumemichel:
The function below is not deterministic and only has a 50% chance to return an error when the context is done.

https://github.com/ipfs/boxo/blob/04cf63e49f78fe22e68be0af0a660ab12c69db6e/provider/internal/queue/queue.go#L57-L65

I rewrote it, feel free to update if required

https://github.com/ipfs/boxo/blob/08e06807d1d8df735e217e8448a5594d31c4b415/provider/internal/queue/queue.go#L57-L71

provider/internal/queue/queue.go
```diff
@@ -79,8 +84,6 @@ func (q *Queue) work() {
 		defer func() {
 			// also cancels any in-progess enqueue tasks.
 			q.close()
-			// unblocks anyone waiting
-			close(q.dequeue)
```
guillaumemichel, line 83: Why?

provider/noop.go
```diff
@@ -0,0 +1,35 @@
...
+
+var _ System = (*noopProvider)(nil)
+
+// NewNoopProvider creates a ProviderSystem that does nothing.
+func NewNoopProvider() System {
+	return &noopProvider{}
+}
+
+func (op *noopProvider) Run() {
+}
```
guillaumemichel: Is `Run()` really necessary?
https://github.com/ipfs/boxo/blob/08e06807d1d8df735e217e8448a5594d31c4b415/provider/provider.go#L17-L36

provider/provider.go
```diff
@@ -3,25 +3,121 @@ package provider
...
+	return func(ctx context.Context) (<-chan cid.Cid, error) {
+		return bstore.AllKeysChan(ctx)
+	}
+}
+
+// Pinner interface defines how the simple.Reprovider wants to interact
+// with a Pinning service
+type Pinner interface {
+	DirectKeys(ctx context.Context) ([]cid.Cid, error)
+	RecursiveKeys(ctx context.Context) ([]cid.Cid, error)
```
guillaumemichel: I am not sure to understand what `RecursiveKeys` is. Comments explaining what `DirectKeys` and `RecursiveKeys` would be welcome.

Is `RecursiveKeys` a superset of `DirectKeys`?

provider/reprovider.go
```diff
@@ -390,31 +473,46 @@ func (s *BatchProvidingSystem) getLastReprovideTime() (time.Time, error) {
...
-	TotalProvides, LastReprovideBatchSize     int
+type ReproviderStats struct {
+	TotalProvides, LastReprovideBatchSize     uint64
 	AvgProvideDuration, LastReprovideDuration time.Duration
 }
 
 // Stat returns various stats about this provider system
-func (s *BatchProvidingSystem) Stat(ctx context.Context) (BatchedProviderStats, error) {
+func (s *reprovider) Stat() (ReproviderStats, error) {
 	// TODO: Does it matter that there is no locking around the total+average values?
```
guillaumemichel: Comment still relevant?

guillaumemichel (approved):

guillaumemichel:
I relocated some magic numbers in [reprovider.go](https://github.com/ipfs/boxo/blob/c1d114c92aa08aafb4d52a6e2767b4328b740b1b/provider/reprovider.go)

https://github.com/ipfs/boxo/blob/c1d114c92aa08aafb4d52a6e2767b4328b740b1b/provider/reprovider.go#L21-L37

Note that I updated the `DefaultReprovideInterval` from 24h to 22h (see https://github.com/ipfs/kubo/pull/9326 and https://github.com/ipfs/kubo/blob/da28fbc65a2e0f1ce59f9923823326ae2bc4f713/config/reprovider.go#L5)

provider/internal/queue/queue.go
```diff
@@ -56,6 +56,12 @@ func (q *Queue) Close() error {
 
 // Enqueue puts a cid in the queue
 func (q *Queue) Enqueue(cid cid.Cid) error {
+	select {
+	case <-q.ctx.Done():
+		return fmt.Errorf("failed to enqueue CID: shutting down")
+	default:
+	}
+
```
Jorropo: Why ? This is confusing and do nothing intresting.

guillaumemichel: The initial function was non-deterministic and had a 50% chance to return an error when the context is done.

Jorropo: is that a problem ? Provider is guarded by a `sync.WaitGroup`, so the close is synchronous, this will converge on a closed state quickly enough.

guillaumemichel: If we don't care about returning an error, and it converges quickly enough on a close state, then the function could simply be:
```go
 func (q *Queue) Enqueue(cid cid.Cid) { 
 	q.enqueue <- cid
 } 
```

I don't see the point of throwing an error, if it only works 50% of the time.

guillaumemichel: But you probably know better the codebase so I trust you on this one

Jorropo: @guillaumemichel we can't just write to `Enqueue` because it could deadlock, we still have to check the context.

My point is that `q.enqueue` will be backed up since the worker will exit and the writes attempt to it would fail forcing select to go through the context.

The only way this code is non deterministic is if you race closing with enqueuing which sounds fine to have non deterministic behaviour there (we couldn't even fully remove it).

provider/noop.go
Jorropo: No, this is an old remanent, before you had to create a provider and then call Run on it.
I found this confusing so I replace it with a `Online` option. Will remove

provider/reprovider.go
Jorropo: No thx will remove

provider/internal/queue/queue.go
Jorropo: Added a comment:
```go
	// We don't close dequeue because the provider which consume this get caught in
	// an infinite loop dequeing cid.Undef if we do that.
	// The provider has it's own select on top of dequeue and will handle this by itself.
```

provider/provider.go
Jorropo: I replaced this with `github.com/ipfs/boxo/pinning/pinner.Pinner`.

guillaumemichel (approved):

BigLep:
@Jorropo : this is good to merge right?

Jorropo:
@BigLep I'll submit changelog update first

PR #278, opened 2023-04-12
feat(routing): allow-offline with routing put

laurentsenta:
sister PR in kubo: https://github.com/ipfs/kubo/pull/9667

### DIFF
--- a/coreiface/tests/routing.go
+++ b/coreiface/tests/routing.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gogo/protobuf/proto"
 	iface "github.com/ipfs/boxo/coreiface"
+	"github.com/ipfs/boxo/coreiface/options"
 	ipns_pb "github.com/ipfs/boxo/ipns/pb"
 )
 
@@ -20,6 +21,7 @@ func (tp *TestSuite) TestRouting(t *testing.T) {
 
 	t.Run("TestRoutingGet", tp.TestRoutingGet)
 	t.Run("TestRoutingPut", tp.TestRoutingPut)
+	t.Run("TestRoutingPutOffline", tp.TestRoutingPutOffline)
 }
 
 func (tp *TestSuite) testRoutingPublishKey(t *testing.T, ctx context.Context, api iface.CoreAPI) iface.IpnsEntry {
@@ -28,7 +30,12 @@ func (tp *TestSuite) testRoutingPublishKey(t *testing.T, ctx context.Context, ap
 		t.Fatal(err)
 	}
 
-	entry, err := api.Name().Publish(ctx, p)
+	// TODO: make sure we are fine with that implicit allow offline
+	opts := []options.NamePublishOption{
+		options.Name.AllowOffline(true),
+	}
+
+	entry, err := api.Name().Publish(ctx, p, opts...)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


Jorropo:
Is https://github.com/ipfs/kubo/pull/9667 correct ? I don't see how this relates

laurentsenta:
@Jorropo yup: https://github.com/ipfs/kubo/pull/9667/commits/290f29afb33123d9d9bf98138b5a835942a56d3a

BigLep:
@hacdias : can you please look at this again so we can get it merged?

hacdias (changes requested):
Thanks @laurentsenta. This looks good to me.

I would like to see a test for this in [coreiface/tests/routing.go](https://github.com/ipfs/boxo/blob/main/coreiface/tests/routing.go) but I'm not sure how easy that would be. I think by using `tp.MakeAPISwarm(ctx, true, 1)` you'll be able to spin an offline node and therefore `Put` should work with `AllowOffline(true)` but not without. Note that this tests are run in Kubo's repo.

laurentsenta:
@hacdias thanks for taking the time to review both PRs, 

From a quick look + tests:

- There is some coupling between online and full identity in the [node init](https://github.com/ipfs/kubo/blob/53821c202f14df2116459ba8729d61248342e15a/core/coreapi/test/api_test.go#L85)

I'll dig some more, but let me know if you've seen this before: when I try to start a node with full identity AND offline, I get a sigsev

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40 pc=0x10131ab08]

goroutine 773 [running]:
github.com/ipfs/kubo/core/bootstrap.bootstrapRound({0x102732a98?, 0x14000a88640?}, {0x0, 0x0}, {0x14000315fb0?, 0x0?, 0x140003a6330?, 0x14000322fe0?})
	/Users/laurent/dev/plabs/ipfs/kubo/core/bootstrap/bootstrap.go:118 +0x68
github.com/ipfs/kubo/core/bootstrap.Bootstrap.func1({0x102741478?, 0x140004f5320?})
	/Users/laurent/dev/plabs/ipfs/kubo/core/bootstrap/bootstrap.go:89 +0xac
github.com/jbenet/goprocess.(*process).Go.func1()
	/Users/laurent/dev/go/pkg/mod/github.com/jbenet/goprocess@v0.1.4/impl-mutex.go:134 +0x3c
created by github.com/jbenet/goprocess.(*process).Go
	/Users/laurent/dev/go/pkg/mod/github.com/jbenet/goprocess@v0.1.4/impl-mutex.go:133 +0x204
FAIL	github.com/ipfs/kubo/core/coreapi/test	0.856s
FAIL
```

- Currently, the "sane" way to create an IPNS entry is by going through Publish + Get (that's what the test for Put does); that might be the right time to extract that code from the Publish function. We had a similar discussion about [verify and unmarshall](https://github.com/ipfs/gateway-conformance/pull/37#discussion_r1165450704).

hacdias:
@laurentsenta I've never seen this error before, and it might be a bug? It seems like `host` is nil, which seems to make sense since we do not set `PeerHost` in `IpfsNode`. It seems like we'll need to make a few changes to the test function in Kubo to ensure we can have an offline and full identity node. But first, do we need full identity here?

> Currently, the "sane" way to create an IPNS entry is by going through Publish + Get (that's what the test for Put does); that might be the right time to extract that code from the Publish function.

Do you mean a `Create` function?

coreiface/tests/routing.go
```diff
@@ -90,3 +97,36 @@ func (tp *TestSuite) TestRoutingPut(t *testing.T) {
...
+	err = api.Routing().Put(ctx, "/ipns/"+ipnsEntry.Name(), data)
+	if err == nil {
+		t.Fatal("this operation should fail because we are offline")
+	}
+
+	err = api.Routing().Put(ctx, "/ipns/"+ipnsEntry.Name(), data, options.Put.AllowOffline(true))
+	if err != nil {
+		t.Fatal(err)
+	}
+}
```
laurentsenta, line 127: @hacdias I fixed the error by having a finer "isOnline" boolean in the `MakeAPI` code,

>> Currently, the "sane" way to create an IPNS entry is by going through Publish + Get (that's what the test for Put does); that might be the right time to extract that code from the Publish function.
>
> Do you mean a Create function?

Yes, see the code above I have to call `Publish`, then `Get`, just to get an IPNS payload that I can use with `Put`. If possible, I would have liked to be able to create the IPNS entry directly.

But seeing the result, it's less ugly than I thought, so maybe we're fine sticking with this approach. Let me know.

coreiface/tests/routing.go
```diff
@@ -28,7 +30,12 @@ func (tp *TestSuite) testRoutingPublishKey(t *testing.T, ctx context.Context, ap
 		t.Fatal(err)
 	}
 
-	entry, err := api.Name().Publish(ctx, p)
+	// TODO: make sure we are fine with that implicit allow offline
```
hacdias: Maybe you could add `options.NamePublishOption` in the arguments and let the caller decide. Just to be more flexible in the future.

coreiface/tests/routing.go
hacdias: I think it's fine keeping this. Please just look at my comment below.

coreiface/tests/routing.go
laurentsenta: That make sense, fixed!

hacdias (approved):

PR #281, opened 2023-04-16
feat(verifcid): AddGoodHash

Wondertan:
Allows Boxo users to expand the set of secure hash functions without having to fork Boxo.

Users may introduce new hashes to multishash registry but won't be able to use them as Blockservice layer is strict about which hashes are allowed via `verifcid` pkg. 

Now anyone can add new hashes globally, potentially degrading security, which is a tradeoff of this solution. Open to alternatives that do not involve forking.

I didn't add any godocing as none of the funcs in the pkg has it, but happy to populate all of them, as a side quest.

### DIFF
--- a/verifcid/validate.go
+++ b/verifcid/validate.go
@@ -33,6 +33,13 @@ var goodset = map[uint64]bool{
 	mh.SHA1: true, // not really secure but still useful
 }
 
+func AddGoodHash(code uint64) {
+	good, found := goodset[code]
+	if !found || !good {
+		goodset[code] = true
+	}
+}
+
 func IsGoodHash(code uint64) bool {
 	good, found := goodset[code]
 	if good {


Jorropo:
I don't like this, using globals for configuration is bad. I am not saying what we have right now is good, but it can't get *that* bad because it's not configurable.

If you want to make it configurable I really think we should refactor this to be dependency injected.

Other maintainers might disagree.

hacdias:
I will agree with @Jorropo as this will influence all the user's of this package in the same application. If we add configurability, it'd be better to ensure that it is only applied to the usage context.

Jorropo:
Feel free to send a PR that refactors this in some dependency injectable interface we would like this.

But no globals configs, this was a mistake but let's not make it worst.

Wondertan:
Please fix this in the way you see the best. It feels aligned with all the maintenance and chore work you've been doing recently.

I have a brief idea, but it is unlikely to follow your opinions and practices, and I will need more time to get into the review round loops than I can afford.

Jorropo:
@Wondertan I don't mind fixing this. I don't want it like this.

Globals are *bearable* as long as they are read only after the `init` of the package.
They are not a valid place to put configuration options, we recently just fixed an other case like #291 #328.

You would first need to create an interface:
```go
// better name wanted
type Validator interface{
 IsGoodHash(code uint64) bool
 ValidateCid(c cid.Cid) error
}
```
Then you move the current global to a struct:
```go
type validator struct{
 goodset map[uint64]bool
}

func NewValidator(allowedCodes map[uint64]bool) Validator {
 return validator{allowedCodes}
}

func (v validator) IsGoodHash(code uint64) bool {
	good, found := v.goodset[code]
	if good {
		return true
	}

	if !found {
		if code >= mh.BLAKE2B_MIN+19 && code <= mh.BLAKE2B_MAX {
			return true
		}
		if code >= mh.BLAKE2S_MIN+19 && code <= mh.BLAKE2S_MAX {
			return true
		}
	}

	return false
}

func (v validator) ValidateCid(c cid.Cid) error {
	pref := c.Prefix()
	if v.IsGoodHash(pref.MhType) {
		return ErrPossiblyInsecureHashFunction
	}

	if pref.MhType != mh.IDENTITY && pref.MhLength < minimumHashLength {
		return ErrBelowMinimumHashLength
	}

	if pref.MhType != mh.IDENTITY && pref.MhLength > maximumHashLength {
		return ErrAboveMaximumHashLength
	}

	return nil
}
```
Then you change all consumers of verifcid from calling the global functions to taking a `Validator` interface as input and calling the interface instead.

Note: for backward compatibility you should have a `DefaultVerifier` global variable that holds a validator with the current list and keep the current global `IsGoodHash` and `ValidateCid` functions but have they query `DefaultVerifier`.

After doing this:
1. There is an interface so you don't actually need to get stuff merged, you can make your own implementation and pass it in to bitswap, ipld, ... whoever uses verifcid.
2. If you want to add `AddGoodHash` you I would be ok merging this in boxo given this is made by make a new extending interface and add the feature (but not on the global verifier).

PR #286, opened 2023-04-26
refactor: remove badger, leveldb dependencies

hsanjuan:
Boxo should not bring in such external dependencies only because some benchmarks that no one looks at anymore decided to use those deps. They have too many strings attached (i.e. badger doesn't build on some platforms etc).


lidel (approved):
thank you! removing these deps sounds sensible 


(:thought_balloon:  we've pulled in a lot of things when we migrated repos, and i suspect we have a lot of deps that are only used in tests. these are more tricky to cleanup -- we run tests, but have no time to refactor to use less deps. probably something that needs to happen organically over time)

pinning/pinner/dspinner/pin_test.go
```diff
@@ -890,11 +889,9 @@ func (d *batchWrap) Batch(_ context.Context) (ds.Batch, error) {
 }
 
 func makeStore() (ds.Datastore, ipld.DAGService) {
-	ldstore, err := lds.NewDatastore("", nil)
-	if err != nil {
-		panic(err)
-	}
-	dstore := &batchWrap{ldstore}
+	md := ds.NewMapDatastore()
```
Jorropo, line 892: This have terrible Query performance and is totally unusable with the dht server code.

Jorropo (dismissed):
It may looks like we just use it as a KV store but this is wrong.
In multiple places we incorrectly implement a graph dB by listing and filtering keys.
This data store implement this by listing everything all the keys using go's map range iteration syntax which is not fast and `O(N)` based on the full size, not bucketed.
Also dht server store GiBs of data which is nice to do on disk.

guseggert:
Are you having a concrete issue w/ this? 

In general, consumers won't get this dependency in their build list--even if they directly use the dspinner package--since it's a test dependency.

pinning/pinner/dspinner/pin_test.go
guseggert: I don't understand how this is relevant, given these are tests?

Jorropo: I was under the impression this was a Kubo PR. It's fine here.

hsanjuan:
> Are you having a concrete issue w/ this?
> 
> In general, consumers won't get this dependency in their build list--even if they directly use the dspinner package--since it's a test dependency.

These rather superfluous tests make Badger appear in go.mod. That automatically directs dependency resolution when nothing overrides it, and I think can cause conflicts sometimes or cause importing old versions of things. This comes after https://github.com/ipfs/kubo/issues/9839, although I think this could stay just fine (because it's just tests as you say). I just thought there's no need for boxo to pull in these deps at all in its dependency tree, particularly given the low value of doing so (it's not even tests, it's benchmarks).

hacdias (approved):
I think this is a reasonable change. I'm not very versed in how Go dependency management works, but assuming what @hsanjuan is true, this is likely a good change and should not affect anything as we're not really removing any tests.

PR #290, opened 2023-05-05
Streamed pins

MichaelMure:
Corresponding kubo PR: https://github.com/ipfs/kubo/pull/9859

The rational is that if the pin list get big, a synchronous call to get the complete list can delay handling unnecessarily. For example, when listing indirect pins, you can start walking the DAGs immediately with the first recursive pin instead of waiting for the full list.

This matters even more on low power device, of if the pin list is stored remotely.

Also in this PR, a DAG generator to simplify testing (used in the kubo PR to test the GC).

replace https://github.com/ipfs/interface-go-ipfs-core/pull/109
replace https://github.com/ipfs/go-ipfs-pinner/pull/24
replace https://github.com/ipfs/go-ipfs-provider/pull/48
replace https://github.com/ipfs/go-ipfs-blocksutil/pull/18

### DIFF
--- a/coreiface/tests/pin.go
+++ b/coreiface/tests/pin.go
@@ -198,6 +198,9 @@ func (tp *TestSuite) TestPinRecursive(t *testing.T) {
 	}
 	n := 0
 	for r := range res {
+		if r.Err() != nil {
+			t.Error(r.Err())
+		}
 		if !r.Ok() {
 			t.Error("expected pin to be ok")
 		}
@@ -208,7 +211,7 @@ func (tp *TestSuite) TestPinRecursive(t *testing.T) {
 		t.Errorf("unexpected verify result count: %d", n)
 	}
 
-	//TODO: figure out a way to test verify without touching IpfsNode
+	// TODO: figure out a way to test verify without touching IpfsNode
 	/*
 		err = api.Block().Rm(ctx, p0, opt.Block.Force(true))
 		if err != nil {
--- a/coreiface/tests/pin.go
+++ b/coreiface/tests/pin.go
@@ -198,8 +198,8 @@ func (tp *TestSuite) TestPinRecursive(t *testing.T) {
 	}
 	n := 0
 	for r := range res {
-		if r.Err() != nil {
-			t.Error(r.Err())
+		if err := r.Err(); err != nil {
+			t.Error(err)
 		}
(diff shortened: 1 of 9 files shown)


BigLep:
@MichaelMure : I know this has been open for awhile.  Apologies.  Maintainers spoke on 2023-05-25.  We'll make sure this goes out with Kubo 0.21.

Slack context: https://filecoinproject.slack.com/archives/C03L0G3B4RX/p1684507196595209

coreiface/tests/pin.go
```diff
@@ -198,6 +198,9 @@ func (tp *TestSuite) TestPinRecursive(t *testing.T) {
 	}
 	n := 0
 	for r := range res {
+		if r.Err() != nil {
+			t.Error(r.Err())
+		}
```
Jorropo:
```suggestion
		if err := r.Err(); err != nil {
			t.Error(err)
		}
```

pinning/pinner/dspinner/pin.go
```diff
@@ -20,6 +18,9 @@ import (
 	"github.com/polydawn/refmt/cbor"
 	"github.com/polydawn/refmt/obj/atlas"
 
+	"github.com/ipfs/boxo/ipld/merkledag"
+	"github.com/ipfs/boxo/ipld/merkledag/dagutils"
+
```
Jorropo:
```suggestion
```

pinning/pinner/dspinner/pin_test.go
```diff
@@ -199,10 +200,21 @@ func TestPinnerBasic(t *testing.T) {
...
 
-	cids, err := p.RecursiveKeys(ctx)
-	if err != nil {
-		t.Fatal(err)
+	allCids := func(ch <-chan ipfspin.StreamedCid) (cids []cid.Cid) {
+		for {
+			val, ok := <-ch
+			if !ok {
+				break
+			}
```
Jorropo:
```suggestion
		for val := range ch {
```

Jorropo (changes requested):
I'll do the rebase on `main`, LGTM.

Jorropo (approved):
LGTM I'll test on Kubo

hacdias:
@Jorropo what's the status?

PR #292, opened 2023-05-09
feat: reusable ipns verify

laurentsenta:
Sister PR in kubo: https://github.com/ipfs/kubo/pull/9867

Extracts the ipns verification code out of kubo to boxo. This makes the verification code reusable, for example, with gateway-conformance testing (https://github.com/ipfs/gateway-conformance/pull/37)

Relates to https://github.com/ipfs/specs/issues/376

- [ ] ⚠️ update with the correct boxo commit when the sister PR is merged

### DIFF
--- a/gateway/handler_ipns_record.go
+++ b/gateway/handler_ipns_record.go
@@ -10,9 +10,8 @@ import (
 	"time"
 
 	"github.com/cespare/xxhash/v2"
-	"github.com/gogo/protobuf/proto"
 	ipath "github.com/ipfs/boxo/coreiface/path"
-	ipns_pb "github.com/ipfs/boxo/ipns/pb"
+	"github.com/ipfs/boxo/ipns"
 	"github.com/ipfs/go-cid"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -50,8 +49,7 @@ func (i *handler) serveIpnsRecord(ctx context.Context, w http.ResponseWriter, r 
 		return false
 	}
 
-	var record ipns_pb.IpnsEntry
-	err = proto.Unmarshal(rawRecord, &record)
+	record, err := ipns.UnmarshalIpnsEntry(rawRecord)
 	if err != nil {
 		webError(w, err, http.StatusInternalServerError)
 		return false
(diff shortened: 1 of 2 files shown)


hacdias (changes requested):
I think this is reasonable. This package will likely be refactored at some point in the near future when the lean records get more attention: https://github.com/ipfs/specs/issues/376

Small ask: can we change the following lines to use `UnmarshalIpnsEntry`?

https://github.com/ipfs/boxo/blob/8059f183d86648f69ad562546b4797bc148c0984/gateway/handler_ipns_record.go#L53-L54

hacdias (approved):

hacdias (approved):
As I stated previously, I think this is a reasonable change. It will likely be [refactored](https://github.com/ipfs/specs/issues/376) until the end of the year, but it's good to keep this functionality together.

hacdias:
Sharness failing because something changed in Kubo. We need to update Boxo there first, then we can merge this: https://github.com/ipfs/kubo/pull/9868

Jorropo:
This makes sense but I am suprised that this is being merged with 0 coverage in `entries.go`.

hacdias:
@Jorropo indeed, I was thinking of reverting this.

hacdias:
@laurentsenta I merged this too impulsively, reverted in #293. I only realised how coupled all the names were to Kubo's `name inspect` functionality once I looked more carefully at https://github.com/ipfs/kubo/pull/9867. Do you mind opening this again with this change. Here's some notes:

## `Verify`

1. Rename to `ValidateWithPeerID`
2. Change signature to `func ValidateWithPeerID(id peer.ID, entry *ipns_pb.IpnsEntry) error`
3. Remove `IpnsInspectValidation`. That should stay in Kubo's `name inspect` logic.
4. Move it above `Validate`
5. Add tests

## `IpnsInspectEntry` and `InspectIpnsRecord` 

Only include these if there's a clear benefit. I do think that they can be useful since the Protobuf type is definitely not the most user-friendly to make operations. However, it needs to make sense. The name of the types are coupled to the command `ipfs name inspect`. Also, the types of the different fields were also chosen according to what we needed in the command. They may not make sense for a generalisation.

---

`UnmarshalIpnsEntry` should also stay.

PR #293, opened 2023-05-10
Revert "feat: reusable ipns verify"

hacdias:
Reverts ipfs/boxo#292.

Let's add some tests and improve the interface. The names are too coupled with `ipfs name inspect` command and may not make sense in Boxo. I didn't notice this when I was going over the Boxo PR, only later when looking at https://github.com/ipfs/kubo/pull/9867


Jorropo (approved):
If the interface needs to be changed a revert sounds good. I havn't checked the content but if it's indeed `git revert` produced, LGTM.

PR #294, opened 2023-05-10
feat(ipns): helper ValidateWithPeerID and UnmarshalIpnsEntry

hacdias:
After my overly excited merge and revert of #292, I looked at the code and saw we were already duplicating existing code. E.g. the key extraction logic already existed and was being duplicated. This PR adds two helper functions `ValidateWithPeerID` and `UnmarshalIpnsEntry`. They already use existing functionality. I also added tests.

I did not include `IpnsInspectEntry` for the reasons here: https://github.com/ipfs/boxo/pull/292#issuecomment-1541581206 - it doesn't make sense for us to be adding it here since it is tightly related to the type we use in `ipfs name inspect`. We only created it in first place to simplify the type used by the RPC client in Kubo.

I also enabled a test that has been disabled for 5 years in 032060566e194f69bb3fd22c6d238eed9a10dcf6. It seems to pass now, and reading the commit description, I don't think it is a problem due to the way Peer IDs are encoded nowadays.

Tested in Kubo here: https://github.com/ipfs/kubo/pull/9867

### DIFF
--- a/ipns/validate_test.go
+++ b/ipns/validate_test.go
@@ -19,6 +19,7 @@ import (
 	pstore "github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoremem"
 	"github.com/multiformats/go-multicodec"
+	"github.com/stretchr/testify/assert"
 )
 
 func testValidatorCase(t *testing.T, priv crypto.PrivKey, kbook pstore.KeyBook, key string, val []byte, eol time.Time, exp error) {
@@ -143,8 +144,6 @@ func TestEmbeddedPubKeyValidate(t *testing.T) {
 }
 
 func TestPeerIDPubKeyValidate(t *testing.T) {
-	t.Skip("disabled until libp2p/go-libp2p-crypto#51 is fixed")
-
 	goodeol := time.Now().Add(time.Hour)
 	kbook, err := pstoremem.NewPeerstore()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+	pid, err := peer.IDFromPublicKey(pk)
+	assert.Nil(t, err)
+
+	entry, err := Create(sk, path, 1, eol, 0)
+	assert.Nil(t, err)
+
+	t.Run("valid peer ID", func(t *testing.T) {
+		t.Parallel()
+		err = ValidateWithPeerID(pid, entry)
+		assert.Nil(t, err)
```
Jorropo:
```suggestion
		assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+	rnd := rand.New(rand.NewSource(42))
+
+	sk, pk, err := crypto.GenerateEd25519Key(rnd)
+	assert.Nil(t, err)
+
+	pid, err := peer.IDFromPublicKey(pk)
+	assert.Nil(t, err)
+
+	entry, err := Create(sk, path, 1, eol, 0)
+	assert.Nil(t, err)
```
Jorropo:
```suggestion
	assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+	path := []byte("/ipfs/bafkreifjjcie6lypi6ny7amxnfftagclbuxndqonfipmb64f2km2devei4")
+	eol := time.Now().Add(time.Hour)
+
+	rnd := rand.New(rand.NewSource(42))
+
+	sk, pk, err := crypto.GenerateEd25519Key(rnd)
+	assert.Nil(t, err)
+
+	pid, err := peer.IDFromPublicKey(pk)
+	assert.Nil(t, err)
```
Jorropo:
```suggestion
	assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
 }
+
+func TestValidateWithPeerID(t *testing.T) {
+	path := []byte("/ipfs/bafkreifjjcie6lypi6ny7amxnfftagclbuxndqonfipmb64f2km2devei4")
+	eol := time.Now().Add(time.Hour)
+
+	rnd := rand.New(rand.NewSource(42))
+
+	sk, pk, err := crypto.GenerateEd25519Key(rnd)
+	assert.Nil(t, err)
```
Jorropo:
```suggestion
	assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+		t.Parallel()
+		err = ValidateWithPeerID(pid, entry)
+		assert.Nil(t, err)
+	})
+
+	t.Run("invalid peer ID", func(t *testing.T) {
+		t.Parallel()
+
+		_, pk2, err := crypto.GenerateEd25519Key(rnd)
+		assert.Nil(t, err)
```
Jorropo:
```suggestion
		assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+
+		pid2, err := peer.IDFromPublicKey(pk2)
+		assert.Nil(t, err)
```
Jorropo:
```suggestion
		assert.NoError(t, err)
```

ipns/validate_test.go
```diff
@@ -413,3 +412,38 @@ func genKeys(t *testing.T) (crypto.PrivKey, peer.ID, string) {
...
+
+		err = ValidateWithPeerID(pid2, entry)
+		assert.NotNil(t, err)
```
Jorropo: Please check the actual error returned (I guess this error type or var is public and we can `errors.Is` it).

Jorropo (changes requested):
LGTM once thoses minors things are fixed.

Jorropo (approved):

PR #297, opened 2023-05-11
feat: add deprecator

guseggert:
This utility adds deprecation comments to all exported types in a given module. It's intended to be used by Boxo maintainers when moving repos into Boxo.

For example, if your module is "github.com/ipfs/go-unixfs", then checkout that repo and, in its root, run:

``` 
deprecator --path github.com/ipfs/boxo/ipld/unixfs
``` 

Then on the func "github.com/ipfs/go-unixfs/mod.NewDagModifier", the deprecator will add the following comment:

``` 
// Deprecated: use github.com/ipfs/boxo/ipld/unixfs/mod.NewDagModifier
```


cmd/deprecator/main.go
```diff
@@ -0,0 +1,183 @@
...
+	"strings"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+	app := &cli.App{
+		Name:  "migrate",
```
hacdias:
```suggestion
		Name:  "deprecator",
```

hacdias (approved):

PR #298, opened 2023-05-12
feat(gateway): improved templates, user friendly errors

### LINKED ISSUE #262: gateway: return user-friendly HTML with HTTP errors
Currently, we return `text/plain` with context of `err.Error()` of some sort. This if fine for `curl`, but very user hostile in web browser context.

### basic human needs (#298)

- [ ] add HTML template for errors to `/gateway/assets`
- [ ] detect when request had `User-Agent` header starting with `Mozilla`  and return `text/html` responses for errors instead (same message, but presented nicely)
- [ ] short sentence: inform user what IPFS is, what gateway is, what content routing is and what failed based on the error code
- [ ] include links to project, docs, and discussion forums (explicitly link to [help forum category](https://discuss.ipfs.tech/c/help/13), so people can find common answer/problem threads)
- [ ] Include basic pointers on how to diagnose 4XX 5XX errors
  - [ ] suggest testing with https://docs.ipfs.tech/reference/diagnostic-tools/
  - [ ] suggest self-hosting, and running client that verifies data (https://docs.ipfs.io/install/command-line/, https://github.com/ipfs/ipfs-desktop, https://brave.com/ipfs-support/)
  - [ ] if 400 or 404, suggest inspecting CID (https://cid.ipfs.tech/) or DAG (http://explore.ipld.io) (or installing IPFS Desktop)

### nice to have 

below could be a follow-up PR. may not be a good idea to hardcode another HTTP endpoint and use of https://pl-diagnose.onrender.com for every gateway deployment, perhaps better to wait for https://github.com/ipfs/in-web-browsers/issues/207 and webtransport transport rollout. or we could use https://pl-diagnose.onrender.com and https://cid.contact/ as default, with config options to change/disable them.

- [ ] built-in self-service when content routing failed
  - [ ] present a button that runs DHT/IPNI checks against failed CID (https://pl-diagnose.on.fleek.co/#/diagnose/access-content, https://cid.contact/)
  - [ ] present field and a button to tests if CID can be fetched from specific Peer (https://check.ipfs.network/)

hacdias:
Closes #262. This PR does the following:

- Adds a new human-friendly error page when something wrong goes on.
- Reworks templates a bit, making the process to develop faster and easier.
	- Common header file.
	- No more bash scripts or `go generate`.
	- Cleaned up CSS.
	- Removed `twitter` and `opengraph` meta tags and just use standard `<title>` and `<meta description`.
- Updated some file icons for the directory to be coherent.
- Removed "Preview as JSON" from DAG template. We removed that function in https://github.com/ipfs/kubo/pull/9566, but forgot to update the template.
	- Preview for DAG-JSON and DAG-CBOR in https://github.com/ipfs/boxo/pull/315.
- Removed hardcoded Kubo issues page and replaced by customisable menu items in the config. Each implementation can now add their own links that will show up to the right of "About" and "Install".

As usual, instructions on how to test the templates are in `gateway/assets/README.md`.

image

image


hacdias:
I'm opening this for review. Please give me feedback regarding error messages and other things.

gateway/assets/error.html
```diff
@@ -0,0 +1,86 @@
...
+<body>
+  <div id="page-header">
+  <div id="page-header-logo" class="ipfs-logo">&nbsp;</div>
+  <div id="page-header-menu">
+    <div class="menu-item-wide"><a href="https://ipfs.tech" target="_blank" rel="noopener noreferrer">About IPFS</a></div>
+    <div class="menu-item-wide"><a href="https://ipfs.tech#install" target="_blank" rel="noopener noreferrer">Install IPFS</a></div>
+    <div class="menu-item-narrow"><a href="https://ipfs.tech" target="_blank" rel="noopener noreferrer">About</a></div>
+    <div class="menu-item-narrow"><a href="https://ipfs.tech#install" target="_blank" rel="noopener noreferrer">Install</a></div>
+    <div>
+      <a href="https://github.com/ipfs/kubo/issues/new/choose" target="_blank" rel="noopener noreferrer" title="Report a bug">
```
BigLep: I don't think creating Kubo issues here makes sense.

gateway/assets/error.html
```diff
@@ -0,0 +1,86 @@
...
+    <div class="table-responsive">
+      <table>
+        <tr>
+          <td>
+            {{ if eq .StatusCode 400 }}
+              <p>Your request is invalid. Please check the error below for more information.</p>
+            {{ else if eq .StatusCode 404 }}
+              <p>The content path you requested cannot be found. There's likely an invalid or missing DAG node.</p>
+            {{ else if eq .StatusCode 410 }}
+              <p>The content path you tride to request is correct, but this gateway refuses to return the requested data.</p>
```
BigLep:
```suggestion
              <p>The content path you tried to request is correct, but this gateway refuses to return the requested data.</p>
```

gateway/assets/error.html
```diff
@@ -0,0 +1,86 @@
...
+            {{ else if eq .StatusCode 412 }}
+              <p>This gateway is unable to return the requested data under the conditions sent by the client.</p>
+            {{ else if eq .StatusCode 429 }}
+              <p>You have made too many requests. Please wait some time and try again.</p>
```
BigLep:
```suggestion
              <p>There are too many pending requests. Please wait some time and try again.</p>
```

It may not be the requester that has too many requests right?  The user could be getting throttled because the server has received too many requests in general right?

gateway/assets/error.html
```diff
@@ -0,0 +1,86 @@
...
+            <p><code>{{ .Error }}</code></p>
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <p>How you can proceed:
+              <ul>
+                <li>Check the <a href="https://discuss.ipfs.tech/c/help/13" rel="nofollow">Discussion Forums</a> for similar errors</li>
+                <li>Try diagnosing your request with the <a href="https://docs.ipfs.tech/reference/diagnostic-tools/" rel="nofollow">diagnostic tools</a></li>
+                <li>Self-host and run a client that verifies your data, such as <a href="https://docs.ipfs.io/install/command-line/" rel="nofollow">Kubo</a>, <a href="https://github.com/ipfs/ipfs-desktop" rel="nofollow">IPFS Desktop</a>, or <a href="https://brave.com/ipfs-support/" rel="nofollow">Brave Browser</a></li>
```
BigLep: Rather than being opinionated about an installation to use, maybe just link to https://docs.ipfs.tech/concepts/ipfs-implementations/ ?

gateway/assets/src/page-header.html
```diff
@@ -0,0 +1,14 @@
+<div id="page-header">
+  <div id="page-header-logo" class="ipfs-logo">&nbsp;</div>
+  <div id="page-header-menu">
+    <div class="menu-item-wide"><a href="https://ipfs.tech" target="_blank" rel="noopener noreferrer">About IPFS</a></div>
+    <div class="menu-item-wide"><a href="https://ipfs.tech#install" target="_blank" rel="noopener noreferrer">Install IPFS</a></div>
+    <div class="menu-item-narrow"><a href="https://ipfs.tech" target="_blank" rel="noopener noreferrer">About</a></div>
+    <div class="menu-item-narrow"><a href="https://ipfs.tech#install" target="_blank" rel="noopener noreferrer">Install</a></div>
+    <div>
+      <a href="https://github.com/ipfs/kubo/issues/new/choose" target="_blank" rel="noopener noreferrer" title="Report a bug">
```
BigLep: I would remove the kubo issue opening from here.  It's not high signal that a Kubo issue should be opened (especially since this is in Boxo).

gateway/assets/error.html
```diff
@@ -0,0 +1,86 @@
...
+  <div id="content">
+    <div id="content-header">
+      <strong>{{ .StatusCode }} {{ .StatusText }}</strong>
+    </div>
+
+    <div class="table-responsive">
+      <table>
+        <tr>
+          <td>
+            {{ if eq .StatusCode 400 }}
```
BigLep: Maybe useful to also print out the HTTP status code given those have meaning in their own right?

Is it documented what "context" a template creator will have passed in (e.g., .StatusCode, .StatusText)?

BigLep (commented):
Thanks for working on this.  A few things:

1. This needs a changelog entry.
2. How does someone customize this for their own needs?   3 immediate examples I see:
- (most important) how does ipfs.io gateway customize this for their product
- (maybe) Kubo default (which can encourage someone to install Kubo for themselves)
- (likely not as important) bifrost-gateway default

gateway/assets/error.html
lidel: I feel that docs page is not the best from the perspective of a regular user reading this in a web browser.

Things in "Popular" are all backend daemons, normal people will just bounce off that website, and we lose potential users who could do just fine with IPFS Desktop or Brave.

Maybe add  "Looking for User Friendly Implementation?" section at the top with IPFS Desktop+IPFS Companion and  [Brave](https://brave.com/ipfs-support/)  options?

gateway/assets/src/page-header.html
lidel: +1 for removing this from default template. 

As part of customization, we could add it as opt-in via config (`Gateway.SupportURL`) and set it to on ipfs.io and dweb.link  to boxo repo with some gateway-specific template, but we should not expose it on third-party gateways, which may be misconfigured.

gateway/assets/error.html
lidel: Add `noopener noreferrer` everywhere, we don't want to leak referers of people's gateways to third-party websites like github or brave.

gateway/assets/src/error.html
```diff
@@ -0,0 +1,72 @@
...
+  <meta property="og:description" content="{{ .Path }}">
+  <meta property="og:type" content="website">
+  <meta property="og:image" content="https://gateway.ipfs.io/ipfs/QmSDeYAe9mga6NdTozAZuyGL3Q1XjsLtvX28XFxJH8oPjq">
+  <meta name="twitter:title" content="{{ .Path }}">
+  <meta name="twitter:description" content="A directory of files hosted on the distributed, decentralized web using IPFS">
+  <meta name="twitter:card" content="summary_large_image">
+  <meta name="twitter:image" content="https://gateway.ipfs.io/ipfs/QmSDeYAe9mga6NdTozAZuyGL3Q1XjsLtvX28XFxJH8oPjq">
+  <meta name="twitter:creator" content="@ipfs">
+  <meta name="twitter:site" content="@ipfs">
+  <meta name="image" content="https://gateway.ipfs.io/ipfs/QmSDeYAe9mga6NdTozAZuyGL3Q1XjsLtvX28XFxJH8oPjq">
```
lidel: Directory description and other social media items should be removed from error page, or customized to reflect something went wrong, so if someone pastes URL with truncated CID on twitter, they get useful error in  preview.

gateway/assets/src/error.html
```diff
@@ -0,0 +1,72 @@
...
+    <div class="table-responsive">
+      <table>
+        <tr>
+          <td>
+            {{ if eq .StatusCode 400 }}
+              <p>Your request is invalid. Please check the error below for more information.</p>
+            {{ else if eq .StatusCode 404 }}
+              <p>The content path you requested cannot be found. There's likely an invalid or missing DAG node.</p>
+            {{ else if eq .StatusCode 410 }}
+              <p>The content path you tride to request is correct, but this gateway refuses to return the requested data.</p>
```
lidel:
```suggestion
              <p>This gateway refuses to return the requested data.</p>
```

gateway/assets/src/error.html
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+
+    <div class="table-responsive">
+      <table>
+        <tr>
+          <td>
+            {{ if eq .StatusCode 400 }}
+              <p>Your request is invalid. Please check the error below for more information.</p>
+            {{ else if eq .StatusCode 404 }}
+              <p>The content path you requested cannot be found. There's likely an invalid or missing DAG node.</p>
```
lidel:
```suggestion
            {{ else if eq .StatusCode 404 }}
              <p>The content path you requested cannot be found. There's likely an invalid or missing DAG node.</p>
            {{ else if eq .StatusCode 406 }}
              <p>This gateway is unable to return the data in the format requested by the client.</p>
```

gateway/assets/src/error.html
```diff
@@ -0,0 +1,72 @@
...
+            {{ else if eq .StatusCode 412 }}
+              <p>This gateway is unable to return the requested data under the conditions sent by the client.</p>
+            {{ else if eq .StatusCode 429 }}
+              <p>You have made too many requests. Please wait some time and try again.</p>
+            {{ else if eq .StatusCode 451 }}
+              <p>This gateway is not allowed to return the requested data due to legal reasons.</p>
+            {{ else if eq .StatusCode 500 }}
+              <p>This gateway was unable to return the requested data due to an internal error. Please check the error below for more information.</p>
+            {{ else if eq .StatusCode 504 }}
+              <p>This gateway was unable to fullfil your request within internally defined time limits.</p>
```
lidel: These two will be constantly shown on Rhea / bifrost-gateway, so worth giving them useful messages, but at the same time we should not make then about Saturn:

- 502 is when IPNI (cid.contact) fails to find providers
- 504 is when Saturn L1  failed to retrieve data under the timeout 


Perhaps:
```suggestion
            {{ else if eq .StatusCode 502 }}
              <p>The gateway backed was unable to fullfil your request due to an error.</p>
            {{ else if eq .StatusCode 504 }}
              <p>The gateway backend was unable to fullfil your request due to a timeout.</p>
```

gateway/errors.go
```diff
@@ -153,7 +155,18 @@ func webError(w http.ResponseWriter, err error, defaultCode int) {
 		code = gwErr.StatusCode
 	}
 
-	http.Error(w, err.Error(), code)
+	acceptsHTML := strings.Contains(r.Header.Get("Accept"), "text/html")
+	if acceptsHTML {
+		w.WriteHeader(code)
+		_ = assets.ErrorTemplate.Execute(w, assets.ErrorTemplateData{
+			Path:       r.URL.Path,
```
lidel: :warning:  accessing URL path directly here may cause bugs – if not today, then in the future, after refactors (it happened already a while ago). 

Rule of thumb for `boxo/gateway` is to avoid `r.URL.Path` and only acting on already parsed `contentPath` -- this way there is no room for bugs caused by content path normalization (now or in future).

So.. we need to pass `contentPath ipath.Path` to  `webError` and use `contentPath.String()` here, just like we do in the dir listing template  in `handler_unixfs_dir.go`

lidel (changes requested):

gateway/assets/error.html
hacdias: @BigLep I think it can be both. 429 just means "Too Many Requests" - I'm not sure what we, as PL, are doing, but most APIs I've used throttle per user (also seems to be the usual meaning: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429). Do you know @lidel?

gateway/assets/error.html
hacdias: We do that above.

gateway/assets/error.html
hacdias: This is the header that is showed in all pages from the gateway. Not the error page itself. Please take a look at the `assets/src` files.

gateway/assets/src/page-header.html
hacdias: This is part of the common header between all pages. If "but we should not expose it on third-party gateways" is the reasoning, we should remove it from **everywhere** (directories, dag listings, errors). Is that the case?

gateway/assets/src/error.html
hacdias: I removed the social media from all the three pages for the following reasons:

1. The linked image gives `410 Gone` so it's not accessible.
2. Social media will fallback on default `<meta name=description` and `<title` tags, which are kept.
3. We probably don't want all the content shared by random people to be associated with `@ipfs` on Twitter. Who knows what people are sharing around

I updated the remaining pages' descriptions to better reflect what's happening.

gateway/errors.go
hacdias: We actually don't need it for this page. Removed. Besides, `webError` is called many times before `contentPath` even exists. If we want to pass it, we have to pass it around in many places we don't right now and see what to do when it doesn't exist yet. But I'll remove for now as we don't use it on the error page.

hacdias:
@BigLep 

> How does someone customize this for their own needs?

Very good question. And I think the answer depends on _what_ you want to customize. What do you need to customize? Right now none of the templates are or have been customizable (directory index, DAG index, and now error). I removed the Kubo hardcoded link and added customisable links for the menu.

gateway/assets/src/page-header.html
hacdias: I removed Kubo's URL and replaced it by a customisable `.SupportURL`. Whoever implements the gateway can now pass that information. If it is set, it will show up with the custom URL. If not, it simply doesn't show up.

gateway/assets/error.html
hacdias: @lidel  this suggestion is to update the Docs page or to update the error template?

hacdias: @lidel I assume that your idea is to update the "IPFS Implementations" page in [ipfs-docs](https://github.com/ipfs/ipfs-docs) to have a "User Friendly Implementations" section at the top. If so, I do agree with it. I updated the link on this PR to the docs. I can open a PR on the docs if I understood it correctly.

gateway/assets/error.html
lidel, line 1: Could we move this to `./out` and/or make sure it is always collapsed in review https://thoughtbot.com/blog/github-diff-supression ? (same for `./diretory-index.html` and `./dag-index.html`).


This way people will not comment on output, but focus on source.

hacdias: @lidel I moved them to `out` for now. I had also considered building them at upon initialisation (during `func init`) instead so we wouldn't need the extra build steps. What do you think?

hacdias: @lidel templates are now built at `init` time (very quickly as everything is in-memory anyways). This means that running `go generate` is no longer necessary either. Testing also becomes easier.

gateway/assets/error.html
hacdias: Opened issue: https://github.com/ipfs/ipfs-docs/issues/1606

gateway/assets/header.html
```diff
@@ -0,0 +1,12 @@
+<header id="header">
+  <div class="ipfs-logo">&nbsp;</div>
+  <nav>
+    <a href="https://ipfs.tech" target="_blank" rel="noopener noreferrer">About<span class="dn-mobile"> IPFS</span></a>
+    <a href="https://ipfs.tech#install" target="_blank" rel="noopener noreferrer">Install<span class="dn-mobile"> IPFS</span></a>
+    {{ with .SupportURL }}
+    <a href="{{ . }}" target="_blank" rel="noopener noreferrer" title="Support">
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18.4 21"><circle cx="7.5" cy="4.8" r="1"/><circle cx="11.1" cy="4.8" r="1"/><path d="M12.7 8.4c-0.5-1.5-1.9-2.5-3.5-2.5 -1.6 0-3 1-3.5 2.5H12.7z"/><path d="M8.5 9.7H5c-0.5 0.8-0.7 1.7-0.7 2.7 0 2.6 1.8 4.8 4.2 5.2V9.7z"/><path d="M13.4 9.7H9.9v7.9c2.4-0.4 4.2-2.5 4.2-5.2C14.1 11.4 13.9 10.5 13.4 9.7z"/><circle cx="15.7" cy="12.9" r="1"/><circle cx="15.1" cy="15.4" r="1"/><circle cx="15.3" cy="10.4" r="1"/><circle cx="2.7" cy="12.9" r="1"/><circle cx="3.3" cy="15.4" r="1"/><circle cx="3.1" cy="10.4" r="1"/></svg>
```
hacdias, line 8:
```suggestion
      Support
```

We may want to replace this 🐞 icon by a more generic message, such as "Support". Another possibility is to allow the implementer to add as many more menu buttons as they want. So each gateway using this library can add as many more things next to "About IPFS" and "Install IPFS" as they want. This would replace `SupportURL` by something like `[]MenuItem`. What do you think @lidel?

hacdias: Done in [ffe46ea](https://github.com/ipfs/boxo/pull/298/commits/ffe46ea5facad5751627b6b1c9f87288bcb9b3be).

gateway/assets/error.html
```diff
@@ -0,0 +1,54 @@
...
+        <p>You have made too many requests. Please wait some time and try again.</p>
+      {{ else if eq .StatusCode 451 }}
+        <p>This gateway is not allowed to return the requested data due to legal reasons.</p>
+      {{ else if eq .StatusCode 500 }}
+        <p>This gateway was unable to return the requested data due to an internal error. Please check the error below for more information.</p>
+      {{ else if eq .StatusCode 502 }}
+        <p>The gateway backed was unable to fullfil your request due to an error.</p>
+      {{ else if eq .StatusCode 504 }}
+        <p>The gateway backend was unable to fullfil your request due to a timeout.</p>
+      {{ end }}
```
lidel, line 40:
```suggestion
      {{ else if eq .StatusCode 506 }}
        <p>The gateway backend was unable to fullfil your request at this time. Try again later.</p>
      {{ end }}
```

gateway/assets/error.html
lidel: Semantically, 4XX are request (user) errors, so HTTP 429 signals per-user throttling.
For global ones one should 503 Service Unavailable, but for practical reasons we often don't do that in our infra, and always return 429. 

In practice, switching to what @BigLep suggested is safer (user may get this error from Saturn because of unrelated throttling)

gateway/assets/error.html
```diff
@@ -0,0 +1,54 @@
...
+      {{ else if eq .StatusCode 404 }}
+        <p>The content path you requested cannot be found. There's likely an invalid or missing DAG node.</p>
+      {{ else if eq .StatusCode 406 }}
+        <p>This gateway is unable to return the data in the format requested by the client.</p>
+      {{ else if eq .StatusCode 410 }}
+        <p>This gateway refuses to return the requested data.</p>
+      {{ else if eq .StatusCode 412 }}
+        <p>This gateway is unable to return the requested data under the conditions sent by the client.</p>
+      {{ else if eq .StatusCode 429 }}
+        <p>You have made too many requests. Please wait some time and try again.</p>
```
lidel:
```suggestion
        <p>There are too many pending requests. Please wait some time and try again.</p>
```

gateway/assets/README.md
```diff
@@ -1,27 +1,20 @@
...
-> DAG and Directory HTML for HTTP gateway
+> HTTP Gateway Templates.
 
 ## Updating
 
-When making updates to the templates, please note the following:
-
-1. Make your changes to the (human-friendly) source documents in the `src` directory.
-2. Before testing or releasing, go to `assets/` and run `go generate .`.
+To update the templates, make changes to the files in `html/` and `css/`.
```
lidel: I dont see these dirs, needs updating?

gateway/assets/dag.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <meta name="description" content="A DAG hosted on IPFS: {{ .Path }}.">
```
lidel: `IPFS: /ipfs` will look redundant, maybe:

```suggestion
  <meta name="description" content="Content-addressed {{.CodecName}} document hosted on IPFS.">
```

lidel (approved):
@hacdias thank you for cleaning this up :pray: 

Mind resolving conflicts and creating a PR draft in Kubo before merging?
If Kubo is green, please merge and rebase #307 so we can continue  there :pray:

hacdias:
Checks are green in Kubo: https://github.com/ipfs/kubo/pull/9904. Merging.

PR #299, opened 2023-05-12
feat: pass records limit on routing.FindProviders

hacdias:
While working on https://github.com/ipfs/kubo/pull/9877, I noticed that the `ContentRouter` interface required in order to implement the Routing V1 server does not provide enough information to make an education decision about how many providers to return.

This did not use to be an issue because the API was non-streamed so we could hardcode a value on the implementation (or timeout). Now that the server also supports streaming requests, it would be great to be able to differentiate both.

After talking to @lidel, we talked about adding a `count int` parameter in the `FindProviders`. By default, `count` will be 20 for non-streamable requests and 0 (unbounded) for streaming requests. ~However, I find that that introduces unnecessary complexity here. So I just added a `stream bool` parameter to `FindProviders` instead, and we let the implementer of `ContentRouter` decide what to do with it.~ See https://github.com/ipfs/boxo/pull/299#discussion_r1203134561.

Note that existing tests that I updated test this too.

### DIFF
--- a/routing/http/server/server.go
+++ b/routing/http/server/server.go
@@ -28,6 +28,9 @@ const (
 	mediaTypeJSON     = "application/json"
 	mediaTypeNDJSON   = "application/x-ndjson"
 	mediaTypeWildcard = "*/*"
+
+	DefaultRecordsLimit          = 20
+	DefaultStreamingRecordsLimit = 0
 )
 
 var logger = logging.Logger("service/server/delegatedrouting")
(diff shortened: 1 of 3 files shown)


routing/http/server/server.go
```diff
@@ -41,7 +41,9 @@ type FindProvidersAsyncResponse struct {
 }
 
 type ContentRouter interface {
-	FindProviders(ctx context.Context, key cid.Cid) (iter.ResultIter[types.ProviderResponse], error)
+	// FindProviders searches for peers who are able to provide a given key. Stream
+	// indicates whether or not this request will be responded as a stream.
+	FindProviders(ctx context.Context, key cid.Cid, stream bool) (iter.ResultIter[types.ProviderResponse], error)
```
lidel: Not familiar with straming plans Gus/Antonio had here, but I have a gut feeling it may be safer if we _also_ have a `limit`, [like the one in libp2p's routing interface here](https://github.com/libp2p/go-libp2p/blob/01651d206f2e17c1c009d578baa48910e5d09051/core/routing/routing.go#L32-L36).
Streaming would have unbouded 0, non-streaming would have 20 (same as `ipfs routing findprovs` atm)

This way we avoid breaking go API if we ever want to add support for streaming with  `/routing/v1/providers/{{cid}}?limit=100`  exposed in Kubo ([code here](https://github.com/ipfs/kubo/pull/9877/files#diff-8de01cacf78a664b433563e3cafd0bc509f61b74d7b7d36a713778d451da8c92R36))

hacdias: I'm also not sure what the plans were. I changed the parameter (see my commit). I also added options such that the implementer is free to choose which value will be passed. Defaults are in constants. I did not want to force users of this server to be forced to have a fixed limit that they cannot change themselves.

lidel (approved):
Configurable limit of returned responses  lgtm, makes it more useful as a library, thanks!

PR #300, opened 2023-05-15
fix(routing/v1): add newline in NDJSON responses

hacdias:
The **ND** in NDJSON was missing.


routing/http/server/server.go
```diff
@@ -254,6 +254,13 @@ func (s *server) findProvidersNDJSON(w http.ResponseWriter, provIter iter.Result
 			logger.Warn("FindProviders ndjson write error", "Error", err)
 			return
 		}
+
+		_, err = w.Write([]byte("\n"))
```
Jorropo:
```suggestion
		_, err = w.Write([]byte{'\n'})
```

lidel (approved):

PR #301, opened 2023-05-15
add test cases for tar module

fulldecent:


Jorropo (approved):
Thx a lot, more tests is always nice.

Just so you know we want to discontinue this package, users who want a similar feature would be better served by a content aware unixfs chunker as this code and format (ipld tar) has some pitfalls.

Jorropo:
Fuck I wanted to enable automerge not merge.
I'm surprised this was disabled.

I'll submit a revert if this breaks CI on master.

PR #303, opened 2023-05-16
feat(gateway): implement IPIP-402 extensions for gateway CAR requests

### LINKED ISSUE #283: Boxo gateway should support IPIP-402 parameters
# What

Feature request. Boxo's gateway module should be updated to support the parameters and behavior described in https://github.com/ipfs/specs/pull/402.

Specifically, for CAR requests, the gateway should:
- [ ] return intermediate blocks for the path when the request is for a cid+path
- [ ] support the `car-scope` parameter (including `block`, `file`, `all`)
- [ ] support the `bytes` parameter (lower priority -- not needed for the use case below)

# Why
1. Lassie would like to retrieve from Boost over HTTP. Boost supplies an HTTP gateway by incorporating Boxo's gateway module. For Lassie to retrieve from Boost trustlessly, and serve all of the neccesary requests, the gateway must support the changes in [IPIP-402](https://github.com/ipfs/specs/pull/402), or at least everything except the `bytes` parameter.  While we can certainly redirect to a shim for CAR requests, it seems like that would be very temporary solution and it would be better to just do it in Boxo.
2. [CAR-based improvements to bifrost-gateway](https://github.com/ipfs/bifrost-gateway/issues/62) will be helped having a stable implementation to make calls against so can iterate against.  Testing directly against not-yet-verified-if-spec-compliant Saturn adds in variables.
3. Implementations of IPIP-402 (e.g., Boxo, Kubo, bifrost-gateway, Lassie, Boost) want/need conformance tests.  Given EngRes IP Stewards are driving the conformance tests, it's good to have a place they can quickly iterate on for validating the conformance tests.
4. Merging https://github.com/ipfs/specs/pull/402 requires having an accompanying client/server implementation to validate the work.  Boxo is a natural place to do this work.

aschmahmann:
Fixes #283.

- [x] Requires https://github.com/ipfs/go-unixfsnode/pull/50
- [ ] IPIP: https://github.com/ipfs/specs/pull/402
- [x] Kubo PR: https://github.com/ipfs/kubo/pull/9914
- [x] Conformance Tests: https://github.com/ipfs/gateway-conformance/pull/56


gateway/gateway.go
```diff
@@ -131,7 +159,7 @@ type IPFSBackend interface {
 	// that may contain a single error for if any errors occur during the streaming. If there was an initial error the
 	// error channel is nil
 	// TODO: Make this function signature better
-	GetCAR(context.Context, ImmutablePath) (ContentPathMetadata, io.ReadCloser, <-chan error, error)
+	GetCAR(context.Context, ImmutablePath, CarParams) (io.ReadCloser, <-chan error, error)
```
aschmahmann: This is an interface breaking change which was needed to accept the CAR params anyway given that we likely don't want to add any more confusion to the different "path" types until we can unify them better.

I also dropped the return of `ContentMetadataPath` since we now return all the blocks in the path instead of skipping the path blocks.

TBD: While we're breaking this anyway we should probably better handle the error signatures here. Suggestions welcome.

gateway/blocks_gateway.go
```diff
@@ -223,38 +231,208 @@ func (api *BlocksGateway) Head(ctx context.Context, path ImmutablePath) (Content
 	return md, fileNode, nil
 }
 
-func (api *BlocksGateway) GetCAR(ctx context.Context, path ImmutablePath) (ContentPathMetadata, io.ReadCloser, <-chan error, error) {
-	// Same go-car settings as dag.export command
-	store := dagStore{api: api, ctx: ctx}
+func (api *BlocksGateway) GetCAR(ctx context.Context, p ImmutablePath, params CarParams) (io.ReadCloser, <-chan error, error) {
+	contentPathStr := p.String()
+	if contentPathStr[0:6] != "/ipfs/" {
```
Jorropo:
```suggestion
	if !strings.HasPrefix(contentPathStr, "/ipfs/") {
```
This never risk a panic if we somehow get a path that is too short.

gateway/blocks_gateway.go
```diff
@@ -223,38 +231,208 @@ func (api *BlocksGateway) Head(ctx context.Context, path ImmutablePath) (Content
...
+	}
+	firstSegment, _, _ := strings.Cut(contentPathStr[6:], "/")
+	rootCid, err := cid.Decode(firstSegment)
+	if err != nil {
+		return nil, nil, err
+	}
 
-	// TODO: When switching to exposing path blocks we'll want to add these as well
-	roots, lastSeg, err := api.getPathRoots(ctx, path)
+	r, w := io.Pipe()
```
Jorropo: I know this isn't new but this is meh, I think this should be refactored as `WriteCar` and take an `io.Writer` as input.

Feel free to ignore this comment.

gateway/blocks_gateway.go
aschmahmann: Sure, although IIRC this shouldn't happen anyway and even if it did any sane gateway implementation would recover.

gateway/blocks_gateway.go
```diff
@@ -223,38 +231,208 @@ func (api *BlocksGateway) Head(ctx context.Context, path ImmutablePath) (Content
...
-	r, w := io.Pipe()
+	lsys := cidlink.DefaultLinkSystem()
+	unixfsnode.AddUnixFSReificationToLinkSystem(&lsys)
 
 	errCh := make(chan error, 1)
 	go func() {
-		carWriteErr := c.Write(w)
+		// TODO: support selectors passed as request param: https://github.com/ipfs/kubo/issues/8769
+		// TODO: this is very slow if blocks are remote due to linear traversal. Do we need deterministic traversals here?
+		carWriteErr := walkGatewaySimpleSelector(ctx, ipfspath.Path(contentPathStr), params, &lsys, pathResolver)
```
aschmahmann: In case anyone is wondering why IPLD selectors were not used instead... I tried (including looking at the code from [Lassie](https://github.com/filecoin-project/lassie/blob/e5fd951d00810c93700bdf9a5a3a27568a789e26/pkg/types/types.go#L298-L310) that implements some of this). Unfortunately, there is no way to get IPLD selectors to properly work on /ipfs/bafy-unix-fs-root?format=car&dag-scope=entity&entity-bytes=0-1023 when it is not known ahead of time if the entity is a file or directory.

gateway/blocks_gateway.go
```diff
@@ -223,38 +231,215 @@ func (api *BlocksGateway) Head(ctx context.Context, path ImmutablePath) (Content
...
+
+		blockGetter := merkledag.NewDAGService(api.blockService).Session(ctx)
+
+		blockGetter = &nodeGetterToCarExporer{
+			ng: blockGetter,
+			cw: cw,
+		}
+
+		// Setup the UnixFS resolver.
+		f := newNodeGetterFetcherSingleUseFactory(ctx, blockGetter)
```
aschmahmann: This new struct is an unfortunate piece of glue. AFAIK the only fetcher implementation that exists is https://github.com/ipfs/go-fetcher/blob/99235cc120519afe2bddcb6b1f7a6844ecc3707f/impl/blockservice/fetcher.go#L32.

This is unfortunate as it means we can't use `ResolveToLastNode` anywhere which is needed because we have no mechanisms in go-ipld-prime/go-unixfsnode to figure out if the thing we got at the end of the path is UnixFS or not.

In short what we're dealing with is that the abstractions here make it easy to generically work with different types of data but if I want to figure out what the data is once we're in the abstraction (e.g. to check for HTTP Gateway API spec compliance) things become at least one of difficult or brittle.

Depending on feedback could either leave this code here, extract it to `go-fetcher` (which could also be copied into boxo), or extract the relevant code in the `resolver` package to not depend on `go-fetcher`

gateway/blocks_gateway.go
```diff
@@ -223,38 +231,215 @@ func (api *BlocksGateway) Head(ctx context.Context, path ImmutablePath) (Content
...
+	} else {
+		switch unixfsFieldData.FieldDataType().Int() {
+		case data.Data_Directory, data.Data_Symlink:
+			// These types are non-recursive so we're done
+			return nil
+		case data.Data_Raw, data.Data_Metadata:
+			// TODO: Is Data_Raw actually different from Data_File or is that a fiction?
+			// TODO: Deal with UnixFS Metadata being handled differently in boxo/ipld/unixfs than go-unixfsnode
+			// Not sure what to do here so just returning
+			return nil
```
aschmahmann: Some TODOs here. IIUC both of these are dealt with differently in go-unixfsnode vs boxo/ipld/unixfs. Need to figure out what to do here... and have some fixtures to illustrate.

@Jorropo @rvagg @ribasushi any insight you can give me that might save me time poking around here?

gateway/gateway.go
hacdias: @aschmahmann can we remove `<-chan error`, wrap `io.ReadCloser` and add the error there (on `.Close`  or `.Read`, if multiple errors, send a multierror)?

gateway/handler_car.go
```diff
@@ -5,14 +5,21 @@ import (
...
 	ipath "github.com/ipfs/boxo/coreiface/path"
+	"github.com/ipfs/go-cid"
+
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/multierr"
 )
 
+const carRangeBytesKey = "entity-bytes"
+const carTerminalElementTypeKey = "dag-scope"
```
hacdias:
```suggestion
const (
	carRangeBytesKey          = "entity-bytes"
	carTerminalElementTypeKey = "dag-scope"
)
```

nit

gateway/blocks_gateway.go
Jorropo: I have yet to see a usecase where `Raw` is different from `File`.
In my spec PR I say it's a file with a metadata like links that is hard coded to nothing and length that is derived from the block lengths.

gateway/handler.go
```diff
@@ -263,6 +263,7 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Detect when If-None-Match HTTP header allows returning HTTP 304 Not Modified
+	// TODO: Handle If-None-Match for CAR files once what goes in the ETag is resolved
```
aschmahmann: @lidel any thoughts on what to do here with CAR files and what the new etag should be (although if  you feel that's a separate flag then feel free to comment on the TODO line in handler_car)

gateway/gateway.go
hacdias: Ok, I checked the code more carefully. This should be doable and it makes sense, since the channel is for streaming errors. What's better for that the the Reader?

gateway/handler_car.go
```diff
@@ -55,6 +96,7 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
 	// Set Cache-Control (same logic as for a regular files)
 	addCacheControlHeaders(w, r, contentPath, rootCid)
```
aschmahmann: Perhaps related to what etag to use which we also need is what the name of the file in the attachment should be. Seems like it could be the root CID, or we could invent some syntax/transform to turn the request into a valid file name. Root cid certainly seems easier to implement although it makes it harder for the user to remember what was in the file (recall that the CAR header doesn't contain this information either).

Thoughts?

lidel: +1,  spec-wise,  it is fine to keep name in `Content-Disposition` simple (`cid.car`) -- it matters only in web browser, and if user wants to customize it, they can do it during Save As dialog or override via `filename` URL param.

If we want to be a bit more user-friendly for boxo/kubo users,  could tweak it  to have suffix `cid_[block|entity|all][_from-to].car` if `dag-scope` or `entity-bytes` are present

gateway/handler.go
lidel: The only real requirement for proper HTTP caching to work is for the`Etag` to be different and deterministic for each parameter permutation.

But it is a nice practice to use CID of the terminating element and append some suffix: 

```
Etag: W/"bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4.car.SUFFIX"
```


it allows people to purge cache keys starting with the same CID (I know Cloudflare does advanced cache eviction).

- We could append human-readable representation of values in `CarParams{dag-scope=entity;entity-bytes=from:to}.String()` → `entity.from.to` 
  - my vote would be on this, better UX if someone inspects headers, tries to debug
- or  do what we do  for [IPNS Record](https://github.com/ipfs/boxo/blob/9b8e63bf94a3f1cad759d3de942ffde77557bc54/gateway/handler_ipns_record.go#L62-L63)  and [unixfs dir listing HTML](https://github.com/ipfs/boxo/blob/9b8e63bf94a3f1cad759d3de942ffde77557bc54/gateway/handler_unixfs_dir.go#L223) ([generated here](https://github.com/ipfs/boxo/blob/9b8e63bf94a3f1cad759d3de942ffde77557bc54/gateway/assets/assets.go#L60)) and use  hash `xxhash(CarParams) → 7qi2369577vui`

gateway/handler_car.go
```diff
@@ -30,18 +37,52 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
+		}
+		params.Range = &rng
+	}
+
+	if hasScope {
+		switch s := DagScope(scopeStr); s {
+		case dagScopeEntity, dagScopeAll, dagScopeBlock:
+			params.Scope = s
+		default:
+			err := fmt.Errorf("unsupported dag-scope %s", scopeStr)
```
lidel: in case the client  got too creative
```suggestion
			err := fmt.Errorf("unsupported dag-scope %q", scopeStr)
```

gateway/blocks_gateway.go
rvagg: I was wondering exactly this while looking at the code; just to be clear—you're saying that the blocker here is the byte ranges? I haven't played with it myself yet (hasn't quite risen to our priority list in lassie) but it's supposed to be something we can do:

* https://github.com/ipld/ipld/pull/184
* https://github.com/ipld/go-ipld-prime/pull/375

There's an awful lot of code in here that's already handle by unixfsnode that would be good to offload. If there's things we can address to make this more usable then we should try.

Pinging @hannahhoward @willscott.

Here's a full form of the traverser and CAR generator that's used in some integration tests, for reference on how the whole thing can fit together, it doesn't have byte ranges though: https://github.com/filecoin-project/lassie/blob/b5aa6be20126af642eb7a86d96144e0f92dd2976/pkg/internal/itest/testpeer/generator.go#L366-L438

gateway/blocks_gateway.go
rvagg: unixfsnode handles raw like a file and metadata like a symlink: https://github.com/ipfs/go-unixfsnode/blob/f4d6da0dbdf3614269b3580eeebe2bb5245f09bc/reification.go#L54-L87

I think those are up for changing if the spec clarifies it differently since I don't believe they're particularly common and the kinds of changes we're likely to make make probably won't show up under common use of those code paths.

rvagg: like a symlink - essentially a unixfs pathable structure, i.e. directory

gateway/handler.go
rvagg: here's what we're currently doing, if it helps, etag as stringified CID: https://github.com/filecoin-project/lassie/blob/main/pkg/server/http/ipfs.go#L190-L199

gateway/handler_car.go
rvagg: what we're currently doing: `<root cid>.car` by default, also accepting a `filename` param to transform it https://github.com/filecoin-project/lassie/blob/main/pkg/server/http/ipfs.go#L149-L154

gateway/blocks_gateway.go
hacdias: Yeah, this would also simplify the [function signature](https://github.com/ipfs/boxo/pull/303#discussion_r1195179883). I would prefer to keep the `GetCAR` but I really don't like that we're returning a `chan error` and an `error`.

gateway/gateway.go
hacdias: Maybe we can somehow use pipe's `CloseWithError` for this?

hacdias: Yes, we can. Pipe's writer.Close never returns an error if we look at the source code. So using that is useless.

gateway/blocks_gateway.go
Jorropo: If we do my `WriteCar` this wouldn't need a chan error anymore.
You would synchronously write to the reader and simply return an error. This code wouldn't spawn a background goroutine.
And in the current code we can use `.CloseWithError` on the pipe instead of `<-chan error`.

gateway/gateway.go
hacdias: So, I've done some digging. Using `io.PipeWriter` is very nice for error handling! `io.PipeWriter.Close` calls `io.PipeWriter.CloseWithError(nil)` which always... returns `nil`: https://pkg.go.dev/io#PipeWriter.CloseWithError

We're checking for the error from `w.Close()` which is guaranteed to always be non-existing. We can do is call `CloseWithError` directly with the error we want, which will make subsequent reads from the pipe return the error we specified. Therefore removing the need for our wonky channel.

hacdias: @aschmahmann see my commit (37c40dd4d6b434561dfa5639e1f32de912823020). Feel free to discard it if it has problems.

gateway/blocks_gateway.go
hacdias: @Jorropo yeah, I already changed that.

Jorropo: I still think we should *consider* removing the go-routine and writing synchronously.

gateway/handler.go
rvagg: not sure why @lidel's comment is showing up now for me, after i posted mine; from an unposted review perhaps?

anyway, human readable etag suffix would be nice, and easier across limitations; it'd be nice to lock down the form in a spec and give ourselves room to manoeuvre when we add/remove parameters. Having optional pieces of a `.` separated string is probably not a good idea for scalability so default values for missing params would be good. Is `0` OK for both `from` and `to`?

aschmahmann: The proposal above handles params, but doesn't say anything about the pathing. Any thoughts on that?

Previously this was easy because it was just the whole graph under a root. Now it's the path to some terminal element and then the operations (e.g. dag-scope, entity-bytes) on that.

gateway/handler_car.go
aschmahmann: SGTM, will use the root CID (at the top of the path, as opposed to before where it was the root CID of the terminal element).

gateway/blocks_gateway.go
aschmahmann: > just to be clear—you're saying that the blocker here is the byte ranges

Not quite, there's a couple different issues exposed in the request below.

> /ipfs/bafy-unix-fs-root?format=car&dag-scope=entity&entity-bytes=0-1023 when it is not known ahead of time if the entity is a file or directory.

The first issue here is with what selector syntax supports. The request above means "if the data is a file give me the first 1024 bytes, **otherwise** give me the entire logical object (i.e. the directory whether sharded or not)".

The word **otherwise** is where selectors fall apart. The `OR` selector hasn't been defined or implemented yet so there's no way to express this as a single query where we try one, and if it fails try the other.

It's possible this request isn't an "optimal" one to use anyway given that without support for things like optionally fetching `index.html` in queries a request for `/ipfs/bafy-wikipedia-root/wiki?format=car&dag-scope=entity&entity-bytes=0-1023` will end up enumerating the entire sharded directory under `wiki` instead of just the first kiB of the index.html file. However, it's currently supported by the spec and works fine for smaller directories.

The second is that the `Slice` selector implementation in go-ipld-prime (not sure if this matches the spec or not) does not handle negative integers. I'm not sure if/how it handles "read until end of file" maybe just passing it a MaxInt64 will work though, I haven't checked.

> There's an awful lot of code in here that's already handle by unixfsnode that would be good to offload. If there's things we can address to make this more usable then we should try.

Happy to move whatever makes sense to go-unixfsnode or to expose this function more publicly if it's likely to be reused elsewhere. Let me know if you need anything from me.

If there's changes that make selectors work here that's fine too, that's how I started and only switched away to write more code because it didn't work out.

gateway/handler.go
lidel: The newly introduced pathing behavior (including parent nodes if requesting subpath) is not configurable (we are doing what we should be always doing), so it does not impact caching /  `Etag` going forward.


Implicit defaults like `0:*` range can be skipped, as the data will match vanilla `cid.car`

gateway/handler_car.go
lidel: Hm.. I mean, the root cid makes sense as filename when there is no meaningful path segment. 
In other cases, we do have a meaningful label in form of terminating path segment:

`Content-Disposition`:
- `/ipfs/cid → cid.car`
- `/ipfs/cid/cats → cats.car` (makes way more sense than parent root cid)

rvagg: Minor quibble with this argument is that it's `root`->`cats`, not just `cats`; it'll retain that in the roots of the CAR but it's not _quite_ `cats` and the `root` in this case could be lower, or even higher in the DAG that leads to `cats`, so it is a bit significant. `<wikipedia cid>/wiki/Cat` as `Cat.car` might be nice, but there's multiple `<wikipedia cid>` values that lead there, so it's really "this particular wikipedia root -> Cat".

Could we combine? `<root>_<last-segment>.car` So `bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze/wiki/Cat` would yield `bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze_Cat.car`.

gateway/blocks_gateway.go
lidel: `Data_Raw` is probably ok, but for `Data_Metadata`, returning a loud `HTTP 501 Not Implemented` error is safer for Gateway. Better to fail loudly, than to break users silently in the future, when we change undocumented behaviors. 

We play things safe already for things like IPLD Paths: [pathing into IPLD Kind other than Link (CID) is not supported on Gateways](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor).

We can revisit such 501 TODOs when we have time to finish UnixFS specs.

gateway/handler_car.go
lidel: Yes, it is sensible convention. 
Since we include all parent nodes, for deeper requests, let's include all parents:  `/ipfs/cid/wiki/Cat` → `cid_wiki_Cat.car`

gateway/handler_car.go
```diff
@@ -96,3 +140,39 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
 	i.carStreamGetMetric.WithLabelValues(contentPath.Namespace()).Observe(time.Since(begin).Seconds())
 	return true
 }
+
+func rangeStrToByteRange(rangeStr string) (DagEntityByteRange, error) {
+	rangeElems := strings.Split(rangeStr, ":")
+	if len(rangeElems) != 2 {
+		return DagEntityByteRange{}, fmt.Errorf("invalid range")
```
lidel:
```suggestion
		return DagEntityByteRange{}, fmt.Errorf("range must have two numbers separated with ':'")
```

gateway/handler_car.go
```diff
@@ -30,18 +39,52 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
-	pathMetadata, carFile, errCh, err := i.api.GetCAR(ctx, imPath)
+	queryParams := r.URL.Query()
+	rangeStr, hasRange := queryParams.Get(carRangeBytesKey), queryParams.Has(carRangeBytesKey)
+	scopeStr, hasScope := queryParams.Get(carTerminalElementTypeKey), queryParams.Has(carTerminalElementTypeKey)
+
+	params := CarParams{}
+	if hasRange {
+		rng, err := rangeStrToByteRange(rangeStr)
+		if err != nil {
+			webError(w, err, http.StatusBadRequest)
```
lidel:
```suggestion
            rangeErr := fmt.Errorf("invalid entity-bytes: %w", err)
			webError(w, rangeErr, http.StatusBadRequest)
```

gateway/handler_car.go
```diff
@@ -96,3 +140,39 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
+		}, nil
+	}
+
+	to, err := strconv.ParseInt(rangeElems[1], 10, 64)
+	if err != nil {
+		return DagEntityByteRange{}, err
+	}
+
+	if from >= 0 && to >= 0 && from > to {
+		return DagEntityByteRange{}, fmt.Errorf("cannot have a range where 'from' is after 'to'")
```
lidel:
```suggestion
		return DagEntityByteRange{}, fmt.Errorf("cannot have an entity-bytes range where 'from' is after 'to'")
```

gateway/handler_car.go
```diff
@@ -96,3 +140,39 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
+	}
+
+	if from < 0 && to < 0 && from > to {
+		return DagEntityByteRange{}, fmt.Errorf("cannot have a range where 'from' is after 'to'")
```
lidel:
```suggestion
		return DagEntityByteRange{}, fmt.Errorf("cannot have an entity-bytes range where 'from' is after 'to'")
```

gateway/handler_car.go
hacdias: @lidel considering file names (and file paths) have a maximum length size in most modern OSes, I think it's sensible to only use the last element as @rvagg mentioned. At least then we ensure users won't have issues downloading files deeply nested. I think that's an unlikely event, but it may happen. Another option would be to cut the filename after 256 characters, but that seems like complicating.

gateway/handler.go
hacdias: This has been addressed.

gateway/blocks_gateway.go
hacdias: Any update on this @rvagg @aschmahmann?

gateway/blocks_gateway.go
hacdias: I think for now we can return a 501 Not Implemented. However, I want to note that this could looks like something that could potentially be re-used and could potentially go elsewhere. In general, it would be great to keep code that interacts with CARs with UnixFS somewhere and well documented.

gateway/handler_car.go
hacdias: Updated as `{CID}[_{last segment}].car`.

gateway/blocks_gateway.go
hacdias: I was also [wondering the same](https://github.com/ipfs/boxo/pull/303#discussion_r1213007373). Anyone has ideas here?

hacdias: I was also [wondering the same](https://github.com/ipfs/boxo/pull/303#discussion_r1213007373). Anyone has ideas here?

gateway/blocks_gateway.go
rvagg: Last discussion was agreement that we can't do enough of it with go-ipld-prime, that there are some awkward corners. I think more could be less manual than what's done here but the functionality gaps are enough to spur additional work. For now this implementation will have to do I suppose.
Currently @hannahhoward and I are just figuring out who's going to free up first enough to dive in to making this properly work through selectors, or some other combination of pieces that makes this more straightforward, because we'd like it for Lassie too, or at least the CAR verifier we have that we want to reuse in a bunch of places that these CARs flow through.

gateway/handler_car.go
```diff
@@ -177,3 +176,37 @@ func rangeStrToByteRange(rangeStr string) (DagEntityByteRange, error) {
...
+	firstSegment, _, _ := strings.Cut(imPathStr[6:], "/")
+	rootCid, err := cid.Decode(firstSegment)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	return rootCid, err
+}
+
+func getCarEtag(r *http.Request, imPath ImmutablePath, params CarParams, rootCid cid.Cid) string {
```
rvagg: @hacdias @lidel did this get into a spec doc anywhere; or alternatively do you have tests that demonstrate what this output is going to be? I'm not entirely clear on what `ImmutablePath#String` does, but I think we end up with something like this from this function, would that be correct?

`"W/\"Qmfoobar.car." + xxhash("/ipfs/Qmfoobar/path/to/file.entity.100.400") + "\""`

Not strictly necessary to get this consistent between implementations I think, but if we end up with the same form then it might be quite useful. With strict ordering I believe we can ditch the `W/` and put in `.dfs` to the hash string.

Can you confirm whether `ImmutablePath#String` includes both the `/ipfs/` prefix and the CID? If so, could we leave that out and write the rest of this up into the path spec?

I recall having a discussion about this but can't find it so am not sure if it progressed!

hacdias: @rvagg afaik, the `Etag` is not going to be in the spec. It must be set, and as long as it's unique per request (path + arguments), it should be fine. We also don't have specs for other Etags.

Nevertheless, your assumption is correct.

rvagg: Here's my go at emulating this, adding `dups` too and removing the `W/` prefix: https://github.com/filecoin-project/lassie/pull/285 - could we encode this in the trustless and/or path spec as a recommendation and optionally some fixtures? There's a pretty important point about which bits are variable that cause different-byte output that might be worth making.

gateway/gateway.go
```diff
@@ -116,6 +116,34 @@ func (i ImmutablePath) IsValid() error {
...
+// From and To mostly follow [HTTP Byte Range] Request semantics.
+// From >= 0 and To = nil: Get the file (From, Length)
+// From >= 0 and To >= 0: Get the range (From, To)
+// From >= 0 and To <0: Get the range (From, Length - To)
+// From < 0 and To = nil: Get the file (Length - From, Length)
+// From < 0 and To >= 0: Get the range (Length - From, To)
+// From < 0 and To <0: Get the range (Length - From, Length - To)
+//
+// [HTTP Byte Range]: https://httpwg.org/specs/rfc9110.html#rfc.section.14.1.2
+type DagEntityByteRange struct {
```
hacdias, line 134: @aschmahmann why not reuse `ByteRange`? I see the `From` is `uint64` on `ByteRange`. Can we change that and unify both?

gateway/handler_car.go
lidel: We had no guidance for Etags because old CAR responses were unspecified, no byte-for-byte guaranteee, Etag always weak.

 @rvagg  I agree this changes with IPIP-412 (which adds optional `dups` and `order`) – weak/strong Etag recommendation and some fixtures should be added there, when we work on tests for 412 that. 
 
cc @hacdias, for when we switch focus to 412

gateway/blocks_gateway.go
aschmahmann: > I think for now we can return a 501 Not Implemented.

IIUC we can't do this in a meaningful way, we're sending out a CAR file stream so are back in lack of HTTP streaming errors/trailers land. 

So this seems to imply just returning (like we do with basic directories).

There is some difficulty here around upgrading in the future to support traversals where we didn't before and figuring out which gateways have upgraded vs not. Might just be a future problem though 🤷.

aschmahmann: > I have yet to see a usecase where Raw is different from File.

So this means we should recurse into both of them, right? @rvagg seems to also imply switching the `TRaw` preloader to be recursive in go-unixfsnode. Sound good?

Don't feel so strongly about what we do as long as we're all doing the same thing 😄

gateway/handler_car.go
```diff
@@ -96,3 +116,109 @@ func (i *handler) serveCAR(ctx context.Context, w http.ResponseWriter, r *http.R
...
+			return CarParams{}, err
+		}
+	} else {
+		params.Scope = DagScopeAll
+	}
+
+	return params, nil
+}
+
+func rangeStrToByteRange(rangeStr string) (DagByteRange, error) {
```
lidel: nit: move this to be  DagByteRange constructor?

gateway/handler_car_test.go
```diff
@@ -0,0 +1,96 @@
...
+				} else {
+					assert.Nil(t, params.Range.To)
+				}
+			}
+		}
+	})
+}
+
+func TestCarEtag(t *testing.T) {
+	t.Run("Etag with entity-bytes=0:* is the same as without query param", func(t *testing.T) {
```
lidel, line 88: can we add one with `dag-scope` and one with `entity-bytes=1:*` and confirm both are different than the vanilla one?

gateway/blocks_backend.go
```diff
@@ -0,0 +1,779 @@
...
+	}
+	firstSegment, _, _ := strings.Cut(contentPathStr[6:], "/")
+	rootCid, err := cid.Decode(firstSegment)
+	if err != nil {
+		return nil, err
+	}
+
+	r, w := io.Pipe()
+	go func() {
+		cw, err := storage.NewWritable(w, []cid.Cid{rootCid}, car.WriteAsCarV1(true))
```
lidel: TODO: either replace `rootCid` with the CID of final path segment or pass empty list (see https://github.com/ipfs/specs/pull/402#discussion_r1221517832).
The former is preferable, as it will keep the old behavior (before IPIP-402).

lidel (dismissed):

gateway/blocks_gateway.go
hacdias: I'm closing this for now, left a comment in-code. This can be handled later.

hacdias (approved):

hacdias:
Plan is:

1. @hacdias Merge this, update in Kubo.
2. @laurentsenta merge conformance, release conformance.
3. @hacdias release Boxo update.

BigLep:
2023-06-08 maintainer conversation: 
Expanding https://github.com/ipfs/boxo/pull/303#issuecomment-1582789926
Adin will handle the proactive communication to bedrock about what is happening.
@hacdias intends to do a 0.9.1 Boxo release on 2023-06-09 with this work included.

We'll let the [spec conversation around CAR roots](https://github.com/ipfs/specs/pull/402#pullrequestreview-1465206498) drive if there are any CAR spec changes and ipld/go-car updates required.

PR #308, opened 2023-05-18
bitswap/client: add basic traceable blocks

### LINKED ISSUE #209: Bitswap: better diagnostics and observability on client
When I call GetBlocks on the Bitswap client, I get back a channel and a synchronous error that is almost always nil. The primary way bitswap "errors" is by the context timing out. I don't know if we count find peers with the block, or we couldn't connect to them, or when we did we found they responded with don't have. So I have no way to track "what went wrong". I also don't have much of a way to track what went right -- if I get a block back, I don't know whom I got it from, the latency, the throughput, anything really.

This makes it very hard for consumers of the Bitswap library to measure performance or debug errors. Right now I am just turning on a bunch of debug logs, which is fine but means I can't do anything with those errors in code, or report them for metrics.

I realize as asynchronous code that ultimately talks to a pubsub protocol, returning errors or even channels of errors on GetBlocks may be hard. However, the client should at least be provided hooks so they can see more about why their requests are failing.

This hook seems like a good start: https://github.com/ipfs/go-libipfs/blob/main/bitswap/client/client.go#LL78C8-L78C8

For starters, we could have a hook that offered the same information, but also exposed Haves and Don't Haves.

Additionally though, it would be nice to be able to observe:
- FindProvider calls to the DHT and what they returned
- Message sending errors at the network level

Maybe such hooks exist somewhere? At the top level bitswap there is https://github.com/ipfs/go-libipfs/blob/main/bitswap/tracer/tracer.go#L10 but that seems not to exist for just the client? Even the tracer appears to only track successfully sent messages, not errors.

Jorropo:
Fixes #209

### DIFF
--- a/bitswap/client/traceability/block.go
+++ b/bitswap/client/traceability/block.go
@@ -7,7 +7,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
-// Block is a block whos provenance has been tracked.
+// Block is a block whose provenance has been tracked.
 type Block struct {
 	blocks.Block
 
--- a/bitswap/client/traceability/block.go
+++ b/bitswap/client/traceability/block.go
@@ -1,0 +1,21 @@
+package traceability
+
+import (
+	"time"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+// Block is a block whos provenance has been tracked.
+type Block struct {
+	blocks.Block
+
+	// From contains the peer id of the node who sent us the block.
+	// It will be the zero value if we did not downloaded this block from the
+	// network. (such as by getting the block from NotifyNewBlocks).
+	From peer.ID
+	// Delay contains how long did we had to wait between when we started being
+	// intrested and when we actually got the block.
+	Delay time.Duration
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The following emojis are used to highlight certain changes:
   * The only change to the default behavior on CAR responses is that we follow
     IPIP-412 and make `order=dfs;dups=n` explicit in the returned
     `Content-Type` HTTP header.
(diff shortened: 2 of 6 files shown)


Jorropo:
This needs tests !

BigLep:
Also a changelog update?

bitswap/client/traceability/block.go
```diff
@@ -0,0 +1,21 @@
+package traceability
+
+import (
+	"time"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+// Block is a block whos provenance has been tracked.
```
hacdias:
```suggestion
// Block is a block whose provenance has been tracked.
```

hacdias:
> This needs tests !

Any ETA to add the tests?

hannahhoward:
@Jorropo @hacdias I've rebased and added testing of the functionality. Note the tests are not actually running on CI cause of #327 . I have run them locally however and verified they pass.

hannahhoward:
also added Changelog, though I don't know if you all have a preferred format.

davidd8:
@Jorropo any updates on getting this PR merged? Based on @BigLep's comment in https://github.com/ipfs/boxo/issues/209, I assume this will happen this week by 2023-07-14, but correct me if I'm wrong.

hannahhoward:
Following up: @BigLep this still isn't merged. It's not clear what's blocking -- can we move this along?

Jorropo:
@hannahhoward my time is blocking, this is the next thing on the list https://github.com/orgs/ipfs/projects/16/views/17 and will get in for the next boxo release.

BTW This PR is not needed for the requirement that we passed in the slack thread, this was under nice to have.
The required per peer statistics are doable with the Tracer API. I'll implement an example of how you do that with peer-buckets in Kubo with prometheus.

BigLep:
I know there has been chatter about this in Slack.  This PR will get landed this week as we'll include it in Kubo 0.22, which we're pushing to get an RC out ASAP.

hacdias (approved):

PR #310, opened 2023-05-19
bitswap/client: fix PeerResponseTrackerProbabilityOneKnownOneUnknownPeer

Jorropo:
This was a bug in the implementation where it would incorrectly assume that 0 and 1 block received are equivalent scores.
I think this test used to not be flaky because prior to go1.20 `math/rand` would initialise to a zero state and the zero state would deterministically pass even tho this was 50/50 odds.

It also unmark all some tests in bitswap/client/internal/session non flaky since they are not flaky.

### DIFF
--- a/bitswap/client/internal/session/peerresponsetracker.go
+++ b/bitswap/client/internal/session/peerresponsetracker.go
@@ -57,14 +57,9 @@ func (prt *peerResponseTracker) choose(peers []peer.ID) peer.ID {
 }
 
 // getPeerCount returns the number of times the peer was first to send us a
-// block
+// block plus one (in order to never get a zero chance).
 func (prt *peerResponseTracker) getPeerCount(p peer.ID) int {
-	count, ok := prt.firstResponder[p]
-	if ok {
-		return count
-	}
-
 	// Make sure there is always at least a small chance a new peer
 	// will be chosen
-	return 1
+	return prt.firstResponder[p] + 1
 }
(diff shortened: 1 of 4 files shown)


Jorropo:
I'll open more PR making bitswap's code coverage better in the future, trying to take this as bite sized pieces.

Jorropo:
Turns out fixing tests in the session code is non trivial, I'll do that later.
I've unflakied trivial non timing based tests.

bitswap/client/internal/session/peerresponsetracker.go
```diff
@@ -57,14 +57,9 @@ func (prt *peerResponseTracker) choose(peers []peer.ID) peer.ID {
...
 func (prt *peerResponseTracker) getPeerCount(p peer.ID) int {
-	count, ok := prt.firstResponder[p]
-	if ok {
-		return count
-	}
-
 	// Make sure there is always at least a small chance a new peer
 	// will be chosen
-	return 1
+	return prt.firstResponder[p] + 1
```
hacdias, line 64: What if `p` is not in `firstResponder`? Or do we have guarantees that it is?

Jorropo: Reading a value that is missing from a map returns the zero of the type so we get `0 + 1`, this is done on purpose.

hacdias: Perfect, thought it'd panic.

Jorropo: @hacdias does that means I can merge ?

hacdias (approved):
@Jorropo the PR looks fine to me. I looked at the rest of the code where this is used and I think it's fine. Thanks!

Issue #288, opened 2023-05-02
Gateway metric tracking partial failures


willscott:
The gateway already has some tracking of whether each request finished [successfully](https://github.com/ipfs/boxo/blob/main/gateway/handler.go#L307)
It would be useful to get a metric tracking the amount of successful vs unsuccessful requests, labeled by the type of request (car vs block vs unixfs vs codec).

BigLep:
@lidel : this one missed our planning.  Are you aware of it?  Is this something we could squeeze in support of Rhea?

I agree we should emit a metric for each request and that each metric emitted should have dimensionality on:
* HTTP response code (particularly was it successful or not)
* request type (car, block, etc.)

lidel:
- we already have metrics for all these things, but response code is not enough for streaming responses (could fail mid-stream).
- we can add more resolution to count failures mid-CAR-stream, initial plan from  https://github.com/ipfs/bifrost-gateway/issues/104#issuecomment-1552072033 below, i'll try to prioritize it 

--- 

_(late on my end, but wanted to write this down, read with pinch of salt)_

afaik we have one place where client may get 200 and then  experience an interrupted stream: `?format=car` 
(problem discussed in length, multiiple ways of dealing with this, but [none of solutions got accepted so far](https://www.notion.so/pl-strflt/Streaming-CARs-bfa54196ed0740eea3c23dd70efbe07b))

we already count successful fetches in `carStreamGetMetric` (produces `ipfs_http_gw_car_stream_get_duration_seconds_count`).

 what we could do  is to  update Vec produced by  `newHistogramMetric` and add "status"  [here](https://github.com/ipfs/boxo/blob/4fb455dd307a85a3d6ac6579400b4f81b80b5b16/gateway/metrics.go#L253), and then update all metrics we have to explicitly update"success". 

then, we can count success and failure separately. in case of CAR, we would increase failure counter inside or [this block](https://github.com/ipfs/boxo/blob/4fb455dd307a85a3d6ac6579400b4f81b80b5b16/gateway/handler_car.go#L82-L89), and then compare with the number of successes in `ipfs_http_gw_car_stream_get_duration_seconds_count`

@hacdias do you know if any other places than CAR and [TAR](https://github.com/ipfs/boxo/blob/4fb455dd307a85a3d6ac6579400b4f81b80b5b16/gateway/handler_tar.go#L84)?

BigLep:
[...]

PR #311, opened 2023-05-22
feat(gw): add ipfs_http_gw_request_types metric

lidel:
This PR adds an explicit request type metric, a counter of how many requests of a specific format hit the gateway.
When no explicit format is requested, the type is empty string `""`.

This is useful on its own, allowing gateway operators to reason what request types are the most popular, but also helps with debugging described in  https://github.com/ipfs/boxo/issues/288.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -67,6 +67,7 @@ type handler struct {
 	api    IPFSBackend
 
 	// response type metrics
+	requestTypeMetric            *prometheus.CounterVec
 	getMetric                    *prometheus.HistogramVec
 	unixfsFileGetMetric          *prometheus.HistogramVec
 	unixfsDirIndexGetMetric      *prometheus.HistogramVec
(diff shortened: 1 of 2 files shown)


Jorropo:
This is a non trivial change, I am surprised this was merged without review or comment explaining why merging as-is.

lidel:
This is a trivial change -- it only adds new counter metric.

Jorropo (commented):
LGTM

Jorropo:
It's code that do stuff. I don't think that trivial. You could had made a wrong copy paste and did `i.getMetric.WithLabelValues(contentPath.Namespace(), responseFormat).Inc()` instead (I know my example wouldn't work because `getMetric` is a histogram, it's more to show who knows what could happen in 20 lines of code ?).

Having someone quickly read with a different state of mind can catch simple mistakes that happen when you code quickly.
Anyway this looks good.

PR #312, opened 2023-05-22
feat(gw): add ipfs_http_gw_car_stream_fail_duration_seconds

### LINKED ISSUE #288: Gateway metric tracking partial failures

lidel:
This PR  adds failure metrics to streaming response types (CAR and TAR)
to give us better visibility (histograms) into how long it takes for CAR to fail mid-stream.

Closes  #288


Jorropo (approved):
<3

PR #315, opened 2023-05-26
feat(gateway): visualize dag-cbor and dag-json

### LINKED ISSUE #307: More useful html with dag-cbor
I have been working with @linuskendall on onboarding [solana blockchain data to filecoin](https://github.com/rpcpool/yellowstone-faithful). The data is encoded with dag-cbor.

We are thinking of supporting retrievals (backed by Saturn) in an IPFS gateway. The intended UX was similar to what a user gets with a unixfs dag in ipfs gateway — they see directories that they can click into till they get to a file, at which point they can decide to download it.

Unfortunately this doesn’t work with dag-cbor — [relevant slack thread](https://filecoinproject.slack.com/archives/CRLC5H4J0/p1682179609351529) . With Kubo 0.18, it recognises the block as dag-cbor and gives the user a helpful message, however, the user still has to download the block and then decode it themselves. 

Screenshot 2023-05-15 at 2 34 46 PM

A user should be able to start at the epoch level, and go down the dag to the point where they want to start downloading, and then download everything under — similar to unixfs.

hacdias:
Closes #307. Builds on top of #298.

- Adds minimal DAG preview for DAG-CBOR and DAG-JSON
- You can preview how it looks by following `gateway/assets/README.md`
	- For live version, clone Kubo and this PR and then `go mod edit -replace=github.com/ipfs/boxo=path/to/your/local/boxo`
- Sharness fails here due to a bug fix: see https://github.com/ipfs/kubo/pull/9904.

image

image


hacdias:
I'm opening this for review to get feedback and for others to take it for a spin. I was quite conservative on how we handle the preview. If some error happens, no preview is shown (this usually means that the DAG isn't valid). If it all works, we show a preview.

gateway/assets/dag.html
```diff
@@ -28,6 +28,33 @@
         <li><a href="?format=dag-cbor" rel="nofollow">Valid DAG-CBOR</a> (specs at <a href="https://ipld.io/specs/codecs/dag-cbor/spec/" target="_blank" rel="noopener noreferrer">IPLD</a> and <a href="https://www.iana.org/assignments/media-types/application/vnd.ipld.dag-cbor" target="_blank" rel="noopener noreferrer">IANA</a>)</li>
       </ul>
     </section>
+    {{ with .Node }}
+      <section class="full-width">
+        <header>
+          <strong>DAG Preview</strong>
```
lidel: nit: this is not a DAG :)  lets use codec name here
```suggestion
          <strong><span translate="no" style="text-transform: uppercase;">{{.CodecName}}</span> preview</strong>
```

gateway/assets/node.go
```diff
@@ -0,0 +1,110 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			vd, err := ParseNode(v)
+			if err != nil {
+				return nil, err
+			}
+
+			dag.Keys = append(dag.Keys, &ParsedNode{Value: fmt.Sprint(k)})
```
lidel: :thought_balloon:  `k` could be bytes that translate to `<script></script>` is that sanitized later when we call `template "node"` ? or do we need to replace all Sprint like this one with explicit sanitizations from [html/template](https://pkg.go.dev/html/template) just to be safe?

gateway/assets/node.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return nil, err
+		}
+		dag.Value = template.HTMLEscapeString(v)
+	case datamodel.Kind_Bytes:
+		v, err := node.AsBytes()
+		if err != nil {
+			return nil, err
+		}
+		dag.Value = fmt.Sprint(v)
```
lidel: Same question about string sanitization

lidel (changes requested):
Thank you @hacdias I think this is good enough to ship, but needs:

- [x] Kubo PR  to be green, fine to reuse one from https://github.com/ipfs/boxo/pull/298#pullrequestreview-1449906423 after that PR lands: https://github.com/ipfs/kubo/pull/9904
- [x] Add a regression test that confirms CBOR bytes that get turned into strings in HTML get sanitized (just to be sure)

gateway/assets/node.go
hacdias: @lidel:

- `k` is an `int64`. I changed it to `fmt.Sprintf("%d", k)` to be explicit.
- Bytes were being printed as `[00 00 00 00 ...]`. I changed it to use `hex.EncodeToString`, which is supposedly faster than `fmt.Sprintf("%x", v)`. We may also consider using [`hex.Dump`](https://pkg.go.dev/encoding/hex#Dump) instead. 
- Floats, booleans, integers, CIDs are fine.
- String type was being sanitised using `template.HTMLEscapeString`. However, `html/template` does auto sanitisation for strings (see https://pkg.go.dev/html/template#hdr-Security_Model), so our string was getting doubly-sanitized, becoming non-readable. I added a test and removed the call to `template.HTMLEscapeString`.

I updated the screenshot with a DAG-CBOR that tests all the types. I used that CID for the tests, added to the fixtures package.

gateway/assets/node.go
hacdias: See https://github.com/ipfs/boxo/pull/315#discussion_r1210051985.

gateway/assets/node.go
hacdias: We may also consider using [`hex.Dump`](https://pkg.go.dev/encoding/hex#Dump) instead.

hacdias:
Kubo is green: https://github.com/ipfs/kubo/pull/9904

gateway/assets/node.go
lidel: @hacdias ooh, `hex.Dump` would be significantly better: mirrors `hexdump -C` and removes the need for someone to download raw block -- can inspect both HEX and ASCII representation right in the browser, something @anjor  hinted at.

https://play.golang.org/p/jj5itycHpKi

Could we swithc to it and put the produced string in `<pre>` tag?

hacdias: Of course.

hacdias: Done.

lidel (changes requested):
@hacdias while testing in Kubo, I've found some things we should address before merging:
- [x]  subdomain gateways should never return dag.html  from a path, only return redirect: http://localhost:8080/ipfs/bafyreihnpl7ami7esahkfdnemm6idx4r2n6u3apmtcrxlqwuapgjsciihy MUST  redirect to http://bafyreihnpl7ami7esahkfdnemm6idx4r2n6u3apmtcrxlqwuapgjsciihy.ipfs.localhost:8080
- [x] while at this, for HTML responses, we should also ensure the generated  HTML document is returned with a scope limited to the CID, the same way dir listings are. On path gateway http://127.0.0.1:8080/ipfs/bafyreihnpl7ami7esahkfdnemm6idx4r2n6u3apmtcrxlqwuapgjsciihy MUST return HTTP 301 redirect to http://127.0.0.1:8080/ipfs/bafyreihnpl7ami7esahkfdnemm6idx4r2n6u3apmtcrxlqwuapgjsciihy/ (normalize with `/` at the end, just like unixfs dirs)
  - this is an important precaution in case there is ever a bug in sanitization and/or our [service worker protection is ever removed](https://github.com/ipfs/boxo/blob/c23df3837fbe224259a60e8eed9d77e2656f2dce/gateway/handler.go#L803-L810)

hacdias:
@lidel I just pushed a fix for both, including tests. In addition, I have a few notes:

- The codec handler was never redirecting properly. I changed the converted and the raw to wrap the `ResponseWriter` with `statusResponseWriter` so we preserve the redirection status. Similar to UnixFS files.
- HTML response now redirects to `/`. Therefore, some sharness tests in Kubo fail where we used to check if we got HTML back. I fixed them by simply requesting the path with trailing slash. The redirection is tested in the new tests in this PR. https://github.com/ipfs/kubo/pull/9904

hacdias:
@lidel regarding our redirects and body conversations, I did not include it here because we have some tests currently checking for that: https://github.com/ipfs/boxo/pull/326 - let's keep it separate. However, I do think we should clean it up asap.

lidel (approved):
Thank you, lgtm, merging. I will bubble up to https://github.com/ipfs/kubo/pull/9904 and review there after CI is green.

Issue #291, opened 2023-05-08
go-merkledag duplicates and global variables


aschmahmann:
## Background

1. There are currently two packages of the `merkledag` library. There is the one here in boxo and go-merkledag.
2. There is a library https://github.com/ipfs/go-ipld-legacy that is used to create nodes that satisfy the go-ipld-format and go-ipld-prime interfaces
3. That library uses a global variable to manage specific types of conversions, such as getting a DagPB specific type rather than a generic untyped node. https://github.com/ipfs/go-ipld-legacy/blob/219032a33b3037ac7754ea9f04c4b1ead533f364/decode.go#L32
4. There exist places in the code that do type checks against specific types such as https://github.com/ipfs/boxo/blob/8059f183d86648f69ad562546b4797bc148c0984/ipld/unixfs/file/unixfile.go#L155

## Bug

These interactions collectively mean that depending on whether go-merkledag gets initialized before or after boxo/ipld/merkledag you may/may not get an error if you call boxo/ipld/merkledag/DagService.Get and pass the result into boxo/ipld/unixfs/file/NewUnixFSFile.

## Proposed fixes

There are obviously lots of problems here that could resolve this particular issue given the many unfortunate practices that have led us here (two versions of similar code, using global variables to manage state, doing explicit type checks, etc.).

Probably the easiest is to copy go-ipld-legacy into boxo. At that point the entry paths into go-ipld-legacy will be a little more obvious/contained and we can figure out the best way to go from there.

Another option that might not be too bad would be to remove the global variables from go-ipld-legacy, and force an instance to be instantiated with a converter registry, which would likely be a breaking change. That repo doesn't appear to be highly utilized explicitly (as opposed to via merkledag) so it probably wouldn't be too rough on consumers. That being said it's certainly more disruptive than a copy.

Open to better suggestions if folks have them.

## Tracking steps

Stage the changes:
- [x] https://github.com/ipfs/boxo/pull/323
- [x] https://github.com/ipfs/boxo/pull/318
- [x] https://github.com/ipfs/boxo/pull/322
- [x] https://github.com/ipfs/boxo/pull/324
- [x] https://github.com/ipfs/go-ipld-format/pull/80

Validate the changes with consumers:: 
- [x] Kubo: https://github.com/ipfs/kubo/pull/9907
- [x] Lotus: https://github.com/filecoin-project/lotus/pull/10921
- [x] Boost: https://github.com/filecoin-project/boost/pull/1492
[...]

Issue #309, opened 2023-05-18
Kubo/Boxo: pre PR signal that a contributor needs to run "go mod tidy"


BigLep:
## Done Criteria
Before a contributor makes a PR to Boxo or Kubo they should have indication if their PR will fail CI checks due to `go mod tidy` checks.

## Why Important
1. This trips up most new contributors to the codebase
2. Maintainers end up copy/pasting messages like https://github.com/ipfs/kubo/pull/9787#issuecomment-1532313618

## Notes
1. We're not going to run mutating code by default, but during the make/build process we can highlight the issue and point someone to the steps they need to take.

gammazero:
CI check is sufficient.

PR #322, opened 2023-05-30
feat: remove ipld legacy and format globals in ipld/merkledag

aschmahmann:
Related to https://github.com/ipfs/boxo/issues/291


ipld/merkledag/merkledag.go
```diff
@@ -19,16 +18,14 @@ import (
...
-	format.Register(cid.DagProtobuf, DecodeProtobufBlock)
-	format.Register(cid.Raw, DecodeRawBlock)
-	format.Register(cid.DagCBOR, ipldcbor.DecodeBlock)
+var ipldLegacyDecoder *legacy.Decoder
 
-	legacy.RegisterCodec(cid.DagProtobuf, dagpb.Type.PBNode, ProtoNodeConverter)
-	legacy.RegisterCodec(cid.Raw, basicnode.Prototype.Bytes, RawNodeConverter)
+// TODO: Don't require global registries
+func init() {
+	d := legacy.NewDecoder()
```
aschmahmann, line 25: @willscott WDYT about explicitly passing the go-ipld-prime linksystem here instead of requiring the default? This might make it easier for people to get out of the "global defaults" game.

ipld/merkledag/merkledag.go
```diff
@@ -19,16 +18,14 @@ import (
...
+	d.RegisterCodec(cid.DagProtobuf, dagpb.Type.PBNode, ProtoNodeConverter)
+	d.RegisterCodec(cid.Raw, basicnode.Prototype.Bytes, RawNodeConverter)
+	ipldLegacyDecoder = d
```
aschmahmann, line 28: @willscott thanks for doing https://github.com/ipfs/go-merkledag/pull/104. I mostly copied it. Although I left the globals up in the global space as a reminder for us to try and exit globals land when it becomes possible 😅

ipld/merkledag/merkledag.go
willscott: is there a reason to have this as a global versus the approach i took in https://github.com/ipfs/go-merkledag/pull/104 and making the decoder in the `NewDAGService` below?

Jorropo:
https://github.com/ipfs/kubo/pull/9787#issuecomment-1532313618 :rofl: #309 has some merits I see

ipld/merkledag/merkledag.go
willscott: > as a reminder for us to try and exit globals land when it becomes possible 

I'm confused by this - in 104 i did exit globals land. - is there a reason not to stay consistent with that in the copy here, since it's the direction we want to go anyway?

aschmahmann: > I'm confused by this - in 104 i did exit globals land

We're still stuck with globals:
1. The merkledag libraries still have globals. IMO this is bad for the same reasons the other ones were bad, it just turns out the other ones are basically unused independently so for now we can at fairly low cost coalesce here (and solve a problem we currently have)
2. There are still go-ipld-prime globals used implicitly in go-ipld-legacy unless we fix that https://github.com/ipfs/go-ipld-legacy/pull/12/files#r1211104340
    - Ran out of time to do the kubo PR exercising this (https://github.com/ipfs/kubo/pull/9907). Maybe it'll be fine to leave this merkledag PR mostly alone (aside from passing in something like a linksystem to go-ipld-legacy). However, to start wrestling control away from the globals we might also want to make the legacy decoder used by merkledag publicly settable (albeit with the various warnings about being careful about using it, and trying to only use it in top-level binaries).
    
>  is there a reason not to stay consistent with that in the copy here

If we go with making the go-ipld-legacy decoder publicly settable (and therefore not tied to go-ipld-prime's global registry) then this approach looks much better rather than just being a stylistic difference. If it's just a stylistic difference then 🤷.

ipld/merkledag/merkledag.go
aschmahmann: Initially it was just a stylistic preference that I had since a) it's still a global in 104, it's just a little more subtle b) it's a bit of a DRY thing where I didn't want to accidentally change one instead of the other.

However, I'm now also thinking that given we still have this global we might want to make it settable until we can make relying on a merkledag global registry no longer a thing (especially since it also implicitly hides the go-ipld-prime global registry)

willscott (approved):

aschmahmann:
merging without approval from the other kubo maintainers since I feel like @willscott's review is enough 😄. If someone objects we can fix it up before the boxo release tag.

PR #333, opened 2023-06-05
feat(routing/http): delegated IPNS server and client, IPIP 379

hacdias:
This implements the Boxo client and server for https://github.com/ipfs/specs/pull/379.

- [x] Requires #339 
- [x] Depends on fix: https://github.com/ipld/go-ipld-prime/pull/525
- [x] **Requires release** of https://github.com/ipld/go-ipld-prime (@rvagg)


hacdias:
Note that this PR will have to be rebased if #339 is merged first.

routing/http/server/server.go
```diff
@@ -49,6 +58,13 @@ type ContentRouter interface {
 	FindProviders(ctx context.Context, key cid.Cid, limit int) (iter.ResultIter[types.ProviderResponse], error)
 	ProvideBitswap(ctx context.Context, req *BitswapWriteProvideRequest) (time.Duration, error)
 	Provide(ctx context.Context, req *WriteProvideRequest) (types.ProviderResponse, error)
+
+	// GetIPNSRecord searches for an IPNS Record for the given peer ID.
+	GetIPNSRecord(ctx context.Context, pid peer.ID) (*ipns_pb.IpnsEntry, error)
+
+	// PutIPNSRecord stores the provided IPNS Record for the given peer ID. It is
+	// guaranteed that the record matches the provided peer ID.
+	PutIPNSRecord(ctx context.Context, pid peer.ID, record *ipns_pb.IpnsEntry) error
```
hacdias: @lidel here I just named them `GetX` and `PutX`. I don't know how to feel about this considering the others are called `Find` and `Provide`. Perhaps, we should (in a separate PR), make the names consistent: `GetProviders`, `PutProviders`, etc. And I don't think there should be a specific `ProvideBitswap`. The implementer of `Provide` should be able to distinguish records and only use the ones they want.

routing/http/client/client.go
```diff
@@ -324,3 +328,68 @@ func (c *client) provideSignedBitswapRecord(ctx context.Context, bswp *types.Wri
 
 	return 0, nil
 }
+
+func (c *client) GetIPNSRecord(ctx context.Context, pid peer.ID) (*ipns_pb.IpnsEntry, error) {
+	url := c.baseURL + "/routing/v1/ipns/" + peer.ToCid(pid).String()
```
lidel: :thought_balloon: I think its a good idea to land https://github.com/ipfs/boxo/pull/339 first and switch to `ipns.Name` here, removing direct dep. on libp2p

routing/http/client/client_test.go
```diff
@@ -441,3 +457,91 @@ func TestClient_Provide(t *testing.T) {
 		})
 	}
 }
+
+func makePeerID(t *testing.T) (crypto.PrivKey, peer.ID) {
```
lidel:
```suggestion
func makeIPNSName(t *testing.T) (crypto.PrivKey, ipns.Name) {
```

routing/http/server/server.go
lidel: - I like `Get` and `Put`, but consistency is more important. Maybe use `Find` and `Provide` for now, and cleanup everything later?
- +1 on  `ProvideBitswap` –  needs to be removed (or made protocol-agnostic): `boxo/routing/http/contentrouter/contentrouter.go#Provide*`  uses it blindly :see_no_evil:

lidel (changes requested):

hacdias:
I will address the feedback once we have the IPNS PRs complete then (or I might rebase it earlier if I have time).

hacdias:
I rebased this on top of the IPNS PR to make the required changes. This also depends on a fix of `go-ipld-prime`, otherwise the mocking tests will fail due to an inconsistent behaviour between creating new integer nodes and decoding integer nodes: https://github.com/ipld/go-ipld-prime/pull/525.

Gateway sharness will fail until IPNS PR is merged.

routing/http/server/server.go
hacdias: @lidel yeah, so `Provide` is already the generic version of `ProvideBitswap`. The problem with the naming `Provide` and `Find` is that now `Provide` should be renamed `ProvideProviders` to be consistent with `ProvideIPNSRecord`. 

Nevertheless, I think that cleanup can happen when we do the peer routing implementation. For now, the IPNS delegation is very nice because it's very contained.

examples/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/ipfs/go-block-format v0.1.2
 	github.com/ipfs/go-cid v0.4.1
 	github.com/ipfs/go-datastore v0.6.0
-	github.com/ipld/go-ipld-prime v0.20.0
+	github.com/ipld/go-ipld-prime v0.20.1-0.20230619045951-499e9b9c2024
```
lidel: @hacdias we are waiting for release as noted in https://github.com/ipld/go-ipld-prime/pull/525#issuecomment-1608818203, right?  (don't want to land unreleased commit in `main`, feels like asking for trouble, given breaking changes Rod  mentioned in linked comment)

routing/http/server/server_test.go
```diff
@@ -117,6 +123,126 @@ func TestResponse(t *testing.T) {
...
+	sk, _, err := crypto.GenerateEd25519Key(rand.Reader)
+	require.NoError(t, err)
+
+	pid, err := peer.IDFromPrivateKey(sk)
+	require.NoError(t, err)
+
+	return sk, ipns.NameFromPeer(pid)
+}
+
+func makeIPNSRecord(t *testing.T, cid cid.Cid, sk crypto.PrivKey) (*ipns.Record, []byte) {
```
lidel: While we are at this, can we add test for both V1+V2 and V2-only ([IPIP-428](https://github.com/ipfs/specs/pull/428))?
Will ensure server works with both and will save us headache in the future.

lidel (changes requested):
Thanks for fixing tests, small nits below.

examples/go.mod
hacdias: @lidel yeah, that'd be great, but I have no idea if there's an ETA. cc @rvagg

routing/http/server/server_test.go
hacdias: Updated both server and client tests.

examples/go.mod
rvagg: Yeah, sorry, this is a pressing concern for me too; I wanted to squeeze one more item into go-ipld-prime (https://github.com/ipld/go-ipld-prime/pull/530) before cutting a release since the releases there are a bit more involved than most other simpler libraries and we have a bit of a backlog of items already. I'm slowly getting back into some work this week but it probably won't be until next week to get this out; I hope this is OK.

rvagg:
https://github.com/ipld/go-ipld-prime/pull/534

will get this out tomorrow unless something gets in the way

hacdias:
@lidel we now have a new release of `go-ipld-prime` (thanks @rvagg) so I think this is ready for final review.

lidel (approved):
Thank you @hacdias, I think its good enough to merge and iterate further in https://github.com/ipfs/boxo/pull/422 +  https://github.com/ipfs/kubo/pull/9877.

ps. on windows [`TestClientOverMax` is flaky](https://github.com/ipfs/boxo/actions/runs/5931852593/job/16084571137#step:13:312), i've re-run it.

PR #334, opened 2023-06-06
feat(path)!: consolidated path libraries

### LINKED ISSUE #198: Consolidate IPFS Path libraries under `boxo/path`

hacdias:
This is the long-awaited consolidation of the multiple path libraries we've had, which closes #198. For more information, see the changelog in the PR.

Kubo PR: https://github.com/ipfs/kubo/pull/10063.


hacdias:
@ipfs/kubo-maintainers I think this is something important to get out of the door for many reasons (see https://github.com/ipfs/boxo/issues/198). I just rebased this work I did in June on top of the recent changes.  I can only imagine this having large repercussions thorough our other packages, especially Kubo. Therefore, before continuing, I would like to ask for some feedback regarding the `Path` interface.

One of the biggest changes is the removal of the `IsValid() bool` function from the path type. The idea is that if a path exists, then it _must_ be valid. There's no point on passing around a path that will be invalid and always calling for `IsValid()`. Now the verification is done at creation time. This is what _scares_ me the most when bubbling this to Kubo, as processes that would otherwise fail later in the process will now fail earlier. Or maybe not the case.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
-type Path string
-
-// ^^^
-// TODO: debate making this a private struct wrapped in a public interface
-// would allow us to control creation, and cache segments.
+const (
+	IPFSNamespace = "ipfs"
+	IPNSNamespace = "ipns"
+	IPLDNamespace = "ipld"
+)
```
Jorropo, line 16:
```suggestion
type Namespace uint

func (namespace Namespace) String() string {
 switch namespace {
 case IPFSNamespace:
  return "ipfs"
 case IPNSNamespace:
  return "ipns"
 case IPLDNamespace:
  return "ipld"
 default:
  return "unknown path namespace: " + strconv.FormatUint64(uint64(namespace), 10)
 }
}

const (
	IPFSNamespace Namespace = iota
	IPNSNamespace
	IPLDNamespace
)
```
Moar type safety.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
 
-// Segments returns the different elements of a path
-// (elements are delimited by a /).
-func (p Path) Segments() []string {
-	cleaned := path.Clean(string(p))
-	segments := strings.Split(cleaned, "/")
+// ImmutablePath is a [Path] which is guaranteed to return "false" to [Mutable].
+type ImmutablePath struct {
+	Path
+}
```
Jorropo:
```suggestion
var _ Path = ImmutablePath{}

// ImmutablePath is a [Path] which is guaranteed to return "false" to [Path.Mutable].
type ImmutablePath struct {
	path Path
}

func (ip ImmutablePath) String() string {
 return ip.path.String()
}

func (ip ImmutablePath) Namespace() Namespace {
 return ip.path.Namespace()
}

func (ip ImmutablePath) Mutable() bool {
 return false
}

func (ip ImmutablePath) Root() cid.Cid {
 return ip.path.Root()
}

func (ip ImmutablePath) Segments() []string {
 return ip.path.Segments()
}
```
Prevents consumers from manually wrapping as an immutable path without going through `NewImmutablePath`.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
-func FromString(s string) Path {
-	return Path(s)
+// Path is a generic path. Paths must be prefixed with a valid prefix.
+type Path interface {
+	// String returns the path as a string.
+	String() string
+
+	// Namespace returns the first component of the path. For example, the namespace
+	// of "/ipfs/bafy" is "ipfs".
+	Namespace() string
```
Jorropo, line 35:
```suggestion
	Namespace() Namespace
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return ImmutablePath{p}, nil
+}
+
+type path struct {
+	str       string
+	root      cid.Cid
+	namespace string
+}
+
+func (p *path) String() string {
```
Jorropo:
```suggestion
func (p path) String() string {
```
This is never mutated so no need to use a pointer.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return p.str
+}
+
+func (p *path) Namespace() string {
```
Jorropo:
```suggestion
func (p path) Namespace() string {
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return p.namespace
 }
 
-// String converts a path to string.
-func (p Path) String() string {
-	return string(p)
+func (p *path) Mutable() bool {
```
Jorropo:
```suggestion
func (p path) Mutable() bool {
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return p.namespace == IPNSNamespace
 }
 
-// IsJustAKey returns true if the path is of the form <key> or /ipfs/<key>, or
-// /ipld/<key>
-func (p Path) IsJustAKey() bool {
-	parts := p.Segments()
-	return len(parts) == 2 && (parts[0] == "ipfs" || parts[0] == "ipld")
+func (p *path) Root() cid.Cid {
```
Jorropo:
```suggestion
func (p path) Root() cid.Cid {
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return p.root
 }
 
-// PopLastSegment returns a new Path without its final segment, and the final
-// segment, separately. If there is no more to pop (the path is just a key),
-// the original path is returned.
-func (p Path) PopLastSegment() (Path, string, error) {
+func (p *path) Segments() []string {
```
Jorropo:
```suggestion
func (p path) Segments() []string {
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
 	}
 
-	return segments
+	return ImmutablePath{p}, nil
+}
+
+type path struct {
+	str       string
+	root      cid.Cid
+	namespace string
```
Jorropo, line 56:
```suggestion
	namespace Namespace
```

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return &resolvedPath{
+		path: path{
+			str:       fmt.Sprintf("/%s/%s", IPFSNamespace, cid.String()),
+			root:      cid,
+			namespace: IPFSNamespace,
+		},
+		cid:       cid,
+		remainder: "",
 	}
+}
```
Jorropo, line 123:
```suggestion
type resolvedPath struct {
	path
}

func (p resolvedPath) Remainder() string {
	return ""
}

// NewIPFSPath returns a new "/ipfs" path with the provided CID.
func NewIPFSPath(cid cid.Cid) ResolvedPath {
	return resolvedPath{
		path: path{
			str:       fmt.Sprintf("/%s/%s", IPFSNamespace, cid.String()),
			root:      cid,
			namespace: IPFSNamespace,
		},
		cid:       cid,
		remainder: "",
	}
}
```
you are never returning something different from the underlying path, and the remainder is always empty so you can hardcode that

Jorropo (dismissed):
I shouldn't look at the implementation before the API so the code review is half done.

I don't understand what is the difference between `ResolvedPath` and `ImmutablePath` ?

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	return strings.Split(str, "/")
+}
+
+type resolvedPath struct {
+	path
+	cid       cid.Cid
+	remainder string
+}
+
+func (p *resolvedPath) Cid() cid.Cid {
```
Jorropo:
```suggestion
func (p resolvedPath) Cid() cid.Cid {
```
no mutation here too (same for other methods)

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+
+	// Mutable returns false if the data pointed to by this path is guaranteed to not
+	// change. Note that resolved mutable paths can be immutable.
+	Mutable() bool
+
+	// Root returns the [cid.Cid] of the root object of the path. Root can return
+	// [cid.Undef] for Mutable IPNS paths that use [DNSLink].
+	//
+	// [DNSLink]: https://dnslink.dev/
+	Root() cid.Cid
```
aschmahmann: Should this (or a similarly named function) be a property of immutable path instead of on the interface? That way IPNS IDs, DNSLink, etc. can live independently.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+
+	// Segments returns the different elements ofd a path, which are delimited
+	// by a forward slash ("/"). The leading slash must be ignored, that is, no
+	// segment should be empty.
```
aschmahmann: Seems like we should make it a little clearer what's going on here:
1. Is the namespace always going to show up as the first element?
2. If so then is `0` the way we express errors in processing the segments (e.g. what should an empty path object return here, what about if I made an `ipfsPath` one where the namespace was built-in)?
3. Are paths like `/ipfs/bafyfoo/////////a` supposed to be disallowed at construction? Are they ok, but `Segments` is supposed to clean them, etc?

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+			namespace: IPLDNamespace,
+		},
+		cid:       cid,
+		remainder: "",
 	}
+}
 
-	return newPath, segs[len(segs)-1], nil
+// NewIPNSPath returns a new "/ipns" path with the provided CID.
+func NewIPNSPath(cid cid.Cid) Path {
```
aschmahmann: IPNS IDs are not inherently CIDs. At the moment they're closer to PeerIDs (which IIRC is what we tend to use with them), although I suspect at some point (not in this PR though) we're going to want them to have their own type.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
-	parts := strings.Split(txt, "/")
-	if len(parts) == 1 {
-		kp, err := ParseCidToPath(txt)
+// NewPath returns a well-formed [Path]. The returned path will always be prefixed
+// with a valid namespace (/ipfs, /ipld, or /ipns). The prefix will be added if not
+// present in the given string. The rules are:
+//
+//  1. If the path has a single component (no slashes) ans it is a valid CID,
+//     an /ipfs path is returned. If the CID is encoded with the Libp2pKey codec,
+//     then a /ipns path is returned.
```
aschmahmann: The idea of not requiring a namespace is bad enough (you could fix this with "NewIPFSPath" if you wanted implied namespaces) let's not add more magic with IPNS here. An example of how this could be particularly bad if at some point we have mutable namespaces where the identifier is a CID which points to a block that tells you the "rules" of the mutation/how to find the latest version.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
-		return "", &ErrInvalidPath{error: err, path: txt}
+// NewPathFromSegments creates a new [Path] from the provided segments. This
+// function simply calls [NewPath] internally with the segments concatenated
+// using a forward slash "/" as separator.
+func NewPathFromSegments(segments ...string) (Path, error) {
+	if len(segments) > 1 {
+		if segments[0] == "" {
+			segments = segments[1:]
+		}
 	}
```
aschmahmann: Why are we doing this? It seems like either we're relying on `NewPath` to do things like `gopath.Clean` and erroring on paths without enough segments, or we're not. Will removing this result in problems?

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+// with a valid namespace (/ipfs, /ipld, or /ipns). The prefix will be added if not
+// present in the given string. The rules are:
+//
+//  1. If the path has a single component (no slashes) ans it is a valid CID,
+//     an /ipfs path is returned. If the CID is encoded with the Libp2pKey codec,
+//     then a /ipns path is returned.
+//  2. If the path has a valid CID root but does not have a namespace, the /ipfs
+//     namespace is automatically added.
+//
+// This function returns an error when the given string is not a valid path.
```
aschmahmann: We should explain here (or on the interface) what a valid path is. Can it have `////` in it? How about `\n`, or `NUL` or other weird characters?

Maybe the answer is that the interface is more flexible and the structs are less, but if so that should be documented as well.

path/path.go
```diff
@@ -3,188 +3,301 @@ package path
...
+	// by a forward slash ("/"). The leading slash must be ignored, that is, no
+	// segment should be empty.
+	Segments() []string
 }
 
-// FromCid safely converts a cid.Cid type to a Path type.
-func FromCid(c cid.Cid) Path {
-	return Path("/ipfs/" + c.String())
+// ResolvedPath is a [Path] which was resolved to the last resolvable node.
+type ResolvedPath interface {
```
aschmahmann: Maybe out of scope for this PR, but hopefully not. I'd like to better document what "last resolvable node" means.

For example:
1. A path that's been fully resolved down to its final IPLD block (i.e. the remainder when traversed does not cross any block boundaries)
2. A path that was resolved as far as the implementation could (e.g. it resolved DNSLink but failed at the IPNS resolution, it resolved a dag-cbor part of the path but couldn't understand a dag-jose part of the path, it couldn't find/access some block in the path, etc.)

It seems like this tries to imply the former, although if true I wonder how much this resolution is particularly needed in comparison to others like:
- I started with `/ipns/` and I want the immutable path
- I want the root CIDs of all the "entities" being traversed
- I want all of the ipns resolution information AND the root CIDs of the entities being traversed

Understanding where/why this is used might help us come up with something better here (and need fewer path wrappers or extra lookups in places like the gateway code).

aschmahmann (commented):
Thank you for starting this 🙏. The current state of the path libraries is .... too much

path/path.go
hacdias: @aschmahmann indeed. I want to change this to `ipns.Name`, which is, in my opinion, the best thing to do here. However, `boxo/ipns` cannot depend on `boxo/path` because import cycles.

What do you think about moving the `NewIPNSPath` to inside of `boxo/ipns`. `ipns.NewPath(name Name)`? I'm slightly concerned here, but I don't see many alternatives. The other alternative I see is making `ipns.Name` its own package to go around the cycles, but that feels a bit iffy too.

> PeerIDs (which IIRC is what we tend to use with them)

We try to default to CIDv1 encoded peer IDs:  https://specs.ipfs.tech/ipns/ipns-record/#string-representation.

path/path.go
hacdias: Ok. So this is the behaviour of the current code in main. I am updating that as I agree this is a bit of a mess. This has been changed in [e688ad9](https://github.com/ipfs/boxo/pull/334/commits/e688ad9a343083239029d59652ea31c92b923063).  I had to make some updates in the namesys package to keep accepting legacy DNSLink records.

hacdias: I do agree with you. I think we should remove all the implied namespaces, including for IPFS. I think that will break some other things, at least within Boxo, but we can patch them where needed. From what I can see, it will only affect the processing of DNSLink entries.

path/path.go
hacdias: Honestly, I'm not sure why I added this. I removed it. I have to add more tests.

path/path.go
hacdias: I pushed 4ba0583b76f365119ddc2e46c22a004a2a81192f, which removes all assumptions.  I do think that's the best way forward.

path/path.go
hacdias: I do agree we should remove this from `Path`. Path is abstract and it may be IPNS or IPFS or IPLD. Therefore, there may not be a Root. I will likely move it to `ImmutablePath`.

path/path.go
hacdias: Could you check the comment update I just made? Another suggestion I would make is to simply remove this. The majority of places where we call `Segments` is to check the namespace. That can now be replaced by `Namespace() == path.Namespace`.

path/path.go
hacdias: @aschmahmann I just updated the description to mention that the path is sanitized with `gopath.Clean`, but final trailing slash is preserved.

path/path.go
hacdias: @aschmahmann Indeed, and I do also think that the current meaning is **2**. I checked and the only places across both Boxo and Kubo where it is used is:

- `gateway.ContentPathMetadata`, to represent the `LastSegment`, which I think we only do because it's the type used by the following two:
  - `kubo/client/rpc.ResolvePath` 
  - `kubo/coreapi.NewResolvedPath` 

I think we have the opportunity here to get rid of `ResolvedPath`. From what I can see, a path, in its generic form, is just `/{namespace}/{root}[/path]`, where:

- `namespace` can be `ipfs`, `ipns` or `ipld`.
- `root` can be a CID for all, can be a `ipns.Name` for `ipns` and a FQDN for `ipns` (DNSLink).

cc @Jorropo because you also had questions about ResolvedPath vs. ImmutablePath

path/path.go
hacdias: I removed IPNS and DNSLink related things for now. They can still be created manually, but we slightly limit the scope here.

path/path.go
hacdias: See [0a71844](https://github.com/ipfs/boxo/pull/334/commits/0a718449a6f1a3a8f8646783e07aa028dbddecd0), where I removed `ResolvedPath` and use `ImmutablePath`. I think this seems like the way to go.

path/path.go
```diff
@@ -3,188 +3,240 @@ package path
...
 	}
+}
 
-	// if nothing, bail.
-	if len(parts) == 0 {
-		return cid.Cid{}, nil, &ErrInvalidPath{error: fmt.Errorf("empty"), path: string(fpath)}
+// NewPath takes the given string and returns a well-forme and sanitized [Path].
+// The given string is cleaned through [gopath.Clean], but preserving the final
+// trailing slash. This function returns an error when the given string is not
+// a valid path.
```
lidel:
```suggestion
// NewPath takes the given string and returns a well-formed and sanitized [Path].
// The given string is cleaned through [gopath.Clean], but preserving the final
// trailing slash. This function returns an error when the given string is not
// a valid content path.
```

path/path.go
```diff
@@ -3,188 +3,240 @@ package path
...
+func NewPath(str string) (Path, error) {
+	cleaned := gopath.Clean(str)
```
lidel: :thought_balloon: Writing down sane normalization like this would benefit other implementations.

:information_source:  fysa I've filled https://github.com/ipfs/specs/issues/432 to document behaviors around pathing on gateways (as this is not limited to unixfs) under a dedicated Web Pathing spec at some point, for now feel free to add any undocumented behavior/ edge case you find to the list there..

path/path.go
```diff
@@ -3,188 +3,240 @@ package path
...
+type immutablePath struct {
+	path Path
+	cid  cid.Cid // TODO
+}
 
-	if len(parts) < 3 {
-		return "", &ErrInvalidPath{error: fmt.Errorf("invalid ipfs path"), path: txt}
+func NewImmutablePath(p Path) (ImmutablePath, error) {
+	if p.Namespace().Mutable() {
+		return nil, fmt.Errorf("path was expected to be immutable: %s", p.String())
```
lidel:
```suggestion
		return nil, fmt.Errorf("path was expected to be immutable: %q", p.String())
```

path/path.go
lidel: +1, :thought_balloon:  @hacdias perhaps  `IPNSName.AsPath()` makes more sense?

hacdias:
This is ready for review. You can see Kubo in https://github.com/ipfs/kubo/pull/10063. Sharness is expected to fail on this PR, as Kubo can't build.

path/path.go
```diff
@@ -3,187 +3,240 @@ package path
...
-	newPath, err := ParsePath("/" + strings.Join(segs[:len(segs)-1], "/"))
-	if err != nil {
-		return "", "", err
-	}
+	// Cid returns the [cid.Cid] of the root object of the path.
+	Cid() cid.Cid
 
-	return newPath, segs[len(segs)-1], nil
+	// Remainder returns the unresolved parts of the path.
+	Remainder() string
```
aschmahmann: Does "unresolved/remainder" here equivalent to "any path component beyond the namespace and CID"?

IIRC historically the "remainder" was for describing the path components that landed inside a block (e.g. pathing into a cbor block). Is the idea that we no longer need this anymore/people are returning `[]string` instead wherever this would've been needed?

path/path.go
aschmahmann: Not necessarily that critical, but how miserable will this end up making things if at some point we want to support pathing through empty names (even if not in places like web pathing)?

path/error.go
```diff
@@ -1,25 +1,32 @@
...
 
 func (e ErrInvalidPath) Unwrap() error {
-	return e.error
+	return e.err
 }
 
 func (e ErrInvalidPath) Is(err error) bool {
 	switch err.(type) {
-	case ErrInvalidPath:
+	case ErrInvalidPath, *ErrInvalidPath:
```
aschmahmann: Is this necessary? Seems like this isn't unique to this case and happens everywhere there is an `errors.Is` check, but I don't think doing this double check is that standard. There might be linters that check for this error as IIRC I've seen Goland (Jetbrains IDE) complain if you use the wrong error type (with or without a pointer) when calling `.Is`

path/path.go
```diff
@@ -3,187 +3,240 @@ package path
...
+		},
+		cid: cid,
+	}
 }
 
-// SplitList splits strings usings /
-func SplitList(pth string) []string {
-	return strings.Split(pth, "/")
+// NewIPLDPath returns a new "/ipld" path with the provided CID.
+func NewIPLDPath(cid cid.Cid) ImmutablePath {
```
aschmahmann: I'm not sure I understand the point of this function. This probably means we need more description around what the IPFS and IPLD namespaces are for.

As I understand it:
- `/ipfs` -> UnixFS pathing semantics with fallback to IPLD data model pathing semantics. Don't recall expectations around empty path segments or awkward characters (e.g. `NUL`).
- `/ipld` -> Just IPLD data model pathing. AFAIK this is form of path has been rarely used so far. Lack of definition for where ADLs, etc. fit in has been rough here. It's also not really defined anywhere so it's in a bit of limbo.

Are there supposed to be semantics for what happens at the end of the path with `/ipfs` and `/ipld` for this interface? Are we defining a resource or just how to get to the terminal element and then what happens with that resource is up to the consumer? From what I can tell it looks like the latter, although at some point we may also need one for the former (e.g. that could include things like byte ranges, IPLD ADLs, etc.).

I understand this mess is already here and we may not want to broach this particular mess right now, but even if not it's probably worth some documentation to help clarify the state of things and point to a GitHub issue for further discussion if people run into problems.

path/path.go
hacdias: Here, `ImmutablePath.Remainder` is what's left to be resolved, or cannot be resolved. Therefore, it may be a path that lands inside a block.

>  Is the idea that we no longer need this anymore/people are returning []string instead wherever this would've been needed?

I don't understand this question. We're not returning a `[]string`.

path/error.go
hacdias: I pushed a commit where I change the error to references, where we do in other places in Boxo. This way we always use the pointer version of it and we don't need to check for both types, as the other does not implement `Error`. I'm not sure if that's the best practice, but at least it's the way [`errors.New` works.](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/errors/errors.go;l=65-72)

Feel free to re-open this conversation if you disagree.

path/path.go
hacdias: > I'm not sure I understand the point of this function.

It just creates a new path under the `/ipld` namespace with that CID, so `/ipld/{cid}`. I just kept it because we're using it in other parts of the code.

> From what I can tell it looks like the latter

Yes.

> I understand this mess is already here and we may not want to broach this particular mess right now, but even if not it's probably worth some documentation to help clarify the state of things and point to a GitHub issue for further discussion if people run into problems.

Yeah, I'm not sure if we should tackle this now. I think it also relates to your other comment about [pathing through empty names](https://github.com/ipfs/boxo/pull/334#discussion_r1315988175).

path/path.go
hacdias: @aschmahmann @lidel I think this is a valid concern. My problem is that the previous behaviour was very inconsistent: (1) we would not clean the path at any point during creation, or validation, (2) we would clean it when splitting it into segments:

https://github.com/ipfs/boxo/blob/574df963653c940ea19b1b8a0d37cd30360c22e9/path/path.go#L36

Therefore, in the `main` branch of Boxo, if you have a certain path, let's say `/ipfs/bafy/a//b/c`, if you convert it to segments and then back to a path, you get `/ipfs/bafy/a/b/c`. My goal here is to have a uniform behaviour in library that makes sense. If you add the following test to the current `main` code, it fails:

```go
func TestInconsistency(t *testing.T) {
	pathStr := "/ipfs/bafkqaaa/a//b/c"
	p, err := ParsePath(pathStr)
	require.NoError(t, err)
	require.Equal(t, pathStr, p.String()) // Passes, preserves "//"

	p2, err := FromSegments("/", p.Segments()...)
	require.NoError(t, err)
	require.Equal(t, pathStr, p2.String()) // Fails, Expected "/ipfs/bafkqaaa/a//b/c", got "/ipfs/bafkqaaa/a/b/c"
}
```

I would also argue that at the moment there are likely no real use cases for traversing through empty namespaces, or for other things that get cleaned up by `gopath.Clean`, such as collapsing `.` and resolving `..`. If we want different behaviour for different namespaces or different path types, we will need more types and likely the `Path` interface itself should have its own `Join` instead of having a global function.

path/path.go
hacdias: @lidel I added that. I think it can still be useful, as you can just `p := name.AsPath()`, instead of doing `p, err := path.NewPath(ipns.NamespacePrefix + name.String())` and then checking for the error.

path/path.go
aschmahmann: > is what's left to be resolved, or cannot be resolved

But this indicates that it's just any path components beyond the CID, doesn't it? https://github.com/ipfs/boxo/pull/334/files#diff-a44166f2b1bbfa22a31664836b3206fbcfb31600bac5a85b0837aa0644131951R142-R147

> I don't understand this question. We're not returning a []string.

We're returning `[]string`  in places like `resolver.ResolveToLastNode` https://github.com/ipfs/boxo/pull/334/files#diff-0731f9658c3343bd65daf65521d04f5064c46ac26dd0d9bce182db016a8f0ba4R82. That's not a change in this PR, but it's somewhere that returning a path with a `Remainder` might be plausible depending on how we've defined remainder.

---

An area where I'm not sure if there are ramifications here, but maybe you can help me understand since you've been going through all of this now, is if it matters now that we:
- Used to have: ResolvedPath has Root CID, Path, Terminal CID, remainder
- Now have: ImmutablePath has Root CID, Path (remainder == path)

This loss of remainder information seems like it could be problematic. For example, f it was meant to convey "hey I tried resolving through this block but there are no more links, just other IPLD data model fields" then that information would need to be recomputed or specified as required in other interface contracts.

hacdias: @aschmahmann there are surprisingly little places where we use the concept of `.Remainder()` in both Boxo and Kubo. Boxo mainly uses it in the [handler codec](https://github.com/ipfs/boxo/blob/32140848b581e44bc044a66af149814732174915/gateway/handler_codec.go#L84-L92) to ensure there's nothing going inside the path, and Kubo only uses it as a result of [`ResolvePath`](https://github.com/search?q=repo%3Aipfs%2Fkubo+Remainder%28%29&type=code).

My suggestion is to completely remove `Remainder` from `ImmutablePath`, and change the interface of `ResolvePath` to return `ImmutablePath, []string, error`, matching `ResolveToLastNode`. This would not require that many changes and would remove the dubious meaning of `.Remainder`, which is already barely used, and every time is used, it's the same as the expected output of `ResolveToLastNode`.

I am going to work out a commit for this and you can take a look.

hacdias: @aschmahmann I made the changes mentioned above:

- Boxo: https://github.com/ipfs/boxo/pull/334/commits/4276346cc566a4033b069d08af9d0ea7323119fc
- Kubo: https://github.com/ipfs/kubo/pull/10063/commits/f5b41755bfe4033167f20c6a787802849a14fed5

Let me know what you think. I think this makes the APIs more consistent, and removes the unclear `.Remainder`. Now the remainder is only returned by resolvers and resolving functions, making it clearer what it means and what it should be used for.

hacdias:
@aschmahmann @lidel I want to ping you for a re-review. I've already addressed the feedback, tests are passing here and [in Kubo](https://github.com/ipfs/kubo/pull/10063). This PR is almost turning 4 months old and I really don't want that to happen.

path/path.go
```diff
@@ -3,187 +3,247 @@ package path
...
+//  3. If namespace is "ipns", "root" may be a [ipns.Name] or a [DNSLink] FQDN.
+//
+// [DNSLink]: https://dnslink.dev/
+type Path interface {
+	// String returns the path as a string.
+	String() string
+
+	// Namespace returns the first component of the path. For example, the namespace
+	// of "/ipfs/bafy" is "ipfs".
+	Namespace() Namespace
```
lidel: Going with  hardcoded list of enums here makes things difficult to extend. `boxo/path` becomes useless to people who want to implement their own `/foo/cid` namespace, forcing them to fork the entire boxo, because other packages use this type.

Could be a conscious design decisions, but we should double confirm this is what we want. I feel we should  have `Path.Namespace()` return a string, so users of this library make decision if the namespace is supported or not.

lidel: Update: this got resolved in https://github.com/ipfs/boxo/pull/334/commits/f0e9c2ff4fd477e3d4f121a5a8e7fc875487e8b5, thanks!

path/resolver/resolver.go
```diff
@@ -24,76 +22,73 @@ import (
...
+// ResolveToLastNode implements [Resolver.ResolveToLastNode].
+func (r *basicResolver) ResolveToLastNode(ctx context.Context, fpath path.ImmutablePath) (cid.Cid, []string, error) {
+	ctx, span := startSpan(ctx, "basicResolver.ResolveToLastNode", trace.WithAttributes(attribute.Stringer("Path", fpath)))
 	defer span.End()
 
-	c, p, err := path.SplitAbsPath(fpath)
-	if err != nil {
-		return cid.Cid{}, nil, err
-	}
+	c, remainder := fpath.Cid(), fpath.Segments()[2:]
```
lidel: :thinking:  it feels we will do this CID + `Segments()[2:]` dance more than once, maybe `ChildSegments()` (if its only in two places here, feel free to ignore)

path/path.go
```diff
@@ -3,187 +3,237 @@ package path
...
 
-// String converts a path to string.
-func (p Path) String() string {
-	return string(p)
+// ImmutablePath is a [Path] which is guaranteed to have an immutable [Namespace].
+type ImmutablePath interface {
+	Path
+
+	// Cid returns the [cid.Cid] of the root object of the path.
+	Cid() cid.Cid
```
lidel: :warning:  The name makes is not obvious what this is.
For `/ipfs/cid/dir/terminus` Could be  CID of the terminating element.

Let's make it clear this is the root one:

```suggestion
	// RootCid returns the [cid.Cid] of the root identifier of the path.
	RootCid() cid.Cid
```

path/path.go
```diff
@@ -3,187 +3,237 @@ package path
...
-// /ipld/<key>
-func (p Path) IsJustAKey() bool {
-	parts := p.Segments()
-	return len(parts) == 2 && (parts[0] == "ipfs" || parts[0] == "ipld")
+var _ Path = immutablePath{}
+var _ ImmutablePath = immutablePath{}
+
+type immutablePath struct {
+	path Path
+	cid  cid.Cid
```
lidel:
```suggestion
	rootCid  cid.Cid
```

path/path.go
```diff
@@ -3,187 +3,237 @@ package path
...
+	cleaned := SegmentsToString(segments...)
+	if strings.HasSuffix(str, "/") {
+		// Do not forget to preserve the trailing slash!
+		cleaned += "/"
 	}
 
-	return FromCid(c), nil
-}
+	switch segments[0] {
+	case "ipfs", "ipld":
```
lidel: :warning:  :thought_balloon:  not sure what is the value of `Namespace` type if we still use hardcoded strings in places like here.  

Calling `namespace.String()` everywhere feels like noise, maybe just go with `IPFSNamespace` and others being a plain `string`? 
Simplifies `path.Path` interface, allows for reuse of segments directly, etc.

path/path.go
```diff
@@ -3,187 +3,237 @@ package path
...
 
-func decodeCid(cstr string) (cid.Cid, error) {
-	c, err := cid.Decode(cstr)
-	if err != nil && len(cstr) == 46 && cstr[:2] == "qm" { // https://github.com/ipfs/go-ipfs/issues/7792
-		return cid.Cid{}, fmt.Errorf("%v (possible lowercased CIDv0; consider converting to a case-agnostic CIDv1, such as base32)", err)
+// StringToSegments converts a string into an array of segments. This function follows
+// the rules of [Path.Segments]: the path is first cleaned through [gopath.Clean] and
+// no empty segments are returned.
+func StringToSegments(str string) []string {
+	str = gopath.Clean(str)
```
lidel, line 212: :thought_balloon: I suspect formally this is  a breaking change (some types of paths are no longer allowed), but don't think it impacts real world use.

See my suggestion to mention it in the CHANGELOG.md

CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ The following emojis are used to highlight certain changes:
     `files.Node`.
 * `boxo/routing/http/client.Client` is now exported. This means you can now pass
   it around functions, or add it to a struct if you want.
+* 🛠 The `path` package has been massively refactored. With this refactor, we have
+  condensed the different path-related packages under a single one. Therefore, there
```
```suggestion
  condensed the different path-related and/or Kubo-specific packages under a single generic one. Therefore, there
```

CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ The following emojis are used to highlight certain changes:
...
+  are many breaking changes. Please consult the [documentation](https://pkg.go.dev/github.com/ipfs/boxo/path)
+  for more details on how to use the new package.
```
lidel, line 30: :thought_balloon: would not hurt to clearly state we now normalize paths:
Maybe also add this to godoc for `Path`?

```suggestion
  for more details on how to use the new package.
  * Note: content paths created with `boxo/path` are automatically normalized:
    - Replace multiple slashes with a single slash.
    - Eliminate each `.` path name element (the current directory).
    - Eliminate each inner `..` path name element (the parent directory) along with the non-`..` element that precedes it.
    - Eliminate `..` elements that begin a rooted path: that is, replace "`/..`" by "`/`" at the beginning of a path.
    - [TODO: anything else?]
```

gateway/gateway_test.go
```diff
@@ -664,8 +672,8 @@ func TestDeserializedResponses(t *testing.T) {
 		t.Parallel()
 
 		backend, root := newMockBackend(t, "fixtures.car")
-		backend.namesys["/ipns/trustless.com"] = path.FromCid(root)
-		backend.namesys["/ipns/trusted.com"] = path.FromCid(root)
```
lidel: :thought_balloon: the old version read better, new one compromises readability only to have `NewIPFSPath` and `NewIPLDPath`, and  people want the first one 99,9% of time (change my mind? :))

path/path.go
lidel: :warning:  :thought_balloon: these `NewFOOPath` helpers feel like bit more noise than signal, either only used in legacy tests (and could be replaced by `NewPath`) or hide the use case how it is different than `NewPath`. The `FromCid` we had in past felt better.

> AFAIK this is form of path [/ipld] has been rarely used so far

@aschmahmann one would think so, but Kubo uses `/ipld` paths for all `ipfs block` and `ipfs dag` commands ([example](https://github.com/ipfs/kubo/blame/6dbae768180d1f454f4d9d24920be4bc161216b6/client/rpc/dag.go#L24))

Not sure why  IPLD team preferred to use `/ipld` instead of `/ipfs` as the implicit default? Perhaps there is some low level difference in how UnixFS named paths ([logical pathing](https://ipld.io/specs/codecs/dag-pb/spec/#logical-format) vs raw IPLD pathing)? 


@hacdias  

-  `NewIPFSPath` →  `FromCid`  (`/ipfs` is the standard for getting path for a CID)
-  `NewIPLDPath` →  remove, create `ipldFromCid`  inside of test cases that can benefit from it, and also in RPC/Commands in Kubo (try changing to `/ipfs` and see if tests in https://github.com/ipfs/kubo/pull/10161 pass, if not fill issue for this as follow-up), but don't pollute `boxo/path`

lidel (dismissed):
Some feedback inline, mostly about making interface more clear, not tied to our namespaces (allowing people to use boxo with `/foo` without having to fork)  and avoiding cognitive noise caused by legacy decisions in Kubo.

To avoid making waves this difficult month, I think we should minimize risk and ship this in Kubo 0.24, and we should wait with merge until 0.23 ships.

path/path.go
hacdias: Removed, and `FromCid` restored. I will check Kubo side when updating the PR there.

CHANGELOG.md
hacdias: I will add it, but for the go docs i think is fine: I already link to `path.CLean` from Go which includes that documentation. If anything changes there [very unlikely we wouldn't have to recopy/paste.

path/resolver/resolver.go
hacdias: 3 places only. I think we can add it in the future if we feel the need for it.

hacdias:
> To avoid making waves this difficult month, I think we should minimize risk and ship this in Kubo 0.24, and we should wait with merge until 0.23 ships.

It already has: https://github.com/ipfs/kubo/releases/tag/v0.23.0!

hacdias:
I am going to go ahead and merge this. I think it has marinated for enough time and I've already addressed the feedbacks we had. Tests are passing here and in Kubo, and Kubo 0.23 has been released. So merging this at the beginning of the new release cycle (0.24) will allow us to test it until the final release of 0.24.

PR #337, opened 2023-06-07
fix(gateway): ensure 'X-Ipfs-Root' header is valid

### LINKED ISSUE #221: Gateway: Possible Invalid 'X-Ipfs-Roots' Header
Extracted from https://github.com/ipfs/boxo/pull/176#discussion_r1140854153 to avoid blocking #176.

> :warning: since `pathMetadata` is based on `maybeResolvedImPath`, if the path is indeed resolved to the final `/ipfs/cid`, we will produce an invalid `X-Ipfs-Roots` header that does not include logical parents from `immutableContentPath`.
> 
> Example: 
> 
> - `immutableContentPath` – `/ipfs/cid/foo/bar`
> - `maybeResolvedImPath` – `/ipfs/bar-cid` 
> - `X-Ipfs-Roots`
>   - expected:  `cid, foo-cid, bar-cid` 
>   - produced: `bar-cid`

## Idea

Maybe some struct that gets passed around, like:

```go
type requestData struct {
	// Defined for all requests.
	begin             time.Time
	logger            *zap.SugaredLogger
	contentPath       ipath.Path
	contentType       string
	contentTypeParams map[string]string

	// Defined for non IPNS Record requests.
	immutablePath ImmutablePath

	// Defined if resolution has already happened.
	pathMetadata *ContentPathMetadata
	resolvedPath *ImmutablePath
}

func (rq *requestData) maybeResolvedPath() ImmutablePath {
	if rq.resolvedPath != nil {
		return *rq.resolvedPath
	}
	return rq.immutablePath
}
```

```[tasklist]
### Tasks
- [ ] confirm header works as expected or there is a bug (if so, fix it)
- [ ] add regression test to https://github.com/ipfs/gateway-conformance for the Example above
```

hacdias:
Fixes #221.

My idea is to create this struct `requestData` that is passed onto the handlers in order to keep the most possible information around. The other option would be to make `handleIfNoneMatch` return one more variable (the path metadata) and then pass it around. However, our functions already have a lot of parameters.

I also cleaned up the existing tests and added a bunch of new tests for `Etag`, `If-None-Match` and `X-Ipfs-Roots`. Some of these tests should be also done in the gateway conformance. However, these have to be slightly different since we can't expect specific `Etag` values. See https://github.com/ipfs/gateway-conformance/issues/73#issuecomment-1582441267.


gateway/gateway_test.go
```diff
@@ -504,317 +116,614 @@ func TestPretty404(t *testing.T) {
...
+		fileCID   = "bafkreiba3vpkcqpc6xtp3hsatzcod6iwneouzjoq7ymy4m2js6gc3czt6i"
+		filePath  = "/ipfs/" + root.String() + "/subdir/fnord"
+		fileRoots = dirRoots + "," + fileCID
+
+		dagCborCID   = "bafyreiaocls5bt2ha5vszv5pwz34zzcdf3axk3uqa56bgsgvlkbezw67hq"
+		dagCborPath  = "/ipfs/" + root.String() + "/subdir/dag-cbor-document"
+		dagCborRoots = dirRoots + "," + dagCborCID
+	)
+
+	t.Run("Control-Cache-Immutable is not immutable for directories", func(t *testing.T) {
```
lidel:
```suggestion
	t.Run("Cache-Control is not immutable on generated /ipfs/  HTML dir listings", func(t *testing.T) {
```

gateway/gateway_test.go
```diff
@@ -504,317 +116,614 @@ func TestPretty404(t *testing.T) {
...
+		if ok {
+			for _, hdr := range hdrs {
+				assert.NotContains(t, hdr, "immutable", "unexpected Cache-Control: immutable on directory listing")
+			}
+		}
+	})
 
-		req, err := http.NewRequest(http.MethodGet, ts.URL+"/ipfs/"+root.String()+"/nonexisting-link", nil)
-		assert.Nil(t, err)
+	t.Run("ETag contains expected values", func(t *testing.T) {
```
lidel:
```suggestion
	t.Run("ETag is based on CID and response format", func(t *testing.T) {
```

gateway/handler_block.go
```diff
@@ -5,23 +5,25 @@ import (
...
 
-	pathMetadata, data, err := i.backend.GetBlock(ctx, imPath)
-	if !i.handleRequestErrors(w, r, contentPath, err) {
+	pathMetadata, data, err := i.backend.GetBlock(ctx, rq.maybeResolvedPath())
+	if !i.handleRequestErrors(w, r, rq.contentPath, err) {
 		return false
 	}
+	if rq.pathMetadata == nil {
+		rq.pathMetadata = &pathMetadata
+	}
```
lidel: nit: this repeats, we have this in multiple places (e.g. here, `handler_codec.go` and `handler_defaults.go`),  could we make `setIpfsRootsHeader` accept both `rq` and `&pathMetadata`? 
and do the opportunistic nil check and set inside?

gateway/handler.go
```diff
@@ -171,6 +172,29 @@ func (i *handler) optionsHandler(w http.ResponseWriter, r *http.Request) {
...
+	contentPath    ipath.Path
+	responseFormat string
+	responseParams map[string]string
+
+	// Defined for non IPNS Record requests.
+	immutablePath ImmutablePath
+
+	// Defined if resolution has already happened.
+	pathMetadata *ContentPathMetadata
+	resolvedPath *ImmutablePath
```
lidel: Hm.. this feels a bit awkward, isn't the `resolvedPath` the same as `*ContentPathMetadata.LastSegment.String()`, just  different type?
Can we live with only one of them? (`pathMetadata.LastSegment`) ?

gateway/handler.go
```diff
@@ -171,6 +172,29 @@ func (i *handler) optionsHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (rq *requestData) maybeResolvedPath() ImmutablePath {
```
lidel: Would this be a better description of that this does?
I really want to avoid `maybe` as is pretty ambiguous :dizzy:  

```suggestion
// mostlyResolvedPath is an opportunistic optimization that returns the mostly resolved version of ImmutablePath available.
// It does not guarantee it is fully resolved, nor that it is the original.
func (rq *requestData) mostlyResolvedPath() ImmutablePath {
```

gateway/gateway_test.go
```diff
@@ -504,317 +116,614 @@ func TestPretty404(t *testing.T) {
...
+
 	ts, _, root := newTestServerAndNode(t, nil)
-	t.Logf("test server url: %s", ts.URL)
 
-	t.Run("plain error if request does not have Accept: text/html", func(t *testing.T) {
-		t.Parallel()
+	var (
+		dirCID   = "bafybeihta5xfgxcmyxyq6druvidc7es6ogffdd6zel22l3y4wddju5xxsu"
+		dirPath  = "/ipfs/" + root.String() + "/subdir/"
+		dirRoots = "bafybeifhvgr4ufgwpoj2iyrymrigjhpexvsyfm2elafebkmrei4skunihe," + dirCID
```
lidel: :thought_balloon: The setup with single `./testdata/fixtures.car` starts to feel a bit brittle.

Every time `./testdata/fixtures.car` is updated when we add new tests  → the  `root.String()` CID changes here  → the first CID  in `dirRoots` below will no longer match, and someone will have to update these to new correct values, which introduces risk of test regression if not careful.

Would be less brittle if we split `fixtures.car` and have a separate `./testdata/roots-header.car` fixture for this + hardcoded explicit CIDs here, so we never have to change this, so if tests fail in the future, we know it's a bug, and not  maintenance related to updating `root.String()`?

gateway/gateway_test.go
```diff
@@ -504,317 +116,614 @@ func TestPretty404(t *testing.T) {
...
+
+		fileCID   = "bafkreiba3vpkcqpc6xtp3hsatzcod6iwneouzjoq7ymy4m2js6gc3czt6i"
+		filePath  = "/ipfs/" + root.String() + "/subdir/fnord"
+		fileRoots = dirRoots + "," + fileCID
+
+		dagCborCID   = "bafyreiaocls5bt2ha5vszv5pwz34zzcdf3axk3uqa56bgsgvlkbezw67hq"
+		dagCborPath  = "/ipfs/" + root.String() + "/subdir/dag-cbor-document"
+		dagCborRoots = dirRoots + "," + dagCborCID
```
lidel, line 150: We need to add `X-Ipfs-Roots` test for HAMT.
We should have test for `/ipfs/cid/hamt/file` and only have logical roots (one per path segment, and not all internal HAMT ones)

lidel (changes requested):
Thank you @hacdias, especially for adding a lot of missing tests. :heart: 

Looks good, but  we should add one test for HAMT (usual edge case) just to be sure + see some questions/style suggestions inline.

gateway/handler.go
hacdias: We nede `pathMetada`. I kept that.

gateway/gateway_test.go
```diff
@@ -496,325 +102,660 @@ func TestPretty404(t *testing.T) {
...
+		dirPath  = "/ipfs/" + rootCID + "/subdir/"
+		dirRoots = rootCID + "," + dirCID
 
-		body, err := io.ReadAll(res.Body)
-		assert.Nil(t, err)
-		assert.NotContains(t, string(body), "<!DOCTYPE html>")
-	})
+		hamtCID   = "bafybeidbclfqleg2uojchspzd4bob56dqetqjsj27gy2cq3klkkgxtpn4i"
+		hamtPath  = "/ipfs/" + rootCID + "/hamt/"
+		hamtRoots = rootCID + "," + hamtCID
```
lidel: nit: this will not show issue with X-Roots-Path, we need `/hamt` that is more than one level dep, and load file that is in `/ipfs/bafybeidbcy4u6y55gsemlubd64zk53xoxs73ifd6rieejxcr7xy46mjvky/hamt/some-file` and then confirm the `X-Ipfs-Roots` includes only 3 CIDs for each path segment, and no internal ones from `/hamt/`.


There is some prior art fixture [here in gateway conformance](https://github.com/ipfs/gateway-conformance/blob/43c0842a1041a2748a74a196e370b84ef02b12a5/tests/trustless_gateway_car_test.go#L44) if that helps.

@hacdias  I don't want to block this PR on this, as it already brings a lot of other useful tests, so if you want fine to merge as-is and fill issue to fix this test in future PR.

lidel (approved):
lgtm (only a small hamt testing nit inline). Improved test coverage here is really valuable, thanks! :+1:

gateway/gateway_test.go
hacdias: Ah, yes. I updated the test as I even used the same fixture. Will merge once CI is green.

PR #338, opened 2023-06-07
chore: do not rely on deprecated logger

hacdias:
Replace `go-log` by `go-log/v2` everywhere.


Jorropo (approved):
idk what the implications are but the actual code changes make sense.

PR #339, opened 2023-06-07
feat(ipns): refactored IPNS package with lean records

### LINKED ISSUE #335: Improvements to the `ipns` package
This is a tracking issue to improve the `ipns` package, such as cleaning the verification process, types, etc. This can be done at the same time of https://github.com/ipfs/specs/issues/376 for example, or not.

## Things to take into account

- https://github.com/ipfs/specs/issues/376
- https://github.com/ipfs/go-ipns/issues/13 (link specs)

hacdias:
Closes #335. Progress towards https://github.com/ipfs/specs/issues/376.

- [x] Kubo PR: https://github.com/ipfs/kubo/pull/9932
- [x] Updated https://github.com/libp2p/go-libp2p-kad-dht to fix dependency on old `go-ipns`, which was causing problems with the `namesys` tests.


ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
-	// KeyBook, if non-nil, will be used to lookup keys for validating IPNS
-	// records.
-	KeyBook pstore.KeyBook
+// Record represents an [IPNS Record].
+//
+// [IPNS Record]: https://specs.ipfs.tech/ipns/ipns-record/
+type Record struct {
+	pb   ipns_pb.IpnsEntry
+	node datamodel.Node
 }
```
hacdias: The way I implemented the `Record` abstraction keeps the Protobuf in memory. Therefore, `NewRecord` takes the option `CompatibleWithV1`. I decided to do this because it makes it much easier for testing.

However, it could be interesting to represent `Record` and only make the decision of compatibility with V1 when marshaling. This is still up for debate. It will make testing harder, but still possible.

```suggestion
// Record represents an [IPNS Record].
//
// [IPNS Record]: https://specs.ipfs.tech/ipns/ipns-record/
type Record struct {
	// For validation of unmarshalled records.
	signatureV1 []byte
	signatureV2 []byte
	
	pk   ic.PubKey
	node datamodel.Node
}
```

ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
+func RecordKey(pid peer.ID) string {
+	cid := peer.ToCid(pid)
+	encoded, err := cid.StringOfBase(multibase.Base36)
+	if err != nil {
+		panic("cid.StringOfBase was called with wrong parameters")
 	}
 
-	return i, nil
+	return "/ipns/" + encoded
 }
```
hacdias, line 468: This will likely create confusion. `RecordKey` now returns the human readable record key. Before, it returned what the `RoutingRecordKey` now does return. Since the arguments are the same, it's hard to spot. I added it clearly in the Changelog. We are publishing a breaking change after all.

If someone else prefers to make this more safe, I would suggest just renaming to `Key(pid peer.ID)` and `RoutingKey(pid peer.ID)`. That would force people to look at the new options and see the difference.

hacdias:
Kubo doesn't build (https://github.com/ipfs/boxo/pull/339) because it depends on the old delegated routing package (https://github.com/ipfs/go-delegated-routing) because it still supports the wonders of Reframe. Either we fix that package to support this new IPNS package, or (preferred) we completely remove the dependency in Kubo, removing Reframe. 

Another option is to simply keep the name of the function `RecordKey` and change `Key` to `HumanKey`. But I don't like that option as it doesn't make it clear. I want the routing-specific key to have "routing" in the function name.

ipns/validation.go
```diff
@@ -0,0 +1,256 @@
...
+		return err
+	}
+
+	return Validate(r, pk)
+}
+
+// Validates validates the given IPNS Record against the given Public Key, following
+// the [Record Verification] specification.
+//
+// [Record Verification]: https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format
```
lidel:
```suggestion
// [Record Verification]: https://specs.ipfs.tech/ipns/ipns-record/#record-verification
```

ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
-		log.Debugf("failed to parse ipns record key %s into peer ID", pidString)
-		return ErrKeyFormat
+		return "", err
 	}
 
-	pubk, err := v.getPublicKey(pid, entry)
+	// Maybe it's an old style record.
+	if mh, err := multihash.Cast(value); err == nil {
+		return path.FromCid(cid.NewCidV0(mh)), nil
+	}
```
lidel: :thought_balloon:  I wonder if this is still necessary, these were pre-V1 things, not even sure if JS ever supported this?  Raw Multihash is not part of [specs](https://specs.ipfs.tech/ipns/ipns-record/#ipns-record) which state Value is a path (string).

If this was never supported by JS, we probably want to remove this, rather than add to the spec, since parsing the value as either  String path or  `[]byte` Multihash may cause bugs in languages other than GO.

ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
+		return 1, nil
+	} else if bt.After(at) {
+		return -1, nil
+	}
+
+	return 0, nil
+}
+
+// EmbedPublicKey embeds the given public key in the given [Record]. While not
+// strictly required, some notes (e.g., DHT servers), may reject IPNS Records
```
lidel:
```suggestion
// strictly required, some nodes (e.g., DHT servers), may reject IPNS Records
```

ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
+func processOptions(opts ...Option) *options {
+	options := &options{}
+	for _, opt := range opts {
+		opt(options)
+	}
+	return options
+}
+
+// NewRecord creates a new IPNS [Record] and signs it with the given private key.
+// This function does not embed the public key. To do so, call [EmbedPublicKey].
```
lidel: @hacdias would calling EmbedPublicKey be up to library user, such as Kubo?
Non-PL users may not know/care when to do it and why.

Perhaps:
- by default, call `EmbedPublicKey` inside `NewRecord` for keys other than `ed25519`  (ensures RSA records are always self-contained and verifiable).
- for cases where someone wants to explicitly disable embedding of RSA, we would have `WithPublicKey(bool)` `Option` which overrides the default

ipns/record.go
```diff
@@ -3,124 +3,466 @@ package ipns
...
+		return pid.ExtractPublicKey()
+	}
+}
+
+// RoutingKey returns the IPNS routing key for the given [peer.ID]. Note
+// that the intended use of this function is for routing only. The output of this
+// function is not human readable. For a human readable version, see [Key].
+func RoutingKey(pid peer.ID) string {
+	return "/ipns/" + string(pid)
+}
```
lidel: Could we make this return `byte[]`  and not `string`, just to make it extra clear this is binary and people who want string should use the `Key` one?

```suggestion
// RoutingKey returns the binary IPNS routing key for the given [peer.ID]. Note
// that the intended use of this function is for routing only. The output of this
// function is binary, not human readable. For a string, human readable version, see [Key].
func RoutingKey(pid peer.ID) byte[] {
    var buffer bytes.Buffer
	buffer.WriteString("/ipns/")
	buffer.Write(pid)
	return buffer.Bytes()
}
```

ipns/validation.go
```diff
@@ -0,0 +1,256 @@
...
+		return fmt.Errorf("could not compute signature data: %w", err)
+	}
+
+	// (6) Verify signature against concatenation.
+	if ok, err := pk.Verify(sig2Data, r.pb.GetSignatureV2()); err != nil || !ok {
+		return ErrSignature
+	}
+
+	// (5) Ensure that CBOR data matches Protobuf, only if Signature V1 is present.
+	if len(r.pb.GetSignatureV1()) != 0 {
```
lidel: Would this be safer? Just in case someone produces `r.pb` with Value but no SignatureV1. 
There could be some legacy code that validates record using library, but then uses the Value from PB instead of CBOR. 

```suggestion
	// (5) Ensure that CBOR data matches Protobuf, only if non-CBOR Value or SignatureV1 are present.
	if len(r.pb.GetSignatureV1()) != 0 || len(r.pb.GetValue()) != 0 {
```

ipns/validation.go
```diff
@@ -0,0 +1,256 @@
...
+
+	r, err := UnmarshalRecord(value)
+	if err != nil {
+		return ErrBadRecord
+	}
+
+	// Get the public key defined by the ipns path
+	pid, err := peer.IDFromBytes([]byte(pidString))
+	if err != nil {
+		log.Debugf("failed to parse ipns record key %s into peer ID", pidString)
```
lidel: we sometimes have it in binary form, sometimes string, using `%q` removes need to think about it
```suggestion
		log.Debugf("failed to parse ipns record key %q into peer ID", pidString)
```

ipns/validation.go
```diff
@@ -0,0 +1,256 @@
...
+	switch pk, err := ExtractPublicKey(r, pid); err {
+	case peer.ErrNoPublicKey:
+	case nil:
+		return pk, nil
+	default:
+		return nil, err
+	}
+
+	if v.KeyBook == nil {
+		log.Debugf("public key with hash %s not found in IPNS record and no peer store provided", pid)
```
lidel:
```suggestion
		log.Debugf("public key with hash %q not found in IPNS record and no peer store provided", pid)
```

ipns/validation.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil, ErrPublicKeyNotFound
+	}
+
+	pk := v.KeyBook.PubKey(pid)
+	if pk == nil {
+		log.Debugf("public key with hash %s not found in peer store", pid)
```
lidel:
```suggestion
		log.Debugf("public key with hash %q not found in peer store", pid)
```

lidel (changes requested):
@hacdias my understanding is that IPNI no longer uses Reframe endpoints (https://github.com/ipfs/kubo/issues/9823#issuecomment-1572198887), they ask their users to set up `/routing/v1` integration, so we can finally remove all Reframe things.

+ some feedback/questions inline

ipns/record.go
hacdias: @lidel I would be all up for this idea, but the libp2p `record.Validator` interface requires a `key string` 😃 If we do return `[]byte`, then the caller would always have to transform it to a `string` 🙃

ipns/validation.go
hacdias: I'm trying to understand why would there be a value but **no signature**? Can you give me more context before I add this?

ipns/record.go
hacdias: I like the idea. In the best case scenario, we could completely remove this function. `EmbedPublicKey` seems to only be used inside `namesys` (no calls from Kubo).

Namesys uses it for:

1. Testing the validation, i.e., embed a _wrong public key_, and ensuring the validator catches that.
2. Ensure that the records put into routing include the public key.

Therefore, it means `ipns.EmbedPublicKey` allows the user to make bad decisions, like embedding the wrong public key. What I'm going to do is to completely remove that function.

For (1) The validator uses `ipns.Validator`, which is already tested by us. Therefore, those tests can be safely removed.

For (2), the public key will be added automatically by default if the key is not Ed25519, so there should not be issues.

hacdias: @lidel please see my commit and see if you agree with this. We can also keep the function and make it clear that it is not **intended** to be used as it does not validate if the public key matches the one from the signature. But I would prefer to simply not have this as a public function that others can use, doesn't make sense. If people need to test, they can generate the protobuf directly with the wrong data.

hacdias:
@lidel  I also just discovered we have a `coreiface.IpnsEntry`. I would like to remove it, but I don't think that's the best idea because it includes the `Name` field, which we do not have in `ipns.Record`. Wdyt?

ipns/validation.go
lidel: I'm thinking malicious scenarios against users of this library. 

Someone could create a record with valid V2 signature, but no V1 signature.
In such case, the `r.pb.Value` could not match the value in CBOR, and validation would not catch it.
At the same time, after validation passes, external some software may use value from not-signed `r.pb.Value` field instead of the signed one in CBOR, because historically the PB was always present.

hacdias: Sounds fair, added.

lidel:
@hacdias try to clean this up and remove it if possible – if we don't clean this up now, it may be long time til someone looks at this.

We have too many abstractions for "ipns record". I think "IpnsEntry" with `Name()` is mostly used in Kubo internals like  `ipfs name` commands ([here](https://github.com/ipfs/kubo/blob/344431d6caeca8238b275a2ba83ad8bf0019a3ac/core/commands/name/name.go#L24-L26), [here](https://github.com/ipfs/kubo/blob/344431d6caeca8238b275a2ba83ad8bf0019a3ac/client/rpc/name.go#L17)).
Probably fine to (re)move `coreiface.IpnsEntry` from boxo to Kubo (or make Kubo use [this one](https://github.com/ipfs/kubo/blob/344431d6caeca8238b275a2ba83ad8bf0019a3ac/core/commands/name/name.go#L24-L26) directly)

ipns/record.go
lidel: I know, I think that is a feature: it highlights the problem to external users, and does not try to hide it inside our library.
If we ever refactor record  Validator interface, it will make it easier to switch to `byte[]` there.

An alternative would be to add a specific helper, like `BinaryRoutingKeyAsString` to make it extra clear and not hide conversion.

Either works, just want to help users understand why there is `/ipfs/Ò̴̺̱̝̹̀̆̃͑ͧ̕͞m̴̡͙̫̹͕̲̱͔̗̹̹̺̺̤̼̺͉͚̙̌ͮ̀̌̓͋̅̏̈́̓̅̍̾̓̚g̛͙̬̬̥̜̫̮͖͉̱̗̟͎̘̓ͪ͊̿̽ͥͮ͗̏̓͐͂ͬ͆ͨͨ͂͝ ̡͖̠͓̭͖̯̲̬͈̗̭̈̆͆̈̒̇̓͋̓̉ͩͭͥ̒̓ͮ͝ͅw̺͓͓̼̬̯̲̪̙̰̗͙̞̳͕͇̅̑ͨ̍ͮ̈́̃͋͂̓ͬ͒͜͜͢͡h̴͖̤̦̱̰̞̲̱͙̭̗̫͚̗͂̏ͧ͌̃ͤ͋̅ͭ̎̌̆͑̔͢ͅa̷̛̰̻͉̟̻̦̟̱͉̘̘̭̩̗͇͙̪͑̐̑ͨ̾ͪ̾̊̒̕͝ͅt͔̰͓̬̼̖̟͙̑̏̇͐̐͆͋ͯ͠ ̷̛͚̣̤̝̜͌͌ͪͣ͋ͤ̃ͮͣͫ̿̇͝i̷̡̜͕̱͍͉̽ͥ̀̾̀̀ş̶̼͕̖͔͎̦͗͋̍ͯ̔ͨͪ̓͑̐͑̈́́ ̸̱̪͈͓̹̏ͩͪ͊͑ͭ͝ţ̷͗̐͋̍ͮ̇̓͑͋̚͏̖̹̹̬̦̻̭͓̯͔̜̼̯̹̯ḩ̴̘̯̠̯̋̎̑̋́ͦ͝ͅi̴̧̾ͣ̀̌̋̆̈́͏̮͇͓͇͕̪̠͇̻͇͖̝̳̞͙s̴̷̐ͥ̌͆ͣͩ̓ͬ̽ͫ҉͏̫̦͈̱̖̭͎̰̺͙̥̩̩͉͙̞̳ͅ` in their logs somehow, especially when they look at Go code when trying to reimplement things in other languages.  :see_no_evil:

ipns/record.go
lidel: Removal of `ipns.EmbedPublicKe` lgtm. Safe defaults  + option to override them is :ok_hand:, thanks!

ipns/record.go
hacdias: I'll change it then, hopefully the validator will be updated at some point 🙃

hacdias:
@lidel cleaning that will be extremely complicated. For example, returning`ipns.Record` is not possible (or is but then we'd nede to call resolve in the publish function) because the requests inside going through namesys, which never returns anything except for an error. What I would suggest is to change the signature of the fucntion to only return `string, error` (essentially ipns record key and error) or just an error. Wdyt?

lidel:
Re  

https://github.com/ipfs/boxo/blob/6f82d77995476b0d85722539230a50a321706ead/coreiface/name.go#LL37C5-L37C5

Notes from 1:1:
- introduce `ipns.Name` type and use it instead of `IpnsEntry` returned by `Publish` here, as only thing that matters / may be unknown is the actual key (publisher uses pet  name like `self`)
  -  make it based on `byte[]`  (we logically decouple ipns names from libp2p peerids;  both are `byte[]`  Multiaddrs, but you can use ipns without leaking libp2p types)

hacdias:
Note to self: see if it's feasible to make the `pb` stuff **internal** such that outsiders of this library can't use it.

hacdias:
@lidel suggestions implemented

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+// this function is binary, not human readable. For a human-readable string, see [Name.Key].
+func (n Name) RoutingKey() []byte {
+	var buffer bytes.Buffer
+	buffer.WriteString("/ipns/")
+	buffer.Write(n.src)
+	return buffer.Bytes()
+}
+
+// String returns the human-readable IPNS Name, prefixed with "/ipns/".
+func (n Name) String() string {
```
lidel: @hacdias  mind removing `/ipns/` prefix from this?

Prefix  is not part of the name, `String()` should return cidv1base36 alone, otherwise everyone will have to trim it.
If we need it, add separate `ContentPath()`  or something.

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+	"strings"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/multiformats/go-multibase"
+)
+
+// Name represents an [IPNS Name].
+//
+// [IPNS Name]: https://specs.ipfs.tech/ipns/ipns-record/#ipns-name
+type Name struct {
```
lidel, line 26: :thought_balloon: Crazy idea: would it be easier to reason if we had  `peer.ID` and   `ipns.ID`? or worse? :upside_down_face: 
We could clean up naming in spec too, if needed. (Fine to keep `Name`, just thought if we want to change the term, it is good time, once people start using `ipns.Name` it will be difficult to change)

```suggestion
type ID struct {
```

ipns/pb/record.proto
```diff
@@ -0,0 +1,21 @@
...
+
+	optional bytes value = 1;
+	optional bytes signatureV1 = 2;
+	optional ValidityType validityType = 3;
+	optional bytes validity = 4;
+	optional uint64 sequence = 5;
+	optional uint64 ttl = 6;
+	optional bytes pubKey = 7;
+	optional bytes signatureV2 = 8;
+	optional bytes data = 9;
```
lidel: Short tldr in case someone eyeballs protobuf definition:

```suggestion
 	// 1-6 are legacy fields used only in V1+V2 records
 	optional bytes value = 1;
	optional bytes signatureV1 = 2;
	optional ValidityType validityType = 3;
	optional bytes validity = 4;
	optional uint64 sequence = 5;
	optional uint64 ttl = 6;
	// 7-9 are V2 records
	optional bytes pubKey = 7;
	optional bytes signatureV2 = 8;
	optional bytes data = 9;
```

ipns/pb/record.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package github.com.boxo.ipns.pb;
+
+option go_package = ".;pb";
+
+message IpnsRecord {
```
lidel, line 8:
```suggestion
// https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format
message IpnsRecord {
```

examples/gateway/proxy/routing.go
```diff
@@ -77,8 +72,8 @@ func (ps *proxyRouting) SearchValue(ctx context.Context, k string, opts ...routi
 	return ch, nil
 }
 
-func (ps *proxyRouting) fetch(ctx context.Context, id peer.ID) ([]byte, error) {
-	urlStr := fmt.Sprintf("%s/ipns/%s", ps.gatewayURL, peer.ToCid(id).String())
+func (ps *proxyRouting) fetch(ctx context.Context, name ipns.Name) ([]byte, error) {
+	urlStr := fmt.Sprintf("%s/%s", ps.gatewayURL, name.String())
```
lidel: :thought_balloon:  ok this diff suggests we add to much implicit magic.  `name ipns.Name`  should not include prefix here, `/ipns/` should be explicit, not implicit. People will want to use this outside of IPFS context, and that means they will remove prefix over and over again. 

Update: lol just noticed https://github.com/ipfs/kubo/pull/9932#discussion_r1229274237 :D  yeah, we should remove prefix.

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/multiformats/go-multibase"
+)
+
+// Name represents an [IPNS Name].
```
lidel:
```suggestion
// Name represents a Multihash of a serialized public key according to [IPNS Name] specs.
```

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+	return Name{src: []byte(pid)}
+}
+
+// RoutingKey returns the binary IPNS Routing Key for the given [Name]. Note that
+// the intended use of this function is for routing purposes only. The output of
+// this function is binary, not human readable. For a human-readable string, see [Name.Key].
+func (n Name) RoutingKey() []byte {
+	var buffer bytes.Buffer
+	buffer.WriteString("/ipns/")
+	buffer.Write(n.src)
```
lidel:
```suggestion
	buffer.Write(n.src) // Note: we append raw multihash bytes (no multibase)
```

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// NameFromPeer creates a [Name] from the given [peer.ID].
+func NameFromPeer(pid peer.ID) Name {
+	return Name{src: []byte(pid)}
+}
+
+// RoutingKey returns the binary IPNS Routing Key for the given [Name]. Note that
+// the intended use of this function is for routing purposes only. The output of
+// this function is binary, not human readable. For a human-readable string, see [Name.Key].
```
lidel:
```suggestion
// RoutingKey returns the binary IPNS Routing Key for the given [Name]. Note that
// the intended use of this function is for [routing purposes] only. The output of
// this function is binary, not human readable. For a human-readable string, see [Name.Key].
//
// [routing purposes]: https://specs.ipfs.tech/ipns/ipns-record/#routing-record
```

ipns/name.go
```diff
@@ -0,0 +1,94 @@
...
+func (n Name) RoutingKey() []byte {
+	var buffer bytes.Buffer
+	buffer.WriteString("/ipns/")
+	buffer.Write(n.src)
+	return buffer.Bytes()
+}
+
+// String returns the human-readable IPNS Name, prefixed with "/ipns/".
+func (n Name) String() string {
+	cid := peer.ToCid(n.pid())
```
lidel: :thought_balloon: thoughts on removing dep. on `peer.ToCid` and adding ipns-specific func?

```go
// Cid encodes IPNS Name as a CID of the public key.
//
// If the IPNS Name is invalid (e.g., empty), this will return the empty CID.
func (n Name) Cid() cid.Cid {
       m, err := mh.Cast([]byte(n))
       if err != nil {
               return cid.Undef
       }
       return cid.NewCidV1(cid.Libp2pKey, m)
}
```

while at it we could also:

```go
// FromCid converts a CID to an IPNS Name, if possible.
func NameFromCid(c cid.Cid) (Name, error) {
       code := mc.Code(c.Type())
       if code != mc.Libp2pKey {
               return "", fmt.Errorf("CID codec %q is not allowed for IPNS Names, use  %q instead", code, mc.Libp2pKey)
       }
       return Name(c.Hash()), nil
}
```

I feel we will need them anyway, see places like https://github.com/ipfs/boxo/pull/333#pullrequestreview-1480332571

lidel (changes requested):
Thanks! Mostly lgtm, some cosmetic changes inline.

Next steps:
- IPIP that documents this, so we can do the same in js-ipns (not a blocker)
- some interop test between boxo/ipns and js-ipns  explicitly doing V1+V2 and  V2-only – unsure where it should live, https://github.com/ipfs/interop/blob/master/test/ipns.js ?

ipns/name.go
hacdias: I like the idea of `Name`, but I'm not opposed to `ID`. My "issue" i that the spec says:

> A good practice is to prefix IPNS Name with /ipns/ namespace, and refer to IPNS addresses as /ipns/{ipns-name} (or /ipns/{libp2p-key}).

Nevertheless, I think IPNS **Name** is better. We don't talk about **IDs** in the specification, and a lot of the code is built around the **Name**.

hacdias:
> some interop test between boxo/ipns and js-ipns explicitly doing V1+V2 and V2-only

Not sure if that's a blocker or not, but the interop tests you linked are for `kubo` and `js-ipfs`. I would assume those should be written once the IPIP is up and/or the `js-ipns` implementation is updated. At the moment, I think `js-ipns` cannot validate v2 only records: https://github.com/ipfs/js-ipns/blob/master/src/validator.ts#L33.

ipns/name.go
lidel: Ack. Let's keep Name.

I remember I've added that sentence to specs without giving much thought, as a best practice, but it seems confusing, maybe it more clear?


```diff
- A good practice is to prefix IPNS Name with /ipns/ namespace, and refer to IPNS addresses as /ipns/{ipns-name} (or /ipns/{libp2p-key}).
+ In context outside IPNS, or when better UX is required due to possibility of content identifiers being presented on the screen, one can prefix IPNS Names with /ipns/ namespace, and refer to IPNS addresses as /ipns/{ipns-name} (or /ipns/{libp2p-key}), differentiating them clearly from immutable `/ipfs/{cid}`.

Comment (hacdias) [reply ipns/name.go:26]: Yes, we can update the spec to make it clearer, and keep `Name` with that name (perhaps pun intended).

**Comment by lidel** [comment]
Yes, I would block this PR on  js-ipns doing the same default + being able to validate the default output.

V2-only is opt-in in Kubo, but  `boxo/ipns` here produces V2 by default.
This is a breaking change, unless js-ipns is marks V2-only as valid, and we had it around for a while. 

@hacdias I think we will have less headache if we do what we did in Kubo, and keep this library producing V1+V2 by default like it did before, and making V2-only opt-in for now. The goal here is to allow existence of V2-only, not make it default output of the library.

```go
func processOptions(opts ...Option) *options {
	options := &options{}options := &options{}
	// TODO: produce V2-only records by default after IPIP-TODO ships with Kubo and Helia for at least 6 months
	options.CompatibleWithV1 = true
```

Once we have IPIP, and implement it in JS, and once we have support for validating V2-only everywhere for a while, then we will make it default, but for now let's keep producing V1+V2.

If we do that, then we can land this PR without waiting for IPIP, as we don't change the default behavior.

hacdias:
@lidel I have updated this PR to produce V1+V2 by default (and adapted tests). Kubo PR updated to run with latest code. I can take care of `js-ipns` and the IPIP. I'll write a note.

ipns/pb/Makefile
lidel, line 1: Since `ipns/pb/Makefile` is removed, how one would regenerate `record.pb.go` if we ever add a new field?
Maybe restore it?

ipns/errors.go
```diff
@@ -4,41 +4,45 @@ import (
...
+// MaxRecordSize is the IPNS Record [size limit].
+//
+// [size limit]: https://specs.ipfs.tech/ipns/ipns-record/#record-size-limit
+const MaxRecordSize int = 10 << (10 * 1)
 
-// ErrUnrecognizedValidity is returned when an IpnsRecord has an
-// unknown validity type.
-var ErrUnrecognizedValidity = errors.New("unrecognized validity type")
+// ErrExpiredRecord is returned when an IPNS [Record] is invalid due to being expired.
+var ErrExpiredRecord = errors.New("the IPNS record is expired")
```
lidel: nit: (here and all below) – we return multierror + most likely will wrap them upstream, perhaps we can drop "the IPNS" from all of them? 
```suggestion
var ErrExpiredRecord = errors.New("record is expired")
```

ipns/errors.go
```diff
@@ -4,41 +4,45 @@ import (
...
+// does not match the expected public key.
+var ErrPublicKeyMismatch = errors.New("the IPNS record public key does not match the expected public key")
 
-// 10 KiB limit defined in https://github.com/ipfs/specs/pull/319
-const MaxRecordSize int = 10 << (10 * 1)
+// ErrPublicKeyNotFound is returned when the public key is not found.
+var ErrPublicKeyNotFound = errors.New("public key not found")
+
+// ErrInvalidPublicKey is returned when an IPNS [Record] has an invalid public key,
+var ErrInvalidPublicKey = errors.New("the IPNS record public key invalid")
```
lidel: we use this outside record (when parsing identity CID), so maybe:
```suggestion
var ErrInvalidPublicKey = errors.New("invalid public key")
```

ipns/errors.go
```diff
@@ -4,41 +4,45 @@ import (
...
+var ErrSignature = errors.New("the IPNS record signature verification failed")
+
+// ErrInvalidName is returned when an IPNS [Name] is invalid.
+var ErrInvalidName = errors.New("the IPNS name is invalid")
 
-// ErrRecordSize should be returned when an ipns record is
-// invalid due to being too big
-var ErrRecordSize = errors.New("record exceeds allowed size limit")
+// ErrInvalidPath is returned when an IPNS Record has an invalid path.
+var ErrInvalidPath = errors.New("the IPNS record path invalid")
```
lidel:
```suggestion
// ErrInvalidPath is returned when an IPNS Record's Value is an invalid path.
var ErrInvalidPath = errors.New("value is not a valid content path")
```

ipns/errors.go
```diff
@@ -4,41 +4,45 @@ import (
...
-var ErrKeyFormat = errors.New("record key could not be parsed into peer ID")
+// ErrRecordSize is returned when an IPNS [Record] exceeds the maximum size.
+var ErrRecordSize = errors.New("the IPNS record exceeds allowed size limit")
 
-// ErrPublicKeyNotFound should be returned when the public key
-// corresponding to the ipns record path cannot be retrieved
-// from the peer store
-var ErrPublicKeyNotFound = errors.New("public key not found in peer store")
+// ErrDataMissing is returned when an IPNS [Record] is missing the data field.
+var ErrDataMissing = errors.New("the IPNS record is missing the data field")
```
lidel:
```suggestion
var ErrDataMissing = errors.New("record is missing the dag-cbor data field")
```

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+* 🛠 The `ipns` package has been refactored. You should no longer use the direct Protobuf
+  version of the IPNS Record. Instead, we have a shiny new `ipns.Record` type that wraps
+  all the required functionality to work the best as possible with IPNS v2 Records. Please
+  check the [documentation](https://pkg.go.dev/github.com/ipfs/boxo/ipns) for more information.
```
```suggestion
  check the [documentation](https://pkg.go.dev/github.com/ipfs/boxo/ipns) for more information,
  and follow [ipfs/specs#376](https://github.com/ipfs/specs/issues/376) for related IPIP.
```

lidel (approved):
Thanks @hacdias!  I think technically it is ready.
Small nits in comments inline, but once addressed, feel free to merge and bubble up to https://github.com/ipfs/kubo/pull/9932 when you feel Kubo is ready :pray: 

ps. I think in cases like this, when we have 40 files modified, fine to just add commits to PR, you dont need to rebase and force push. Smaller deltas make it easier to review too – we then squash at the end during merge.

ipns/pb/Makefile
hacdias: `go generate` does it all 😄 I decided to depend on less external dependencies as possible.

https://github.com/ipfs/boxo/blob/d8155fab846ec97a5c5d18224caaa53f5de6103f/ipns/record.go#L1-L2

PR #347, opened 2023-06-12
New very fast unixfs implementation.

Jorropo:
Depends on https://github.com/ipfs/go-cid/pull/159.

`DagPB + Unixfs` parsing with `gogoproto` is `~1500ns` here it is `~250ns`.
I'll do `boxo/ipld/unixfs` vs `go-unixfsnode` vs `boxo/unixfs` decoding in the future.


Jorropo:
Will release this in a new repo

PR #358, opened 2023-06-19
fix(gateway): allow CAR trustless requests with path

### LINKED ISSUE #356: Trustless Gateway mode rejects CAR requests allowed under IPIP-402
While browsing gateway code I was bringing into boost, I noticed this function, that gets called when DeserializedResponses is false:

https://github.com/ipfs/boxo/blob/main/gateway/handler.go#L334

This appears to reject all requests for trustless gateway that contain paths. As I understand it, if the response format is CAR, then these requests should not get rejected under IPIP-402.

hacdias:
Fixes #356.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -331,7 +331,7 @@ func (i *handler) isTrustlessRequest(contentPath ipath.Path, responseFormat stri
 	// Only allow "/{#1}/{#2}"-like paths.
 	trimmedPath := strings.Trim(contentPath.String(), "/")
 	pathComponents := strings.Split(trimmedPath, "/")
-	if len(pathComponents) != 2 {
+	if responseFormat != carResponseFormat && len(pathComponents) != 2 {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


gateway/handler.go
```diff
@@ -331,7 +331,7 @@ func (i *handler) isTrustlessRequest(contentPath ipath.Path, responseFormat stri
 	// Only allow "/{#1}/{#2}"-like paths.
 	trimmedPath := strings.Trim(contentPath.String(), "/")
 	pathComponents := strings.Split(trimmedPath, "/")
-	if len(pathComponents) != 2 {
+	if responseFormat != carResponseFormat && len(pathComponents) != 2 {
```
lidel, line 334: Can we add a regression test for this when deserialized responses are disabled?
(we need both positive and negative for both trustless formats)
- `/ipfs/cid?format=raw`  → ok
- `/ipfs/cid/subpath?format=raw`  → expect error
- `/ipfs/cid?format=car`  → ok
- `/ipfs/cid/subpath?format=car`  → ok

lidel (approved):
Thank you @hacdias and @hannahhoward  for catching this.
Lgtm, but needs regression test mentioned below.

PR #367, opened 2023-06-19
refactor: golang-lru → golang-lru/v2

### LINKED ISSUE #366: Remove dependency on unused ARC algorithm from hashicorp/golang-lru
`arc_cache.go` seems modified to avoid hitting the patent (https://github.com/ipfs/go-ipfs-blockstore/pull/20) of ARC algorithm, but I was confused as the code comments still state that it uses the ARC algorithm.

https://github.com/ipfs/go-ipfs-blockstore/blob/03acccfc7eae2b977b3ec49663b6e23b6a1bdf9e/arc_cache.go#L16-L29

Could you add code comments to clarify the patent status of this file?

lidel:
The goal of this PR is to emove ARC patent annoyance from boxo users. Even tho we do not use ARC (we use TwoQueue cache instead), automated testing tools still flag boxo as it imports library which has ARC implementatation.

- [x] switch to  v2 of golang-lru without arc code (https://github.com/ipfs/boxo/issues/366#issuecomment-1597253540)
- [x] remove mentions of ARC in comments and func/var names, make it clear we use TwoQueue 
- [x] Update Changelog

Closes #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- Removed mentions of unused ARC algorithm ([#336](https://github.com/ipfs/boxo/issues/366#issuecomment-1597253540))
+
 ### Security
 
 ## [0.10.1] - 2023-06-19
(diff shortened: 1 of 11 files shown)


hacdias (approved):
Great, thanks for this.

PR #369, opened 2023-06-21
feat(gateway): more explicit IPFSBackend and no multi-range

aschmahmann:
Also, switches to only handling single range requests rather than multiples.

 - [x] Merge conformance tests updates: https://github.com/ipfs/gateway-conformance/pull/113
 - [x] Release conformance tests updates
 - [x] Run new conformance tests
 - [x] Kubo PR for removing outdated sharness: https://github.com/ipfs/kubo/pull/9999
 - [x] Kubo PR https://github.com/ipfs/kubo/pull/10024

### DIFF
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -153,7 +153,9 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 		return md, nil, err
 	}
 
-	// Only a single range is supported
+	// Only a single range is supported in responses to HTTP Range Requests.
+	// When more than one is passed in the Range header, this library will
+	// return a response for the first one and ignores remaining ones.
 	var ra *ByteRange
 	if len(ranges) > 0 {
 		ra = &ranges[0]
@@ -170,10 +172,8 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 			return ContentPathMetadata{}, nil, err
 		}
 
-		if ra != nil && ra.From != 0 {
-			if _, err := f.Seek(int64(ra.From), io.SeekStart); err != nil {
-				return ContentPathMetadata{}, nil, err
-			}
+		if err := seekToRangeStart(f, ra); err != nil {
+			return ContentPathMetadata{}, nil, err
 		}
 
 		return md, NewGetResponseFromReader(f, fileSize), nil
@@ -206,10 +206,8 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 			return ContentPathMetadata{}, nil, err
 		}
 
-		if ra != nil && ra.From != 0 {
-			if _, err := file.Seek(int64(ra.From), io.SeekStart); err != nil {
-				return ContentPathMetadata{}, nil, err
-			}
+		if err := seekToRangeStart(file, ra); err != nil {
+			return ContentPathMetadata{}, nil, err
 		}
 
 		if s, ok := f.(*files.Symlink); ok {
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -153,11 +153,30 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 		return md, nil, err
 	}
 
+	// Only a single range is supported
+	var ra *ByteRange
+	if len(ranges) > 0 {
+		ra = &ranges[0]
+	}
+
 	rootCodec := nd.Cid().Prefix().GetCodec()
 	// This covers both Raw blocks and terminal IPLD codecs like dag-cbor and dag-json
 	// Note: while only cbor, json, dag-cbor, and dag-json are currently supported by gateways this could change
 	if rootCodec != uint64(mc.DagPb) {
-		return md, NewGetResponseFromFile(files.NewBytesFile(nd.RawData())), nil
+		f := files.NewBytesFile(nd.RawData())
+
+		fileSize, err := f.Size()
+		if err != nil {
+			return ContentPathMetadata{}, nil, err
+		}
+
+		if ra != nil && ra.From != 0 {
+			if _, err := f.Seek(int64(ra.From), io.SeekStart); err != nil {
+				return ContentPathMetadata{}, nil, err
+			}
+		}
+
+		return md, NewGetResponseFromReader(f, fileSize), nil
 	}
 
 	// This code path covers full graph, single file/directory, and range requests
@@ -179,10 +198,25 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 		if sz < 0 {
 			return ContentPathMetadata{}, nil, fmt.Errorf("directory cumulative DAG size cannot be negative")
 		}
-		return md, NewGetResponseFromDirectoryListing(uint64(sz), dir.EnumLinksAsync(ctx)), nil
+		return md, NewGetResponseFromDirectoryListing(uint64(sz), dir.EnumLinksAsync(ctx), nil), nil
 	}
 	if file, ok := f.(files.File); ok {
-		return md, NewGetResponseFromFile(file), nil
+		fileSize, err := f.Size()
+		if err != nil {
+			return ContentPathMetadata{}, nil, err
+		}
+
+		if ra != nil && ra.From != 0 {
+			if _, err := file.Seek(int64(ra.From), io.SeekStart); err != nil {
+				return ContentPathMetadata{}, nil, err
+			}
+		}
+
+		if s, ok := f.(*files.Symlink); ok {
+			return md, NewGetResponseFromSymlink(s, fileSize), nil
+		}
+
+		return md, NewGetResponseFromReader(file, fileSize), nil
 	}
 
 	return ContentPathMetadata{}, nil, fmt.Errorf("data was not a valid file or directory: %w", ErrInternalServerError) // TODO: should there be a gateway invalid content type to abstract over the various IPLD error types?
@@ -211,7 +245,7 @@ func (bb *BlocksBackend) GetBlock(ctx context.Context, path ImmutablePath) (Cont
 	return md, files.NewBytesFile(nd.RawData()), nil
 }
 
-func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentPathMetadata, files.Node, error) {
+func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentPathMetadata, *HeadResponse, error) {
 	md, nd, err := bb.getNode(ctx, path)
 	if err != nil {
 		return md, nil, err
@@ -219,7 +253,7 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
 
 	rootCodec := nd.Cid().Prefix().GetCodec()
 	if rootCodec != uint64(mc.DagPb) {
-		return md, files.NewBytesFile(nd.RawData()), nil
+		return md, NewHeadResponseForFile(files.NewBytesFile(nd.RawData()), int64(len(nd.RawData()))), nil
 	}
 
 	// TODO: We're not handling non-UnixFS dag-pb. There's a bit of a discrepancy
@@ -229,7 +263,24 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
 		return ContentPathMetadata{}, nil, err
 	}
 
-	return md, fileNode, nil
+	sz, err := fileNode.Size()
+	if err != nil {
+		return ContentPathMetadata{}, nil, err
+	}
+
+	if _, ok := fileNode.(files.Directory); ok {
+		return md, NewHeadResponseForDirectory(sz), nil
+	}
+
+	if _, ok := fileNode.(*files.Symlink); ok {
+		return md, NewHeadResponseForSymlink(sz), nil
+	}
+
+	if f, ok := fileNode.(files.File); ok {
+		return md, NewHeadResponseForFile(f, sz), nil
+	}
+
+	return ContentPathMetadata{}, nil, fmt.Errorf("unsupported UnixFS file type")
 }
 
 // emptyRoot is a CAR root with the empty identity CID. CAR files are recommended
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -162,8 +162,10 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 	}
 
 	rootCodec := nd.Cid().Prefix().GetCodec()
+
 	// This covers both Raw blocks and terminal IPLD codecs like dag-cbor and dag-json
 	// Note: while only cbor, json, dag-cbor, and dag-json are currently supported by gateways this could change
+	// Note: For the raw codec we return just the relevant range rather than the entire block
 	if rootCodec != uint64(mc.DagPb) {
 		f := files.NewBytesFile(nd.RawData())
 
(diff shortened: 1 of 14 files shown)


hacdias:
@aschmahmann can we not support multiple ranges? Isn't that part of the specs? Also note that I'll be working on IPIP-412 soon, sow e might get some _conflicts_.

BigLep:
2023-07-11 maintainer conversation: we shouldn't stress about multiple ranges.  We have a metric for this against ipfs.io and haven't found any actual usage.  This conformance test for multiple ranges can be moved to Kubo.
@aschmahmann will flip it out of draft when ready for review.

hacdias (changes requested):
~Well, we still need the multiple range requests to pass conformance: https://github.com/ipfs/kubo/actions/runs/5655778072/job/15321538379?pr=10024~

Nevermind,  I just saw "This conformance test for multiple ranges can be moved to Kubo.".

hacdias:
> This conformance test for multiple ranges can be moved to Kubo.

@BigLep @aschmahmann I don't understand how that helps. If we move the test to Kubo, it will still fail since we are removing the functionality on this PR and that functionality will simply no longer be available.

aschmahmann:
> If we move the test to Kubo, it will still fail since we are removing the functionality on this PR and that functionality will simply no longer be available.

Correct. IIUC the right thing to do here is:
1. Have the conformance tests test the HTTP guarantees around multiple ranges (e.g. that the server can send back the ranges, the first range, or none as long as it's clear from the response headers)
2. If we end up having this functionality in just one implementation/consumer of boxo (e.g. kubo) then we test there

We can add back in the multiple range functionality, but IIUC from @lidel basically no one is using this so it's not worth the hassle at the moment.

hacdias:
Working on updating the tests here: https://github.com/ipfs/gateway-conformance/pull/113

hacdias:
@aschmahmann @lidel this PR is blocked until https://github.com/ipfs/gateway-conformance/pull/113 is merged and released (ready for review), so that we can have the tests passing here and/or in Kubo's PR.

gateway/blocks_backend.go
```diff
@@ -153,11 +153,30 @@ func (bb *BlocksBackend) Get(ctx context.Context, path ImmutablePath, ranges ...
 		return md, nil, err
 	}
 
+	// Only a single range is supported
```
lidel: @hacdias mind adding a PR (just PR, no need for IPIP since this is only a clarification) against https://specs.ipfs.tech/http-gateways/path-gateway/#range-request-header that adds one sentence that states  the implementation SHOULD support single ranges, and that support for more than one range is optional?

gateway/blocks_backend.go
```diff
@@ -229,12 +263,60 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
...
+		return md, NewHeadResponseForFile(f, sz), nil
+	}
+
+	return ContentPathMetadata{}, nil, fmt.Errorf("unsupported UnixFS file type")
 }
 
 func (bb *BlocksBackend) GetCAR(ctx context.Context, p ImmutablePath, params CarParams) (ContentPathMetadata, io.ReadCloser, error) {
 	pathMetadata, err := bb.ResolvePath(ctx, p)
 	if err != nil {
+		rootCid, err := cid.Decode(strings.Split(p.String(), "/")[2])
```
lidel: :thought_balloon: feels like `ImmutablePath` should have `func` that returns the rootCid, instead of parsing this everywhere like this

gateway/blocks_backend.go
```diff
@@ -229,12 +263,60 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
...
+		if err != nil {
+			return ContentPathMetadata{}, nil, err
+		}
+
+		var buf bytes.Buffer
+		cw, err := storage.NewWritable(&buf, []cid.Cid{cid.MustParse("bafkqaaa")}, car.WriteAsCarV1(true))
```
lidel: nit:  `[]cid.Cid{cid.MustParse("bafkqaaa")` could be a reusable const with comment linking to https://ipld.io/specs/transport/car/carv1/#number-of-roots

gateway/blocks_backend.go
```diff
@@ -229,12 +263,60 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
...
+		pathResolver := resolver.NewBasicResolver(f)
+		ip := ipfspath.FromString(p.String())
+		_, _, err = pathResolver.ResolveToLastNode(ctx, ip)
+
+		if isErrNotFound(err) {
+			return ContentPathMetadata{
+				PathSegmentRoots: nil,
+				LastSegment:      ifacepath.NewResolvedPath(ip, rootCid, rootCid, ""),
+				ContentType:      "",
+			}, io.NopCloser(&buf), nil
```
lidel: :thought_balloon:  TODO:  was not obvious to me how does 404 scenarios work.

When part of the requested content path does not exist, gateway should return CAR with  HTTP status code 200 and a CAR body that includes blocks up to the last available parent entity.

But here we return ReadCloser with `LastSegment: ifacepath.NewResolvedPath(ip, rootCid, rootCid, "")`.
What happens next?

gateway/errors.go
```diff
@@ -178,10 +177,6 @@ func webError(w http.ResponseWriter, r *http.Request, c *Config, err error, defa
 }
 
 func isErrNotFound(err error) bool {
-	if ipld.IsNotFound(err) {
-		return true
-	}
```
lidel: :warning: I realized we have no end-to-end tests for 404 scenarios, just unit tests which do not really reflect final behavior.  Filled:

https://github.com/ipfs/gateway-conformance/issues/126 (car responses)
https://github.com/ipfs/gateway-conformance/issues/127 (deserialized behavior)


I feel we need to add them as part of this work. @hacdias would you have bandwidth to add them?

gateway/handler_codec.go
```diff
@@ -69,10 +69,42 @@ func (i *handler) serveCodec(ctx context.Context, w http.ResponseWriter, r *http
...
+
+	if !i.seekToStartOfFirstRange(w, r, data) {
+		return false
+	}
+
+	return i.renderCodec(ctx, w, r, rq, blockSize, data)
+}
+
+// seekToStartOfFirstRange seeks to the start of the first Range if the request is an HTTP Range Request
+func (i *handler) seekToStartOfFirstRange(w http.ResponseWriter, r *http.Request, data io.Seeker) bool {
```
lidel: nit: move to `serve_http_content.go`

lidel (changes requested):
(submitting first-pass at the end of day)

gateway/blocks_backend.go
hacdias: https://github.com/ipfs/specs/pull/430

gateway/blocks_backend.go
hacdias: I couldn't find many other similar cases like this. I think this is something we can solve with #334.

gateway/blocks_backend.go
hacdias: @lidel as far as I understand, this will create a CAR file with all the blocks to resolve the path up to the last available entity.

gateway/errors.go
hacdias: @lidel not today. Maybe Friday or early next week.

hacdias: @lidel 
- https://github.com/ipfs/gateway-conformance/pull/131
- https://github.com/ipfs/gateway-conformance/pull/132

hacdias (approved):
Tests are passing in https://github.com/ipfs/kubo/pull/10024.

gateway/blocks_backend.go
```diff
@@ -229,12 +263,65 @@ func (bb *BlocksBackend) Head(ctx context.Context, path ImmutablePath) (ContentP
...
 func (bb *BlocksBackend) GetCAR(ctx context.Context, p ImmutablePath, params CarParams) (ContentPathMetadata, io.ReadCloser, error) {
 	pathMetadata, err := bb.ResolvePath(ctx, p)
 	if err != nil {
+		rootCid, err := cid.Decode(strings.Split(p.String(), "/")[2])
+		if err != nil {
+			return ContentPathMetadata{}, nil, err
+		}
+
+		var buf bytes.Buffer
+		cw, err := storage.NewWritable(&buf, emptyRoot, car.WriteAsCarV1(true))
```
lidel: @hacdias @aschmahmann  Could we keep a meaningful root for exports done with `/ipfs/cid` (without any sub-path)?  

This would be a more gentle way of doing this, as we would not break existing users (pre IPIP-402) who rely on exporting full DAG by its CID, and  then doing `ipfs dag import` without extra pinning to keep it around (because `ipfs dag import` pins roots by default).

aschmahmann: Yeah, that seems fine by me.

hacdias: Hmm. The line you highlighted is only for paths that cannot be resolved - the other CARs (line 344) use the CID of the last segment as the root CID. Only if `/ipfs/{cid}` fails to resolve with a `isErrNotFound` is this code path followed. Is that a realistic scenario?

gateway/gateway.go
```diff
@@ -255,21 +255,74 @@ type ByteRange struct {
...
-func NewGetResponseFromFile(file files.File) *GetResponse {
-	return &GetResponse{bytes: file}
+func NewGetResponseFromDirectoryListing(dagSize uint64, entries <-chan unixfs.LinkResult, closeFn func() error) *GetResponse {
+	return &GetResponse{directoryMetadata: &directoryMetadata{dagSize: dagSize, entries: entries, closeFn: closeFn}}
 }
 
-func NewGetResponseFromDirectoryListing(dagSize uint64, entries <-chan unixfs.LinkResult) *GetResponse {
-	return &GetResponse{directoryMetadata: &directoryMetadata{dagSize, entries}}
+type HeadResponse struct {
+	size          int64
```
lidel: nit:  `HeadResponse` field introduced in this PR should match already existing  `GetResponse.byteSize` 
```suggestion
	bytesSize          int64
```

gateway/handler_defaults.go
```diff
@@ -114,36 +97,62 @@ func (i *handler) serveDefaults(ctx context.Context, w http.ResponseWriter, r *h
...
-		}
 		rq.logger.Debugw("serving codec", "path", rq.contentPath)
-		return i.renderCodec(r.Context(), w, r, rq, bytesResponse)
+		var blockSize int64
+		var dataToRender io.ReadCloser
+		if headResp != nil {
+			blockSize = headResp.size
+			dataToRender = nil
+		} else {
+			blockSize = getResp.bytesSize
```
lidel: (example how it looks weird, new type should use the same field name)

gateway/serve_http_content.go
```diff
@@ -0,0 +1,464 @@
...
+	start, length int64
+}
+
+func (r httpRange) contentRange(size int64) string {
+	return fmt.Sprintf("bytes %d-%d/%d", r.start, r.start+r.length-1, size)
+}
+
+// parseRange parses a Range header string as per RFC 7233.
+// errNoOverlap is returned if none of the ranges overlap.
+func parsehttpRange(s string, size int64) ([]httpRange, error) {
```
lidel:
```suggestion
// parseRange parses a Range header string as per RFC 7233.
// errNoOverlap is returned if none of the ranges overlap.
func parseRange(s string, size int64) ([]httpRange, error) {
```

gateway/serve_http_content.go
lidel, line 1: :thought_balloon:  / :warning: something that makes me anxious is that we have no(?) tests for all the low level HTTP funcs related to `httpServeContent` that we've ended up vendoring here. 

In the past we had the protection of unit tests from places like https://cs.opensource.google/go/go/+/master:src/net/http/range_test.go  https://cs.opensource.google/go/go/+/master:src/net/http/fs_test.go (and others in the dir), but now we don't use the same funcs as upstream golang and lose some coverage.

For example (full list would require bit closer look), conformance tests exist for some  If-None-Match scenarios, but we have no tests for behaviors around  If-Match, If-Unmodified-Since, If-Modified-Since, and If-Range.

We could add basic unit tests here, or invest time into better coverage in gateway conformance to benefit implementations other than  boxo/gateway (fill issue for missing tests in conformance repo).

@aschmahmann any preference?

gateway/handler_unixfs_dir.go
```diff
@@ -59,42 +60,60 @@ func (i *handler) serveDirectory(ctx context.Context, w http.ResponseWriter, r *
...
+	appendIndexHtml := func(p ipath.Path) ipath.Path {
+		basePath := p.String()
+		if basePath[len(basePath)-1] == '/' {
+			return ipath.New(basePath + "index.html")
+		}
+		return ipath.New(basePath + "/index.html")
+	}
+
+	idxPath := appendIndexHtml(contentPath)
+	imIndexPath, err := NewImmutablePath(appendIndexHtml(resolvedPath))
```
aschmahmann: @lidel @hacdias how's this for our `//index.html` issues? Kind of gross, but seems like it'd do the job until we have something like #334 to make this less painful.

gateway/blocks_backend.go
aschmahmann: Good call, I wasn't looking too closely when I commented here. We shouldn't be able to hit this codepath with a failed `/ipfs/cid` resolution, and if we did we should be returning an error not writing out an empty CAR file.

lidel (changes requested):

gateway/handler_unixfs_dir.go
lidel: sgtm. we could would flip it to append index.html only in one place:

```suggestion
	appendIndexHtml := func(p ipath.Path) ipath.Path {
		basePath := p.String()
		if basePath[len(basePath)-1] != '/' {
			basePath += "/"
		}
		return ipath.New(basePath + "index.html")
	}

	idxPath := appendIndexHtml(contentPath)
	imIndexPath, err := NewImmutablePath(appendIndexHtml(resolvedPath))
```

gateway/handler_codec.go
```diff
@@ -69,10 +69,20 @@ func (i *handler) serveCodec(ctx context.Context, w http.ResponseWriter, r *http
...
-	return i.renderCodec(ctx, w, r, rq, data)
+
+	blockSize, err := data.Size()
+	if !i.handleRequestErrors(w, r, rq.contentPath, err) {
+		return false
+	}
+
+	if !i.seekToStartOfFirstRange(w, r, data) {
+		return false
+	}
```
aschmahmann: @hacdias I don't know if you or I put this in, but this looks  suspicious. Why would we be doing manipulation of the data range before doing the full rendering/transformation?

hacdias: The blame says you added it - tbf, I do not remember adding it, so it's probably right. (698eccfdefb70d57fc5cab8cbf36f47afb00b25b). I have no idea about what it might be doing.

BigLep:
2023-08-22 conversation: @lidel will review and figure out appropriate testing story (Boxo unit tests vs. conformance tests)

BigLep:
2023-08-24 conversation: 
We are going to lean on conformance tests.  @lidel is going to create a gateway-conformance issues for this.
It's a lesser evil to merge without the conformance tests than keep this open and let it diverge.
@aschmahmann will add some Boxo unit tests and flag the changed lines of potential concerns.
@aschmahmann will make sure we get the merge decision 
(There is also some dance for keeping CI green with sharness.)

BigLep:
@aschmahmann : is this on track to land this week?

BigLep:
2023-09-12 conversation:
- Do a gateway-conformance release as soon as possible: https://github.com/ipfs/gateway-conformance/pull/162 lands and passes.
- To get green CI
- [x]  Conformance release
- [x]  Bubble up to Boxo
- [x]  Bubble this PR into Kubo
- [x]  Run this PR against Kubo version
- [x]  Once Boxo CI passes
- [x]  Switch to target Kubo master
- [x]  Merge this PR
- [x]  Bubble up Boxo commit to Kubo PR
- [x]  Merge Kubo PR
- [x]  Rerun Boxo CI

hacdias:
CI is green in Kubo. As per what was agreed, I'm merging this once CI finishes here and bubbling it up.

PR #370, opened 2023-06-22
feat(gateway): support for order=, dups= parameters from IPIP-412

### LINKED ISSUE #349: Support IPIP-412
This is about supporting https://github.com/ipfs/specs/pull/412

```[tasklist]
### Tasks
- [ ] https://github.com/ipfs/boxo/pull/370
- [ ] https://github.com/ipfs/gateway-conformance/pull/87
```

hacdias:
Implements IPIP-412. Closes #349.

- [x] Depends on #337 (too many conflicts otherwise)
- [x] Kubo PR: https://github.com/ipfs/kubo/pull/9989
- [ ] Conformance: https://github.com/ipfs/gateway-conformance/pull/87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,17 +17,29 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 * ✨ The gateway now supports the optional `order` and `dups` CAR parameters
-  from [IPIP-412](https://github.com/ipfs/specs/pull/412). `BlocksBackend` only
-  DFS ordering. If the request explicitly requests an ordering other than `dfs`
-  and `unk`, the request will return an error.
+  from [IPIP-412](https://github.com/ipfs/specs/pull/412).
+  * The `BlocksBackend` only implements `order=dfs` (Depth-First Search)
+    ordering, which was already the default behavior.
+  * If a request specifies no `dups`, response with `dups=n` is returned, which
+    was already the default behavior.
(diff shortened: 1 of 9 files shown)


lidel:
Pushed a minor refactor in https://github.com/ipfs/boxo/pull/370/commits/c6e10eb5fe098ebb6ae678ae2d0a9a25b21dbf15:

- adds more type safety: `DuplicateBlocksPolicy` is no longer a pointer to   a `bool`, but a real type wit h `String()` and `Bool()`, which simplified code a bit, and together with `DagOrder` both have an explicit `*Unspecified` state that also makes it easier to clearly define implicit default in a single place + find it if needed. 
- removes passing pointer and reliance on mutation. now mutation   happens only once, in buildCarParams func, where implicit defaults   are set when order or dups are unspecified in the request
- moved car version validation and other parameter related logic   into the same build funcs
- fixed a bug where ?format=car eclipsed params from Accept header (would be a problem in many apps, as they often send both)

Will bubble up to Kubo and then add some tests in https://github.com/ipfs/gateway-conformance/pull/87 tomorrow.

lidel (approved):
In case I don't get to this before you are back @hacdias, feel free to bubble this up to https://github.com/ipfs/kubo/pull/9989 and https://github.com/ipfs/gateway-conformance/pull/87

PR #371, opened 2023-06-22
fix(ipns): racy parallel tests writing to the same error variable

hacdias:
🙈 Writing two errors to the same variable in parallel tests (also corrected the name of the variable `pid`, should've been `name`).


willscott (approved):

lidel (approved):

Jorropo (approved):
`LOOPVAR` :disappointed:

PR #379, opened 2023-06-26
util: use subtle for XORing bytes

### LINKED ISSUE #377: util: Optimize XOR
We can usually XOR words instead of bytes for a significant speedup. See:

https://golang.org/src/crypto/cipher/xor.go

Jorropo:
Fixes #377

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -3,6 +3,7 @@
 package util
 
 import (
+	"crypto/subtle"
 	"errors"
 	"io"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


Jorropo:
Revisit when updating to go1.21

util/util.go
```diff
@@ -150,9 +151,9 @@ func IsValidHash(s string) bool {
 
 // XOR takes two byte slices, XORs them together, returns the resulting slice.
 func XOR(a, b []byte) []byte {
+	_ = b[len(a)-1] // keeping same behaviour as previously but this looks like a bug
```
hacdias, line 154: We should revisit this later.

hacdias (approved):

util/util.go
Jorropo: AFAIK this is only ever called with slices of the same length, so we could check that `if len(a) != len(b) { panic() }`.

PR #393, opened 2023-06-26
ipld/unixfs/hamt: catch panic in walkChildren

Jorropo:
From https://github.com/ipfs/go-unixfs/pull/123


Jorropo:
NVM this does not fix any issues, it prints debug info when that happen, I don't see why printf debugs should be commited ?

gammazero (approved):

gammazero:
@lidel Even though the problem should not happen, do we still want this change in case there are any existing HAMTs that would cause the panic? If not, then please close this PR.

PR #395, opened 2023-06-26
fix(gateway): include CORS on subdomain redirects

lidel:
This PR closes regression introduced during one of the refactors since Kubo 0.20 and adds dedicated tests for preflight requests on both path and subdomain gateway.

Context: https://github.com/ipfs/kubo/issues/9983#issuecomment-1599673976

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -320,20 +320,22 @@ func cleanHeaderSet(headers []string) []string {
 	return result
 }
 
-// AddAccessControlHeaders adds default HTTP headers used for controlling
-// cross-origin requests. This function adds several values to the
-// [Access-Control-Allow-Headers] and [Access-Control-Expose-Headers] entries.
+// AddAccessControlHeaders ensures safe default HTTP headers are used for
+// controlling cross-origin requests. This function adds several values to the
+// [Access-Control-Allow-Headers] and [Access-Control-Expose-Headers] entries
+// to be exposed on GET and OPTIONS responses, including [CORS Preflight].
 //
-// If the Access-Control-Allow-Origin entry is missing a value of '*' is
+// If the Access-Control-Allow-Origin entry is missing, a default value of '*' is
 // added, indicating that browsers should allow requesting code from any
 // origin to access the resource.
 //
-// If the Access-Control-Allow-Methods entry is missing a value of 'GET' is
-// added, indicating that browsers may use the GET method when issuing cross
+// If the Access-Control-Allow-Methods entry is missing a value, 'GET, HEAD,
+// OPTIONS' is added, indicating that browsers may use them when issuing cross
 // origin requests.
 //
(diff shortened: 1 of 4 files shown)


gateway/gateway.go
```diff
@@ -346,8 +348,12 @@ func AddAccessControlHeaders(headers map[string][]string) {
...
 	}
 	if _, ok := headers[ACAMethodsName]; !ok {
-		// Default to GET
-		headers[ACAMethodsName] = []string{http.MethodGet}
+		// Default to GET, HEAD, OPTIONS
+		headers[ACAMethodsName] = []string{
+			http.MethodGet,
+			http.MethodHead,
+			http.MethodOptions,
+		}
```
lidel, line 356: :information_source:  we were missing HEAD and OPTIONS, which may not be used a lot today, but will be once HTTP transport becomes more and more popular. Included tests for this too.

gateway/gateway_test.go
```diff
@@ -619,6 +619,91 @@ func TestIpnsBase58MultihashRedirect(t *testing.T) {
 	})
 }
 
+// TestCORSPreflightHeaders ensures CORS headers are present in HTTP OPTIONS responses
+// https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
+func TestCORSPreflightHeaders(t *testing.T) {
```
lidel: :information_source: basic regression test that guards CORS-related headers on path and subdomain gateways

gateway/gateway.go
```diff
@@ -320,20 +320,22 @@ func cleanHeaderSet(headers []string) []string {
...
-// If the Access-Control-Allow-Methods entry is missing a value of 'GET' is
-// added, indicating that browsers may use the GET method when issuing cross
+// If the Access-Control-Allow-Methods entry is missing a value, 'GET, HEAD,
+// OPTIONS' is added, indicating that browsers may use them when issuing cross
 // origin requests.
 //
 // [Access-Control-Allow-Headers]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers
 // [Access-Control-Expose-Headers]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+// [CORS Preflight]: https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
 func AddAccessControlHeaders(headers map[string][]string) {
```
lidel, line 339: @laurentsenta  this is just FYI if you ever need to find where `boxo/gateway` sets default CORS headers :)

Once this ships in Kubo, we can update conformance [TestCors](https://github.com/ipfs/gateway-conformance/blob/43c0842a1041a2748a74a196e370b84ef02b12a5/tests/path_gateway_unixfs_test.go#L455) to include all headers set by `AddAccessControlHeaders` here.

I've also filled https://github.com/ipfs/specs/issues/423 to write this down and add it to specs (currently CORS on gateways is grossly undocumented), but fine to update conformance before specs land.

hacdias (approved):
LGTM, I rebased it. Will merge once CI is green.

PR #396, opened 2023-06-27
feat: update conformance to v0.2

laurentsenta:


hacdias (approved):

hacdias (changes requested):
Was too fast. I think we have to remove other IPNS related tests.

hacdias (approved):

PR #398, opened 2023-06-27
ci: add changelog update checker workflow

### LINKED ISSUE #267: Enforce changelog updates in CI
It'd be nice for CI to enforce changelog updates for each PR, or require explicit opt-out, so that when we start a release train, the release engineer doesn't need to cat-herd everyone to write changelog updates.

I've opened an issue here requesting this feature: https://github.com/pl-strflt/changelog-driven-release/issues/5

galargh:
Resolves https://github.com/ipfs/boxo/issues/267

This PR adds a CI workflow which checks whether the PR contains CHANGELOG.md modifications. 

The check can be skipped by either adding `[skip changelog]` to the PR title or labelling the PR with `skip/changelog`. These instructions are included in the PR template and in the workflow failure message.

The checks succeeds if CHANGELOG.md was modified and fails otherwise.

Tested in https://github.com/galargh/.github/pull/40


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,35 @@
...
+    if: contains(github.event.pull_request.title, '[skip changelog]') == false &&
+        contains(github.event.pull_request.labels.*.name, 'skip/changelog') == false
+    runs-on: ubuntu-latest
+    name: Changelog
+    steps:
+      - id: changelog
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          ENDPOINT: repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files
+          SELECTOR: '.[] | select(.filename == "CHANGELOG.md") | length'
```
galargh:
```suggestion
          SELECTOR: 'map(select(.filename == "CHANGELOG.md")) | length'
```

hacdias (approved):

hacdias:
@galargh question: will we need to add `[skip changelog]` to release-related PRs?

galargh:
> @galargh question: will we need to add [skip changelog] to release-related PRs?

Good call! I'll update the PR titles in kuboreleaser.

galargh:
> @galargh question: will we need to add [skip changelog] to release-related PRs?

Good call! How about we trigger only on PRs that modify Go files (including go.mod/go.sum)? I think that should cover the cases where CHANGELOG update might be required.

hacdias:
@galargh that still worries me: what about a cherry-picked release like https://github.com/ipfs/boxo/pull/401, and a merge back PR like https://github.com/ipfs/boxo/pull/403?

galargh:
#401 modifies Go files so the changelog workflow run would be triggered. It also modifies CHANGELOG.md so the workflow run would have passed.

#403 doesn't modify Go files so with the changes from https://github.com/ipfs/boxo/pull/398/commits/71cac87a0167c2967ab9ec124b5b5c08f4771d7b the changelog workflow wouldn't run at all.

hacdias:
Okay, I think it sounds good. I'm happy to get this merged. We can always make any improvements if we find something.

PR #407, opened 2023-07-09
verifcid: introduce and integrate Allowlist interface

Wondertan:
The successor of https://github.com/ipfs/boxo/pull/281.


verifcid/validator.go
```diff
@@ -0,0 +1,54 @@
+package verifcid
+
+import (
+	"github.com/ipfs/go-cid"
+	mh "github.com/multiformats/go-multihash"
+)
+
+type Allowlist interface {
```
Wondertan: I feel like it should be moved to go-cid instead. Although, it's not that widely used in PL.

verifcid/allowlist.go
```diff
@@ -0,0 +1,58 @@
+package verifcid
+
+import (
+	"github.com/ipfs/go-cid"
+	mh "github.com/multiformats/go-multihash"
+)
+
+// Allowlist defines an interface containing list of allowed multihashes.
+type Allowlist interface {
```
Wondertan, line 11: I feel like it should be moved to go-cid instead. Although, it's not that widely used

verifcid/allowlist.go
```diff
@@ -0,0 +1,58 @@
...
+		return true
+	}
+
+	if !found {
+		if code >= mh.BLAKE2B_MIN+19 && code <= mh.BLAKE2B_MAX {
+			return true
+		}
+		if code >= mh.BLAKE2S_MIN+19 && code <= mh.BLAKE2S_MAX {
+			return true
+		}
```
Jorropo: This is weird, the allow list takes in a map of allowed codecs, but have special case for blake hardcoded ?

I don't personally mind if a whole bunch of new cases are added to the map instead, this isn't high performance code.

verifcid/validate.go
```diff
@@ -33,37 +33,15 @@ var goodset = map[uint64]bool{
...
-	if pref.MhType != mh.IDENTITY && pref.MhLength > maximumHashLength {
-		return ErrAboveMaximumHashLength
-	}
-
-	return nil
+	return DefaultAllowlist.ValidateCid(c)
 }
+
+// DefaultAllowlist is the default list of hashes allowed in IPFS.
+var DefaultAllowlist = NewAllowlist(goodset)
```
Jorropo: Having a `DefaultAllowlist` is good, but I think we should take advantage of it and use a stricter implementation.
Something like `binary.LittleEndian` would be better and ward off allocations by having a concrete type.

So:
```suggestion
var DefaultAllowlist defaultAllowlist
```
And then defining `IsAllowed` to use a big switch statement.

Jorropo (changes requested):

verifcid/allowlist.go
Jorropo: I remember discussing with @aschmahmann about it, he said that back then go-verifcid was supposed to move into go-cid but this didn't happend. This would have positives, very few pieces of code and peoples are aware they probably want to use this package.

I don't know if we want to move it now, it does not really help with the globals problem we are having here.

verifcid/allowlist.go
Wondertan: `mh.BLAKE2B_MAX-mh.BLAKE2B_MIN+19 = 82`

I don't think it's a good idea to type in all 82 cases manually. Alternatively, this could be a loop filling up the goodset, but then it's not any better than current ifs.

verifcid/validate.go
Wondertan: I don't mind following the `binary` pkg here

Wondertan: @Jorropo, but what do you mean by?
> And then defining IsAllowed to use a big switch statement.

We have goodset map, not a switch statement.

Wondertan: > I don't mind following the binary pkg here

Actually, if we do so, we will lose the default implementation of Allowlist and if we still want to keep it, then we will have to repeat the code: in default and public impls.

Jorropo: > Actually, if we do so, we will lose the default implementation of Allowlist and if we still want to keep it, then we will have to repeat the code: in default and public impls.

We can keep it and just not use it here.

I don't see what code would be repeated.

verifcid/allowlist.go
Jorropo: you can add an if just for that, I don't see an issue with 82 cases tho, you just add a comment before that block.

Jorropo:
@Wondertan can you give me your opinion on https://github.com/ipfs/boxo/pull/407/commits/8e41b5324be733c7f2297417e109072e663a1182 please ?

verifcid/validate.go
Wondertan: Ok, your idea to eliminate ValidateCID from the interface resolves my concern!

verifcid/allowlist.go
Wondertan: Ok, so you kept ifs, nice

verifcid/allowlist.go
Wondertan: As we removed ValidateCID, the Allowlist would better be moved to go-multihash then

verifcid/validate.go
```diff
@@ -3,57 +3,65 @@ package verifcid
...
 	"fmt"
 
-	cid "github.com/ipfs/go-cid"
+	"github.com/ipfs/go-cid"
 	mh "github.com/multiformats/go-multihash"
 )
 
 var ErrPossiblyInsecureHashFunction = fmt.Errorf("potentially insecure hash functions not allowed")
 var ErrBelowMinimumHashLength = fmt.Errorf("hashes must be at least %d bytes long", minimumHashLength)
 var ErrAboveMaximumHashLength = fmt.Errorf("hashes must be at most %d bytes long", maximumHashLength)
```
Wondertan: Do we actually need to export those? A good question to ask as we revamping this pkg completely.

verifcid/validate.go
```diff
@@ -3,57 +3,65 @@ package verifcid
...
+func NewOverdingAllowlist(override Allowlist, allowset map[uint64]bool) Allowlist {
+	return allowlist{override, allowset}
+}
+
+type allowlist struct {
+	override Allowlist
+	allowset map[uint64]bool
+}
+
+func (ghr allowlist) IsAllowed(code uint64) bool {
```
Wondertan: That's a relic from when I initially called it GoodHashRegistry. Allowlist is a much better name and fits perfectly!
```suggestion
func (al allowlist) IsAllowed(code uint64) bool {
```

verifcid/validate.go
```diff
@@ -3,57 +3,65 @@ package verifcid
...
+	return allowlist{allowset: allowset}
 }
 
-func IsGoodHash(code uint64) bool {
-	good, found := goodset[code]
-	if good {
-		return true
+// NewOverdingAllowlist is like [NewAllowlist] but it will fallback to an other [AllowList] if keys are missing.
+// If override is nil it will return unsecure for unknown things.
+func NewOverdingAllowlist(override Allowlist, allowset map[uint64]bool) Allowlist {
```
Wondertan:
```suggestion
// NewOverridingAllowlist is like [NewAllowlist] but it will fallback to an other [AllowList] if keys are missing.
// If override is nil it will return unsecure for unknown things.
func NewOverridingAllowlist(override Allowlist, allowset map[uint64]bool) Allowlist {
```

Wondertan (commented):
@Jorropo, I am fond of the simplifications you made! Removal of ValidateCID is neat, and glad you didn't add all the Blake family in the switch case. 

I still don't really get the use case you see for the OverridingAllowlist, though. Something like `NewOverridingAllowlist(DefaultAllowlist, customset)`?

Would be also nice to add unit test for overriding

verifcid/validate.go
Jorropo: This allows to use `errors.Is`, LGTM as-is.

verifcid/validate.go
Jorropo: :+1:

verifcid/validate.go
Jorropo: :+1:

Jorropo:
> Something like `NewOverridingAllowlist(DefaultAllowlist, customset)`?

Exactly

> Would be also nice to add unit test for overriding

Yes

BigLep:
2023-07-11 maintainer conversation: since the number of consumers of low, we should break away from globals/defaults.  @Jorropo can provide more details.

verifcid/validate.go
```diff
@@ -3,57 +3,65 @@ package verifcid
...
+func IsGoodHash(code uint64) bool {
+	return DefaultAllowlist.IsAllowed(code)
+}
+
+// ValidateCid validates multihash allowance behind given CID.
 func ValidateCid(c cid.Cid) error {
+	return ValidateCidWithAllowlist(DefaultAllowlist, c)
+}
+
+func ValidateCidWithAllowlist(allowlist Allowlist, c cid.Cid) error {
```
Jorropo:
```suggestion
func ValidateCid(allowlist Allowlist, c cid.Cid) error {
```

There is a very little amount of places which uses this code, removing the defaults querier will force all future consumers to use the correct dependency.

Jorropo (changes requested):
I am onboard with fixing this, thx for doing it, but I'll be busy this week. Should make it for kubo 0.22, hopefully.

Wondertan:
I won't have time this and the following week either. Mainly due to crypto events in Paris

verifcid/validate.go
Wondertan: Let's also then change the name of the pkg. verifcid should become verifmh or something

Wondertan:
Rebased

Wondertan:
I've implemented the requested changes. One last thing to consider is perhaps renaming the pkg to something more suitable.

blockservice/blockservice.go
```diff
@@ -64,41 +64,61 @@ type BlockService interface {
 	DeleteBlock(ctx context.Context, o cid.Cid) error
 }
 
+// BoundedBlockService is a Blockservice bounded via strict multihash Allowlist.
+type BoundedBlockService interface {
+	BlockService
+
+	Allowlist() verifcid.Allowlist
+}
```
Wondertan, line 72: Mainly to forward Allowlist to Sessions. Can privatize it or anonymize

Wondertan:
@Jorropo, gentle ping

blockservice/blockservice.go
Jorropo: I would embed `verifcid.AllowList` in the `BoundedBlockService` but this is also fine, up to you.

Jorropo:
I've noticed we couldn't have a blockservice that is both writethrough and with an allowlist so I refactored it to the functional option pattern.
I've also added tests.

verifcid/cid.go
```diff
@@ -0,0 +1,33 @@
...
+
+var ErrPossiblyInsecureHashFunction = fmt.Errorf("potentially insecure hash functions not allowed")
+var ErrBelowMinimumHashLength = fmt.Errorf("hashes must be at least %d bytes long", minimumHashLength)
+var ErrAboveMaximumHashLength = fmt.Errorf("hashes must be at most %d bytes long", maximumHashLength)
+
+const minimumHashLength = 20
+const maximumHashLength = 128
+
+// ValidateCid validates multihash allowance behind given CID.
+func ValidateCid(allowlist Allowlist, c cid.Cid) error {
```
Jorropo, line 18: Could we make this non breaking, better name than `ValidateCidWithAllowlist` wanted tho ?
```suggestion
// ValidateCid validates multihash allowance behind given CID.
// It uses [DefaultAllowlist].
func ValidateCid(c cid.Cid) error {
 return ValidateCidWithAllowlist(DefautAllowlist, c)
}

// ValidateCidWithAllowlist is like [ValidateCid] but it uses a custom [Allowlist] instead of [DefaultAllowlist].
func ValidateCidWithAllowlist(allowlist Allowlist, c cid.Cid) error {
```

verifcid/allowlist.go
```diff
@@ -0,0 +1,69 @@
+package verifcid
+
+import (
+	mh "github.com/multiformats/go-multihash"
+)
+
+// DefaultAllowlist is the default list of hashes allowed in IPFS.
+var DefaultAllowlist Allowlist = defaultAllowlist{}
```
Jorropo:
```suggestion
var DefaultAllowlist defaultAllowlist
```
This prevents reassignments and the temptation to override this global for configuration (I think I got wrong originally, mb).

Jorropo (changes requested):
Thx, this looks good to me, I just have minor questions.

verifcid/cid.go
Wondertan: What's the point? I thought we are breaking stuff here. 

The only users of the default allowlist are supposed to be boxo/kubo libs, but now they enable customisable allowlists and will always use ValidateWithAllowlist.

Who is going to use ValidateCid with default?

Jorropo: Ah right, I remember we talked about that, this is good.
I forgot no one is using this code.

verifcid/allowlist.go
Wondertan: Fixed

Wondertan: > I think I got wrong originally

No, you were right originally.

Jorropo (approved):
LGTM thx :heart:

Wondertan:
There is one test failure, which seems unrelated and likely a flake:

```
--- FAIL: TestMfsStress (1.06s)
      mfs_test.go:889: file already exists
```

blockservice/blockservice.go
Wondertan: I like the idea, so pushed a change

Jorropo:
Actually let's not embed it, it's gonna confuse ADI frameworks. Also you didn't embeded in the implementation (just interface) and this is breaking the tests.

Wondertan:
I realized that a bit late and fixed it now. Please take a look at the new proper commit, and if we like the result we can keep it, if not, I'll revert.

Jorropo:
> I realized that a bit late and fixed it now. Please take a look at the commit, and if we like the result we can keep it, if not, I'll revert.

I force pushed it away, this is a type problem.
Basically frameworks like `fx` are gonna look at blockservice and may choose to plug it inside anyone that needs an `Allowlist`, including blockservice. So it's gonna try to instanciate a blockservice in order to instanciate a blockservice, hopefully it errors when it realise the dependency graph is cyclic, or it eats of all your memory explode and catch fire, who knows ?

Wondertan:
FWIW, FX realizes dep cycles, but I see your point and even cycle error would be annoying to deal with, and simple `fx.Provide` would need tags or additional fx shenanigans to make it work. Let's keep it then how I made it originally.

Wondertan:
My last take is that to fully finish this `Allowlist` interface with default impl should move to go-multihash and `ValidateCid` with errors should live inside `blockservice` pkg. Would love to know your thoughts on that.

Jorropo:
If you want to make an other PR about moving stuff around why not, I'm not sure about putting in the blockservice tho, because some of my code use verifcid without using the blockservice.

If anything I think this would be worth moving into `go-cid`, this might break a lot of peoples tho.

Wondertan:
What's the ETA for the next release including this PR?

Jorropo:
https://github.com/ipfs/kubo/issues/10014 Currently September 10.

In theory we could make a Boxo release before a Kubo one, we recently got in the habits of keeping them in sync because it's convenient, but if you need this I don't see why not release boxo earlier.

Wondertan:
I would appreciate an earlier release, so we can finally migrate to boxo as part of our big IPFS deps bump PR

PR #411, opened 2023-07-17
add wasm build option for filewriter flags[skip changelog]

mar1n3r0:

### DIFF
--- a/files/filewriter_unix.go
+++ b/files/filewriter_unix.go
@@ -1,4 +1,4 @@
-//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || wasm
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
 
 package files
 
--- a/files/filewriter_unix.go
+++ b/files/filewriter_unix.go
@@ -1,4 +1,4 @@
-//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || wasm
 
 package files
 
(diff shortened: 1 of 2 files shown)


Jorropo (changes requested):
This looks wrong to me, right now it is a list of compatible GOOSes but you are adding a GOARCH.

If windows added support for running WASM binaries this would be incorrect.

What do you think of:
> ```
> //go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
> ```
?

mar1n3r0:
Yep, totally makes sense.

files/filewriter_unix.go
```diff
@@ -1,4 +1,4 @@
-//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || wasm
```
Jorropo:
```suggestion
//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
```

files/filewriter_unix_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build darwin || linux || netbsd || openbsd
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || wasm
```
Jorropo:
```suggestion
//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
```

Jorropo (approved):
LGTM thx

Jorropo:
I can't apply the suggestions, could you do it please ?

mar1n3r0:
> I can't apply the suggestions, could you do it please ?

All set.

files/filewriter_unix_test.go
hacdias: @mar1n3r0 this suggestion is still missing, can you pls apply it?

mar1n3r0:
Sorry, forgot about that. Done.

PR #412, opened 2023-07-18
fix: handle _redirects for If-None-Match headers

aschmahmann:
Fixes a bug where `If-None-Match` requests would return 404s when _redirects files should've been used.


aschmahmann:
@lidel HEAD requests currently skip _redirects, I forget some of our prior conversation on if they should respect them as well.

lidel (changes requested):
@aschmahmann  according to  HTTP specs the behavior for HEAD should be the same as GET,
but it has such niche utility, that we said it is ok to skip the `_redirect` logic for HEAD if it is too much work in MVP (https://github.com/ipfs/go-libipfs/pull/176). 

iiuc next steps here:
- [x] see if we can make HEAD  the same as GET while we are at it
- [x] add regression test described in https://github.com/ipfs/gateway-conformance/issues/111

BigLep:
2023-07-20 conversation: this will make it in Kubo 0.22.  We're not planning to do a Kubo 0.21 patch release currently with this.  This was reported by Fleek.  

@aschmahmann will create an issue for this and notify #ipfs-operators in case anyone wants to block the header for the time being.

hacdias (approved):
LGTM, added a test. Pinging @lidel for a second look at the test.

lidel (approved):
lgtm, thank you!

aschmahmann:
@hacdias @lidel do we need separate tests for `if-none-match` and HEAD instead of bundling them together or is this enough?

PR #413, opened 2023-07-20
bitswap: peer prom tacker

### LINKED ISSUE #209: Bitswap: better diagnostics and observability on client

Jorropo:
Fixes  #209

- [ ] Peer Prom Tracker example
- [ ] Tracer tests, I think the client's tracer is bugged and it does not record outbound messages
- [ ] Changelog

### DIFF
--- a/bitswap/peer-prom-tracker/peer-prom-tracker.go
+++ b/bitswap/peer-prom-tracker/peer-prom-tracker.go
@@ -1,0 +1,146 @@
+package peerpromtracker
+
+import (
+	"errors"
+	"fmt"
+
+	bsmsg "github.com/ipfs/boxo/bitswap/message"
+	"github.com/ipfs/boxo/bitswap/tracer"
+	"github.com/ipfs/go-log/v2"
+	peer "github.com/libp2p/go-libp2p/core/peer"
+	prom "github.com/prometheus/client_golang/prometheus"
+)
+
+var logger = log.Logger("bitswap/peer-prom-tracker")
+
+var _ tracer.Tracer = (*PeerTracer)(nil)
+
+type PeerTracer struct {
+	bytesSent        *prom.CounterVec
+	messagesSent     *prom.CounterVec
+	bytesReceived    *prom.CounterVec
+	messagesReceived *prom.CounterVec
+
+	reg *prom.Registry
+}
+
+func NewPeerTracer(reg *prom.Registry) (*PeerTracer, error) {
+	pt := &PeerTracer{
+		reg: reg,
+	}
+
+	var good bool
+	defer func() {
+		if !good {
+			pt.Close() // will unregister the sucessfully registered metrics
+		}
+	}()
+
+	peerIdLabel := []string{"peer-id"}
+
+	bytesSent := prom.NewCounterVec(prom.CounterOpts{
+		Namespace: "ipfs",
+		Subsystem: "bitswap/messages",
+		Name:      "bytes-sent",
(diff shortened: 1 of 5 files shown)


bitswap/peer-prom-tracker/peer-prom-tracker.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	var good bool
+	defer func() {
+		if !good {
+			pt.Close() // will unregister the sucessfully registered metrics
+		}
+	}()
+
+	peerIdLabel := []string{"peer-id"}
```
lidel, line 39: :warning:  This will create a separate time series per PeerID, which is ok for debugging, but should NOT be enabled by default. 

High cardinality labels in prometheus are  is considered antipattern. If there was 20K of peers, this will create 20K time series, and that may cause  problems (performance, billing) when Grafana tries to visualize it.

To understand why high cardinality is a problem, see:
- https://stackoverflow.com/questions/46373442/how-dangerous-are-high-cardinality-labels-in-prometheus
- https://grafana.com/blog/2022/10/20/how-to-manage-high-cardinality-metrics-in-prometheus-and-kubernetes/ 

IMO this PR can't land in boxo in this form as it creates footgun for users of this library.

There needs to be either a hard-limit on the number of peers tracked, or an explicit opt-in via constructor option or ENV variable.

lidel (changes requested):

bitswap/peer-prom-tracker/peer-prom-tracker.go
lidel: :thought_balloon:  I think if we wanted to have metrics similar to this, we could measure P95 globally without running into the cardinality problem.

To do so, one would define `Objectives` in `SummaryOpts` to be  P50, P75, P95 etc, and calculate `messages-received`, `messages-sent`, `bytes-received` etc  across all peers, not specific per peer. This way we get useful P95 metric with known error margin, without exploding the time series.

Jorropo:
This was comunicated as not needed and we don't want to run it in production in Kubo due to the cardinality issue.
So I'll close this an leave the branch in case it comes back up later.

PR #422, opened 2023-08-02
feat(routing/http)!: delegated peer routing server and client, IPIP 417

hacdias:
Implementation of https://github.com/ipfs/specs/pull/417. Requires https://github.com/ipfs/boxo/pull/333.

- [x] Removes generic `Provide`.
- [x] Marks `ProvideBitswap` and all related methods and types as deprecated.
- [x] Add Peer Routing functions to the client. Tests.
- [x] Add Peer Routing functions to the server. Tests.
- [x] Chagelog for more updates. 

**Kubo PR** (includes #333): https://github.com/ipfs/kubo/pull/9877


BigLep:
Just wanted to say thanks for jumping on this @hacdias  as this is key for https://github.com/ipfs/helia/issues/182

CHANGELOG.md
```diff
@@ -16,12 +16,27 @@ The following emojis are used to highlight certain changes:
...
-* The `routing/http` client and server now support Delegated IPNS as per [IPIP-379](https://specs.ipfs.tech/ipips/ipip-0379/).
+* ✨ The `routing/http` package has received the following additions:
+  * Supports Delegated IPNS as per [IPIP-379](https://specs.ipfs.tech/ipips/ipip-0379/).
+  * Supports Delegated Peer Routing as per [IPIP-417](https://github.com/ipfs/specs/pull/417).
 
 ### Changed
 
+* 🛠 The `routing/http` package has suffered the following modifications:
+  * Client `FindProviders` has been renamed to `GetProviders`. Similarly, the
+    required function names in the server `ContentRouter` have also been updated
```
lidel: nit: https://github.com/ipfs/boxo/pull/333 adds `FindIPNSRecord`, see [comment](https://github.com/ipfs/boxo/pull/422#discussion_r1300710279) below

routing/http/contentrouter/contentrouter.go
```diff
@@ -3,55 +3,44 @@ package contentrouter
...
-var logger = logging.Logger("service/contentrouting")
-
-const ttl = 24 * time.Hour
+var logger = logging.Logger("routing/http/contentrouter")
 
 type Client interface {
-	ProvideBitswap(ctx context.Context, keys []cid.Cid, ttl time.Duration) (time.Duration, error)
-	FindProviders(ctx context.Context, key cid.Cid) (iter.ResultIter[types.ProviderResponse], error)
+	GetProviders(ctx context.Context, key cid.Cid) (iter.ResultIter[types.Record], error)
+	GetPeers(ctx context.Context, pid peer.ID) (peers iter.ResultIter[types.Record], err error)
```
lidel: :thought_balloon: Looking at this from different perspective, I am unsure if we want to change these to be `Get`.
In practice routing system only _Finds_ _some of them_, it never _Gets_ deterministic set of results.

Also, we have some ossification already for `FindProviders`. The custom Routing in Kubo confing  already uses `Find*`, and these keywords are  used by third-party projects ([example](https://github.com/ipni/index-provider/blob/main/README.md?plain=1#L146-L163)) and has things like:

```json
"Methods": {
      "find-peers": {
        "RouterName": "WanDHT"
      },
      "find-providers": {
        "RouterName": "ParallelHelper"
      },
      "get-ipns": {
        "RouterName": "WanDHT"
      },
      "provide": {
        "RouterName": "ParallelHelper"
      },
      "put-ipns": {
        "RouterName": "WanDHT"
      }
    },
```

`find-providers` is being used in production, and if we have it in config but `GetProviders` in code, it is bad UX/DX. 

@hacdias thoughts?

My initial idea would be to avoid changing `FindProviders` convention, since it is being used already in Kubo config for provider lookups.
 Instead, we can tweak ipns router names in Kubo config, as these are not used much yet, and make them consistent with provider ones.

My thinking is for Kubo config to  match go func names frotm this library, so:
- `find-providers` → `FindProviders
- `provide` → `Provide`
- `find-ipns` → `FindIPNS`
- `provide-ipns` → `ProvideIPNS` (might be a better name than "publish" because it allows us to do third-party providing of signed IPNS records, even when we are not original publisher)

routing/http/contentrouter/contentrouter.go
```diff
@@ -3,55 +3,44 @@ package contentrouter
...
 	"github.com/multiformats/go-multihash"
 )
 
-var logger = logging.Logger("service/contentrouting")
-
-const ttl = 24 * time.Hour
+var logger = logging.Logger("routing/http/contentrouter")
 
 type Client interface {
-	ProvideBitswap(ctx context.Context, keys []cid.Cid, ttl time.Duration) (time.Duration, error)
```
lidel, line 24: :warning:  i've realized that  if we remove this undocumented API, we will break https://github.com/ipni/index-provider/blob/3d81887128d9baa287324be93bda48307f5e8800/delegatedrouting/listener.go#L275

Let's be kinder here and avoid rug-pulling folks from bedrock working on the index-provider sidecar (and any third-party devs that use this)  until https://github.com/ipfs/specs/pull/378 is figured out.

@hacdias  thoughts what would be less work? I feel it might be (A) but lmk.

## (A) Restore the `ProvideBitswap`

- [ ]  restore and mark `ProvideBitswap` as  `// Deprecated: transport-agnostic provide is being worked on in https://github.com/ipfs/specs/pull/378` 
  - [ ] maybe also add generic `Provide` to the interface but make it no-op with comment or error with link to wip IPIP at https://github.com/ipfs/specs/pull/378
- [ ]  restore `ProvideMany` implementation that uses `ProvideBitswap` for now and `// TODO` to switch to `Provide`
 
 The goal here is to not break existing software until we have replacement for PURs, but signal  `ProvideBitswap` is deprecated and `Provide` is work in progress.

## (B)  Add generic  `Provide` ?

- [ ] interface: add `Provide` in place of removed `ProvideBitswap`
- [ ]  restore `ProvideMany` implementation but make it use `Provide` and `peer` schema  

This would  force `ipni/index-provider`  to switch to `peer` schema for puts (?). 
I'm not a fan as it grows scope, but maybe easier than it seems?

## (C)  Ask `/ipni/index-provider` maintainers if ok to break bitswap PUTs for now

We have sync with Bedrock on Tuesday ([content routing workgroup](https://www.notion.so/pl-strflt/Content-Routing-Workgroup-e59fa94a9c3f48d58480b7daf15bd356)) and we also have private sync on Friday. 

If we wan to avoid  (A) and (B) we could ask them if it is acceptable to break bitswap PUTs until we have bandwidth to finish and implement [IPIP-378](https://github.com/ipfs/specs/pull/378) is done.

lidel (changes requested):
Slight concern around inconsitent naming across our stack + breaking PUTs for existign users like  `ipni/index-provider` (details inline).

routing/http/contentrouter/contentrouter.go
hacdias: @lidel I decided to Go with A, even though I would rather prefer a B. I marked a bunch of types as deprecated, and now the CI is complaining because we use them ourselves 😭. I'm gonna see if I can "fix" that.

lidel: Thanks. Filled https://github.com/ipni/index-provider/issues/403 so we don't forget to update it when the IPIP lands.

routing/http/contentrouter/contentrouter_test.go
```diff
@@ -108,26 +128,40 @@ func TestFindProvidersAsync(t *testing.T) {
...
+		},
+		&types.PeerRecord{
+			Schema:    types.SchemaPeer,
+			ID:        &p3,
+			Protocols: []string{"transport-bitswap"},
+		},
+		&types.PeerRecord{
+			Schema:    types.SchemaPeer,
+			ID:        &p4,
+			Protocols: []string{"transport-horse"},
```
lidel, line 154: :horse:

routing/http/contentrouter/contentrouter_test.go
```diff
@@ -108,26 +128,40 @@ func TestFindProvidersAsync(t *testing.T) {
...
 		},
-		&types.UnknownProviderRecord{
-			Protocol: "UNKNOWN",
+		&types.UnknownRecord{
+			Schema: "UNKNOWN",
```
lidel: nit: cid.contact production uses lowercase ([example](https://cid.contact/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi)), lets do the same here
```suggestion
			Schema: "unknown",
```

routing/http/contentrouter/contentrouter_test.go
```diff
@@ -137,7 +171,117 @@ func TestFindProvidersAsync(t *testing.T) {
...
+
+func TestFindPeer(t *testing.T) {
+	ctx := context.Background()
+	client := &mockClient{}
+	crc := NewContentRoutingClient(client)
+
+	p1 := peer.ID("peer1")
+	ais := []types.Record{
+		&types.UnknownRecord{
+			Schema: "Unknown",
```
lidel:
```suggestion
			Schema: "unknown",
```

routing/http/contentrouter/contentrouter_test.go
```diff
@@ -108,26 +128,40 @@ func TestFindProvidersAsync(t *testing.T) {
...
+			Protocols: []string{"transport-bitswap"},
+		},
+		&types.PeerRecord{
+			Schema:    types.SchemaPeer,
+			ID:        &p4,
+			Protocols: []string{"transport-horse"},
 		},
-		&types.UnknownProviderRecord{
-			Protocol: "UNKNOWN",
+		&types.UnknownRecord{
```
lidel, line 156: :warning:  @hacdias mind adding `ID` field here and including it in expects to ensure we are not ignoring responses with `Schema: "unknown"` that have valid `ID` (and optional `Addrs`)?

Rationale: cid.contact returns all nft.storage announcements like this ([example](https://cid.contact/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi)) so it is really important to not have regressions here:

```json
{
  "Protocol":"unknown",
  "Schema":"unknown",
  "ID":"QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
  "Addrs":["/dns4/dag.w3s.link/tcp/443/https"]
}
```

 to ensure smooth transition, the client should detect when `UnknownRecord` has `ID` and `Addrs`  fields and convert it to `PeerRecord`.

routing/http/server/server_test.go
```diff
@@ -73,20 +83,19 @@ func TestResponse(t *testing.T) {
...
-				Schema:   types.SchemaBitswap,
-				ID:       &pid2,
-				Addrs:    []types.Multiaddr{},
+		results := iter.FromSlice([]iter.Result[types.Record]{
+			{Val: &types.PeerRecord{
+				Schema:    types.SchemaPeer,
+				ID:        &pid,
+				Protocols: []string{"transport-bitswap"},
+				Addrs:     []types.Multiaddr{},
 			}},
```
lidel, line 92: nit: in preexisting cases like this let's keep at least one with `Schema: types.SchemaBitswap,` because that is what https://cid.contact/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi currently returns, and we should keep a regression test for it.

lidel (approved):
Thank you @hacdias, I think this cooked long enough to give it a try as opt-in in the next kubo release (https://github.com/ipfs/kubo/pull/9877).
Feel free to merge once we add missing regression tests for `Schema: "unknown"` (details inline)

routing/http/contentrouter/contentrouter_test.go
hacdias: @lidel I don't think there is a regression here. Anything other than `schema: bitswap` has _always_ been ignored:

https://github.com/ipfs/boxo/blob/303595bcdba7c4ff1e4186a8089a550df6da9d28/routing/http/contentrouter/contentrouter.go#L105-L138

If we add support for `unknown` stuff, we'd be entering a can of worms we might not be able to leave. I don't think we should add support here that did not exist. Also, right now, the contentrouter also filters out any peer that does not provide bitswap, as that was its behaviour before. I can change that, but is that what we want?

routing/http/contentrouter/contentrouter_test.go
hacdias: I changed it but it doesn't matter as anything other than bitswap has always been ignored: https://github.com/ipfs/boxo/pull/422#discussion_r1305340163

routing/http/contentrouter/contentrouter_test.go
hacdias: I changed it but it doesn't matter as anything other than bitswap has always been ignored: https://github.com/ipfs/boxo/pull/422#discussion_r1305340163

routing/http/server/server_test.go
hacdias: I re-added it.

hacdias:
@lidel I think this is ready to merge. Can you please take a look at my comment: https://github.com/ipfs/boxo/pull/422#discussion_r1305340163?

routing/http/contentrouter/contentrouter_test.go
hacdias: Also see https://github.com/ipfs/kubo/pull/9877#discussion_r1305482906.

lidel: Thanks for confirming this :pray:  If `Schema: unknown` was ignored before, I agree we should continue ignoring it, just wanted to be sure we don't regress on any behavior.

Filled https://github.com/ipni/indexstar/issues/147 to fix this on the other end, and have cid.contact return these using the `peer` schema.

gateway/gateway.go
lidel, line 1: @hacdias I don't think we should touch `boxo/gateway` in this PR? Changelog says nothing about this breaking change in `IPFSBackend` so i assume this is a bug (result of search and replace)?

routing/http/server/server.go
lidel, line 1: @hacdias  is there a way for the JSON produced by the server to NOT include `null`  fields?

```json
{
      "Addrs": null,
      "ID": "12D3KooWC7RmqL62K3gjxCeEs3Q1NP6Ca3xHN1Ht6ZwQqw86c5B6",
      "Protocols": null,
      "Schema": "peer"
    },
```

hacdias: Updated for `SchemaPeer` and `SchemaBitswap`.

routing/http/server/server.go
```diff
@@ -132,7 +147,145 @@ type server struct {
...
+func (s *server) findPeers(w http.ResponseWriter, r *http.Request) {
+	pidStr := mux.Vars(r)["peer-id"]
+
+	// pidStr must be in CIDv1 format. Therefore, use [cid.Decode]. We can't use
+	// [peer.Decode] because that would allow other formats to pass through.
+	cid, err := cid.Decode(pidStr)
+	if err != nil {
+		writeErr(w, "FindPeers", http.StatusBadRequest, fmt.Errorf("unable to parse peer ID: %w", err))
+		return
+	}
```
lidel: :warning: UX hiccup:  this error will be really, really annoying in Kubo, where commands like `ipfs swarm peers` return non-CID PeerIDs like `12D3KooWHj48PbuAPUf9LeovBJZEXn2A4rseW5z9wbTBwvM82nAq`.
User will copy it, and then get this error without any 


```console
$ curl  http://127.0.0.1:8280/routing/v1/peers/12D3KooWHj48PbuAPUf9LeovBJZEXn2A4rseW5z9wbTBwvM82nAq
unable to parse peer ID: invalid cid: selected encoding not supported
```

@hacdias I agree we should error here, but could you improve error mesaage?

Do what we already do at https://cid.ipfs.tech/#12D3KooWHj48PbuAPUf9LeovBJZEXn2A4rseW5z9wbTBwvM82nAq and return meaningful error that has CIDv1 (with libp2p-key codec) representation and asks user to try again with it (actionable, allows user to self-service and fix without asking on forums, googling).

CHANGELOG.md
```diff
@@ -16,10 +16,26 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+* ✨ The `routing/http` not supports Delegated Peer Routing as per [IPIP-417](https://github.com/ipfs/specs/pull/417).
```
```suggestion
* ✨ The `routing/http` implements Delegated Peer Routing introduced in [IPIP-417](https://github.com/ipfs/specs/pull/417).
```

CHANGELOG.md
```diff
@@ -16,10 +16,26 @@ The following emojis are used to highlight certain changes:
...
+
 ### Changed
 
+* 🛠 The `routing/http` package has suffered the following modifications:
```
```suggestion
* 🛠 The `routing/http` package received the following modifications:
```

CHANGELOG.md
```diff
@@ -16,10 +16,26 @@ The following emojis are used to highlight certain changes:
...
+  * `ReadBitswapProviderRecord` has been renamed to `BitswapRecord` and marked as deprecated.
+    From now on, please use the protocol-agnostic `PeerRecord` for most use cases. The new
+    Peer Schema has been introduced in [IPIP-417](https://github.com/ipfs/specs/pull/417).
+
 ### Removed
 
+* 🛠 The `routing/http` package has suffered the following removals:
+  * Server and client no longer support the generic `Provide` method for content routing.
+    `ProvideBitswap` is still usable, but marked as deprecated. A protocol-agnostic 
+    provide mechanism is being worked on in [IPIP-378](https://github.com/ipfs/specs/pull/378).
```
```suggestion
* 🛠 The `routing/http` package experienced following removals:
  * Server and client no longer support the experimental `Provide` method.
    `ProvideBitswap` is still usable, but marked as deprecated. A protocol-agnostic 
    provide mechanism is being worked on in [IPIP-378](https://github.com/ipfs/specs/pull/378).
```

PR #433, opened 2023-08-15
chore: go 1.20

hacdias:


hacdias:
Close for https://github.com/ipfs/boxo/pull/431

PR #434, opened 2023-08-15
fix!: add escaped abspath header

hacdias:
See https://github.com/ipfs/kubo/issues/10065 and https://github.com/golang/go/issues/60674.

**Problem:** we used to send binary data in `abspath` header when sending multipart requests. This is not compliant to the RFC but used to be allowed by Go (see linked issue). It is no longer possible to read multipart requests with binary data on their headers from Go 1.20.

**Solution:** add a new header with the encoded version of the `abspath`. Unfortunately, we cannot send both the old and the new header, as Go 1.20 parses all headers when reading the multipart "part" and immediately fails. Therefore, we have to be careful. In this PR, we add an option to `NewMultiFileReader`: this option will tell which header to use. See code for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ The following emojis are used to highlight certain changes:
 
 - HTTP Gateway API: Not having a block will result in a 5xx error rather than 404
 - HTTP Gateway API: CAR requests will return 200s and a CAR file proving a requested path does not exist rather than returning an error
+- 🛠 `MultiFileReader` has been updated with a new header with the encoded file name instead of the plain filename, due to a regression found in  [`net/textproto`](https://github.com/golang/go/issues/60674). This only affects files with binary characters in their name. By keeping the old header, we maximize backwards compatibility.
+  |            | New Client | Old Client  |
+  |------------|------------|-------------|
+  | New Server | ✅         | 🟡*         |
+  | Old Server | ✅         | ✅          |
+   *Old clients can only send Unicode file paths to the server.
 
 ### Security
 
--- a/files/multifilereader.go
+++ b/files/multifilereader.go
@@ -29,19 +29,26 @@ type MultiFileReader struct {
 	// if true, the content disposition will be "form-data"
 	// if false, the content disposition will be "attachment"
 	form bool
+
+	// if true, 'abspath' header will be sent with raw (potentially binary) file
+	// name. This must only be used for legacy purposes to talk with old servers.
+	// if false, 'abspath-encoded' header will be sent with %-encoded filename
+	rawAbsPath bool
 }
 
 // NewMultiFileReader constructs a MultiFileReader. `file` can be any `commands.Directory`.
 // If `form` is set to true, the Content-Disposition will be "form-data".
-// Otherwise, it will be "attachment".
-func NewMultiFileReader(file Directory, form bool) *MultiFileReader {
+// Otherwise, it will be "attachment". If `rawAbsPath` is set to true, the
+// "abspath" header will be sent. Otherwise, the "abspath-encoded" header will be sent.
+func NewMultiFileReader(file Directory, form, rawAbsPath bool) *MultiFileReader {
 	it := file.Entries()
 
 	mfr := &MultiFileReader{
-		files: []DirIterator{it},
-		path:  []string{""},
-		form:  form,
-		mutex: &sync.Mutex{},
+		files:      []DirIterator{it},
+		path:       []string{""},
+		form:       form,
+		rawAbsPath: rawAbsPath,
+		mutex:      &sync.Mutex{},
 	}
 	mfr.mpWriter = multipart.NewWriter(&mfr.buf)
 
@@ -114,7 +121,12 @@ func (mfr *MultiFileReader) Read(buf []byte) (written int, err error) {
 
 			header.Set("Content-Type", contentType)
 			if rf, ok := entry.Node().(FileInfo); ok {
-				header.Set("abspath", rf.AbsPath())
+				if mfr.rawAbsPath {
+					// Legacy compatibility with old servers.
+					header.Set("abspath", rf.AbsPath())
+				} else {
+					header.Set("abspath-encoded", url.QueryEscape(rf.AbsPath()))
+				}
 			}
 
 			_, err := mfr.mpWriter.CreatePart(header)
--- a/files/multifilereader_binary_go119_test.go
+++ b/files/multifilereader_binary_go119_test.go
@@ -1,0 +1,10 @@
+//go:build !go1.20
+
+package files
+
+import "testing"
+
+func TestAbspathHeaderWithBinaryFilenameSucceeds(t *testing.T) {
+	// Simulates old client talking to old server (< Go 1.20).
+	runMultiFileReaderToMultiFileTest(t, true, true, false)
+}
--- a/files/multifilereader_binary_go120_test.go
+++ b/files/multifilereader_binary_go120_test.go
@@ -1,0 +1,13 @@
+//go:build go1.20
+
+package files
+
+import (
+	"testing"
+)
+
+func TestAbspathHeaderWithBinaryFilenameFails(t *testing.T) {
+	// Simulates old client talking to new server (>= Go 1.20). Old client will
+	// send the binary filename in the regular headers and the new server will error.
+	runMultiFileReaderToMultiFileTest(t, true, true, true)
+}
--- a/files/multifilereader_test.go
+++ b/files/multifilereader_test.go
@@ -10,91 +10,89 @@ import (
 
 var text = "Some text! :)"
 
-func TestMultiFileReaderToMultiFile(t *testing.T) {
-	do := func(t *testing.T, binaryFileName, rawAbsPath, expectFailure bool) {
-		var (
-			filename string
-			file     File
-		)
+func makeMultiFileReader(t *testing.T, binaryFileName, rawAbsPath bool) (string, *MultiFileReader) {
+	var (
+		filename string
+		file     File
+	)
 
-		if binaryFileName {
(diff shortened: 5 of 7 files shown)


Jorropo (changes requested):
So after discussions with @aschmahmann about concerns of silent breaking changes and @hacdias about fixing this, we are gonna go with this solution for binary file names in the `ipfs add` endpoint:
- leave current `abspath` header as-is, this header wont support binary encoded filepaths anymore due to the change in golang std;
- add a new `abspath-percent` header that is exactly like `abspath` but `url.QueryEscape`ed, kubo will then prefer using `abspath-percent` if present but use `abspath` if not
- new clients will send both headers

This gives us this compatibility table:
|            | New Client                            | Old Client                    |
|------------|---------------------------------------|-------------------------------|
| New Server | Works fully through `abspath-percent` | Binary file paths do not work  |
| Old Server | Works fully through `abspath`         | Works fully through `abspath` |

I don't know if headers are lazily or not decoded, so we might need to add an other response header indicating if `abspath-percent` is supported and do a HEAD request to identify if we are talking to a new or old server.

hacdias:
Note that this will be more involved than initially planned. The current PR is not yet ready. We cannot send both headers at the same time. The `net/textproto` package will decode all headers immediately and therefore fail if it sees the new header. Therefore, we need to be able to detect at the client if the server is "new enough".

My idea is to add a new argument to `NewMultiFileReader` where we indicate if we want to send the raw absolute path or the new encoded one. I don't see where else to do this. This is used in the following places:

- `kubo/client/rpc` in 2 places
- `go-ipfs-cmds` (I would argue this should be part of Kubo) in 2 places
- `go-ipfs-api` in 6 places

In those 3 packages. My idea is to just fetch `/api/v0/version` once at the first time it is required. Check if Kubo version if `>= 0.23.0` (including dev versions). I don't think this will be a necessarily easy change.

Jorropo:
~~I was thinking the server would have a new `Support-Percent-Abspath-Header` then `NewMultiFileReader` would start with a `HEAD` request. I guess this would be less efficient because we would need to check each time, or cache it once.
I think a new option is fine as long as you follow the functional option parameters (so it defaults to `%` encoding unless you explicitly ask it not to)~~

Edit: this code does not do HTTP stuff, it just handles readers and writers, @hacdias options looks the least awful one that is still somewhat backward compatible.

CHANGELOG.md
```diff
@@ -32,6 +32,11 @@ The following emojis are used to highlight certain changes:
 ### Fixed
 
 - Address a Bitswap findpeers / connect race condition that can prevent peer communication ([#435](https://github.com/ipfs/boxo/issues/435))
+- 🛠 `MultiFileReader` has been updated with a new header with the encoded file name instead of the plain filename, due to a regression found in  [`net/textproto`](https://github.com/golang/go/issues/60674). This only affects files with binary characters in their name. By keeping the old header, we maximize backwards compatibility.
+  |            | New Client | Old Client  |
+  |------------|------------|-------------|
+  | New Server | ✅         | ❌          |
+  | Old Server | ✅         | ✅          |
```
Jorropo:
```suggestion
  |            | New Client | Old Client  |
  |------------|------------|-------------|
  | New Server | ✅         | *🟡         |
  | Old Server | ✅         | ✅          |
  *Old clients can only send ASCII file paths to the server.
```

Jorropo: Actually do you know if it is ASCII or unicode ?

Jorropo (approved):
LGTM

files/multifilereader.go
```diff
@@ -114,7 +121,12 @@ func (mfr *MultiFileReader) Read(buf []byte) (written int, err error) {
 
 			header.Set("Content-Type", contentType)
 			if rf, ok := entry.Node().(FileInfo); ok {
-				header.Set("abspath", rf.AbsPath())
+				if mfr.rawAbsPath {
+					// Legacy compatibility with old servers.
+					header.Set("abspath", rf.AbsPath())
+				} else {
+					header.Set("abspath-encoded", url.QueryEscape(rf.AbsPath()))
+				}
```
lidel, line 129: @hacdias unsure if useful, but there is prior art in HTTP stack for doing this kind of thing without introducing a new header.

We could implement signaling convention from [RFC6266](https://httpwg.org/specs/rfc6266.html) and keep a single  `abspath` header + remove need for passing `rawAbsPath bool`. 

This way we have one header and:

- If filename is ascii-only, nothing changes:
  ```
  Abspath:  resume.txt
  ```
- If we have non-ASCII characters, like `résumé.txt` then we do `url.QueryEscape` AND  prefix it with `utf-8''`: 
  ```
  Abspath:  utf-8''r%C3%A9sum%C3%A9.txt
  ```

Receiving end knows if they need to un-escape based on the presence of `utf-8''` prefix.

Just food for thought.

files/multifilereader_test.go
```diff
@@ -19,57 +21,62 @@ func getTestMultiFileReader(t *testing.T) *MultiFileReader {
...
-	}
+	t.Run("With Header 'abspath'", func(t *testing.T) {
+		do(t, true)
+	})
 
-	if it.Next() || it.Err() != nil {
-		t.Fatal("iterator didn't work as expected")
-	}
+	t.Run("With Header 'abspath-encoded'", func(t *testing.T) {
+		do(t, false)
```
lidel: :warning: we only have tests for ASCII, would be good to add  separate one for `résumé.txt` to truly test `abspath-encoded`

lidel (changes requested):
Seems to be missing test for scenario with non-ASCII `abspath` this PR is trying to fix?
(+ might be simplified – see comment inline, but that is fine to disregard)

files/multifilereader.go
hacdias: @lidel I see. We can then go around and use the same header. However, we still need to pass some variable (like `rawAbsPath`). This is client code. The client has to know if they're talking to an old server or not. The old server will not support decoding `utf-8''`-starting-headers. Therefore, we still have to send the right one.

CHANGELOG.md
hacdias: @Jorropo from my tests it seems to be Unicode.

files/multifilereader.go
hacdias: That considered, I don't think this would necessarily simplify. What do you think?

hacdias:
~~@lidel, as asked (makes sense), I added tests with the new header and **binary filenames**. Surprise surprise: our CI is still running on Go 1.19 for `go this`. Therefore, all tests will fail on `go this` and pass on `go next`. Therefore, this is likely blocked by #431 unless we want to be merging failing tests.~~

Fixed with separate tests with tags as per @Jorropo's suggestion.

files/readerfile.go
```diff
@@ -21,6 +21,10 @@ func NewBytesFile(b []byte) File {
 	return &ReaderFile{"", bytesReaderCloser{bytes.NewReader(b)}, nil, int64(len(b))}
 }
 
+func NewBytesFileWithPath(abspath string, b []byte) File {
```
hacdias: Maybe I should unexport is.

files/multifilereader.go
lidel: Agree, keeping two headers covers more edge cases with the smallest amount of work. 
We can use `utf-8''` thing in the future, but for now let's keep `abspath-encoded`.

(iiuc tgus header is only used with [`--nocopy` from Filestore, which is still marked  as experiment](https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md#ipfs-filestore) in Kubo)

files/multifilereader_binary_go119_test.go
```diff
@@ -0,0 +1,10 @@
+//go:build !go1.20
```
lidel, line 1: :thought_balloon:  @hacdias would it be better to have `go1.19` here to avoid the problem when we try to build with `go1.21`? my understanding is that 21 will behave the same way as 20, and we don't need to worry about <19.
```suggestion
//go:build go1.19
```

files/multifilereader_binary_go120_test.go
```diff
@@ -0,0 +1,13 @@
+//go:build go1.20
```
lidel, line 1:
```suggestion
//go:build !go1.19
```

lidel (approved):
The tests look hairy, but acceptable given this header is only used for [experimental filestore feature in Kubo](https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md#ipfs-filestore).  We will have to revisit code related to `--nocopy` at some point anyway.
Small nit inline but lgtm.

files/multifilereader_binary_go119_test.go
Jorropo: the `go1.20` tag will be set in 1.21 too.

files/multifilereader_binary_go120_test.go
Jorropo: This is wrong `go1.19` is set when building with 1.20 or 1.21 so this will only trigger for `1.18` and bellow.
Generally the tag `go1.X` is set for releases of go `>= 1.X`.

files/multifilereader_binary_go119_test.go
hacdias: I'm marking this as resolved due to @Jorropo explanation.

files/multifilereader_binary_go120_test.go
hacdias: I'm marking this as resolved due to @Jorropo explanation.

hacdias:
@Jorropo @lidel we need to **release** this before we bubble up to https://github.com/ipfs/go-ipfs-api/pull/305 and https://github.com/ipfs/go-ipfs-cmds/pull/243 so we don't have commit tags. Only when those two are merged can we also update Kubo.

PR #439, opened 2023-08-17
style: run gofumpt

Jorropo:


Jorropo:
https://github.com/ipfs/kubo/pull/10083#issuecomment-1682338409

PR #440, opened 2023-08-17
fix(gw):  IPIP-402 CARs return useful blocks on not found errors

aschmahmann:
Fix extracted from #369 that could be suitable for a patch release.

### DIFF
--- a/.github/workflows/gateway-conformance.yml
+++ b/.github/workflows/gateway-conformance.yml
@@ -40,7 +40,7 @@ jobs:
 
       # 4. Run the gateway-conformance tests
       - name: Run gateway-conformance tests
-        uses: ipfs/gateway-conformance/.github/actions/test@261cc49391dcd2bddd7704ad28b236dc4a63c424
+        uses: ipfs/gateway-conformance/.github/actions/test@v0.3
         with:
           gateway-url: http://127.0.0.1:8040
           json: output.json
--- a/.github/workflows/gateway-conformance.yml
+++ b/.github/workflows/gateway-conformance.yml
@@ -40,7 +40,7 @@ jobs:
 
       # 4. Run the gateway-conformance tests
       - name: Run gateway-conformance tests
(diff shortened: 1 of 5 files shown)


gateway/gateway_test.go
```diff
@@ -763,7 +763,7 @@ func TestErrorBubblingFromBackend(t *testing.T) {
 		})
 	}
 
-	testError("404 Not Found from IPLD", &ipld.ErrNotFound{}, http.StatusNotFound)
+	testError("500 Not Found from IPLD", &ipld.ErrNotFound{}, http.StatusInternalServerError)
```
aschmahmann, line 766: I think this was a mistake. `ipld.ErrNotFound` is an "I can't find this block"  kind of error which IIUC is supposed to be 5xx as opposed to "this path doesn't exist/IPLD errors" which are 4xx.

I think this error could also reasonably be a `502`

.github/workflows/gateway-conformance.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
       # 4. Run the gateway-conformance tests
       - name: Run gateway-conformance tests
-        uses: ipfs/gateway-conformance/.github/actions/test@v0.3
+        uses: ipfs/gateway-conformance/.github/actions/test@261cc49391dcd2bddd7704ad28b236dc4a63c424
```
lidel: - [x] Change to `@v0.3` (or later)  before merge

lidel (approved):
I've tested this in https://github.com/ipfs/kubo/pull/10084 and looks green :+1:

hacdias (approved):

.github/workflows/gateway-conformance.yml
hacdias:
```suggestion
        uses: ipfs/gateway-conformance/.github/actions/test@v0.3
```

PR #441, opened 2023-08-21
style: run gofumpt

Jorropo:
#439 but not using my fork this time


hacdias (approved):
Very interesting to see how inconsistent our code is regarding empty lines.

PR #443, opened 2023-08-21
routing/http: add more type and WithDynamicProviderInfo

Jorropo:
This is required to fix a bug in Kubo where we advertise the *listening* addresses.
We can't pass our real addresses once since thing like autonat takes time.

### DIFF
--- a/routing/http/client/client.go
+++ b/routing/http/client/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"mime"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


hacdias:
@Jorropo please note that `routing/http` is suffering immense changes in #422.

Jorropo:
@hacdias this is a pretty simple change that adds a callback instead of statically hardcoding the listen addresses, I would be surprised if this can't be forward ported.

hacdias:
@Jorropo feel free to rebase and ping me after.

Jorropo:
I'll probably do that for Kubo v0.24

BigLep:
2023-10-05 conversation: we aren't going to invest time in this until /routing/v1 PUTs get addressed: https://github.com/ipfs/specs/pull/378.  We aren't spending time fixing bugs here.

PR #436, opened 2023-08-17
bitswap/network: refactor connectEventManager more simply in in bsnet 

Jorropo:
I think this is fine because libp2p's .Connect call does not return before the .Connected callback has returned.

This new code also have similar dedup logic and state transitions.

Fixes #432

PR #435, opened 2023-08-16
feat(connecteventmanager): block Connected() until accepted

rvagg:
Fixes: https://github.com/ipfs/boxo/issues/432

Minimal attempt at solving #432

PR #444, opened 2023-08-21
Revert "feat(connecteventmanager): block Connected() until accepted (#435)" and tests

Jorropo:
This reverts commit 7ec68c5e5adfb13e52f93c20e7c5aadf4860a871.
This reverts commit 59a2bca3f0924d5a9cc61bbb6c33dbd483f4a0f4.
This reverts commit 1d2f5e511e9fb0e1ff4c2b93aea421fc2091e3f5.

See https://github.com/ipfs/boxo/pull/436#issuecomment-1683116240

cc @hannahhoward @rvagg if the fix were working out for you, I think you can continue to run the previous commit while a proper fix is made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,6 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
-- Address a Bitswap findpeers / connect race condition that can prevent peer communication ([#435](https://github.com/ipfs/boxo/issues/435))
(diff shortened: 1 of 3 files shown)


aschmahmann:
LGTM, the fix was likely worse than the problem since blocking go-libp2p connections to finish processing in Bitswap can lead to misery across the stack (we've had this problem before). This will have to get fixed another way that doesn't block go-libp2p connections.

aschmahmann (approved):

PR #449, opened 2023-08-22
blockservice/test: passthrough blockservice options [skip changelog]

Wondertan:
Self-explanatory

### DIFF
--- a/blockservice/test/mock.go
+++ b/blockservice/test/mock.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Mocks returns |n| connected mock Blockservices
-func Mocks(n int) []blockservice.BlockService {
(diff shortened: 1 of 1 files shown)


Wondertan:
~~Btw @Jorropo, what's the status of go-merkledag repo? Is it supposed to be migrated to boxo or moved into?~~

~~We use https://github.com/ipfs/go-merkledag/blob/master/test/utils.go#L20 from it and it depends on the old blockservice, so for now I decided to use canonical test utils defined in this pkg~~

Nvm, found https://github.com/ipfs/boxo/tree/main/ipld/merkledag

Jorropo (approved):

BigLep:
> Nvm, found https://github.com/ipfs/boxo/tree/main/ipld/merkledag

There is discussion in the README as well: https://github.com/ipfs/go-merkledag#status

PR #450, opened 2023-08-22
Revert "ci: uci/copy-templates"

hacdias:
Reverts ipfs/boxo#431.

This updates `go-libp2p` and those updates are not yet ready in Kubo. Since we don't have the reciprocal PR in Kubo, it hinders the development of other PRs since Kubo doesn't build with the latest Boxo, and therefore no commits on top of it.

Unless a PR for Kubo magically shows up within a few hours, I think we should revert this to allow work to continue without major issues.


galargh:
Are there any boxo consumers, other than Kubo, that might be waiting for go 1.21 compatibility?

hacdias:
> Are there any boxo consumers, other than Kubo, that might be waiting for go 1.21 compatibility?

@galargh unsure. 

---

I made a PR on Kubo side: https://github.com/ipfs/kubo/pull/10094

I will close this.

Issue #432, opened 2023-08-15
Bitswap peer connection race


rvagg:
Experienced via Lassie's CI suite which has suddenly become flaky with some new workloads. Using the [testing/sws-logging](https://github.com/ipfs/boxo/tree/testing/sws-logging) branch here (started by @hannahhoward, continued by me) to diagnose with additional debug printing. It evidences in an inability to communicate with a peer that we know has the block, and bitswap has been told has the block.

See logs below, which come from [here](https://github.com/filecoin-project/lassie/actions/runs/5865827197/job/15903479085?pr=371) using 715365a23d36ccd0fe2c9cf95ec16afa0d50f39b on the [testing/sws-logging](https://github.com/ipfs/boxo/tree/testing/sws-logging) branch. Here's what I believe happens:

* `session#findMorePeers` -> `ProviderQueryManager#FindProvidersAsync`
* `ProviderQueryManager#findProviderWorker` does its thing, processing the queue of providers to connect to, performs a `pqm.network.ConnectTo()` which is goes through (bitswap/network/ipfs_impl.go) `bsnet#ConnectTo()` to just do a basic `host.Connect()`
* We get a `Connected()` notification which is received by (bitswap/network/ipfs_impl.go) `connectEventManager#Connected()` and it does a `c.setState(p, stateResponsive)` on that peer.
* Meanwhile, the `ProviderQueryManager#FindProvidersAsync` has returned, the `Connect()` returned, we even got a `Connected()`, so the `Session` invokes `sessionWantSender#Update(peer, nil, have, nil)`, i.e. the `peer` has the cid `have`, this is a "change".
* In `sessionWantSender` we go from `onChange()` to `processUpdates()` to `updateWantBlockPresence()` to `wantInfo#setPeerBlockPresence()` to `wantInfo#calculateBestPeer()` - we see these in the logs
* After the call to `processUpdates()`, `sessionWantSender` moves on to `sendNextWants()`, which compiles them and calls `sendWants()` which in turn calls `PeerManager#SendWants()`.
* **BUT** it turns out that `PeerManager` doesn't have a `peerQueues` entry for this peer .. yet ..
[...]

PR #452, opened 2023-08-23
feat(session): do not record erroneous session want sends

hannahhoward:
# Goals

Find acceptable fix for #432 

# Implementation

The basic idea here is there's nothing wrong with the connection event manager -- the problem lies in the fact that the PeerManager never tells the caller it didn't actually send anything cause the peer appeared disconnected. The SessionWantSender in turn updates the state of sent wants based on incorrect assumption something actually happened. This PR deals with the problem by helping the SessionWantSender avoid getting into an incorrect state by not updating its state when SendWants does nothing.

The subsequent connect event should trigger the sending of the WantBlock then.

# For discussion

This is just an experiment and a suggestion for someone else to pick up, to help solve #432 more quickly -- I'm not doing the leg work of testing and change log etc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- Do not erroneously update the state of sent wants when a send a peer disconnected and the send did not happen. [#452](https://github.com/ipfs/boxo/pull/452)
+
 ### Security
 
 ## [v0.24.3]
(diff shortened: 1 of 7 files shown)


rvagg:
seems reasonable, aside from probably needing some form of cleanup for cases where the error doesn't ever resolve - we don't want to continue to try this in perpetuity, unless there is already an alternative cleanup mechanism in place for these wants

hannahhoward:
Presumably, the case where the error does not resolve is one where a peer has disconnected, in which case, we will eventually get a disconnected message.

hsanjuan (approved):
LGTM. Let's try it out.

PR #459, opened 2023-09-04
feat!: namesys refactor, ipns TTL bubbled up to gateway

### LINKED ISSUE #329: Gateway: Cache-Control, max-age=ttl, ETag for /ipns requests
The gateway does not seem to provide a Cache-Control/max-age header or an ETag for `/ipns` URLs.

It might be useful to assume _some_ reasonable non-zero max-age for content on IPNS. A minute? Fifteen seconds?

It would also be good to serve an ETag corresponding to whatever IPFS hash the IPNS path resolved to.

hacdias:
Turns out, we already had the TTL for IPNS Records inside this package. We were just not bubbling it up. This, for IPNS Records. We can manage this without doing it for domains, but I think it would be great if we did. Closes #329.

- [x] Based on #334
- [x] Spec clarification: https://github.com/ipfs/specs/pull/436
- [x] Gateway Conformance: https://github.com/ipfs/gateway-conformance/pull/173
- [x] Kubo PR: https://github.com/ipfs/kubo/pull/10115


hacdias:
@lidel  IPNS TTL is easy. We already have it inside `namesys`, it's just a matter of returning it. DNS not so much. Our DNS resolution is 100% tied to [go-multiaddr-dns](https://github.com/multiformats/go-multiaddr-dns), where it was explicitly decided not to return the [TTL](https://github.com/multiformats/go-multiaddr-dns/issues/28#issuecomment-819675623). My suggestion is to set ETag and Max-Age for IPNS requests where we can return the TTL. If we can't, we just set the `Last-Modified` as we do now. What do you think?

namesys/dns_resolver.go
```diff
@@ -18,53 +18,45 @@ import (
...
 
 	var fqdn string
-	out := make(chan onceResult, 1)
+	out := make(chan ResolveResult, 1)
 	segments := strings.SplitN(name, "/", 2)
 	domain := segments[0]
 
 	if _, ok := dns.IsDomainName(domain); !ok {
-		out <- onceResult{err: fmt.Errorf("not a valid domain name: %s", domain)}
+		out <- ResolveResult{Err: fmt.Errorf("not a valid domain name: %s", domain)}
```
lidel:
```suggestion
		out <- ResolveResult{Err: fmt.Errorf("not a valid domain name: %q", domain)}
```

namesys/ipns_publisher.go
```diff
@@ -249,39 +247,39 @@ func waitOnErrChan(ctx context.Context, errs chan error) error {
...
-func PublishPublicKey(ctx context.Context, r routing.ValueStore, k string, pubk crypto.PubKey) error {
-	ctx, span := StartSpan(ctx, "PublishPublicKey", trace.WithAttributes(attribute.String("Key", k)))
+// PublishPublicKey stores the given [crypto.PubKey] for the given key in the [routing.ValueStore].
+func PublishPublicKey(ctx context.Context, r routing.ValueStore, key string, pubKey crypto.PubKey) error {
+	ctx, span := startSpan(ctx, "PublishPublicKey", trace.WithAttributes(attribute.String("Key", key)))
 	defer span.End()
 
-	log.Debugf("Storing pubkey at: %s", k)
-	pkbytes, err := crypto.MarshalPublicKey(pubk)
+	log.Debugf("Storing pubkey at: %s", key)
```
lidel:
```suggestion
	log.Debugf("Storing pubkey at: %q", key)
```

namesys/mpns.go
```diff
@@ -0,0 +1,310 @@
...
+)
+
+// nameSys is a multi-protocol [NameSystem] that implements generic IPFS naming.
+// It uses several [Resolver]s:
+//
+//  1. IPFS routing naming: SFS-like PKI names.
+//  2. dns domains: resolves using links in DNS TXT records
+//
+// It can only publish to: 1. IPFS routing naming.
+type nameSys struct {
```
lidel: nit:  "namesys" has always been lowercase, either do that or keep old name

```suggestion
type namesys struct {
```

```suggestion
type mpns struct {
```

namesys/namesys.go
lidel, line 1: rename to `interface.go`?

namesys/mpns.go
lidel, line 1: :thought_balloon:  if we rename `namesys.go` to `interfaces.go` we could rename this to `namesys.go`

namesys/interface.go
lidel, line 1: :thought_balloon: having `interface.go` was useful, made it clear what is the contract, see https://github.com/ipfs/boxo/pull/459#discussion_r1315112178

lidel (changes requested):
@hacdias sounds like a sensible compromise. quick thoughts:
- [x] update [specs for `Cache-Control` and `Last-Modified`](http://specs.ipfs.tech.ipns.localhost:8080/http-gateways/path-gateway/#cache-control-response-header)  to reflect that acting on TTL is SHOULD, but implementation that have no access to TTL information SHOULD return `Last-Modified` instead (replace language from legacy/new etc) [https://github.com/ipfs/specs/pull/436]
- [x] testing this will need something like `AnyOf` from https://github.com/ipfs/gateway-conformance/issues/153  i think?

hacdias:
@lidel thanks for the preliminary review. Do you think we should pass this TTL the way we did before for IPNS Records? That is, either the TTL from the record, the time to EOL, or the default. Or do you think it's best to pass strictly the given TTL? So far this was only being used for caching purposes, so I may need to update this if we really only want to return the TTL defined in the record.

https://github.com/ipfs/boxo/blob/d213160c2726a00b0b652dd07d382e875cde06c9/namesys/ipns_resolver.go#L114-L129

namesys/ipns_publisher.go
```diff
@@ -66,9 +64,9 @@ func IpnsDsKey(id peer.ID) ds.Key {
 //
 // This method will not search the routing system for records published by other
 // nodes.
-func (p *IpnsPublisher) ListPublished(ctx context.Context) (map[peer.ID]*ipns.Record, error) {
+func (p *IPNSPublisher) ListPublished(ctx context.Context) (map[peer.ID]*ipns.Record, error) {
```
hacdias: This code isn't used anywhere. It's been sitting here for 5 years without being used. I would rather kill it. If we want to keep it, probably will change the return to `map[*ipns.Name]*ipns.Record`.

namesys/ipns_publisher.go
```diff
@@ -113,7 +111,7 @@ func (p *IpnsPublisher) ListPublished(ctx context.Context) (map[peer.ID]*ipns.Re
 //
 // If `checkRouting` is true and we have no existing record, this method will
 // check the routing system for any existing records.
-func (p *IpnsPublisher) GetPublished(ctx context.Context, id peer.ID, checkRouting bool) (*ipns.Record, error) {
+func (p *IPNSPublisher) GetPublished(ctx context.Context, id peer.ID, checkRouting bool) (*ipns.Record, error) {
```
hacdias: Similarly to the comment to `ListPublished`, I'm not sure why this was exported. It's only used internally. If we remove `ListPublished`, I think this should be unexported.

hacdias:
@lidel thanks for the preliminary review. Here's some notes and further questions:

1. Specification update is in: https://github.com/ipfs/specs/pull/436
2. Kubo update is in: https://github.com/ipfs/kubo/pull/10115, failing at the moment because gateway conformance expects no `Cache-Control` in IPNS land.
3. Do you think we should pass this TTL the way we did before for IPNS Records? That is, either the TTL from the record, the time to EOL, or the default. Or do you think it's best to pass strictly the given TTL? So far this was only being used for caching purposes, so I may need to update this if we really only want to return the TTL defined in the record.

https://github.com/ipfs/boxo/blob/d213160c2726a00b0b652dd07d382e875cde06c9/namesys/ipns_resolver.go#L114-L129

4. Regarding testing, I think it would be useful to have `AnyOf`, but I can probably get around without it for now.
5. Should we add `Cache-Control` in the following situations (important before tackling conformance):
    a. Generated directory listings under `/ipns`?
    b. Directory listings that serve `index.html`?
6. Other comments inline.

namesys/interface.go
```diff
@@ -60,39 +57,152 @@ type NameSystem interface {
...
-	// adjust the limit you can specify it as an option.
-	Resolve(ctx context.Context, name string, options ...opts.ResolveOpt) (value path.Path, err error)
+	// There is a default depth-limit to avoid infinite recursion. Most users will be fine with
+	// this default limit, but if you need to adjust the limit you can specify it as an option.
+	Resolve(ctx context.Context, name string, options ...ResolveOption) (value path.Path, ttl time.Duration, err error)
+
+	// ResolveAsync performs recursive name lookup, like Resolve, but it returns entries as
+	// they are discovered in the DHT. Each returned result is guaranteed to be "better"
+	// (which usually means newer) than the previous one.
+	ResolveAsync(ctx context.Context, name string, options ...ResolveOption) <-chan ResolveResult
```
hacdias: I'm considering to change the `name string` here for the type `p path.Path`. The name here is supposed to be a path. This is unrelated to the routing key in the DHT. These are the human readable names. So I think it would make sense.

namesys/interface.go
```diff
@@ -60,39 +57,152 @@ type NameSystem interface {
...
+	}
+	return resolveOptions
 }
 
 // Publisher is an object capable of publishing particular names.
 type Publisher interface {
 	// Publish establishes a name-value mapping.
 	// TODO make this not PrivKey specific.
-	Publish(ctx context.Context, name ci.PrivKey, value path.Path, options ...opts.PublishOption) error
+	Publish(ctx context.Context, name ci.PrivKey, value path.Path, options ...PublishOption) error
```
hacdias: Similar to https://github.com/ipfs/boxo/pull/459/files#r1315812409, also potentially solve this `TODO`.

lidel:
re:
- (3) Hm.. the TTL is still used for caching purposes, we reuse it for HTTP caching. I think the logic we had was sensible: it was respecting TTL, but if EOL of IPNS Record happened to hit before TTL expires, we picked the lower value (EOL).  Does not matter much today (when we have low ttl), but might be safer to keep this in case people set short expiration and high TTL (being stuck with old value against their will).
- (4) mark these places with `// TODO: switch to AnyOf from https://github.com/ipfs/gateway-conformance/issues/153`, so we dont forget :pray: 
- (5) Yes to both?
  - (5.a) Should be ok to do so:  TTL will usually expire before there is a new release  of Kubo 
  - (5.b) This is the most important change we make, as it will allow IPNS publisher to use TTL value from the record to control how long websites are cached on  gateways and on user's browsers.  Currently its `Last-Modified`-based heuristic which is a blunt instrument. Returning `Cache-Control` when we have TTL will be value added.
  
  
ps. We probably want to KEEP `Last-Modified` for `/ipns` and base it on how long namesys resolution entry has been on the cache, and add `Cache-Control` based on TTL in addition to it (when we know TTL).  See https://github.com/ipfs/specs/pull/436#discussion_r1317996877

hacdias:
> ps. We probably want to KEEP Last-Modified for /ipns and base it on how long namesys resolution entry has been on the cache, and add Cache-Control based on TTL in addition to it (when we know TTL). 

Not sure how we're supposed to achieve that. Somehow we'd have to tap into the resolver cache to get that information out of it. Besides, the time it is in cache is, at most TTL or "time until EOL" if TTL is smaller.

namesys/interface.go
hacdias: I would love to change this, but then it would be important to change the KeyStore to support fetching keys by peer ID. This would involve a migration (which I can potentially add inside `NewFSKeystore` itself), and I wanted to avoid migrations 🤔 What do you think? @lidel

The way this interface is defined nowadays is too IPNS centric. What if we want to build an `DNSPublisher` via DNSLink? That can't possibly take a private key 😃 

This would also allow me to improve the Republisher and rename it to `IPNSRepublisher` for clarity, and instead take an `IPNSPublisher`. Then, we can leverage the fact that the `IPNSPublisher` can access the keys by itself, and also leverage the unused `ListPublished`.

namesys/interface.go
hacdias: The issue here is that right now we support things such as `Resolve(ctx, "dnslink.net")` and `Resolve(ctx, "peer-id")`, which means we do not necessarily need a namespace, or a path. Changing that would break some things. It would still be possible to keep Kubo's CLI with the same API with no changes.

What do you think? @lidel

lidel: Probably better keep `name string` as-is, as allowing paths will complicate implementation and tests.

Currently, the path would effectively always be `/ipns/id`.
But if you allow paths,  we need to decide what happens when someone passes   `/ipns/id/foo/bar`.
Do we keep path remainer and append it to returned path? Do consumers of this library expect that? If not, what is the point in passing `/ipns/` prefix every time?

Keeping `name string` keeps scope limited to the root identifier, and consumer of the library know they always get path in return for the `id` (and if they had any sub-path, they append themselves).

namesys/interface.go
lidel: Unless you feel strongly, I would remove TODO and not change anything.

If we want to change something, maybe replacing `name ci.PrivKey` with  `name string` and a separate `signingKey ci.PrivKey`:

```go
Publish(ctx context.Context, name string, signingKey ci.PrivKey, value path.Path, options ...PublishOption) error
```

This would make API less error prone by requiring user to pass both IPNS name and private key (implementation could check if privkey matches expected name).
Since `name` is a string, this new `Publish` could be used for signing other things than IPNS.

Anything more (migrations etc)  feels out of the scope of his PR. The additional work to refactor things to generalize this interface does not feel worth the effort any time soon. `Publisher` is IPNS centric because there is no other use case than DNSPublisher, which itself is not feasible:

DNSLink does not provide end-to-end integrity; there is no generally accepted DNS protocol for making DNS updates that "just works" across providers, nor there is user need for having it built-in.  The way DNSLink ended up being used in real world   is setting DNS TXT record by hand once, pointing it at `dnslink=/ipns/id` and then using IPNS for publishing updates. This is what ENS users do, since every DNS TXT record update on ENS costs gas,  DNSLink pointing at IPNS saves them money.

lidel:
Namesys has own cache.  The idea here is to leverage that cache and store (value, lastmodified, ttl) instead of (value, ttl).

When ttl expires namesys knows it should check if there is a new updated version of the record.
Once a record with higher sequence number appears,  and the value in the new record is the same, namesys keeps the same value in the cache, lastmodified timestamp also remains the same (indicating the first time the value was resolved), and only bumps TTL.  IF the value changed, we bump lastmodified to the current timestamp (in the future, we could use value from signed IPNS record)

This way, we could have a more meaningful  `Last-Modified` than `time.now()`.  
I think the main ask here is to future-proof the interface and return lastModified timestamp in addition to the ttl that is already added by this PR (this way we break API only once).

namesys/interface.go
hacdias: @lidel well, at the moment there's a few weird and inconsistent things happening:

- `DNSResolver` accepts `[/ipns/]{domain}[{path}]`. Returns appended result + path.
- `IPNSResolver` accepts `[/ipns/]{ipns-name}` (no path accepted). Returns result.
    - `ResolveIPNS` exists so that it detaches the path, sends it to the namesys, and gets back and appends the path. Seems like a completely useless function to be honest and I may cut it if I prove that's the case [tested with Kubo and Boxo, indeed this function is useless except for the fact that it returns an error if namesys is nil].
- `namesys` accepts valid full path, just a cid, and some more magic. Returns appended result + path.

So we already have extremely inconsistent behaviours across them. I want to clean this up. We can potentially tighten `DNSResolver` and `IPNSResolver` to "root identifier" and let `namesys`, which combines both, parse the path and do re-attachments as necessary.

This is why I wanted to accept a `path.Path` - no more complications. Then it would be trivial to append path segments and parse the path.

My two options:

1. Change everything to `path.Path`; or
2. Keep `name string`, but allow full paths everywhere, and we just convert to a `path.Path` and do some magic to not break the current behaviours (like magically accepting a cid).

namesys/interface.go
hacdias: Then I think I'll just remove the TODO, but not change it.

hacdias:
> When ttl expires namesys knows it should check if there is a new updated version of the record.

Btw, I want to note that that is **not** the case at the time. Namesys cache keeps the record value until EOL: 

https://github.com/ipfs/boxo/blob/574df963653c940ea19b1b8a0d37cd30360c22e9/namesys/cache.go#L33-L35

I can definitely change it, because it would make more sense, but just a note.

namesys/interface.go
hacdias: My suggestion: make `.Resolve` strictly only accept `[/ipns/]{name}` and have some utility function like what we already have nowadays (`ResolveIPNS`) that does the path handling. But here's where the problem arises: recursion. What if an IPNS name's value is `/ipns/{name}/a/b`? How will we recursively solve that now that we no longer accept paths?

lidel: Wow, indeed this is worse than I've remembered :see_no_evil:  Thank you for the in-depth context.

It feels like dechnical debt. There was a time when we preferred to design libraries that represent paths as `string`  to reduce dependency hell, but  that has not been the case for a long time (at some point we started  returning them as explicit  Path type and not a `string`).

We already return `path.Path`, so biting the bullet and cleaning it up and using the same type for input (1) sounds logical.
((2) would make sense if we removed `Path` from all interfaces, but there is little value since it lives in boxo now)

lidel:
@hacdias iiuc `cacheEntry.eol` is based on current time + TTL, has nothing to do with IPNS Record's EOL:

https://github.com/ipfs/boxo/blob/574df963653c940ea19b1b8a0d37cd30360c22e9/namesys/cache.go#L48

Mind renaming it to something like `cacheEntry.cacheExpiration` or `cacheTTL`, so it is super clear and nobody confuses it with EOL of IPNS Record?

hacdias:
@lidel I still need to write the conformance tests. Kubo PR is at (https://github.com/ipfs/boxo/pull/459). Conformance failing due to unexpected headers in IPNS requests. Hence, needing writing new tests.

It would be great if I could have some feedback here in the meanwhile. But please prioritise #334 first.

hacdias:
This PR is ready for review:

- Kubo passing PR is at https://github.com/ipfs/kubo/pull/10115 (conformance see below)
- Conformance is at https://github.com/ipfs/gateway-conformance/pull/173.

namesys/ipns_publisher.go
```diff
@@ -189,47 +188,45 @@ func (p *IpnsPublisher) updateRecord(ctx context.Context, k crypto.PrivKey, valu
...
-// PutRecordToRouting publishes the given entry using the provided ValueStore,
-// keyed on the ID associated with the provided public key. The public key is
-// also made available to the routing system so that entries can be verified.
-func PutRecordToRouting(ctx context.Context, r routing.ValueStore, k crypto.PubKey, rec *ipns.Record) error {
-	ctx, span := StartSpan(ctx, "PutRecordToRouting")
+// PublishIPNSRecord publishes the given [ipns.Record] for the provided [crypto.PubKey] in
+// the provided [routing.ValueStore]. The public key is also made available to the routing
+// system if it cannot be derived from the corresponding [peer.ID].
+func PublishIPNSRecord(ctx context.Context, r routing.ValueStore, pubKey crypto.PubKey, rec *ipns.Record) error {
+	ctx, span := startSpan(ctx, "PutRecordToRouting")
```
lidel: :thought_balloon:  @hacdias  you renamed funcs but the span name is the same.
Do we need to rename funcs? Maybe keep the old func name for now? (Or at least keep them in sync).
(here and in other cases like this)

hacdias: I will update them to be consistent. My idea was to make them consistent `PublishIPNSRecord` puts the record (and key when needed) to the routing system. 

Btw, can you also look at https://github.com/ipfs/boxo/pull/459#discussion_r1315162250 and https://github.com/ipfs/boxo/pull/459#discussion_r1315162377?

hacdias: I have pushed a commit with the fix.

namesys/ipns_publisher.go
lidel: fwiw it is mentioned in:  https://github.com/ipfs/boxo/blob/25ae0bd151b43ff3ec56aac44fc8a03337a72f7b/namesys/republisher/repub.go#L98   

@hacdias  I would avoid making waves and keep it, but  use `ipns.Name` like you suggested (make sure we dont use PeerIDs in public APIs that are related to IPNS)

hacdias: @lidel the problem here is that `peer.ID` is a string, therefore it works well for a map. `ipns.Name` won't, as the key cannot be a value, only the pointer. And multiple pointers can have the same name.

namesys/ipns_publisher.go
lidel: @hacdias Not feeling strongly, but feels useful: allows user of the `boxo/namesys` library to do  local lookup without hitting remote routing.
I'd say we keep it but replace `peer.ID` with `ipns.Name` (since we make breaking changes anyway, remove all `peer.ID` from public apis in `boxo/namesys`?

namesys/ipns_publisher.go
lidel: @hacdias ack, i guess the best thing we have for this is CID from go-cid? (with libp2p-key codec)
Still feels better than leaking libp2p PeerID into namesys interface.

hacdias: @lidel I changed `ipns.Name.src` from `[]byte` to `string` and `ipns.Name` is now comparable. I think that is perfectly fine as that variable is only used internally.

namesys/interface.go
```diff
@@ -1,52 +1,50 @@
...
+	DefaultDepthLimit = 32
+
+	// UnlimitedDepth allows infinite recursion in [Resolver]. You probably don't want
+	// to use this, but it's here if you absolutely trust resolution to eventually
+	// complete and can't put an upper limit on how many steps it will take.
+	UnlimitedDepth = 0
+
+	// DefaultIPNSRecordTTL specifies the time that the record can be cached before
+	// checking if its validity again.
+	DefaultIPNSRecordTTL = time.Minute
```
lidel: :information_source:  since this PR is actually using TTLs, we should adjust default to make them more useful by default. 

I've moved this to `boxo/ipns` and  made this  1h, as recommended in https://specs.ipfs.tech/ipns/ipns-record/#ipns-record

namesys/interface.go
```diff
@@ -60,39 +58,168 @@ type NameSystem interface {
...
+	// and verified. A zero value indicates that there is no explicit timeout
+	// (although there is an implicit timeout due to dial timeouts within the DHT).
+	DhtTimeout time.Duration
+}
+
+// DefaultResolveOptions returns the default options for resolving an IPNS Path.
+func DefaultResolveOptions() ResolveOptions {
+	return ResolveOptions{
+		Depth:          DefaultDepthLimit,
+		DhtRecordCount: 16,
```
lidel: :information_source:  made it a constant in c7ca873c82af550eb1561ed89bbd464ad21663f0

namesys/republisher/repub.go
```diff
@@ -21,24 +23,27 @@ import (
...
 
-// FailureRetryInterval is the interval at which we retry IPNS records broadcasts (when they fail)
-var FailureRetryInterval = time.Minute * 5
+	// FailureRetryInterval is the interval at which we retry IPNS records broadcasts (when they fail)
+	FailureRetryInterval = time.Minute * 5
 
-// DefaultRecordLifetime is the default lifetime for IPNS records
-const DefaultRecordLifetime = time.Hour * 24
+	// DefaultRecordLifetime is the default lifetime for IPNS records
+	DefaultRecordLifetime = time.Hour * 24
```
lidel: :information_source: 0fdcc1f4ff09deec29ea2e5f4b548b734b315e50   moved  to `ipns/defaults.go`, changed default to 48h to match `ValidityType = 0` notes at  https://specs.ipfs.tech/ipns/ipns-record/#ipns-record

namesys/interface.go
```diff
@@ -1,52 +1,50 @@
...
+	UnlimitedDepth = 0
+
+	// DefaultIPNSRecordTTL specifies the time that the record can be cached before
+	// checking if its validity again.
+	DefaultIPNSRecordTTL = time.Minute
+
+	// DefaultIPNSRecordEOL specifies the time that the network will cache IPNS
+	// records after being published. Records should be re-published before this
+	// interval expires. We use the same default expiration as the DHT.
+	DefaultIPNSRecordEOL = 48 * time.Hour
```
lidel: :information_source:  This felt like a duplicate of `DefaultRecordLifetime`
Updated  `ipns.DefaultRecordLifetime` to 48h in 0fdcc1f4ff09deec29ea2e5f4b548b734b315e50  and removed this duplicate

namesys/interface.go
```diff
@@ -60,39 +58,168 @@ type NameSystem interface {
...
+	LastMod time.Time
+}
+
+// ResolveAsyncResult is the return type for [Resolver.ResolveAsync].
+type ResolveAsyncResult struct {
+	Path    path.Path
+	TTL     time.Duration
+	LastMod time.Time
+	Err     error
 }
```
lidel: :thought_balloon:  :warning: 

@hacdias this feels like overkill, do we need both? can we have only one `ResolveResult` (or even shorter `Result`) with `Err`, like we did before?

namesys/interface.go
lidel: :information_source:  increased to 1h, see 0fdcc1f4ff09deec29ea2e5f4b548b734b315e50 and https://specs.ipfs.tech/ipns/ipns-record/#ipns-record

ipns/name.go
```diff
@@ -25,7 +25,7 @@ const (
 // [Multihash]: https://multiformats.io/multihash/
 // [IPNS Name]: https://specs.ipfs.tech/ipns/ipns-record/#ipns-name
 type Name struct {
-	src []byte
+	multihash string // binary Multihash without multibase envelope
```
lidel, line 28: :information_source:  I've pushed small change  (8c1d32a0fb01e2832d76a0fe9bf292987cc1bf73) that renamed `src` to meaningful `multihash`  and added  the above comment, so at least a drive-by reader does not make wrong assumptions.

fwiw I looked across ecosystem, our core types are mostly `string`:

- go-multihash `type Multihash []byte`
- go-cid `type Cid struct{ str string }`
- go-libp2p/peer  `type ID string`

I guess `string` here is ok, especially now that we've added comment, but 

@hacdias  did we switch from `[]byte` to `string` here   to allow them in map keys, or was there a deeper reason?

lidel (approved):
@hacdias thank you for cleaning the `coreifrace` and path leftovers and being patient with reviews on this. 

I blocked some time today for review and pushed cosmetic refactors that unify implicit behaviors and docs for changes we should make to the IPNS publishing defaults.

Please check my changes and CHANGELOG I've added, namely the changes to defaults:
We had to increase TTL, now that it impacts caching in multiple places (namesys, gateway responses) 1 minute was too low. While at it, I also increased Lifetime and used it for publishing and reproviding to benefit from longer persistence window in Amino DHT.

Only remaining questions:
- why CI did not run? (it is in pendign for me)
- inline if we need separate `ResolveResult` and `ResolveAsyncResult` or can have one, but not a blocker.

If you are ok with my changes, and CI runs and is  green,  feel free to merge and bubble up – we want to include this in Kubo 0.24.

hacdias:
@lidel I think CI did not run because it has conflicts with the `main` branch. I will take care of the feedback in Kubo, and bubble this up.

ipns/name.go
hacdias: Yes, only to allow `Name` to be a map key.

namesys/interface.go
hacdias: @lidel the issue is that I don't think it makes sense for `Resolve` (non-async) to return a struct that contains an `Err` field. We can always add a comment saying that `Err` is always `nil` for `.Resolve`, but that's a bit beating around the bush. I would however suggest:

```go
// Result is the return type for [Resolver.Resolve].
type Result struct {
	Path    path.Path
	TTL     time.Duration
	LastMod time.Time
}

// AsyncResult is the return type for [Resolver.ResolveAsync].
type AsyncResult struct {
	Result Result
	Err       error
}
```

hacdias: @lidel I have simplified the names, but I did not incorporate `Result` inside `AsyncResult`. I am going to merge the PR as-is. We can change this in a separate PR if we find a better of doing it. As long as it's before final release, it should be fine.

PR #461, opened 2023-09-05
coreiface: deprecate DhtAPI in favor of RoutingAPI

Jorropo:


coreiface/routing.go
```diff
@@ -4,13 +4,25 @@ import (
...
 
-	// Put sets a value for a given key
+	// Put sets a value for a given key.
 	Put(ctx context.Context, key string, value []byte, opts ...options.RoutingPutOption) error
+
+	// FindPeer queries the DHT for all of the multiaddresses associated with a Peer ID.
+	FindPeer(context.Context, peer.ID) (peer.AddrInfo, error)
+
+	// FindProviders finds peers in the DHT who can provide a specific value given a key.
+	FindProviders(context.Context, path.Path, ...options.DhtFindProvidersOption) (<-chan peer.AddrInfo, error)
```
aschmahmann, line 24: options.Routing?

coreiface/routing.go
```diff
@@ -4,13 +4,25 @@ import (
...
+
+	// Provide announces to the network that you are providing given values.
+	Provide(context.Context, path.Path, ...options.DhtProvideOption) error
```
aschmahmann, line 27: options.Routing?

coreiface/tests/routing.go
```diff
@@ -98,3 +102,145 @@ func (tp *TestSuite) TestRoutingPutOffline(t *testing.T) {
...
+	}
+
+	p, err := addTestObject(ctx, apis[0])
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(3 * time.Second)
+
+	out, err := apis[2].Routing().FindProviders(ctx, p, options.Dht.NumProviders(1))
```
aschmahmann, line 176: options.Routing?

aschmahmann (changes requested):
There's a bunch of go-check errors around using the wrong types. After fixing that things look pretty good.

aschmahmann:
@Jorropo why skip changelog here? We're deprecating an API. No we're not removing it or anything but seems reasonable to mention.

Jorropo:
Good question, the API was already deprecated in Kubo and I had tunnel vision.
I'll add changelog here thx.

lidel:
Triage notes:
- needs Kubo PR to show we dont remove/deprecate too much
  - converting to draft, as we need to look at this from Kub oside, afaik `ipfs routing`   in Kubo does the same thing as `ipfs dht` but 
    - does not allow for selecting backend, so not possible to run dht-only query without chaing config to be dht-only
    - `ipfs dht query` is not present in `ipfs routing`

PR #463, opened 2023-09-19
feat(gateway): add DisableHTMLErrors option

hacdias:
This adds a new option `DisableHTMLErrors ` to `gateway.Config`. If this is `true`, then the errors will always be returned as `text/plain` with the raw error message. This will need to be bubbled up to Kubo, which will then allow `ipfs.io` to have a better insight into the error messages.

This also allows users of `boxo/gateway` library to do white-label hosting, without IPFS branding on error pages.

cc @cewood @mcamou

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 * ✨ The `routing/http` implements Delegated Peer Routing introduced in [IPIP-417](https://github.com/ipfs/specs/pull/417).
+* An option `DisableHTMLErrors` has been added to `gateway.Config`. When this option
+  is `true`, pretty HTML error pages for web browsers are disabled. Instead, a
+  `text/plain` page with the raw error message as the body is returned.
 
 ### Changed
 
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -40,6 +40,11 @@ type Config struct {
 	// overridden per FQDN in PublicGateways. To be used with WithHostname.
 	NoDNSLink bool
 
+	// DisableHTMLErrors disables pretty HTML pages when an error occurs. Instead, a `text/plain`
+	// page will be sent with the raw error message. This can be useful if this gateway
+	// is being proxied by other service, which wants to use the error message.
+	DisableHTMLErrors bool
+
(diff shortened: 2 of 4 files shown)


mcamou (approved):

gateway/gateway.go
```diff
@@ -40,6 +40,11 @@ type Config struct {
 	// overridden per FQDN in PublicGateways. To be used with WithHostname.
 	NoDNSLink bool
 
+	// RawError disables pretty HTML pages when an error occurs. Instead, a `text/plain`
+	// page will be sent with the raw error message. This can be useful if this gateway
+	// is being proxied by other service, which wants to use the error message.
+	RawError bool
```
lidel, line 46:
```suggestion
	// ForcePlainTextErrors disables pretty HTML pages when an error occurs. Instead, a `text/plain`
	// page will be sent with the raw error message. This can be useful if this gateway
	// is being proxied by other service, which wants to use the error message.
	ForcePlainTextErrors bool
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 * ✨ The `routing/http` implements Delegated Peer Routing introduced in [IPIP-417](https://github.com/ipfs/specs/pull/417).
+* An option `RawError` has been added to `gateway.Config`. When this option is `true`,
+  pretty HTML error pages are disabled. Instead, a `text/plain` page with the raw error
```
```suggestion
  pretty HTML error pages for web browsers are disabled. Instead, a `text/plain` page with the raw error
```

lidel (changes requested):
@hacdias  I think this is a sensible flag to have here and in Kubo(no matter if this is used by bifrost infra or not).
HTML Error pages include IPFS brand and links to project docs, and users may want to disable that to have whitelabel hosting setup.

Just rename it, so it is more clear what it does, either:
- `ForcePlainTextErrors`
- `DisableHTMLErrors`  ← imo preferable, does what it says, does not corner us into returning plain text error forever (we can start returning JSON instead, and this won't break or look awkward)

lidel (approved):
This feature is now nice and clear, thanks! :heart: 

Mind bubbling up to Kubo?

PR #472, opened 2023-09-26
feat(routing/dht/zikade): add dht routing implementation using zikade

iand:
Add support for the [Zikade](https://github.com/plprobelab/zikade) Amino DHT implementation as a routing choice.


Jorropo (commented):
`routing/dht` does not exists btw, no need to add `v2`

iand:
> `routing/dht` does not exists btw, no need to add `v2`

Understood. I chose this on the assumption that dht v1 would still be needed for a while and might end up in Boxo. Happy to remove the v2 if needed

BigLep:
@iand : just in case this was an accident, the team teans to not look at draft PRs.  Please callout if you're wanting feedback on the current state (or move out of draft).

iand:
@BigLep it's not an accident, I wanted early visibility while I am working on it. Will mark it ready for review when complete.

iand:
@Jorropo would you be able to take another look?

gammazero:
Too much has changed to accept this. Closing.

PR #474, opened 2023-09-29
chore: migrate bootstrap

gammazero:
Migrating bootstrap service from kubo. The bootstrap package is used outside of kubo, and should be relocated into boxo so that kubo is not used as an IPFS library.


bootstrap/bootstrap.go
```diff
@@ -0,0 +1,371 @@
...
+
+	saveBackupBootstrapPeers func(context.Context, []peer.AddrInfo)
+	loadBackupBootstrapPeers func(context.Context) []peer.AddrInfo
+}
+
+// DefaultBootstrapConfig specifies default sane parameters for bootstrapping.
+var DefaultBootstrapConfig = BootstrapConfig{
+	MinPeerThreshold:        4,
+	Period:                  30 * time.Second,
+	ConnectionTimeout:       (30 * time.Second) / 3, // Perod / 3
```
Jorropo:
```suggestion
	ConnectionTimeout:       (30 * time.Second) / 3, // Period / 3
```

Jorropo (commented):
I'm gonna refactor this to work without goprocess before reviewing.

hsanjuan:
> I'm gonna refactor this to work without goprocess before reviewing.

No please... merge and then do whatever... I would like to rely on this as part of upstream, otherwise I will copy-paste the whole thing directly.

Jorropo:
@hsanjuan I'm not sure about the API exposed to consumers either, if I touch this code I can see this.
We want to have a high bar in what goes in boxo, we want to avoid merging code we don't want to maintain.

hsanjuan:
> @hsanjuan I'm not sure about the API exposed to consumers either, if I touch this code I can see this. We want to have a high bar in what goes in boxo, we want to avoid merging code we don't want to maintain.

If I run `staticcheck --checks all ./...` in boxo there are 328 warnings, with tons of undocumented stuff and things not following convention. So I don't think this code is any different from everything else that has been imported into Boxo (and didn't go through refactoring). Anything is better without goprocess in it, but this is nice-to-have as is right now, while it gets refactored.

Anyways, no big deal, I just have different approach but it's just matter of opinion.

BigLep:
2023-10-19 conversation: Jorropo will refactor this to remove goprocess and then merge next week (week of 2023-10-23).

Jorropo:
@gammazero just flagging that I plan to push on this branch, if you ever force push be sure to use `--force-with-lease` and not `--force`, this will error in case I pushed anything in the meanwhile, thx in advance.

hacdias (approved):
This is fine by me. As per the standup, this is going to be merged as-is for tomorrow RC.

PR #478, opened 2023-10-04
feat: export routing/http/client.Client

hacdias:
Exports Routing V1 client so people can pass it around in their implementations. At the moment it can't be added to structs, or, for example, passed to a function, since it is unexported.

### DIFF
--- a/routing/http/client/client.go
+++ b/routing/http/client/client.go
@@ -28,7 +28,7 @@ import (
 )
 
 var (
-	_                 contentrouter.Client = &client{}
+	_                 contentrouter.Client = &Client{}
 	logger                                 = logging.Logger("routing/http/client")
 	defaultHTTPClient                      = &http.Client{
 		Transport: &ResponseBodyLimitedTransport{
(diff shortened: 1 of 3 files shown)


routing/http/client/client.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 var (
-	_                 contentrouter.Client = &client{}
+	_                 contentrouter.Client = &Client{}
```
lidel, line 31: @hacdias Given https://specs.ipfs.tech/routing/http-routing-v1/ maybe call this `DelegatedRoutingClient` ?

I imagine someone integrating IPFS or experimenting in a single `.go` file with some PoC  may have multiple "clients" (DNS, Kubo RPC, Pinning service, and now this), better to have meaningful names to reduce cognitive load?

hacdias: @lidel hmm, that feels a bit long. I think the main issue here is also the package name. It';s just called `client`. If it were called something else, it would be much more explicit, like `delegatedrouting.Client`.

hacdias:
Let's just include this in #479.

PR #479, opened 2023-10-04
Routing V1 CLI Example

hacdias:
- Exposes `routing/http/client.Client` so that users can actually use it across their code.
- Adds Routing V1 CLI example. Added some basic tests.

I will use this example for my talk in IPFS Connect and I thought it was useful to keep here.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,3 +9,4 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 - [Fetching a UnixFS file by CID](./unixfs-file-cid)
 - [Gateway backed by a CAR file](./gateway/car)
 - [Gateway backed by a remote blockstore and IPNS resolver](./gateway/proxy)
+- [Delegated Routing V1 Command Line Client](./routing/delegated-routing-client/)
--- a/examples/routing/delegated-routing-client/main.go
+++ b/examples/routing/delegated-routing-client/main.go
@@ -1,0 +1,148 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"time"
+
+	"github.com/ipfs/boxo/ipns"
+	"github.com/ipfs/boxo/routing/http/client"
+	"github.com/ipfs/boxo/routing/http/types"
+	"github.com/ipfs/boxo/routing/http/types/iter"
+	"github.com/ipfs/go-cid"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+func main() {
+	gatewayUrlPtr := flag.String("e", "", "routing v1 endpoint to use")
+	timeoutPtr := flag.Int("t", 10, "timeout in seconds for lookup")
+	cidPtr := flag.String("cid", "", "cid to find")
+	pidPtr := flag.String("peer", "", "peer to find")
+	namePtr := flag.String("ipns", "", "ipns name to retrieve record for")
+	flag.Parse()
+
+	if err := run(os.Stdout, *gatewayUrlPtr, *cidPtr, *pidPtr, *namePtr, *timeoutPtr); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run(w io.Writer, gatewayURL, cidStr, pidStr, nameStr string, timeoutSeconds int) error {
+	// Creates a new Delegated Routing V1 client.
+	client, err := client.New(gatewayURL)
+	if err != nil {
+		return err
+	}
+
+	timeout := time.Duration(timeoutSeconds) * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	if cidStr != "" {
+		return findProviders(w, ctx, client, cidStr)
+	} else if pidStr != "" {
+		return findPeers(w, ctx, client, pidStr)
+	} else if nameStr != "" {
+		return findIPNS(w, ctx, client, nameStr)
+	} else {
+		return errors.New("cid or peer must be provided")
+	}
+}
+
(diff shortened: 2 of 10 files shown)


examples/routing-v1/client/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Build
+
+```bash
+> go build -o routing-client
+```
+
+## Usage
+
+First, you will need a Routing V1 server. For that, you can potentially use [Kubo],
```
lidel:
```suggestion
First, you will need a HTTP endpoint compatible  with  [Delegated Routing V1 Specification](https://specs.ipfs.tech/routing/http-routing-v1/). For that, you can potentially use [Kubo],
```

examples/routing-v1/client/README.md
```diff
@@ -0,0 +1,68 @@
...
+which supports [exposing](https://github.com/ipfs/kubo/blob/master/docs/config.md#gatewayexposeroutingapi)
+a Routing V1 endpoint. For the commands below, we assume the endpoint is `http://127.0.0.1:8080`.
```
lidel:
```suggestion
which supports [exposing](https://github.com/ipfs/kubo/blob/master/docs/config.md#gatewayexposeroutingapi)
a  `/routing/v1` endpoint. For the commands below, we assume the HTTP server that provides the endpoint  is `http://127.0.0.1:8080`.
```

examples/routing-v1/client/main.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+func main() {
+	gatewayUrlPtr := flag.String("e", "", "routing v1 endpoint to use")
+	timeoutPtr := flag.Int("t", 10, "timeout in seconds for lookup")
+	cidPtr := flag.String("c", "", "cid to find")
+	pidPtr := flag.String("p", "", "peer to find")
+	namePtr := flag.String("n", "", "ipns name to retrieve record for")
```
lidel: :thought_balloon:  since this is an example, using meaningful parameters makes one-liners on slides and blogs easier to parse and understand:
```suggestion
	cidPtr := flag.String("cid", "", "cid to find")
	pidPtr := flag.String("peer", "", "peer to find")
	namePtr := flag.String("ipns", "", "ipns name to retrieve record for")
```

examples/routing-v1/client/README.md
```diff
@@ -0,0 +1,68 @@
...
+This is an example of how to use the Delegated Routing V1 HTTP client from Boxo.
+In this package, we build a small command line tool that allows you to connect to
+a Routing V1 endpoint and fetch content providers, peer information, as well as
+IPNS records for a certain IPNS name.
+
+## Build
+
+```bash
+> go build -o routing-client
+```
```
lidel: :thought_balloon:  I wonder if we could make this example more useful by making it compatible with `go install` CLI? (removes the need to git clone and do manual build)

````suggestion
## Install

```bash
> go install github.com/ipfs/boxo/examples/routing/delegated-routing-client@latest
```

## Build

```bash
> go build -o routing-client
```
````

Comment (hacdias) [reply examples/routing-v1/client/README.md]: @lidel that will not work:

```
The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

We'd have to remove the replace from the `go.mod` and ensure we'd keep it up to date with latest Boxo. Could be done with CI/release-process. But out of scope for this PR.

File: examples/README.md [hunk 5]
```diff
@@ -9,3 +9,4 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 - [Fetching a UnixFS file by CID](./unixfs-file-cid)
 - [Gateway backed by a CAR file](./gateway/car)
 - [Gateway backed by a remote blockstore and IPNS resolver](./gateway/proxy)
+- [Delegated Routing V1 Command Line Client](./routing-v1/client/)
```
lidel:
```suggestion
- [Delegated Routing V1 Command Line Client](./routing/delegated-routing-client/)
```

examples/routing/delegated-routing-client/main.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func findIPNS(w io.Writer, ctx context.Context, client *client.Client, nameStr string) error {
+	// Parses the given name string to get a record for.
+	name, err := ipns.NameFromString(nameStr)
+	if err != nil {
+		return err
+	}
+
+	// Ask for a record. [client.GetIPNS] validates the record for us.
```
lidel: :thought_balloon: This feels like we  teach people to trust remote validation of IPNS, which is not what we want for IPNS.
Someone will copy this code and run it in production :upside_down_face: 

Could we do bare minimum and:
- add local signature check and print `Signature: VALID|INVALID` below `Value`
- show `Value` only when signature is valid, otherwise use placeholder `[INVALID]`

lidel (approved):
lgtm, small nits below

examples/routing/delegated-routing-client/main.go
hacdias: @lidel hmm, it's not remote validation though. We can add a validation __again__, but it will never be invalid because the client code already validates it.

hacdias: @lidel hmm, it is not remote validation though. It is very local validation inside the `GetIPNS` function:

https://github.com/ipfs/boxo/blob/5ae64f335dd4ccedcb5b63108ccc8fe8e4408843/routing/http/client/client.go#L408-L444

I updated the comments to make it 100% clear that the record is validated by Boxo client. Also added that when using other clients or retrieving through other methods, records should always be validated.

I am going ahead and merge, but if you feel like we should really add a _second validation_ that will always succeed for some clarity, I can do it in a following PR.

PR #481, opened 2023-10-06
path: replace ImmutablePath interface with struct

Jorropo:
Let's not repeat https://github.com/ipfs/go-block-format/issues/45 interface for struct with one implementation and no value added.


Jorropo:
Skip changelog since this is not a breaking change yet (we havn't released path yet).

hacdias (approved):
I think this change looks sensible. I'm almost sure this was part of the initial PR for the path but I changed it somewhere. Just want to note that `Path` should stay an interface, and namespaces should be a string. This allows for more experimentation from other people.

But `ImmutablePath` does indeed make sense to be a struct.

PR #482, opened 2023-10-09
docs: how to use and create examples

hacdias:
As asked during the retro.


hacdias:
@codecov yes, of course updating the README will **increase** the coverage.

hacdias:
I'm merging this. We can iterate on the copy later. Better have some information than none.

PR #485, opened 2023-10-12
Simplify fetcher setup

hsanjuan:
The default fetcher already uses the same PrototypeChooser that we were setting explicitally (https://github.com/ipfs/boxo/blob/main/fetcher/impl/blockservice/fetcher.go#L127).

### DIFF
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -137,7 +137,9 @@ func NewBlocksBackend(blockService blockservice.BlockService, opts ...BlocksBack
 	r = compiledOptions.r
 	if r == nil {
 		// Setup the UnixFS resolver.
-		fetcher := bsfetcher.NewFetcherConfig(blockService).WithReifier(unixfsnode.Reify)
+		fetcherCfg := bsfetcher.NewFetcherConfig(blockService)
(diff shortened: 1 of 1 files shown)


gateway/blocks_backend.go
```diff
@@ -145,7 +136,8 @@ func NewBlocksBackend(blockService blockservice.BlockService, opts ...BlocksBack
 
 	r = compiledOptions.r
 	if r == nil {
-		fetcher := fetcherConfig.WithReifier(unixfsnode.Reify)
+		// Setup the UnixFS resolver.
+		fetcher := bsfetcher.NewFetcherConfig(blockService)
```
hacdias: Tests are failing (many 404s instead of 200s). I think you're missing the reifier here:

```suggestion
		fetcher := bsfetcher.NewFetcherConfig(blockService).WithReifier(unixfsnode.Reify)
```

hsanjuan: I mixed PrototypeChooser and Reifier. Yes.

hsanjuan:
Ok, I got confused. We need to add dagpb support on top of the default PrototypeChooser using dagpb.AddSupportToChooser. But yeah given the default is already defined we can just use it rather than rewriting it.

hacdias (approved):

PR #486, opened 2023-10-12
fix(gw): duplicate blocks and range etags

Jorropo:


gateway/gateway.go
```diff
@@ -182,14 +182,16 @@ const (
...
-		return DuplicateBlocksIncluded
+		return DuplicateBlocksIncluded, nil
 	case "n":
-		return DuplicateBlocksExcluded
+		return DuplicateBlocksExcluded, nil
+	case "":
+		return DuplicateBlocksUnspecified, nil
 	}
-	return DuplicateBlocksUnspecified
+	return 0, fmt.Errorf("unsupported application/vnd.ipld.car content type dups parameter: %q", dupsValue)
```
hacdias, line 195: Maybe add a test just to make sure everything's working.

gateway/handler_car.go
```diff
@@ -226,31 +225,43 @@ func getCarRootCidAndLastSegment(imPath path.ImmutablePath) (cid.Cid, string, er
 }
 
 func getCarEtag(imPath path.ImmutablePath, params CarParams, rootCid cid.Cid) string {
-	data := imPath.String()
+	h := xxhash.New()
+	h.WriteString(imPath.String())
+	// be careful with hashes here, we need boundries and per entry salt, we don't want a request that has:
```
hacdias:
```suggestion
	// be careful with hashes here, we need boundaries and per entry salt, we don't want a request that has:
```

hacdias (changes requested):
Looks good to me, but maybe a test would be great.

hacdias (approved):

PR #488, opened 2023-10-13
bitswap/client: explain what the options do

hsanjuan:
Added godoc explanation of what the client options do in more detail, since given information was useless.

### DIFF
--- a/bitswap/client/client.go
+++ b/bitswap/client/client.go
@@ -44,14 +44,20 @@ var log = logging.Logger("bitswap-client")
 // bitswap instances
 type Option func(*Client)
 
-// ProviderSearchDelay overwrites the global provider search delay
+// ProviderSearchDelay sets idle-delay value.  When the session
+// idles (does not receive any blocks) for this duration, a broadcast is
+// triggered. The broadcast finds new peers and sends the wantlist. See
+// [defaults.ProvSearchDelay] for the default (1s as of this writing).
 func ProviderSearchDelay(newProvSearchDelay time.Duration) Option {
 	return func(bs *Client) {
 		bs.provSearchDelay = newProvSearchDelay
 	}
 }
 
-// RebroadcastDelay overwrites the global provider rebroadcast delay
+// RebroadcastDelay sets a custom delay for periodic search of a random want.
+// When the value ellapses, a random CID from the wantlist is chosen and the
+// client attempts to find more peers for it and sends them the single want.
+// Default is set to 1 minute in [New].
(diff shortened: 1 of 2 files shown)


bitswap/client/client.go
```diff
@@ -44,14 +44,20 @@ var log = logging.Logger("bitswap-client")
 // bitswap instances
 type Option func(*Client)
 
-// ProviderSearchDelay overwrites the global provider search delay
+// ProviderSearchDelay sets idle-delay value.  When the session
+// idles (does not receive any blocks) for this duration, a broadcast is
+// triggered. The broadcast finds new peers and sends the wantlist. See
+// [defaults.ProvSearchDelay] for the default (1s as of this writing).
```
Jorropo:
```suggestion
// [defaults.ProvSearchDelay] for the default.
```

bitswap/client/client.go
```diff
@@ -44,14 +44,20 @@ var log = logging.Logger("bitswap-client")
...
 	return func(bs *Client) {
 		bs.provSearchDelay = newProvSearchDelay
 	}
 }
 
-// RebroadcastDelay overwrites the global provider rebroadcast delay
+// RebroadcastDelay sets a custom delay for periodic search of a random want.
+// When the value ellapses, a random CID from the wantlist is chosen and the
+// client attempts to find more peers for it and sends them the single want.
+// Default is set to 1 minute in [New].
```
Jorropo: Please expose a default constant.

Jorropo (changes requested):
Let's not write text that is gonna be outdated once we change this, I people to do 1 click if they want to follow the constant.

Jorropo (approved):

bitswap/internal/defaults/defaults.go
```diff
@@ -34,3 +36,9 @@ const (
 	MaximumHashLength = 128
 	MaximumAllowedCid = binary.MaxVarintLen64*4 + MaximumHashLength
 )
+
+var (
+	// RebroadcastDelay is the default delay to trigger broadcast of
+	// random CIDs in the wantlist.
+	RebroadcastDelay = delay.Fixed(time.Minute)
```
Jorropo: Does this needs to be a var ? I would just use a `time.Duration` if `go-ipfs-delay` can't be constified.

Jorropo (dismissed):

bitswap/internal/defaults/defaults.go
hsanjuan: fair enough, I had also forgotten to actually use this value. Fixed in latest commit.

hacdias (approved):

PR #490, opened 2023-10-17
routing/http: return PeerRecord for FindPeers

hacdias:
`/routing/v1/peers` always returns peer schema conforming objects: https://specs.ipfs.tech/routing/http-routing-v1/#response-body-0.

This makes it easier for the consumers of the library.

### DIFF
--- a/examples/routing/delegated-routing-client/main.go
+++ b/examples/routing/delegated-routing-client/main.go
@@ -69,22 +69,6 @@ func findProviders(w io.Writer, ctx context.Context, client *client.Client, cidS
 	return printIter(w, recordsIter)
 }
 
-func findPeers(w io.Writer, ctx context.Context, client *client.Client, pidStr string) error {
-	// Parses the given Peer ID to lookup the information for.
-	pid, err := peer.Decode(pidStr)
-	if err != nil {
-		return err
-	}
-
-	// Ask for information about the peer with the given peer ID.
-	recordsIter, err := client.FindPeers(ctx, pid)
-	if err != nil {
-		return err
-	}
-	defer recordsIter.Close()
-	return printIter(w, recordsIter)
-}
-
 func printIter(w io.Writer, iter iter.ResultIter[types.Record]) error {
 	// The response is streamed. Alternatively, you could use [iter.ReadAll]
 	// to fetch all the results all at once, instead of iterating as they are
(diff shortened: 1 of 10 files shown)


examples/routing/delegated-routing-client/main.go
```diff
@@ -118,6 +102,44 @@ func printIter(w io.Writer, iter iter.ResultIter[types.Record]) error {
...
+	recordsIter, err := client.FindPeers(ctx, pid)
+	if err != nil {
+		return err
+	}
+	defer recordsIter.Close()
+
+	// The response is streamed. Alternatively, you could use [iter.ReadAll]
+	// to fetch all the results all at once, instead of iterating as they are
+	// streamed.
+	for recordsIter.Next() {
```
aschmahmann, line 122: Can you not reuse `printIter` here since this seems to be a limited copy-paste of that function?

If there are some mechanics here that require this then ok as long as this isn't the kind of thing users are going to regularly stub their toes on.

routing/http/client/client.go
```diff
@@ -160,6 +160,7 @@ func (c *measuringIter[T]) Close() error {
 	return c.Iter.Close()
 }
 
+// FindProviders searches for providers that are able to provide the given [cid.Cid].
```
aschmahmann, line 163: This is how it's most commonly used, but it's perhaps more accurately about coordination/metadata sharing around a given CID.

Some examples:
- Within kubo IPNS over PubSub advertises in the Amino DHT which IDs it's interested in so as to bootstrap the mesh
- Records in IPNI could hold basically anything (e.g. for a large SHA256 block it could map to a CID of a DAG containing an incrementally verifiable manifest as I did in https://github.com/aschmahmann/mdinc)

The text is likely ok as it is for now, but it might be worth adding another line about how this interface is sometimes used as a more general `CID -> []metadata` map.

routing/http/client/client.go
```diff
@@ -387,24 +389,26 @@ func (c *Client) FindPeers(ctx context.Context, pid peer.ID) (peers iter.ResultI
...
 		logger.Errorw("unknown media type", "MediaType", mediaType, "ContentType", respContentType)
 		return nil, errors.New("unknown content type")
 	}
 
-	return &measuringIter[iter.Result[types.Record]]{Iter: it, ctx: ctx, m: m}, nil
+	return &measuringIter[iter.Result[types.PeerRecord]]{Iter: it, ctx: ctx, m: m}, nil
 }
 
+// GetIPNS tries to retrieve the [ipns.Record] for the given [ipns.Name]. The record is
+// validated against the given name. If validation fails, an error is returned.
```
aschmahmann: Is there a specific validation error that's returned if it fails validation that the consumer is supposed to be allowed to check against, or not? If so that error should be called out.

Either way if we're calling out the failed validation -> returned error we may also want to clarify that no record is returned (although that's the standard behavior people implement in Go)

examples/routing/delegated-routing-client/main.go
hacdias: I don't think we can because it's `PeerRecord`, not a generic `Record`. Unless there's some way of weirdly converting that in a single line.

routing/http/client/client.go
hacdias: > Is there a specific validation error that's returned if it fails validation that the consumer is supposed to be allowed to check against, or not?

No. We could add a generic one in `ipns/` that wraps every validation error.

routing/http/client/client.go
hacdias: Updated, wdyt?

aschmahmann: SGTM

examples/routing/delegated-routing-client/main.go
aschmahmann: Isn't peer record a type of record because `Record` is an interface that `PeerRecord` implements?

hacdias: It is, but Go can't infer that. Maybe you can try, I tried and couldn't manage to make it work. Perhaps too many generics and interfaces together 😃

aschmahmann (approved):

routing/http/client/client.go
aschmahmann: Could add it, but fine as a separate PR

PR #491, opened 2023-10-17
feat!: remove util.MultiErr

hacdias:
Self-explanatory. Not used anywhere I could find. Boxo supports Go 1.21+. There are native ways of doing this. Cleanup is always good. It's funny we have our own multi error while we've just been using Uber's and others this whole time.


aschmahmann (approved):
LGTM. Not sure if anyone is actually using this, but if so it's a good opportunity to move them along.

Issue #121, opened 2019-12-09
[ipfs/go-bitswap] Proposal: Streaming GetBlocks


Stebalien:
At the moment, it's a bit tricky to continuously prefetch blocks as one needs to launch a goroutine per batch.

Proposal: Change the signature of `GetBlocks` to take an input channel and return an output channel:

```go
// Fetcher is an object that can be used to retrieve blocks (defined in go-ipfs-exchange-interface)
type Fetcher interface {
	// GetBlock returns the block associated with a given key.
	GetBlock(context.Context, cid.Cid) (blocks.Block, error)

	// GetBlocks returns a stream of blocks, given a stream of CIDs. It will
	// return blocks in any order.
	//
	// To wait for all remaining blocks, close the CID channel and wait for
	// the blocks channel to be closed. A closed channel does not mean that
	// _all_ blocks were retrieved, it just means that the fetcher is done
	// retrieving blocks.
	GetBlocks(context.Context, <-chan cid.Cid) (<-chan blocks.Block, error)
}
```

Additional elements:

* The `go-blockservice.BlockGetter` interface should be replaced with the `Fetcher` interface.
* ~The output channel should have the same buffer as the input channel (I think?).~
* The `error` return type is for indicating that the request couldn't even be started (e.g., closed service). This interface doesn't really have a way to report runtime errors (and there's almost always nothing we can do about them anyways).

This will require changes to go-ipfs-exchange-interface, go-blockservice, and go-bitswap.

aschmahmann:
Two other variations on the streaming proposal above

## Per Block errors

Return `<-chan BlockOption` instead of `<-chan blocks.Block` so that we can handle per-block errors

## Streaming sessions-like requests

A more complex, but more featureful version would be if we could add/remove requests from the channel so that we don't have to spin up one goroutine per block-get (or batch of gets). Could be something like

```
type AddRemoveCid interface {
	IsAdd() bool
	Key() cid.Cid
} // or a similar struct

GetBlocksCh(ctx context.Context, keys <-chan AddRemoveCid) (<-chan blocks.Block, error) // Or BlockOption if that's better
```

I wrote some code for this in this experiment
https://github.com/ipfs/go-bitswap/pull/593/files#diff-b751089c21e2437423d20fbee6c9a01a61c2422d5f7630ab8272d7fdeb63b654R17

[...]

PR #492, opened 2023-10-17
feat: blockstore: GetMany blockstore method

i-norden:
This PR introduces a `GetMany` blockstore method to compliment the existing `PutMany` method. This is for use in a `GetMany` go-ipld-cbor datastore implementation which, in turn, is for use in parallel `ForEach` methods for both go-amt-ipld and go-hamt-ipld. This PR uses the `TxnDatastore` keytransform/namespace wrapper introduced in https://github.com/ipfs/go-datastore/pull/210.

`GetMany` go-ipld-cbor PR that uses this: https://github.com/ipfs/go-ipld-cbor/pull/97

TODO:
Replace replace directives if/when the dependency is merged and released


blockstore/blockstore.go
```diff
@@ -310,6 +317,227 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
...
+	rehash       atomic.Bool
+	writeThrough bool
+	noPrefix     bool
+}
+
+func (bs *getManyBlockStore) HashOnRead(enabled bool) {
+	bs.rehash.Store(enabled)
+}
+
+func (bs *getManyBlockStore) Get(ctx context.Context, k cid.Cid) (blocks.Block, error) {
```
Jorropo: I am surprised this function needs to be duplicated, I think you could use embeding here.

blockstore/blockstore.go
```diff
@@ -310,6 +317,227 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
...
 }
 
+// GetManyOption is a getManyBlockStore option implementation
+type GetManyOption struct {
+	f func(bs *getManyBlockStore)
+}
+
+// NewGetManyBlockstore returns a default GetManyBlockstore implementation
+// using the provided datastore.TxnDatastore backend.
+func NewGetManyBlockstore(d ds.TxnDatastore, opts ...GetManyOption) GetManyBlockstore {
```
Jorropo: You implement both GetMany and PutMany but only return an interface which only expose GetMany.

Jorropo (commented):
Note for future me:
We maybe want automatic forwarding using type assertions in the default constructor function too.

blockstore/blockstore.go
i-norden: You are 100% correct- will adjust.

blockstore/blockstore.go
i-norden: Ahh thanks, good catch. This is a bit of testament to how much the `PutMany` is an afterthought, as we are actually only using the `GetMany` method at higher levels currently. Fixing now.

blockstore/blockstore.go
i-norden: This is possible but would require passing in a `ds.Batching` to the constructor function in addition to the`ds.TxnDatastore` (or some composite interface of the two) as `ds.TxnDatastore` can't be used to construct a regular `Blockstore` for embedding because it doesn't satisfy the `ds.Batching` interface, even though the `Put` and `Get` methods we would like to fall through to don't use the portions of that interface that it doesn't support...

i-norden: Another option would be to break the normal `Blockstore` up into the components that need `ds.Batching` and the ones that just need `ds.Datastore`.`ds.TxDatastore` satisfies `ds.Datastore` and we only need `ds.Datastore` to fulfill the `Get` and `Put` methods on either blockstore.

blockstore/blockstore.go
i-norden: Oh actually `PutMany` is already a method of the regular `Blockstore` interface so it is a part of the `GetManyBlockstore` interface returned here since it extends `Blockstore`. But, this can be made much more clear by declaring it explicitly in the new interface and also by adjusting the name of this interface and its implementation to reflect the fact that the `PutMany` method operates differently (through `ds.TxnFeature` instead of `ds.Batching`) such that the new `GetMany` method isn't the only difference vs the regular blockstore.

blockstore/blockstore.go
i-norden: Let me know what you think the best/cleanest approach is here

blockstore/blockstore.go
i-norden: Proposing to change the name to `TxnBlockstore`

Jorropo: :facepalm: mb, didn't knew we already had `PutMany`, `GetManyBlockstore` is fine.

blockstore/blockstore.go
```diff
@@ -64,6 +65,13 @@ type Blockstore interface {
 	HashOnRead(enabled bool)
 }
 
+// TxnBlockstore is a blockstore interface that supports GetMany and PutMany methods using ds.TxnDatastore
+type TxnBlockstore interface {
```
Jorropo: I think `TxnBlockstore` is wrong because the blockstore does not provide transactions, `GetManyBlockstore` was fine mb.

blockstore/blockstore.go
Jorropo: I would make it take the union of batching and txn datastores so it has access to all that is needed.
You could also implement a stub, a batch is inferior to a transaction while providing the same features:
> Batching datastores support deferred, grouped updates to the database. `Batch`es do NOT have transactional semantics: updates to the underlying datastore are not guaranteed to occur in the same iota of time. Similarly, batched updates will not be flushed to the underlying datastore until `Commit` has been called. `Txn`s from a `TxnDatastore` have all the capabilities of a `Batch`, but the reverse is NOT true. 

https://pkg.go.dev/github.com/ipfs/go-datastore#Batching

So if you accept a TxnDatastore you can use a simple stub that implements batches using transactions, this would make thee code work in the only place it's used right now.

Jorropo (commented):
Thx I'll try to look at this before the freeze on friday but I can't promise anything.

blockstore/blockstore.go
```diff
@@ -64,6 +65,13 @@ type Blockstore interface {
 	HashOnRead(enabled bool)
 }
 
+// GetManyBlockstore is a blockstore interface that supports GetMany and PutMany methods using ds.TxnDatastore
+type GetManyBlockstore interface {
+	Blockstore
+	PutMany(ctx context.Context, blocks []blocks.Block) error
+	GetMany(context.Context, []cid.Cid) ([]blocks.Block, []cid.Cid, error)
```
aschmahmann, line 72: Two recommendations:
1. It seems like there's no need for `[]blocks.Block` and `[]cid.Cid` since `Block` contains a `.Cid()` method https://github.com/ipfs/go-block-format/blob/v0.2.0/blocks.go#L19-L25
2. You may want to consider a streaming interface so that you don't have to buffer all the blocks in memory
   - The simplest option is to return either `(<-chan blocks.Block, error)` or `(<-chan BlockOption`) depending on how you want errors returned
   - More complicated, but more featureful, would be if this had a more "session-like" quality to it where you could add/remove requests from the channel so that you don't have to spin up one goroutine per-block get (or per batch of gets). Could be something like this https://github.com/ipfs/go-bitswap/pull/593/files#diff-b751089c21e2437423d20fbee6c9a01a61c2422d5f7630ab8272d7fdeb63b654R17

If returning an asynchronous object (e.g. channel or iterator) might be worth taking a look at https://github.com/ipfs/kubo/issues/4592 to make sure you don't run into some common pitfalls. With Go generics now iterators may also make this easier than it used to be.

blockstore/blockstore.go
```diff
@@ -310,6 +318,123 @@ func (bs *blockstore) AllKeysChan(ctx context.Context) (<-chan cid.Cid, error) {
...
+type batchingTxnDatastoreStub struct {
+	ds.BatchingFeature
+	ds.TxnDatastore
+}
+
+func (bs *getManyBlockstore) GetMany(ctx context.Context, cs []cid.Cid) ([]blocks.Block, []cid.Cid, error) {
+	if len(cs) == 1 {
+		// performance fast-path
+		block, err := bs.Get(ctx, cs[0])
+		return []blocks.Block{block}, nil, err
```
aschmahmann, line 359: Doesn't make sense to not return the CID here given it's in the signature, but also it doesn't seem like `[]cid.Cid` needs to be in the return signature

aschmahmann (changes requested):
Thanks @i-norden 🙏. IIUC this is related to your work landing a usable alternative to https://github.com/filecoin-project/go-hamt-ipld/pull/103 which is great to see.

Left some suggestions.

gammazero:
This does not seem needed at the current time. Closing this in favor of providing a streaming interface as a more flexible way of handling Get/GetMany.

aschmahmann:
Note: Existing issue on a streaming interface https://github.com/ipfs/boxo/issues/121.

PR #493, opened 2023-10-18
fix(gw): frugal DNSLink lookups on subdomains

lidel:
This PR refactors the rushed ipfs-companion Manifest V3 fix from https://github.com/ipfs/boxo/pull/462 to be more efficient and with better UX.

The old version was not very smart and wasted DNS lookup per every inlined DNSLink label (`my-v--long-example-com`) before retrying with un-inlined one (`my.v-long.example.com`).

This version is optimized to avoid unnecessary DNS lookups for both positive and negative scenarios.

It also ensures that in case of missing DNSLink, real FQDN (un-inlined domain name) is used in error path, which makes better UX AND simplifies the flow related to content blocking @hsanjuan  is working on.

### Positive scenario

```console
$ curl http://en-wikipedia--on--ipfs-org.ipns.localhost:8080 -H 'X-Forwarded-Proto: https'
```

DNS lookups made by gateway:
1. `en-wikipedia--on--ipfs-org` (💢 we tried the unlikely inlined version for every name first, and it always failed)
2. `en.wikipedia-on-ipfs.org`  

### Negative scenario (no DNSLink)

```console
$ curl http://google-com.ipns.localhost:8080 -H 'X-Forwarded-Proto: https'
failed to resolve /ipns/google.com/: could not resolve name: _dnslink subdomain at "google-com." is missing a TXT record (https://docs.ipfs.tech/concepts/dnslink/)
```

💢 **Note:** error message is about inlined notation `google-com` and not the real domain `google.com` :-1: 

DNS lookups made by gateway:
1. `google-com` (💢 less likely, but we checked it first)
2. `google.com` (more  likely,  but we checked this second)

### Positive scenario

```console
$ curl http://google-com.ipns.localhost:8080 -H 'X-Forwarded-Proto: https'
```

DNS lookups made by gateway:
1. `en.wikipedia-on-ipfs.org` (:green_circle:  no wasted lookups)

### Negative scenario (no DNSLink)

```console
$ curl http://google-com.ipns.localhost:8080 -H 'X-Forwarded-Proto: https'
failed to resolve /ipns/google.com/: could not resolve name: _dnslink subdomain at "google.com." is missing a TXT record (https://docs.ipfs.tech/concepts/dnslink/)
```

:green_circle: **Note:** error message is about real domain `google.com` and not inlined notation `google-com` :+1:

DNS lookups made by gateway:
1. `google.com` (:green_circle: we try this first as it is more likely to have dnslink, as seen in positive scenario above)
2. `google-com` (:green_circle: less likely, but we have to check to avoid false-negatives)

### DIFF
--- a/gateway/hostname.go
+++ b/gateway/hostname.go
@@ -165,15 +165,32 @@ func NewHostnameHandler(c Config, backend IPFSBackend, next http.Handler) http.H
 				// can be loaded from a subdomain gateway with a wildcard
 				// TLS cert if represented as a single DNS label:
 				// https://my-v--long-example-com.ipns.dweb.link
-				if ns == "ipns" && !strings.Contains(rootID, ".") {
-					// if there is no TXT recordfor rootID
-					if !hasDNSLinkRecord(r.Context(), backend, rootID) {
-						// my-v--long-example-com → my.v-long.example.com
-						dnslinkFQDN := toDNSLinkFQDN(rootID)
-						if hasDNSLinkRecord(r.Context(), backend, dnslinkFQDN) {
-							// update path prefix to use real FQDN with DNSLink
-							pathPrefix = "/ipns/" + dnslinkFQDN
-						}
+				if ns == "ipns" && !strings.Contains(rootID, ".") && strings.Contains(rootID, "-") {
+					// If there are no '.' but '-' is present in rootID, we most
+					// likely have an inlined DNSLink (like my-v--long-example-com)
+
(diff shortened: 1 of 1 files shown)


aschmahmann (approved):
LGTM.

We may still want to have more specific error checking for `hasDNSLinkRecord` than just checking if the resolution error is nil to prevent accidentally falling through to check the non-fqdn name. However, this is lower priority since it's mostly covering edge cases in the less common scenario of non-fqdn DNSLinks

hacdias (approved):

PR #494, opened 2023-10-18
feat(gw): improved InlineDNSLink and UninlineDNSLink

lidel:
@hsanjuan your version from https://github.com/ipfs-shipyard/nopfs/blob/5433362b11ac21130f84e5b1761c2c15aa87fef5/denylist.go#L24 highlighted the need for reusable implementation. 

This PR exposes `InlineDNSLink` and `UninlineDNSLink` and also includes my first stab at faster implementation (since this is DNS, we only care bout ASCII, no need for Unicode runes).

### Benchmark

"Simple" version is the old one:

```console
$ cd gateway
$ go test -bench=.

BenchmarkUninlineDNSLinkSimple-8   	 1687743	      1009.0 ns/op
BenchmarkUninlineDNSLink-8         	 3390468	       334.3 ns/op

BenchmarkInlineDNSLinkSimple-8     	 1915861	       672.4 ns/op
BenchmarkInlineDNSLink-8           	 2148770	       495.2 ns/op
```

### DIFF
--- a/gateway/hostname.go
+++ b/gateway/hostname.go
@@ -172,7 +172,7 @@ func NewHostnameHandler(c Config, backend IPFSBackend, next http.Handler) http.H
 					// We un-inline and check for DNSLink presence on domain with '.'
 					// first to minimize the amount of DNS lookups:
 					// my-v--long-example-com → my.v-long.example.com
-					dnslinkFQDN := toDNSLinkFQDN(rootID)
+					dnslinkFQDN := UninlineDNSLink(rootID)
 
 					// Does _dnslink.my.v-long.example.com exist?
 					if hasDNSLinkRecord(r.Context(), backend, dnslinkFQDN) {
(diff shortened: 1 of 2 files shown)


gateway/hostname.go
```diff
@@ -306,22 +306,59 @@ func isHTTPSRequest(r *http.Request) bool {
 
 // Converts a FQDN to DNS-safe representation that fits in 63 characters:
 // my.v-long.example.com → my-v--long-example-com
-func toDNSLinkDNSLabel(fqdn string) (dnsLabel string, err error) {
+// Implements specification from https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header
+func InlineDNSLink(fqdn string) (dnsLabel string, err error) {
```
hacdias:
```suggestion
// InlineDNSLink implements specification from https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header
func InlineDNSLink(fqdn string) (dnsLabel string, err error) {
```

Just Go convention-stuff.

gateway/hostname.go
```diff
@@ -306,22 +306,59 @@ func isHTTPSRequest(r *http.Request) bool {
...
+		return "", fmt.Errorf("inlined DNSLink incompatible with DNS label length limit of 63: %q", result)
+	}
+	return string(result), nil
 }
 
 // Converts a DNS-safe representation of DNSLink FQDN to real FQDN:
 // my-v--long-example-com → my.v-long.example.com
-func toDNSLinkFQDN(dnsLabel string) (fqdn string) {
+// Implements specification from https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header
+func UninlineDNSLink(dnsLabel string) (fqdn string) {
```
hacdias:
```suggestion
// UninlineDNSLink implements specification from https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header
func UninlineDNSLink(dnsLabel string) (fqdn string) {
```

Just Go convention-stuff.

gateway/hostname.go
```diff
@@ -306,22 +306,59 @@ func isHTTPSRequest(r *http.Request) bool {
...
+	/* What follows is an optimized version this three-liner:
 	fqdn = strings.ReplaceAll(dnsLabel, "--", "@") // @ placeholder is unused in DNS labels
 	fqdn = strings.ReplaceAll(fqdn, "-", ".")
 	fqdn = strings.ReplaceAll(fqdn, "@", "-")
 	return fqdn
+	*/
+	result := make([]byte, 0, len(dnsLabel))
```
hacdias, line 363: I'm not sure what happens in Go in this case: `result` is the size of `len(dnsLabel)`. However, when replacing `--` by `@`, it loses one character in length. Is that not potentially problematic?

hacdias (approved):
I have a question (see inline), but other than that it looks good.

gateway/hostname.go
lidel: Last argument of `make` defines byte slice capacity (how much space is allocated upfront in memory to accommodate future elements), the `len(string(result))` will reflect actually used bytes, and can be shorter than capacity.

PR #497, opened 2023-10-28
fix(gw): blocked content produces http error 410

lidel:
This quick and dirty  minimum we need right now to make content blocking from  https://github.com/ipfs/kubo/pull/10161 return HTTP 410 on rule match, and have tests that guard this pass.

- End-to-end test in `kubo/test/cli/content_blocking_test.go` (https://github.com/ipfs/kubo/pull/10161)
- We may want to refine this in future PR, but its enough for tests to pass and RC to ship.
- This should be merged with merge commit, so we have the same commit in the main branch, and we are not blocked by incompatible peering commits that follow.

### DIFF
--- a/gateway/errors.go
+++ b/gateway/errors.go
@@ -149,6 +149,8 @@ func webError(w http.ResponseWriter, r *http.Request, c *Config, err error, defa
 		code = http.StatusBadRequest
 	case isErrNotFound(err):
 		code = http.StatusNotFound
+	case isErrContentBlocked(err):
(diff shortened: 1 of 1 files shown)


lidel:
I'm merging this to have 182e86f5bb9b89c0b17f23c7f5739463ffbdf5bb in main boxo branch directly after a50f784985dd.
This allows us to land https://github.com/ipfs/kubo/pull/10161 into Kubo master without havign to deal with changes after a50f784985dd.

PR #504, opened 2023-11-03
chore: release 0.14.1

hacdias:


Jorropo (commented):
I'm pretty sure this should be `v0.15.0` because this is a breaking change `peer.ID.Pretty` was removed.

Issue #507, opened 2023-11-05
namesys: error when more than one IPFS-related DNSLink record exist


lidel:
Ref. https://github.com/ipfs/distributions/issues/1053

In cases like the above, we should not silently grab the first record, but instead, 
- error when a domain name has more than one DNSLink TXT record starting with one of IPFS namespaces (`dnslink=/ipfs/` or `dnslink=/ipns/`)
  - caveat: ignore records that use different namespaces (I've seen folks leveraging DNSLink for things other than IPFS).

```
Error: DNSLink lookup for "{domain}" returned more than one IPFS content path; ask domain owner to remove duplicate TXT record
```

lidel:
@hacdias since you've recently been looking into `boxo/namesys`, would appreciate your insight: is this an easy fix or a more involved refactor? (I've put this as 0.25 item for now)

Jorropo:
Triage notes: if this takes more than a few hours to do then it's P2.

hacdias:
I think this might be easier than expected, and I also want to raise another issue.

## We check both `_dnslink.{fqdn}` and `{fqdn}`, giving priority to the former

See: https://github.com/ipfs/boxo/blob/a966a74de9d2300c8a370758a62ec90fcb1c4329/namesys/dns_resolver.go#L72-L76

The specification does not define DNSLink entries in the root domain. Should we remove this? This was likely done for some backwards compatibility before the DNSLink specification was created: https://dnslink.dev/

## Duplicated entries for same namespace

Seems like we can just check here:

https://github.com/ipfs/boxo/blob/a966a74de9d2300c8a370758a62ec90fcb1c4329/namesys/dns_resolver.go#L153-L159

It seems that, at the moment, we're just parsing the first TXT entry we find, no matter the namespace. It seems to me that if a domain had 3 DNSLink records for 3 different namespaces (valid), we'd only pick one and not even choose the one we support.

This brings another question: _which_ namespace does Boxo prioritize? We support IPNS and IPFS. In case it has __both__, which do we choose?

lidel:
> The specification does not define DNSLink entries in the root domain. Should we remove this? This was likely done for some backwards compatibility before the DNSLink specification was created: https://dnslink.dev/

[...]

PR #510, opened 2023-11-08
fix(namesys): handling of multiple DNSLink TXT records

hacdias:
This addresses #507. I made individual commits to be easy to follow.

I removed the legacy support for DNSLink entries at the root of the domain. This made it easier to plug in the errors from the DNS. My issue was that if both `rootCh` and `subCh` returned an error, and the error was different, which one to choose? Probably `subCh`, but that still felt fragile. I can maintain legacy support if we **want it**, but that will result in a bit more work.

I also added two new errors to the package, to make it easier to test against specific types of errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,12 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+* 🛠 `boxo/namesys`: now fails when multiple valid DNSLink entries are found for the same domain. This used to cause undefined behavior before. Now, we return an error, according to the [specification](https://dnslink.dev/).
+
 ### Removed
 
 * 🛠 `boxo/gateway`: removed support for undocumented legacy `ipfs-404.html`. Use [`_redirects`](https://specs.ipfs.tech/http-gateways/web-redirects-file/) instead.
+* 🛠 `boxo/namesys`: removed support for legacy DNSLink entries at the root of the domain. Use [`_dnslink.` TXT record](https://docs.ipfs.tech/concepts/dnslink/) instead.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


lidel (approved):
LGTM. Removing undocumented legacy lookup is ok. People had multiple years to migrate to `_dnslink`, and the error message will be clear:

```
Error: DNSLink lookup for "google.com." failed: could not resolve name: DNSLink lookup could not find a TXT record (https://docs.ipfs.tech/concepts/dnslink/)
```

Added positive test for valid duplicates that we don't care about (making sure that is ignored correctly – we don't want to gatekeep DNSLink, people should be able to use it for non-IPFS namespaces if they want)

Tested it end-to-end against latest Kubo and works as expected:

```console
$ dig +short TXT _dnslink.invalid-duplicate-dnslink-test.example.com.
"dnslink=/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR"
"dnslink=/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi"

$ ipfs resolve /ipns/invalid-duplicate-dnslink-test.example.com
Error: DNSLink lookup for "invalid-duplicate-dnslink-test.example.com." failed: could not resolve name: DNSLink lookup returned more than one IPFS content path; ask domain owner to remove duplicate TXT records

$ dig +short TXT _dnslink.valid-duplicate-dnslink-test.example.com.
"dnslink=/foo/bar"
"dnslink=/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi"

$ ipfs resolve /ipns/valid-duplicate-dnslink-test.example.com
/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
```

PR #513, opened 2023-11-15
fix: all go-log import to v2

Dreamacro:
Seems like an oversight when migrating `bootstrap` and `peering`.


hacdias (approved):
thank you for this!

PR #515, opened 2023-11-21
fix(gateway): do not act on template errors

### LINKED ISSUE #487: http: superfluous response.WriteHeader
There are many error logs in the IPFS gateway.
```
http: superfluous response.WriteHeader call from github.com/ipfs/boxo/gateway.webError (errors.go:175)

http: superfluous response.WriteHeader call from github.com/ipfs/boxo/gateway.webError (errors.go:165)
```
Calling writeError after writeHeader will trigger this error. I'm not sure if this part of the code is causing it, but this code will trigger this issue.
https://github.com/ipfs/boxo/blob/6602207a8fa3624bfd52d8859c544abdb1ef0324/gateway/handler.go#L921-L928

hacdias:
Fixes #487. When we try to execute a template, we have already written the header. In some places, we were already ignoring the template execution error. In others, we were calling `webError`. That would lead to a duplicate header writing warning.

There is not much that can be done if a template fails to execute: we could perhaps log it, but so far what we've done in other places is to ignore. A way to circumvent this would be to write to a buffer and then write to `http.ResponseWriter`. That would allow us to catch the template error, but writing can always fail (e.g., broken connection).

I wouldn't care much about this and I think just ignoring the template writing error is probably fine. If errors happen, they will most likely be writing errors, so broken connections and other network issues, most likely.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -912,13 +912,11 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
 	// - redirects to intendedURL after a short delay
 
 	w.WriteHeader(http.StatusBadRequest)
-	if err := redirectTemplate.Execute(w, redirectTemplateData{
+	_ = redirectTemplate.Execute(w, redirectTemplateData{
 		RedirectURL:   intendedURL,
 		SuggestedPath: intendedPath.String(),
 		ErrorMsg:      fmt.Sprintf("invalid path: %q should be %q", r.URL.Path, intendedPath.String()),
-	}); err != nil {
-		i.webError(w, r, fmt.Errorf("failed to redirect when fixing superfluous namespace: %w", err), http.StatusBadRequest)
-	}
+	})
 
 	return true
 }
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -912,11 +912,14 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
 	// - redirects to intendedURL after a short delay
 
 	w.WriteHeader(http.StatusBadRequest)
-	_ = redirectTemplate.Execute(w, redirectTemplateData{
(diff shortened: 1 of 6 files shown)


gateway/handler.go
```diff
@@ -912,13 +912,11 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
 	// - redirects to intendedURL after a short delay
 
 	w.WriteHeader(http.StatusBadRequest)
-	if err := redirectTemplate.Execute(w, redirectTemplateData{
+	_ = redirectTemplate.Execute(w, redirectTemplateData{
 		RedirectURL:   intendedURL,
 		SuggestedPath: intendedPath.String(),
 		ErrorMsg:      fmt.Sprintf("invalid path: %q should be %q", r.URL.Path, intendedPath.String()),
-	}); err != nil {
-		i.webError(w, r, fmt.Errorf("failed to redirect when fixing superfluous namespace: %w", err), http.StatusBadRequest)
```
lidel, line 920: @hacdias instead of silently failing in all these places and ignoring `err`, could we still write the error to the `http.ResponseWriter`? 

Even tho we can't set error status code, this will be useful for end user who tries to figure out what went wrong, unlike the empty body.

My thinking is to replace `i.webError(` calls with:

```go
_, _ = w.Write([]byte(fmt.Sprintf("error during body generation: %v", err)))
```

Thoughts?

lidel (changes requested):

gateway/handler.go
hacdias: @lidel I did, I think it is fine. My initial idea was that we know that the templates are correct. They are tested, we send the right structs. Thus, having an error _executing_ the template is extremely unlikely. Most of these errors will likely be network errors, and attempting to write again will also fail.

But yes, by printing the error we will also catch the extremely unlikely template executing errors.

lidel (approved):
Thanks! Lgtm.

PR #516, opened 2023-11-23
fix: recursive directory filtering in `ShouldExclude`

crackcomm:
Ensure ignore rule `bar/` matches directories by appending `/`. Addresses issue where `bar/` didn't catch directories. After this change, both `bar` and `bar/` rules filter out the `bar` directory and its contents.

Note: This is required because git would skip empty directories and use full paths relative to repo.
I'm not sure how much I could break the API but it's still just using file names as opposed to full relative paths for filtering which is not ideal.

### DIFF
--- a/files/filter.go
+++ b/files/filter.go
@@ -1,6 +1,7 @@
 package files
 
 import (
+	"fmt"
 	"os"
 
 	ignore "github.com/crackcomm/go-gitignore"
@@ -44,6 +45,9 @@ func (filter *Filter) ShouldExclude(fileInfo os.FileInfo) (result bool) {
 	path := fileInfo.Name()
 	if !filter.IncludeHidden && isHidden(fileInfo) {
 		return true
+	}
+	if fileInfo.IsDir() {
+		path = fmt.Sprintf("%s/", path)
 	}
 	return filter.Rules.MatchesPath(path)
 }


hacdias (changes requested):
@crackcomm just to understand: you're saying the current implementation is not fully compatible with https://git-scm.com/docs/gitignore, correct? If so, do you mind fixing the tests (adding the `.IsDir` method to the [mock](https://github.com/crackcomm/boxo/blob/fe0a8237ce7ca56e314928152a18886c0b147fce/files/filter_test.go#L9-L22)), as well as adding a regression test to ensure that this is working as expected?

crackcomm:
> @crackcomm just to understand: you're saying the current implementation is not fully compatible with https://git-scm.com/docs/gitignore, correct?

Yes, this PR fixes:

> `frotz/` matches `frotz` and `a/frotz` that is a directory (all paths are relative from the `.gitignore` file)

Nevertheless even after this change it's not fully compatible. To be fully compatible, the paths have to be matched against a relative path, right now it's just a [file name](https://github.com/ipfs/boxo/blob/fe0a8237ce7ca56e314928152a18886c0b147fce/files/filter.go#L45).

crackcomm:
@hacdias I can add the function to the mock, no problem but could you please guide me where to add the test?

I would also want to add another change (not sure if separate PR or should I change this one): do NOT add the hidden files when they match ignore rules even if `filter.IncludeHidden` is true. Do you think that is a reasonable change?

hacdias:
@crackcomm as for adding a test, I think `filter_test.go` is the place to go.

> Do you think that is a reasonable change?

I think so. To me, that sounds like something that should've already been happening. If it is explicitly ignored as per the file, `IncludeHidden` should not override it. 

For me you can include it with this PR, but please keep the commits separate just in case.

lidel (changes requested):
Triage notes:
- Rebased, but CI fails
- [`TestFileFilter`](https://github.com/ipfs/boxo/actions/runs/8437998187/job/23109097738?pr=516#step:13:81) seems to fail on CI, needs to be updated  (and cI needs to be green)
- We need a PR against Kubo that confirms [these sharness tests that use filtering via `--ignore*` param in `ipfs add`](https://github.com/search?q=repo%3Aipfs%2Fkubo+%22--ignore%22&type=code) pass


@crackcomm are you able to address the above?

lidel:
I've converted this into a bug report: https://github.com/ipfs/boxo/issues/648
Closing this one, as it does not fully fix, nor test.  T
Thank you for bringing this to our attention.
(If you have time to work on this in the future, feel free to open a new PR (with tests).)

PR #517, opened 2023-11-29
coreiface: remove and move to Kubo

hacdias:
Removes `coreiface` package which is an intrinsic part of Kubo. We haven't done it before because of the path package, which has now been completely revamped. This will simplify new features when we need to update or add new thing to the API, such as https://github.com/ipfs/kubo/pull/10235.


Jorropo (approved):
Thx

PR #523, opened 2023-12-08
fix(gw): entity-bytes with negative indexes beyond file size

hannahhoward:
# Goal

Fix the following problem:
- I pass a pair of negative indexs to the `entity-bytes` parameter of a trustless HTTP request.
- The start index is MORE negative than the length of the underlying entity (i.e. file) -- so `entity-bytes=-1000,-50` on a 100 byte file
- The code will see the start index is negative and add the entity length, but that will still leave it negative (https://github.com/ipfs/boxo/blob/main/gateway/blocks_backend.go#L510) -- in our example case from is no `-900`
- The calculation of bytes to read will be higher than it should be -- https://github.com/ipfs/boxo/blob/main/gateway/blocks_backend.go#L533 -- in our example case, `1 + 50 - -900` = 951
- When executed, this will simply read all 100 bytes of the underlying entity, instead of the intended behavior to read only the first 50 (i.e everything up to the -50 index)

# Implementation

Very simple fix -- if after adding the entity length, the from is still negative, simply set it to 0.

This is the logic already implemented in the IPLD selector for ranges -- https://github.com/ipld/go-ipld-prime/blob/master/traversal/selector/matcher.go#L49

# For discussion

This wil cause a gateway conformance test to fail, which is due to a failure in the way the test is setup -- the issue is documented here: https://github.com/ipfs/gateway-conformance/issues/189

### DIFF
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -508,6 +508,9 @@ func walkGatewaySimpleSelector(ctx context.Context, p path.ImmutablePath, params
 					return err
 				}
 				from = fileLength + entityRange.From
+				if from < 0 {
+					from = 0
+				}
 				foundFileLength = true
 			}
 
--- a/gateway/blocks_backend.go
+++ b/gateway/blocks_backend.go
@@ -524,13 +524,15 @@ func walkGatewaySimpleSelector(ctx context.Context, p path.ImmutablePath, params
 			}
 
 			to := *entityRange.To
-			if (*entityRange.To) < 0 && !foundFileLength {
-				fileLength, err = f.Seek(0, io.SeekEnd)
-				if err != nil {
-					return err
+			if (*entityRange.To) < 0 {
+				if !foundFileLength {
+					fileLength, err = f.Seek(0, io.SeekEnd)
+					if err != nil {
+						return err
+					}
+					foundFileLength = true
 				}
 				to = fileLength + *entityRange.To
-				foundFileLength = true
 			}
 
 			numToRead := 1 + to - from
--- a/.github/workflows/gateway-conformance.yml
+++ b/.github/workflows/gateway-conformance.yml
@@ -16,18 +16,18 @@ jobs:
     steps:
       # 1. Download the gateway-conformance fixtures
       - name: Download gateway-conformance fixtures
-        uses: ipfs/gateway-conformance/.github/actions/extract-fixtures@v0.4
+        uses: ipfs/gateway-conformance/.github/actions/extract-fixtures@v0.5
         with:
           output: fixtures
           merged: true
 
       # 2. Build the car-gateway
       - name: Setup Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
           go-version: 1.21.x
       - name: Checkout boxo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           path: boxo
       - name: Build car-gateway
@@ -40,7 +40,7 @@ jobs:
 
       # 4. Run the gateway-conformance tests
       - name: Run gateway-conformance tests
-        uses: ipfs/gateway-conformance/.github/actions/test@v0.4
+        uses: ipfs/gateway-conformance/.github/actions/test@v0.5
         with:
           gateway-url: http://127.0.0.1:8040
           json: output.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ The following emojis are used to highlight certain changes:
 
 ### Removed
 
+### Fixed
+
+* `boxo/gateway`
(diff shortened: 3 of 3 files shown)


rvagg (approved):

gateway/blocks_backend.go
```diff
@@ -508,6 +508,9 @@ func walkGatewaySimpleSelector(ctx context.Context, p path.ImmutablePath, params
 					return err
 				}
 				from = fileLength + entityRange.From
+				if from < 0 {
+					from = 0
+				}
```
lidel, line 513: @rvagg @hannahhoward adding a guard here makes sense, but implicit fixup to 0 makes me a bit uneasy.

If the `fileLength` is 100, and user explicitly requested `entity-bytes=-1000:-75`, 
should we ignore the fact the user expects to way more than 25?

We've been trying to avoid surprises in Gateway semantics, and follow conventions from HTTP RFCs where possible. When designing `entity-bytes`,  we've been trying to follow HTTP Range Requests semantics, and those do not silently fixup ranges for the user, but return HTTP Error 416 Range Not Satisfiable every time things are out of range, removing surface for hard to debug errors and allowing user to inspect situation and retry with correct range.


Could we return below error instead?

```suggestion
				from = fileLength + entityRange.From
				if from < 0 {
					return fmt.Errorf("tried to read beyond the file length from the end")
				}
```

lidel (changes requested):
Thank you, adding a guard here makes sense, but since this is HTTP Gateway, [initial instinct] I feel is we should error instead of silently fixing up – rationale below.

_(we will engage on this in january, after holiday break, and will include a fix in Kubo 0.26, but posting today so we have more time to think about this)_

rvagg:
As an implementer I don't want to have to make status code decisions after I've successfully processed the first block; I know boxo is fine buffering all of the blocks at the beginning but that's not what we do elsewhere and the transport is currently designed for maximal streaming. So introducing a requirement to status code at some point >1 block down into a DAG is going to introduce further incompatibilities between implementations.

I think I see the negative start value landing before the start of the file in a similar way to the end value being beyond the end of the file, it's like setting the end to MaxInt64, or `*`. It's not quite the same, because in most cases `0` would do the job for you. But perhaps you want to have a query that always grabs the last >= 100 bytes for a file - let's say Station has a reason to do this for any unixfs file it can find indexed in filecoin for some weird reason, so it has a generic query that adds `entity-bytes=-100:*` - I think I'd expect that to just truncate the start if it encountered a 50 byte file rather than error out that it's unsatisfiable.

gateway/blocks_backend.go
hannahhoward: Sorry for failing to engage -- I'm fine with either outcome. Though I should note if you do this one of your gateway tests uses a From with a very negative value -- the very one I corrected. It will need to be changed to produce an error, and this will put it out of alignment with the IPLD prime implementation.

hannahhoward: IOW, that means that either this test (the one I submitted the Gateway Conformance PR for) should be removed entirely, or erroring should be added to the actual trustless spec.

Personally, I don't think an obscure edge case is important enough to warrent being in the spec, so personally I'd just remove that gateway conformance test.

hannahhoward: cc: @rvagg @lidel 

Also, I'm pretty immersed in Saturn stuff FYI so just to say my engagement may be a bit slow.

lidel:
(looked deeper with a bit more fresh eyes today) I agree, truncation proposed here will be less headache across systems than HTTP-style error that is hard to enforce across implementations.

(Given the non-HTTP abstractions we will be dealing with, and the fact we already have a situation where we return CAR With HTTP 200 on error cases such as sub-path not found in a DAG, it is not as big precedent as I originally thought, and truncation instead of error makes implementations easier)

@hannahhoward agree, this is a niche thing, but I do think we should have one sentence in spec and basic tests for edge cases like this because someone may build things that depend on them, like @rvagg noted above in https://github.com/ipfs/boxo/pull/523#issuecomment-1858737243

Created https://github.com/ipfs/kubo/pull/10320 to test this PR against https://github.com/ipfs/gateway-conformance/pull/190. 

Will try to land it this week, next steps are:

- [x] Once CI is green on the Kubo PR  https://github.com/ipfs/kubo/pull/10320 (which tests end-to-end), 
- [x] I will land conformance release with https://github.com/ipfs/gateway-conformance/pull/190  first
- [x] then bubble up to this PR  (boxo)
- [x] and then bubble up release versions to https://github.com/ipfs/kubo/pull/10320

lidel (approved):
- switched to tests released in https://github.com/ipfs/gateway-conformance/releases/tag/v0.5.0
  - kubo https://github.com/ipfs/kubo/pull/10320 runs against both gateway-conformance@v0.5 and this PR, and CI is green there (ci in this PR might fail until that is merged upstream, so ok to disregard)
- DSL for deterministic CID assertions is out of scope here, will track it in https://github.com/ipfs/gateway-conformance/issues/112
- Bubbling up, thanks all!

PR #525, opened 2023-12-15
feat(pinning): add support for names

hacdias:
Adds the long awaited support for labels for pins (https://github.com/ipfs/kubo/issues/4586). The support was already kinda there, just needed to be plumbed out. This might not be the best approach and I think the pinning code is quite convoluted in general. But it works.

See https://github.com/ipfs/kubo/pull/10261

### DIFF
--- a/pinning/pinner/dspinner/pin.go
+++ b/pinning/pinner/dspinner/pin.go
@@ -665,16 +665,16 @@ func (p *pinner) loadPin(ctx context.Context, pid string) (*pin, error) {
 }
 
 // DirectKeys returns a slice containing the directly pinned keys
-func (p *pinner) DirectKeys(ctx context.Context) <-chan ipfspinner.StreamedPin {
-	return p.streamIndex(ctx, p.cidDIndex)
+func (p *pinner) DirectKeys(ctx context.Context, detailed bool) <-chan ipfspinner.StreamedPin {
+	return p.streamIndex(ctx, p.cidDIndex, detailed)
 }
 
 // RecursiveKeys returns a slice containing the recursively pinned keys
-func (p *pinner) RecursiveKeys(ctx context.Context) <-chan ipfspinner.StreamedPin {
-	return p.streamIndex(ctx, p.cidRIndex)
+func (p *pinner) RecursiveKeys(ctx context.Context, detailed bool) <-chan ipfspinner.StreamedPin {
+	return p.streamIndex(ctx, p.cidRIndex, detailed)
 }
 
-func (p *pinner) streamIndex(ctx context.Context, index dsindex.Indexer) <-chan ipfspinner.StreamedPin {
+func (p *pinner) streamIndex(ctx context.Context, index dsindex.Indexer, detailed bool) <-chan ipfspinner.StreamedPin {
 	out := make(chan ipfspinner.StreamedPin)
 
 	go func() {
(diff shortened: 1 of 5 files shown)


pinning/pinner/dspinner/pin.go
```diff
@@ -688,21 +688,32 @@ func (p *pinner) streamIndex(ctx context.Context, index dsindex.Indexer) <-chan
...
-				out <- ipfspinner.StreamedCid{Err: err}
+				out <- ipfspinner.StreamedPin{Err: err}
 				return false
 			}
+
+			pp, err := p.loadPin(ctx, value)
+			if err != nil {
+				out <- ipfspinner.StreamedPin{Err: err}
+				return false
+			}
```
lidel: ⚠️ lmk if I got this wrong, but calling `loadPin` here adds overhead of loading CBOR representation of every pin from datastore, and deserializing it.

Could be a problem in a bigger repo with a lot of pins, especially if someone does not use names at all.

I wonder if we could keep the old default behavior, and not call `loadPin`, and do this additional deserialization only if someone explicitly called `ipfs pin ls --names`?

Or, we could also do a benchmark and see if this causes any significant slowdown with millions of pins.

lidel (changes requested):

pinning/pinner/dspinner/pin.go
hacdias: I think the benchmark + names option might be a good idea indeed.

hacdias: I added a benchmark. I think we should indeed add a `--names` or `--long` (to match the files API?) option.

Benchmark results (always ~2.7 times slower):

```
BenchmarkDetails/Keys-NoDetails-128-10         	    5762	    211920 ns/op	  118836 B/op	    1096 allocs/op
BenchmarkDetails/Keys-Details-128-10           	    2272	    580459 ns/op	 1475367 B/op	    3912 allocs/op
BenchmarkDetails/Keys-NoDetails-256-10         	    2958	    398473 ns/op	  223355 B/op	    2122 allocs/op
BenchmarkDetails/Keys-Details-256-10           	    1118	   1079467 ns/op	 2936426 B/op	    7756 allocs/op
BenchmarkDetails/Keys-NoDetails-512-10         	    1538	    779022 ns/op	  432658 B/op	    4175 allocs/op
BenchmarkDetails/Keys-Details-512-10           	     536	   2293866 ns/op	 5858794 B/op	   15441 allocs/op
BenchmarkDetails/Keys-NoDetails-1024-10        	     792	   1517402 ns/op	  933169 B/op	    8280 allocs/op
BenchmarkDetails/Keys-Details-1024-10          	     288	   4107997 ns/op	11785236 B/op	   30811 allocs/op
BenchmarkDetails/Keys-NoDetails-2048-10        	     402	   2959785 ns/op	 1926020 B/op	   16486 allocs/op
BenchmarkDetails/Keys-Details-2048-10          	     154	   7757229 ns/op	23629631 B/op	   61544 allocs/op
BenchmarkDetails/Keys-NoDetails-4096-10        	     204	   5878773 ns/op	 3846053 B/op	   32895 allocs/op
BenchmarkDetails/Keys-Details-4096-10          	      72	  15350311 ns/op	47253256 B/op	  123010 allocs/op
BenchmarkDetails/Keys-NoDetails-8192-10        	      94	  12786774 ns/op	 8202228 B/op	   65710 allocs/op
BenchmarkDetails/Keys-Details-8192-10          	      37	  33523060 ns/op	95016018 B/op	  245937 allocs/op
BenchmarkDetails/Keys-NoDetails-16384-10       	      44	  24266083 ns/op	17029126 B/op	  131392 allocs/op
BenchmarkDetails/Keys-Details-16384-10         	      16	  63185271 ns/op	190656195 B/op	  491842 allocs/op
```

lidel (approved):
Tested with https://github.com/ipfs/kubo/pull/10261 and lgtm.

Issue #83, opened 2021-08-30
[ipfs/go-bitswap] Failure to fetch content from connected nodes


vdamle:
### Environment

IPFS Version:

```
go-ipfs version: 0.9.1-dc2715a
Repo version: 11
System version: amd64/linux
Golang version: go1.15.2
```

We are running a private IPFS cluster with 2 or more nodes (docker/k8s)

### Description

As an example, we've got a network with 2 nodes - `node1` and `node2`. After startup, we observe that `node2` cannot fetch content from `node1` (a bootstrap peer) because it is not registered as a `Partner`. We see on `node1` that `node2` is registered as a `Peer/Partner`. I tried adding some additonal debug logs and best I could tell, the steps to register a `BitSwapNetwork  Receiver` to receive events from the network was not complete before `node2` received a "connected" notification from `node1`. It seems to me like this is a race condition and we see this occur 90% of the time, when we spin up nodes (ref: https://github.com/ipfs/go-bitswap/blob/0e81f7c1fef795a8d34cdd494fa2d8fdc045d199/bitswap.go#L215 ). As a throw of dice, I tried to move up the call _before_ any options/decision engine is setup, this did not make a whole lot of difference to the outcome.

### Workaround

* Execute `ipfs swarm peer disconnect <address>`
* Execute `ipfs swarm peer connect <address>`

We can see that the [PeerConnected](https://github.com/ipfs/go-bitswap/blob/0e81f7c1fef795a8d34cdd494fa2d8fdc045d199/bitswap.go#L528) handler is invoked from  [Connected](https://github.com/ipfs/go-bitswap/blob/0e81f7c1fef795a8d34cdd494fa2d8fdc045d199/network/ipfs_impl.go#L428), which adds the Peer to the peer list. After this, `WANT_HAVE` is sent as expected to the peer and content is retrieved.

PS: Originally, I created the following [issue](https://github.com/ipfs/go-ipfs/issues/8346) in `go-ipfs`, but after further investigation, it feels like `go-bitswap` is the right place to report this.

I'd love to hear any suggestions/thoughts for how this can be addressed since I have a limited view of the architecture/intent for how connections are supposed to work. I have attached logs and output of commands from both nodes.

[ipfs-config-node1.txt](https://github.com/ipfs/go-bitswap/files/7079480/ipfs-config-node1.txt)
[ipfs-config-node2.txt](https://github.com/ipfs/go-bitswap/files/7079481/ipfs-config-node2.txt)
[...]

PR #526, opened 2023-12-16
Improve peer connection handling

peterargue:
This PR was originaly submitted to https://github.com/ipfs/go-bitswap/pull/590

Currently, the bitswap network only adds peers to its client/server peer lists that connected after the network has started. This means the network must be started before enabling the libp2p listeners, otherwise some peers may connect too early and never be added to the network.

To address this, this PR updates the bitswap network's `Start()` method to first, listen for new peer connection events, and then iterate through all peers and add any that support the bitswap protocols. The issue and solution were originally described by @Stebalien in https://github.com/ipfs/boxo/issues/83.

### DIFF
--- a/bitswap/network/ipfs_impl.go
+++ b/bitswap/network/ipfs_impl.go
@@ -356,13 +356,32 @@ func (bsnet *impl) Start(r ...Receiver) {
 		bsnet.host.SetStreamHandler(proto, bsnet.handleNewStream)
 	}
 
+	// try to subscribe to libp2p events that indicate a change in connection state
+	// if this fails, continue as normal
+	err := bsnet.trySubscribePeerUpdates()
+	if err != nil {
+		log.Errorf("failed to subscribe to libp2p events: %s", err)
+	}
+
+	// listen for disconnects and start processing the events
+	bsnet.host.Network().Notify((*netNotifiee)(bsnet))
+	bsnet.connectEvtMgr.Start()
+}
+
+func (bsnet *impl) Stop() {
+	bsnet.connectEvtMgr.Stop()
+	bsnet.host.Network().StopNotify((*netNotifiee)(bsnet))
+	bsnet.cancel()
+}
+
+func (bsnet *impl) trySubscribePeerUpdates() error {
 	// first, subscribe to libp2p events that indicate a change in connection state
 	sub, err := bsnet.host.EventBus().Subscribe([]interface{}{
 		&event.EvtPeerProtocolsUpdated{},
 		&event.EvtPeerIdentificationCompleted{},
 	})
 	if err != nil {
-		panic(err)
+		return err
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -379,15 +398,7 @@ func (bsnet *impl) Start(r ...Receiver) {
 		}
 	}
 
-	// finally, listen for disconnects and start processing the events
-	bsnet.host.Network().Notify((*netNotifiee)(bsnet))
-	bsnet.connectEvtMgr.Start()
-}
-
-func (bsnet *impl) Stop() {
-	bsnet.connectEvtMgr.Stop()
-	bsnet.host.Network().StopNotify((*netNotifiee)(bsnet))
-	bsnet.cancel()
+	return nil
 }
 
 func (bsnet *impl) peerUpdatedSubscription(ctx context.Context, sub event.Subscription) {
--- a/bitswap/network/ipfs_impl.go
+++ b/bitswap/network/ipfs_impl.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/core/connmgr"
+	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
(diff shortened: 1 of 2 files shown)


peterargue:
Hey @Jorropo, I ported this change over.

bitswap/network/ipfs_impl.go
```diff
@@ -354,15 +357,85 @@ func (bsnet *impl) Start(r ...Receiver) {
...
 		bsnet.host.SetStreamHandler(proto, bsnet.handleNewStream)
 	}
+
+	// first, subscribe to libp2p events that indicate a change in connection state
+	sub, err := bsnet.host.EventBus().Subscribe([]interface{}{
+		&event.EvtPeerProtocolsUpdated{},
+		&event.EvtPeerIdentificationCompleted{},
+	})
+	if err != nil {
+		panic(err)
```
Stebalien: Well, that's fun. This can fail if the libp2p node is stopped...

I'd:

1. Log an error.
2. Start `bsnet.connectEventMgr`.
3. Ignore all the libp2p stuff.

For now. Ideally we'd return an error higher up, but it honestly doesn't matter all that much.

bitswap/network/ipfs_impl.go
```diff
@@ -354,15 +357,85 @@ func (bsnet *impl) Start(r ...Receiver) {
...
 	bsnet.connectEvtMgr.Start()
 }
 
 func (bsnet *impl) Stop() {
 	bsnet.connectEvtMgr.Stop()
 	bsnet.host.Network().StopNotify((*netNotifiee)(bsnet))
+	bsnet.cancel()
+}
+
+func (bsnet *impl) peerUpdatedSubscription(ctx context.Context, sub event.Subscription) {
```
Stebalien, line 404: This will race with the "old" version of notifications. Instead, we need to:

1. Drop the old version (the net notifiee).
2. Subscribe to `EvtPeerConnectednessChanged` to get connectivity changes.
3. We care about "disconnected" events, but ignore "connected" events. Instead, rely on "protocol changed" and "identify" notifications to tell you that the peer supports bitswap (but check this with @Jorropo).

bitswap/network/ipfs_impl.go
```diff
@@ -354,15 +357,85 @@ func (bsnet *impl) Start(r ...Receiver) {
...
+		}
+	}
+}
+
+func (bsnet *impl) peerSupportsBitswap(peerID peer.ID) bool {
+	protocols, err := bsnet.host.Peerstore().SupportsProtocols(peerID, bsnet.supportedProtocols...)
+	return err == nil && len(protocols) > 0
+}
+
+func (bsnet *impl) hasBitswapProtocol(protos []protocol.ID) bool {
```
Stebalien, line 437: This should be a free function.

bitswap/network/ipfs_impl.go
Stebalien: Ah, but once you do this, you _also_ have to move the existing connection walk to the top of this function. That is, you need to:

1. Subscribe to connect/disconnect events.
2. Process all existing connections.
3. Start processing the events.

Failure modes:

* If you do 2 before 1, you could miss new connections.
* If you do 3 concurrent with 2, you could process a connect/disconnect event in step-3, then re-add the peer in the step-2 loop.

If you do them in order (1,2,3), you _may_ end up processing a "connect" event for a disconnected peer in step 3, but then you'll process the disconnect event.

You could _additionally_ check the peer's "connectedness" before updating a peer's status, but I don't think that is necessary.

Stebalien: Hm. Although, I _think_ we need to ask libp2p for all connections to the peer to make sure that we have a non-transient one. I wonder if that'll cause any issues? I assume we'll re-identify when we create a new connection... That'll require some testing.

Ideally libp2p would let us filter protocols by connection type, but that's a larger change.

Thoughts @Jorropo? @aschmahmann?

bitswap/network/ipfs_impl.go
peterargue: this was needed because the protocol IDs are generated based on the network config

bitswap/network/ipfs_impl.go
peterargue: addressed in [decd8d9](https://github.com/ipfs/boxo/pull/526/commits/decd8d9277e2d9ffcd7eb30bafcc4e0fed28b181)

Stebalien:
I've filed two related libp2p issues:

- https://github.com/libp2p/go-libp2p/issues/2692 (easier, but may cause unexpected issues).
- https://github.com/libp2p/go-libp2p/issues/2693 (harder, but highly desirable and is likely to _fix_ hidden issues and simplify quite a bit of code just about everywhere).

bitswap/network/ipfs_impl.go
Stebalien: Oh, I see.

Stebalien:
I believe https://github.com/libp2p/go-libp2p/pull/2696 should unblock this from the libp2p side.

peterargue:
@Jorropo any chance you can take a look and provide any remaining feedback?

lidel:
Triage notes:
- unsure if this is still needed, https://github.com/libp2p/go-libp2p/issues/2692 landed, and the stack is being refactored (e.g. https://github.com/libp2p/go-libp2p-kad-dht/pull/976) to take advantage of it
  - thoughts @sukunrt  @MarcoPolo @2color?

peterargue:
@lidel are you saying that on the latest version, peers who connect before the bitswap network is started will be included as peers?

PR #531, opened 2024-01-01
Update tracing.md

0xbasar:
It appears there's a typo in the package name. It should be opentelemetry-go instead of opentelemtry-go.


Jorropo (approved):
Thx

PR #535, opened 2024-01-03
Remove content routing from bitswap network

Jorropo:
Based on #534

PR #536, opened 2024-01-03
Remove provider query manager

Jorropo:
Based on #535


Jorropo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/ipfs/boxo/536?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#536** <a href="https://app.graphite.dev/github/pr/ipfs/boxo/536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ipfs/boxo/536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#535** <a href="https://app.graphite.dev/github/pr/ipfs/boxo/535?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#534** <a href="https://app.graphite.dev/github/pr/ipfs/boxo/534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#578** <a href="https://app.graphite.dev/github/pr/ipfs/boxo/578?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

bitswap/client/internal/session/session.go
```diff
@@ -396,12 +409,68 @@
...
 	go func(k cid.Cid) {
-		for p := range s.providerFinder.FindProvidersAsync(ctx, k) {
-			// When a provider indicates that it has a cid, it's equivalent to
-			// the providing peer sending a HAVE
-			s.sws.Update(p, nil, []cid.Cid{c}, nil)
+		ctx, span := internal.StartSpan(ctx, "Session.findMorePeers")
+		defer span.End()
+		if span.IsRecording() {
+			span.SetAttributes(attribute.Stringer("cid", c))
+		}
```
hacdias, line 418: Can't this be done directly? AFAIK, it won't call `.String` unless it's indeed recording. I think it could indeed be more problematic if directly passed as a string and needed to encode it every time. But here it seems fine. Is there another reason?

hacdias (approved):
as on the previous stack, it looks like it does what it says, just a comment inline

bitswap/client/internal/session/session.go
Jorropo: It literally always call `.String` instantly:
https://github.com/open-telemetry/opentelemetry-go/blob/98b32a6c3a87fbee5d34c063b9096f416b250897/attribute/kv.go#L85
```go
// Stringer creates a new key-value pair with a passed name and a string
// value generated by the passed Stringer interface.
func Stringer(k string, v fmt.Stringer) KeyValue {
	return Key(k).String(v.String())
}
```

gammazero:
ProviderQueryManager has been moved and refactored sufficiently where we do not need this PR, and if we want to make changes like this in the future it should be done in a new PR.

PR #539, opened 2024-01-08
routing/http!: support for IPIP-378 (delegated content and peer providing)

hacdias:
This is the client and server implementation for the delegated PUTs for peers and providers. See the specification on https://github.com/ipfs/specs/pull/378. There are breaking changes. Check the changelog to know more about them.

Since the specification is not yet finished and set in stone, there are quite a few questions that I need answered in order to finish the code. They are written as TODO comments in line and I left them also in the specs PR.

- [x] Solve in-code `TODO`, which are mostly related to specifications
- [x] Client tests for `Provide` (also test server)
- [x] Client tests for `ProvidePeer` (also test server)
- [x] Simple server tests for `Provide` output
- [x] Simple server tests for `ProvidePeer` output
- [x] Changelog

Not needed to merge this:

- [ ] `someguy`: https://github.com/ipfs/someguy/pull/40
- [ ] `kubo`: https://github.com/ipfs/kubo/pull/10420
    - Only dependency update, no support for POST per se
- [ ] `ipni`: https://github.com/ipni/index-provider/pull/441


hacdias:
There's still a few tests added, but I'm opening this as ready for review @lidel. It would be useful to go through the code and checking the TODOs that I added inline.

routing/http/client/client.go
```diff
@@ -408,6 +430,50 @@ func (c *Client) FindPeers(ctx context.Context, pid peer.ID) (peers iter.ResultI
...
+// ProvidePeer provides information regarding your own peer, setup with [WithProviderInfo].
+func (c *Client) ProvidePeer(ctx context.Context, ttl time.Duration, metadata []byte) (iter.ResultIter[*types.AnnouncementRecord], error) {
+	if err := c.canProvide(); err != nil {
+		return nil, err
+	}
+
+	record := &types.AnnouncementRecord{
+		Schema: types.SchemaAnnouncement,
+		Payload: types.AnnouncementPayload{
+			// TODO: CID, Scope not present for /routing/v1/peers, right?
```
lidel: Correct. But the fact you had to ask, means everyone implementing it will have to think about too :thinking: 

Should we reconsider and have two separate schemas instead of one (`provider-record` and `peer-record`) ?

routing/http/server/server.go
```diff
@@ -40,10 +40,11 @@ const (
 var logger = logging.Logger("routing/http/server")
 
 const (
-	providePath       = "/routing/v1/providers/"
+	providePath       = "/routing/v1/providers" // TODO: with trailing slash?
 	findProvidersPath = "/routing/v1/providers/{cid}"
+	providePeersPath  = "/routing/v1/peers" // TODO: with trailing slash?
```
lidel: No trailing slash. Updated https://github.com/ipfs/specs/pull/378 to do POST to `/providers and /peers, instead of putting and overriding specific one like we do for IPNS.

routing/http/server/server_test.go
```diff
@@ -126,17 +137,100 @@ func TestProviders(t *testing.T) {
...
+			Timestamp: rec2.Payload.Timestamp,
+			TTL:       rec2.Payload.TTL,
+			ID:        pid2,
+			Addrs:     []multiaddr.Multiaddr{},
+			Protocols: rec2.Payload.Protocols,
+		}).Return(time.Minute, nil)
+
+		urlStr := serverAddr + "/routing/v1/providers"
+
+		httpReq, err := http.NewRequest(http.MethodPut, urlStr, bytes.NewReader(body))
```
lidel: TODO: we need to switch /providers and /peers from PUT to POST to match https://github.com/ipfs/specs/pull/378

routing/http/types/json/requests.go
```diff
@@ -1,51 +1,11 @@
...
-				return err
-			}
-			r.Providers = append(r.Providers, &prov)
-		}
-	}
-	return nil
-}
+// AnnouncePeersRequest is the content of a PUT Peers request.
+// TODO: is the the same? Shouldn't Providers be Peers?
+type AnnouncePeersRequest = AnnounceProvidersRequest
```
lidel: Good catch. It is pretty  awkward. We have shared "announcement" schema for individual records, but the request envelope is be different:
- /providers expects result list in `Providers` field
- /peers expects result list in `Peers` field

@hacdias switch to separate envelope structs here?

routing/http/types/json/requests.go
```diff
@@ -1,51 +1,11 @@
...
-)
-
-// Deprecated: protocol-agnostic provide is being worked on in [IPIP-378]:
-//
-// [IPIP-378]: https://github.com/ipfs/specs/pull/378
-type WriteProvidersRequest struct {
-	Providers []types.Record
+// AnnounceProvidersRequest is the content of a PUT Providers request.
+type AnnounceProvidersRequest struct {
+	Error     string
```
lidel: I think `Error` is only in response, no?

routing/http/types/record_announcement.go
```diff
@@ -0,0 +1,368 @@
...
+	// Make final IPLD node.
+	nd, err := makeIPLDMap(m)
+	if err != nil {
+		return nil, err
+	}
+
+	// Encode it with the DAG-JSON encoder, which automatically sorts the fields
+	// in a deterministic manner.
+	var b bytes.Buffer
+	err = dagjson.Encode(nd, &b)
```
lidel: @hacdias mind switching to DAG-CBOR to match https://github.com/ipfs/specs/pull/378?

routing/http/types/record_announcement.go
```diff
@@ -0,0 +1,368 @@
...
+func (ap AnnouncementPayload) MarshalJSON() ([]byte, error) {
+	m := make(map[string]ipld.Node)
+	var err error
+
+	// TODO: or goes empty? Affects signature.
+	if ap.CID.Defined() {
+		m["CID"] = basicnode.NewString(ap.CID.String())
+	}
+
+	// TODO: is scope not set when empty? Affects signature.
```
lidel: We should skip fields when unset or empty. `""` is a golang idiom, but not that obvious in other languages.

lidel (changes requested):
Pushed some changes to IPIP-378 PR, tldr in https://github.com/ipfs/specs/pull/378#issuecomment-1928601164
Some feedback on TODOs inline.

routing/http/client/client.go
hacdias: @lidel maybe it wouldn't hurt to have two schemas, making them more specific: instead of _Announcement Record_, maybe _Peer Announcement Record_ and _Provider Announcement Record_.

Another question is that I see you introduced the Error Record (Schema). I'm not yet sure how I'll make this work well with the current types we have, but I'll figure it out.

hacdias: @lidel I still think one is fine, I'm not exactly sure why I asked this. As long as it is stated in the specs, it is fine. I think it did not say anything when I wrote this.

I also see you introduced the error schema. That will make the code more complicated. Instead, I want to suggest a `Announcement Response Schema`. It is a mix of the announcement schema and the error schema:

```json
{
  "Schema": "announcement-response",
  "Error": "error in case there was error",
  "TTL": 17280000
}
```

This perhaps allows better coupling between the announcement and its individual response. We can even add the CID and/or PID to have more context.

lidel: @hacdias sounds like a sensible change. Are you able to apply this (remove error and andd this response one) to https://github.com/ipfs/specs/pull/378?

hacdias: @lidel made an update

lidel:
Closing as the IPIP-378 never landed, see https://github.com/ipfs/specs/pull/378#issuecomment-3604361178

PR #569, opened 2024-01-18
gateway: extract CORS to headers middleware

hacdias:
To help with https://github.com/ipshipyard/waterworks-community/issues/4. Sharness won't pass here because Kubo won't build.

Kubo PR: https://github.com/ipfs/kubo/pull/10311

### DIFF
--- a/gateway/headers.go
+++ b/gateway/headers.go
@@ -12,7 +12,7 @@ type Headers struct {
 
 // NewHeaders creates a new [Headers] middleware that applies the given headers
 // to all requests. If you call [Headers.ApplyCors], the default CORS configuration
-// will also be applied.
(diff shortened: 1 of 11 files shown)


gateway/headers.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// Headers is an HTTP middleware that sets the configured headers in all requests.
+type Headers struct {
+	headers map[string][]string
+}
+
+// NewHeaders creates a new [Headers] middleware that applies the given headers
+// to all requests. If you call [Headers.ApplyCors], the default CORS configuration
+// will also be applied.
```
lidel:
```suggestion
// will also be applied, if any of the CORS headers is missing.
```

lidel (approved):

Issue #567, opened 2024-01-16
Difficult to use one session per request (e.g. gateway request)


Jorropo:
## Background

The point of data transfer sessions is to handle the logic around discovering peers that might have the data you need and which peers to ask for which subset of the data. It does this by making assumptions/guesses around the association of various requests (i.e. these 10 block requests are all associated with the same HTTP gateway request). Historically getting the plumbing right here to actually have one session in code per logical session has been a pain resulting in multiple sessions being created within single requests (e.g. gateway requests).

On top of this long standing issue we're also seeing some issues with blockservice wrappers not working nicely with sessions https://github.com/ipfs-shipyard/nopfs/issues/34 due to how the abstraction layers are entangled.

## Options

There have been a few recent and older proposals, most of them involve embedding session information in the context (e.g. https://github.com/ipfs/kubo/issues/7198, #549 although it's incomplete on it's own).

Some of the concrete approaches some of them (along with a brief description) are:

1. https://github.com/ipfs/boxo/pull/563 -> Put the concept of sessions in the blockservice interface
2. https://github.com/ipfs/boxo/pull/561 -> Remove the abstraction of a blockservice interface and whenever someone needs more functionality it means more options passed into the boxo.BlockService struct's constructor
3. https://github.com/ipfs/kubo/issues/7198 -> Make the concept of sessions global, but basically just a uint64, individual layers of the stack (e.g. bitswap or something higher level if that emerges) can then choose to locally associate state with
4. Do nothing (pre #549) -> Make the user create a shallow clone of the stack of components for every session

## Analysis

- Doing nothing/asking the user to create one "stack" per request isn't working
- #561 is less flexible than #563

The main difference between #561 and https://github.com/ipfs/kubo/issues/7198 is a tradeoff between cleanup complexity and adding more propagation of sessions everywhere.
- #563
  - When the context gets cleaned up by Go's GC the session object will get cleaned up as well
  - Sessions need to be handled by: Session consumers (e.g. bitswap), Session creators (e.g. gateway), everything in between (e.g. boxo.blockservice, nopfs, dagservice, fetcher, etc.)
- https://github.com/ipfs/kubo/issues/7198
  - Needs to manually manage and synchronise state (`uint64` → `*session`) in all session consumers
  -  A goroutine would need to get spun up to handle cleanup on context cancellation.
     - Could be a long lived one (as is currently the case in the bitswap implementation) or a short lived one via https://pkg.go.dev/context#AfterFunc  
[...]

PR #549, opened 2024-01-12
blockservice: add `NewSessionContext` and `EmbedSessionInContext`

Jorropo:

PR #570, opened 2024-01-22
session: add new session package

Jorropo:
This replaces #563 and #561.

TODO:
- [ ] Tests
- [ ] Wire bitswap to unwrap the session
- [ ] Revert 22fa8b16a3e530b86542286bebba77c08e17487e and wire the gateway to use this new package.
- [ ] Changelog (@Jorropo will do later)
- [ ] ¿ tracing ?

Questions:
- Right now consumer is a simple `comparable` to ensure it doesn't panic the map, it might be interesting to enforce something like `interface[S any]{ comparable; SessionBadge() S }` because this would enforce at compile time that the session consumer is not mixing different session types together (which would not work).
- Should `GetOrCreate` return `(S, error)` instead of `S, bool` ?
  Passing errors can always be done efficiently through closures as `create` does not leak, the important part is to not save in the map failed sessions.
- Should it be `ContextWithSession(context.Context, ...Option)` ? We might want to add something like a request root in there.

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -1,0 +1,94 @@
+// session implements a finality independent session system. It let session producers transparently integrate with any session consumer without prequired shared state.
+// A session bundles a logical set of related requests, it let download clients more efficiently send requests in the network, the assumption is that the same nodes should host most of the data in the related requests.
+// To create a new session use [ContextWithSession], compatible consumers like bitswap will automatically pick it up from the context.
+package session
+
+import (
+	"context"
+	"sync"
+)
+
+type contextKey struct{}
+
+// ContextWithSession attach a session to the context, it may reuse an existing session if one already exists in the context.
+// The context must be canceled in some way to ensure all sessions are cleaned up.
+func ContextWithSession(ctx context.Context) context.Context {
+	if _, ok := getRegistryFromContext(ctx); ok {
+		return ctx
+	}
+
+	return context.WithValue(ctx, contextKey{}, &sessionRegistry{sesCtx: ctx})
+}
+
+// sessionRegistry is meant for session consumers.
+// This type allows session consumers to inject their own session object within an existing context.
+// This exists as it allows to satisfy theses two requirements:
+// - Session producers must not need to hold some prexisting session factory object to inject into their contextes.
+// - Session consumers must be able to not to maintain any state to track inflight sessions, instead they can store their session object in the sessionRegistry and rely on GC to cleanup.
+// It is threadsafe.
+type sessionRegistry struct {
+	sesCtx   context.Context
+	lk       sync.Mutex // protects sessions
+	sessions map[any]any
+}
+
+func getRegistryFromContext(ctx context.Context) (_ *sessionRegistry, ok bool) {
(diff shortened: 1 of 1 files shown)


Jorropo:
We ultimately went that route because it satisfy both:
- Session producers don't need a pointer to some already shared object (like `blockservice`) to inject a session in their context.
- It permits session consumers which don't need to maintain more state than in the session object to only inject their session in the context instead maintaining their own `uint64 → session` structures for it. This allows such state contained sessions to be cleaned up by the GC automatically.

session/session.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// sessionRegistry is meant for session consumers.
+// This type allows session consumers to inject their own session object within an existing context.
+// This exists as it allows to satisfy theses two requirements:
+// - Session producers must not need to hold some prexisting session factory object to inject into their contextes.
+// - Session consumers must be able to not to maintain any state to track inflight sessions, instead they can store their session object in the sessionRegistry and rely on GC to cleanup.
+// It is threadsafe.
+type sessionRegistry struct {
+	sesCtx   context.Context
```
Stebalien, line 30: Why store this?

Jorropo: For the same reason you originally did in your few years old PR.
Let's assume the consumer does this:
```go
sesCtx, cancel := context.WithCancel(context.Background())
defer cancel()

sesCtx = sessions.ContextWithSession(sesCtx)

var wg sync.WaitGroup
wg.Add(2)
go func() {
 defer wg.Done()
 ctx := context.WithTimeout(sesCtx, time.Second)
 useRequest1(bitswap.GetBlocks(ctx, request1))
}()
go func() {
 defer wg.Done()
 ctx := context.WithTimeout(sesCtx, time.Second*10)
 useRequest2(bitswap.GetBlocks(ctx, request2))
}()
wg.Wait()
```
Which ever one of the two request gets ran first will try to load the session from the session registry, it wont exists, create will be called and at this point within `GetBlocks` create need to use `sesCtx` not `ctx` because `GetBlocks` can have a timeout or cancel policy smaller than the whole session context.

If I passed the local context in the example above, if request one is scheduled first the whole session would be created with a one second timeout, then request two would have a timeout of one second too instead of the 10 it expects.

Jorropo:
```go
	ss, ok := create(r.sesCtx)
```

Stebalien: Oh, I see. Yeah, makes sense.

lidel:
Closing as stale - this has been open for 2 years with incomplete TODOs and no activity since Aug 2025.

The approach here (context-embedded session registry with GC cleanup) seemed sound and was agreed upon in #567. If someone wants to pick this up, the design discussion in #567 may be still
relevant.

For now, we have the partial fix from #549 (`NewSessionContext`/`EmbedSessionInContext`). The nopfs incompatibility (ipfs-shipyard/nopfs#34) remains, but would need this or a similar solution to fully fix.

Happy to reopen (as a new PR) if there's renewed interest in completing this.

Issue #545, opened 2024-01-11
ipns: only use cache for prefetching then do confirmation with actual dht


Jorropo:
I had a report from a user complaining about stale IPNS records.

After trying out a small repro on my machine, it is fairly easy to have 1h+ sync times because we will cache IPNS records for 1h without ever challenging them.

The IPNS code is surprising, it implements rollback based resolution, so it can get bad candidates early, start next recursive resolution, then when it get better candidates it can rollback and reconciliate the previous name resolution with the newer better results (or not if better candidates lines up in the existing timeline).
This allows to optimistically start with bad candidates and validate or cancel and restart with better ones later.

I think the cache should be used as a first untrusted candidate, so when a cache entry is found, we add as the first candidate in the rollback resolution process, but then we proceed with the rollback resolution as usual.

This would also combine nicely with #397 (altho any of the two is still a good improvement)

Jorropo:
For people interested to look at fixing it:
The whole rollback resolution code is based at:
https://github.com/ipfs/boxo/blob/76d9292da780aaa8c1904cc73e3e03d87a46557e/namesys/utilities.go#L34

Through interfaces it then callback into:
https://github.com/ipfs/boxo/blob/76d9292da780aaa8c1904cc73e3e03d87a46557e/namesys/namesys.go#L156

Here is when the cache is checked:
https://github.com/ipfs/boxo/blob/76d9292da780aaa8c1904cc73e3e03d87a46557e/namesys/namesys.go#L175-L181

See `return out`, I think the early return should be removed and the code should continue execution bellow (after writing cached entry to the channel).

lidel:
> it is fairly easy to have 1h+ sync times because we will cache IPNS records for 1h without ever challenging them.

When one publishes IPNS record with `ipfs name publish` the implicit default for caching hint is `--ttl 1h`, 
which seems to align with your observation ("stale" record is cached for ~1h before checking for update). 

What was the TTL of the record you used in your test?
Do you still see "~1h sync problem" if you resolve IPNS name published with `--ttl` set to `1m`?

My mental model is that IPNS caching based on TTL is a feature similar to TTL in DNS records.
Services should leverage TTL for caching, avoid work and only check for updates when TTL window expires.

> I think the cache should be used as a first untrusted candidate, so when a cache entry is found, we add as the first candidate in the rollback resolution process, but then we proceed with the rollback resolution as usual.

[...]

PR #571, opened 2024-01-25
namesys: add WithMaxCacheTTL

hacdias:
Boxo part of #545.

I want to note that this TTL does not apply to DNS **ONLY** because the DNSResolver does not provide TTL values. It's important to note that the Namesys cache is oblivious regarding which resolver is backing it. In practice, it only caches IPNS records.

Kubo https://github.com/ipfs/kubo/pull/10321


lidel (approved):
lgtm, if this global ceiling overrides DNS-specific [`DNS.MaxCacheTTL`](https://github.com/ipfs/kubo/blob/master/docs/config.md#dnsmaxcachettl) we can clarify that in Kubo docs in https://github.com/ipfs/kubo/pull/10321.

hacdias:
> lgtm, if this global ceiling overrides DNS-specific [DNS.MaxCacheTTL](https://github.com/ipfs/kubo/blob/master/docs/config.md#dnsmaxcachettl) we can clarify that in Kubo docs in https://github.com/ipfs/kubo/pull/10321.

It does not because the DNS Resolver does not return the TTL. So that is fine. If it ever happens to return a TTL - consequently caching the DNSLink entries too in Namesys - we can update the docs.

PR #572, opened 2024-01-25
gateway: tests for simpler superfluous namespace handling

### LINKED ISSUE #344: Gateway: `handleSuperfluousNamespace` doesn't work with Hostnames
When the gateway is wrapped with the Hostname Handler, the superfluous namespace redirect doesn't work. This is because HostnameHandler will immediately say that the path is invalid. I suspect this may be happening with other things. It has probably happened before, but we've never noticed.

Solutions:

1. Remove it. Do we really need it?
2. Add it to the hostname handler. Then, we'll be verifying this twice which contains path validation. We already check the URI query twice, but that's cheap: if it doesn't exist, nothing is done. Here we always run the path.IsValid, which costs more.
3. Rework it to make it simpler: just check if URL starts with `/ipfs/ipfs` or `/ipfs/ipns` and do the redirect. Do not do path validation. That will happen after redirection anyways. Then add it to the Hostname handler again.

- [ ] Add conformance test, hint should say it's quality of life improvement.

hacdias:
Closes #344. See comments below.

Kubo: https://github.com/ipfs/kubo/pull/10322

### DIFF
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -577,6 +577,92 @@ func TestRedirects(t *testing.T) {
 		do(http.MethodGet)
 		do(http.MethodHead)
 	})
+
+	t.Run("Superfluous namespace", func(t *testing.T) {
+		t.Parallel()
+
+		backend, root := newMockBackend(t, "fixtures.car")
+		backend.namesys["/ipns/dnslink-gateway.com"] = newMockNamesysItem(path.FromCid(root), 0)
+		backend.namesys["/ipns/dnslink-website.com"] = newMockNamesysItem(path.FromCid(root), 0)
+
+		ts := newTestServerWithConfig(t, backend, Config{
+			NoDNSLink: false,
+			PublicGateways: map[string]*PublicGateway{
+				"dnslink-gateway.com": {
+					Paths:                 []string{"/ipfs", "/ipns"},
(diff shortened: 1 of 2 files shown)


lidel:
I did not look too deep, but I think this breaks valid content paths like `/ipfs/ipfs/file.txt` in places like subdomains – see the failing  test of `{{scheme}}://{{cid}}.ipfs.{{host}}/ipfs/ipns/` at https://github.com/ipfs/gateway-conformance/blob/193833b91f2e9b17daf45c84afaeeae61d9d7c7e/tests/subdomain_gateway_ipfs_test.go#L244-L259


Not sure how feasible, but it may be easier to move handleSuperfluousNamespace check from very front of the request processign pipeline, to very end: trigger it on "not found" code path (e.g. the place where we look for `_redirects` or produce HTTP 404 response)

hacdias:
@lidel I have actually realised #344 is actually not an issue. When I initially tested, they didn't redirect because I hadn't set the `Paths` field in the configuration 🤦 

Nevertheless, we can use this PR as a simplification and also a bunch of tests.

hacdias:
@lidel it seems that we tested this in Kubo too. See here: https://github.com/ipfs/kubo/pull/10322 too.

lidel (approved):
lgtm, thank you for cleaning this up!

Issue #186, opened 2023-02-25
Broken WASM builds


hsanjuan:
Using latest versions of things I discovered that WASM builds are now broken.

```
# github.com/ipfs/go-libipfs/files
../../../../pkg/mod/github.com/ipfs/go-libipfs@v0.6.0/files/filewriter.go:25:13: undefined: createNewFile
../../../../pkg/mod/github.com/ipfs/go-libipfs@v0.6.0/files/filewriter.go:47:6: undefined: isValidFilename
```

I would normally be able to just use an older version of go-ipfs-files, but it was deprecated and now dependencies are only pulling the broken code present in libipfs.

I cannot think of a workaround in terms of dependency management that can fix this, is there one?

aschmahmann:
> I cannot think of a workaround in terms of dependency management that can fix this, is there one?

Off the top of my head Go vendoring almost always lets you fix the problem. However, given how small the fix is here I would just make that patch rather than trying to use an older version of the library that has other bugs in it.

WASM builds of this library aren't supported, but (as with other unsupported platforms) fixes are definitely welcome. If you'd like to fix the issue you can do so in a similar manner to https://github.com/ipfs/go-ipfs-files/pull/50.

lidel:
Triage notes:
- Maintainers are concerned how future-proof this is, and IF we want to officially support WASM target  
  - We need a policy on WARM.
    - If we are to merge https://github.com/ipfs/boxo/pull/573, we need to officially support wasm builds, and if we do that, we need to have regression tests  / CI build that confirms WASM target always works on PRs that add new things.

PR #573, opened 2024-01-28
Make GOOS=js GOARCH=wasm a possible build target

bahner:
As writing files is not really applicable in WebAssembly create a noop filewriter to make compilation possible.

### DIFF
--- a/files/filewriter_unix_test.go
+++ b/files/filewriter_unix_test.go
@@ -1,4 +1,6 @@
-//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || (js && !wasm) || wasip1
+// +build darwin linux netbsd openbsd freebsd dragonfly js,!wasm wasip1
+
 
 package files
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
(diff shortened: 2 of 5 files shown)


bahner:
Incidentally, I believe this closes issue #186

bahner:
@Jorropo when you take a look at this, then feel free to suggest actual writing to in-memory files. I just don't see any reason, but if there is then just holler.

Jorropo:
I would like to support WASM too but I want to you know that you havn't added anything to the test suite.
If we don't test WASM in CI I could merge it but with expectations that it's not maintained, experimental and could break at any moment.

Jorropo (changes requested):
I don't think a new `filewriter_wasm` file is correct.
The STD emulates a unix filesystem on top of JS:
https://cs.opensource.google/go/go/+/master:src/os/file_unix.go;l=5?q=openFileNolog&ss=go%2Fgo

files/filewriter_unix_test.go
```diff
@@ -1,4 +1,6 @@
-//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || js || wasip1
+//go:build darwin || linux || netbsd || openbsd || freebsd || dragonfly || (js && !wasm) || wasip1
+// +build darwin linux netbsd openbsd freebsd dragonfly js,!wasm wasip1
```
Jorropo, line 2: Btw, it's been since 1.18 that `+build` isn't used anymore.
I think you should upgrade your version of go. :wink: :slightly_smiling_face:

bahner:
Hepp! Good feedback. A little busy here, but I'll follow up.

lidel:
Superseded by #935

PR #577, opened 2024-02-12
examples: verified CAR file fetcher

hacdias:
This adds a simple example on how you can use Boxo to verifiably download a file from a trustless gateway. It does not cover the case where we have directories. I think having this, albeit not supporting directories, is better than nothing. With time, we should convert this into its own library module.

Based on #594 to reuse some code.


examples/verified-fetch/fetcher.go
```diff
@@ -0,0 +1,271 @@
...
+	return nil
+}
+
+func (f *fetcher) resolvePath(ctx context.Context, p path.Path) (path.ImmutablePath, error) {
+	for p.Mutable() {
+		// Download IPNS record and verify through the gateway, or resolve the
+		// DNSLink with the default DNS resolver.
+		name, err := ipns.NameFromString(p.Segments()[1])
+		if err == nil {
+			p, err = f.resolveIPNS(ctx, name)
```
aschmahmann: What about `/ipns/myKey/subpath` seems like we skip the subpath?

examples/verified-fetch/fetcher.go
```diff
@@ -0,0 +1,271 @@
...
+		} else {
+			p, err = f.resolveDNSLink(ctx, p)
```
aschmahmann: Neither of these handle recursion (e.g. DNSLink -> IPNS being the most common, but other nesting is also valid).

Maybe it'd be easier here to use the namesys module but stub out IPNS resolution by using a "router" that's backed by a gateway rather than a DHT. Likely depends on if the idea of this example to show people some of the guts around gateway usage + verification or how to assemble boxo components to do the job even when we've stripped out all of the p2p networking layer.

examples/verified-fetch/fetcher.go
```diff
@@ -0,0 +1,271 @@
...
+	if f.limit > 0 {
+		rc = newLimitReadCloser(resp.Body, f.limit)
+	} else {
+		rc = resp.Body
+	}
+
+	return rc, nil
+}
+
+func carToFileStream(ctx context.Context, r io.ReadCloser, imPath path.ImmutablePath) (io.ReadCloser, error) {
```
aschmahmann: This doesn't do incremental verification from the gateways which might be fine, it depends what we're trying to do with this example. So not necessarily a blocker here but we should clarify what the plan is.

If we were trying to make this a library we should be able to turn on (or have by default) incremental verification. Code that does similar work should exist in #587, but at a high level:
1. If using DFS + duplicates: Have a blockstore that pulls data from the stream on demand and validates the multihash is the one being asked for, else it fails
2. If using DFS + no duplicates: Have a blockstore that pulls validated data from the stream (or the existing blockstore), if the requested block is neither the next block in the stream nor local then fail
3. If type is unknown either fail, or guess as 2

examples/verified-fetch/fetcher.go
```diff
@@ -0,0 +1,271 @@
...
+// withLimit sets the limit for the [Fetcher].
+func withLimit(limit int64) fetcherOption {
+	return func(f *fetcher) error {
+		f.limit = limit
+		return nil
+	}
+}
+
+// newFetcher creates a new [Fetcher]. Setting the gateway is mandatory.
+func newFetcher(gateway string, options ...fetcherOption) (*fetcher, error) {
```
aschmahmann: What's the reason we're doing this as a test rather than a library component? e.g. we don't have any consumers immediately setup to use it, not enough time to productionize it yet, trying to show people the guts more than give them the tools, etc.

aschmahmann (changes requested):

examples/verified-fetch/fetcher.go
hacdias: We can replace this by a NameSystem with the `RemoteValueStore` from https://github.com/ipfs/boxo/pull/594: 
https://github.com/ipfs/boxo/blob/2ee63056241a09dd7dae603419ab5461a824ad8e/gateway/value_store.go

examples/verified-fetch/fetcher.go
hacdias: If we were to merge #587 before, it would actually be easy to reuse the graph gateway backend: you could just instantiate it and call `GetCAR`, or even `Get` and stream directly to the file system. Leeching from the gateway code, but would do exactly what we want.

examples/verified-fetch/fetcher.go
hacdias: @aschmahmann well, this isn't a test. It's an example. But if we want to make it a part of Boxo itself, then we need to improve it massively and allow it to work with UnixFS directories too and not only files.

examples/verified-fetch/fetcher.go
lidel: @hacdias https://github.com/ipfs/boxo/pull/587 landed, would you be able to refactor this example to use that, and have incremental verification (if so, fetch my changes before you start)?

If this is too much before 29th, let's park this and convert this example to a draft. 
Adin has a point, if we publish this as an example, people will be blindly doing this, and have no incremental verification in their production apps.

hacdias: @lidel in retrospective, I'm not sure if we have the best code to do this at the moment. Using `NewRemoteCarBackend` would make it very simple to resolve the given IPFS Path and downloading a CAR file. The problem then becomes extracting the CAR file. We don't have any good library to do it unfortunately. For that, I would need to resort to basically copying all of the code in this file: https://github.com/ipld/go-car/blob/master/cmd/car/extract.go

Or... we can make it such that this example simply downloads a CAR file and we can suggest the `go-car` CLI tool to then unpack the file. That would make it much simpler. What do you think?

hacdias: Yes! Will do :)

hacdias:
@lidel @aschmahmann I'm mesmerised by how short this code snippet is, and it does more than what we had before. It is stupidly simple and works for both regular files and directories. We already had all the necessary pieces in Boxo. No need to be meddling with `go-car` for this example.

examples/car-file-fetcher/main.go
```diff
@@ -0,0 +1,121 @@
...
+	// backend gateway, this call will internally fetch a CAR file from the remote
+	// gateway and ensure that all blocks are present and verified.
+	_, file, err := backend.GetAll(context.Background(), imPath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	// Write the returned UnixFS file or directory to the file system.
+	return files.WriteTo(file, outputPath)
```
lidel, line 85: :thought_balloon:  sweet!

lidel (approved):
Thank you @hacdias, hiding all complexity in behind boxo APIs is really nice!

PR #579, opened 2024-02-15
routing/http/client: avoid race by not using global http.Client

hacdias:
Fixes the race condition observed in the tests in https://github.com/ipfs/rainbow/pull/87.

It could've been fixed in rainbow, but that doesn't fix the source issue: we use the same global default HTTP client, while giving an option that can override its user agent. This option changes the user agent of the global variable, which may be used by multiple clients.

In addition, I updated the tests and documentations to reflect how `WithUserAgent` and `WithHTTPClient` work. The tests were passing until now by mistake and only because we were using the **global** `defaultHTTPClient `.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The following emojis are used to highlight certain changes:
 
 - 🛠 `boxo/gateway`: when making a trustless CAR request with the "entity-bytes" parameter, using a negative index greater than the underlying entity length could trigger reading more data than intended
 - 🛠 `boxo/gateway`: the header configuration `Config.Headers` and `AddAccessControlHeaders` has been replaced by the new middleware provided by `NewHeaders`.
+- 🛠 `routing/http/client`: a potential race condition has been fixed when creating multiple clients within the same process. With this change, `WithHTTPClient` and `WithUserAgent` will no longer work with each other. If you use your custom HTTP Client, you are also responsible for configuring the user agent that it uses.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ The following emojis are used to highlight certain changes:
 
 - 🛠 `boxo/gateway`: when making a trustless CAR request with the "entity-bytes" parameter, using a negative index greater than the underlying entity length could trigger reading more data than intended
 - 🛠 `boxo/gateway`: the header configuration `Config.Headers` and `AddAccessControlHeaders` has been replaced by the new middleware provided by `NewHeaders`.
-- 🛠 `routing/http/client`: a potential race condition has been fixed when creating multiple clients within the same process. With this change, `WithHTTPClient` and `WithUserAgent` may not always work with each other. More details can be found in the documentation.
+- 🛠 `routing/http/client`: the default HTTP client is no longer a global singleton. Therefore, using `WithUserAgent` won't modify the user agent of existing routing clients. This will also prevent potential race conditions. In addition, incompatible options will now return errors instead of silently failing.
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


routing/http/client/client.go
```diff
@@ -79,12 +82,18 @@ func WithIdentity(identity crypto.PrivKey) Option {
 	}
 }
 
+// WithHTTPClient sets a custom HTTP Client to be used with [Client]. If you use
+// this option, [WithUserAgent] will no longer work, as it is made to work with
+// the default HTTP Client.
```
aschmahmann: Is this true? Seems like setting the user agent will work as long as the implementation `httpClient` is `*http.Client` and it has a `ResponseBodyLimitedTransport`.

routing/http/client/client.go
```diff
@@ -79,12 +82,18 @@ func WithIdentity(identity crypto.PrivKey) Option {
...
 func WithHTTPClient(h httpClient) Option {
 	return func(c *Client) {
 		c.httpClient = h
 	}
 }
 
+// WithUserAgent sets a custom user agent to use with the HTTP Client. This will
+// only work if you use the default HTTP Client. If you use [WithHTTPClient], you
+// should set the user agent yourself.
```
aschmahmann: Either this should only work if `WithHTTPClient` isn't set or it should clarify that "default HTTP Client" means `http.Client`  (with a ResponseBodyLimitedTransport transport) rather than the default one for the routing client.

routing/http/client/client.go
```diff
@@ -79,12 +82,18 @@ func WithIdentity(identity crypto.PrivKey) Option {
...
+// this option, [WithUserAgent] will no longer work, as it is made to work with
+// the default HTTP Client.
 func WithHTTPClient(h httpClient) Option {
 	return func(c *Client) {
 		c.httpClient = h
 	}
 }
 
+// WithUserAgent sets a custom user agent to use with the HTTP Client. This will
+// only work if you use the default HTTP Client. If you use [WithHTTPClient], you
```
aschmahmann: Most of the time we use options the options are allowed to return errors, for some reason that isn't the case here. Given there likely aren't (m)any options implemented not in this package WDYT about changing Option to return errors and then having `WithUserAgent` error rather than failing silently if the types are wrong?

aschmahmann (changes requested):

hacdias:
@aschmahmann I updated the options to return an error, and simplified the documentation of the options.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The following emojis are used to highlight certain changes:
 
 - 🛠 `boxo/gateway`: when making a trustless CAR request with the "entity-bytes" parameter, using a negative index greater than the underlying entity length could trigger reading more data than intended
 - 🛠 `boxo/gateway`: the header configuration `Config.Headers` and `AddAccessControlHeaders` has been replaced by the new middleware provided by `NewHeaders`.
+- 🛠 `routing/http/client`: a potential race condition has been fixed when creating multiple clients within the same process. With this change, `WithHTTPClient` and `WithUserAgent` may not always work with each other. More details can be found in the documentation.
```
aschmahmann: > With this change, `WithHTTPClient` and `WithUserAgent` may not always work with each other. More details can be found in the documentation.

Wasn't this always the case? If you called `WithHTTPClient` and used a non-compatible client `WithUserAgent` would be silently ignored. AFAICT the changes are:
1. Incompatible options will now return errors rather than silently failing
2. The default HTTP client is no longer a singleton and therefore passing `WithUserAgent` won't modify the user agent of existing routing clients with default http clients.

routing/http/client/client.go
```diff
@@ -67,53 +60,72 @@ var defaultUserAgent = moduleVersion()
...
+// WithHTTPClient sets a custom HTTP Client to be used with [Client].
 func WithHTTPClient(h httpClient) Option {
-	return func(c *Client) {
+	return func(c *Client) error {
 		c.httpClient = h
+		return nil
 	}
 }
 
+// WithUserAgent sets a custom user agent to use with the HTTP Client. This only
```
aschmahmann: WDYT about specifically noting that this modifies the underlying http.Client (i.e. so that people don't try and pass the same HTTP Client to two routing clients with different user agents and get confused as to why both user agents end up as the second one)

aschmahmann (commented):
Code changes look good, left some documentation suggestions/clarifications

PR #585, opened 2024-03-05
fix(routing/http): support lookups with legacy peerid notation

lidel:
### Summary

This PR adds support for legacy Base58-encoded PeerIDs on `/routing/v1/peers/{peerid}` endpoint.

### Rationale 

Robustness and consistency.

We already return base58 in responses, but we did not support base58 notation as input, hoping this would force initial implementers to do the right thing. But returning error here likely hurts routing in general, as someone may learn about peer  using legacy ID, do a naive lookup, fail due to using old encoding,  and then assume the peer is offline.

Another reason: [libp2p specs here](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation) state implementations MUST support both, 
it is also better for end users if we are liberal in inputs and support both notations: 

Ref. 

- https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- 🛠️`routing/http/server`: delegated peer routing endpoint now supports both [PeerID string notaitons from libp2p specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation).
+
 ### Security
 
 ## [v0.18.0]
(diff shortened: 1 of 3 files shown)


hacdias (approved):

lidel:
For the record, anothr bug fixed by this PR is case when a legacy RSA key is used:

```console
$ curl https://delegated-ipfs.dev/routing/v1/peers/QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC
unable to parse peer ID: can't convert CID of type "dag-pb" to a peer ID
```

PR #586, opened 2024-03-06
ipns: fix test by using global random reader

hacdias:
I think this may fix the test that was intermittently failing on Windows. By using the global `crypto/rand.Reader` we avoid creating two different random generators seeded with the same time, and therefore likely generating the same key.


hacdias:
This is a relatively small change. I will merge in hopes this will indeed fix the issue. We shall see.

PR #587, opened 2024-03-07
gateway: remote car gateway backend

### LINKED ISSUE #576: gateway:  remote backend implementations from bifrost-gateway
## About gateway backends

Right now, `boxo/gateway` comes with only one implementation of local backend (`BlocksBackend` in  [`gateway/blocks_backend.go`](https://github.com/ipfs/boxo/blob/main/gateway/blocks_backend.go)).

Remote backends that follow https://specs.ipfs.tech/http-gateways/trustless-gateway/ as data transfer protocol were created in 2023 as part of Project Rhea, but are hard to discover and use outside the no longer actively maintained [bifrost-gateway](https://github.com/ipfs/bifrost-gateway).

## Proposed improvement

We should salvage that work and make these backends useful to boxo users by moving remote backend implementations from `biforst-gateway` to `boxo/gateway`

Namely:
- `remote_blocks_backend.go` (porting https://github.com/ipfs/bifrost-gateway/blob/main/blockstore_proxy.go)  
- `remote_car_backend.go`  (porting https://github.com/ipfs/bifrost-gateway/blob/main/lib/graph_gateway.go)

This will not only benefit boxo users, but also allow us to add these backends to `rainbow` (https://github.com/ipfs/rainbow/issues/88) as alternative to libp2p one, allowing us to sunset  and archive [bifrost-gateway](https://github.com/ipfs/bifrost-gateway).

hacdias:
Closes #576. Generally, this imports the graph gateway, here named `CarBackend`, from our old [bifrost-gateway](https://github.com/ipfs/bifrost-gateway) repository. The following updates were also done:

- Run Conformance Tests against the Remote Car Gateway
- Re-use options between `CarBackend` and `BlocksBackend`
- Update `CarBackend` such that CARs return useful blocks when paths are not not resolvable

Depends on:

- https://github.com/ipfs/gateway-conformance/pull/202


hacdias:
@aschmahmann there are quite some tests failing for _Gateway Conformance (Graph Gateway)_. I've already fixed some tests but maybe you have a better intuition for what might be failing here. I had to update the code quite a bit considering all the updates we've had in Boxo.

gateway/backend_graph_test.go
```diff
@@ -0,0 +1,1077 @@
...
+	carv2 "github.com/ipld/go-car/v2"
+	carbs "github.com/ipld/go-car/v2/blockstore"
+	"github.com/ipld/go-car/v2/storage"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed testdata/directory-with-multilayer-hamt-and-multiblock-files.car
+var dirWithMultiblockHAMTandFiles []byte
+
+func TestGraphBackendTar(t *testing.T) {
```
hacdias: @aschmahmann why was this named `TestTar`?

gateway/backend_graph_test.go
```diff
@@ -0,0 +1,1077 @@
...
+	assertNextEntryNameEquals(t, hamtDirIter, "exampleC-hamt-collide-exampleA-seed-52")
+	assertFileEqual(t, "bafkreidqhbqn5htm5qejxpb3hps7dookudo3nncfn6al6niqibi5lq6fee", hamtDirIter.Node().(files.File))
+
+	assertNextEntryNameEquals(t, hamtDirIter, "exampleA")
+	assertFileEqual(t, "bafybeigcisqd7m5nf3qmuvjdbakl5bdnh4ocrmacaqkpuh77qjvggmt2sa", hamtDirIter.Node().(files.File))
+
+	require.False(t, rootDirIter.Next() || basicDirIter.Next() || hamtDirIter.Next())
+}
+
+func TestGraphBackendTarAtEndOfPath(t *testing.T) {
```
hacdias: @aschmahmann why was this named `TestTarAtEndOfPath`?

gateway/backend_graph.go
lidel, line 1: nit: remove "graphs" from file and and types, will cause problems in future if we have other means of fetching graphs. 

This gateway backend is specific to CARs, let's call it `backend_cars.go` as that is the primary  transport :pray:

hacdias:
We have now 10 failing conformance tests with the remote CAR gateway backend. I've managed to solve the other ones in the meanwhile.

## `Cache-Control: only-if-cached`

This are currently failing because the Car Gateway does not have a local blockstore and does not cache anything. Therefore, we always return `false` to `IsCached`.

I think we should disable this test for this backend.

## Range requests for DAG-CBOR converted to HTML

And this is a very interesting one. For the range requests, we initially do a request without a range and then accept either the correct range, the first range, or the whole content as response. But here's where it gets tricky!

When converting to HTML, Boxo does some magic to nicely display DAG-CBOR and DAG-JSON. When we have the whole data, we do it:

image

However, the Car Backend fetches a CAR with only the necessary blocks from the range request. Therefore, it does not have the necessary data to pretty print the DAG into HTML format:

image

And this makes the test fail (content of range request does not match range, nor the whole request).

I don't have a clear solution here: we could potentially try doing resolution first, and then acting depending on the codec of the CID. If the codec is not dag-pb and request accepts `text/html` (information currently not passed into `.Get`), then fetch everything without ranges. But this creates another problem: we'd be doing two requests _and_ potentially more data than necessary for exotic formats.

Wdyt? I temporarily solved it with [6fe39c8](https://github.com/ipfs/boxo/pull/587/commits/6fe39c8c29e09cad5238961a4e2f0b19a76edbfc) by adding a small exception.

cc @aschmahmann @lidel

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,45 @@
+# Gateway as Proxy for Trustless CAR Remote Backend
+
+This is an example of building a Gateway that uses `application/vnd.ipld.car`
+responses from another gateway acting as a remote Trustless Gateway and IPNS resolver.
+
```
lidel, line 10:
```suggestion

**NOTE:** the CAR backend MUST implement [IPIP-0402: Partial CAR Support on Trustless Gateways](https://specs.ipfs.tech/ipips/ipip-0402/)

```

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,45 @@
...
+> go build -o gateway
+```
+
+## Usage
+
+First, you need a compliant gateway that supports both [CAR requests](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) and IPNS Record response
+types. Once you have it, run the proxy gateway with its address as the host parameter:
+
+```
+./gateway -g https://ipfs.io -p 8040
```
lidel:
```suggestion
./gateway -g https://trustless-gateway.link -p 8040
```

examples/gateway/proxy-blocks/README.md
```diff
@@ -28,7 +28,7 @@ types. Once you have it, run the proxy gateway with its address as the host para
 
 
 ```
-./verifying-proxy -g https://ipfs.io -p 8040
+./gateway -g https://ipfs.io -p 8040
```
lidel:
```suggestion
./gateway -g https://trustless-gateway.link -p 8040
```

examples/gateway/proxy-car/main.go
```diff
@@ -0,0 +1,50 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	gatewayUrlPtr := flag.String("g", "", "gateway to proxy to")
+	port := flag.Int("p", 8040, "port to run this gateway from")
+	flag.Parse()
+
+	// Setups up tracing. This is optional and only required if the implementer
+	// wants to be able to enable tracing.
+	tp, err := common.SetupTracing(ctx, "Graph Gateway Example")
```
lidel:
```suggestion
	tp, err := common.SetupTracing(ctx, "CAR Gateway Example")
```

examples/gateway/proxy-car/main.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Fatal(err)
+	}
+	defer (func() { _ = tp.Shutdown(ctx) })()
+
+	carFetcher, err := gateway.NewRemoteCarFetcher([]string{*gatewayUrlPtr})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Creates the gateway with the remote graph backend.
```
lidel:
```suggestion
	// Creates the gateway with the remote car (IPIP-402) backend.
```

lidel (commented):
Thank you for salvaging this work @hacdias :heart: 

### On only-if-cached

Disabling tests for backend_car" for now sgtm. 

<details>
  <summary>💭 Click for Rationale</summary>

In theory, we could wire it up to in-memory cache, and pre-warm the cache so it passes the test, but it feels like a time sink not worth the effort.

In practice, `Cache-Control: only-if-cached` woudl be handled by CDN/cache in front of gateway, so it is reasonable for gateway backed by a remote backend to always return [HTTP 412](https://specs.ipfs.tech/http-gateways/path-gateway/#412-precondition-failed). HTTP 200s will be returned only by the cache that sits in front of it.

</details>

--- 

### On Range requests for DAG-CBOR converted to HTML

Thank you for surfacing this one. Took me a moment to understand what is happening here.

The way I see it, the conformance test of (dag-cbor + `Accept` with `text/html` + `Range`) in [path_gateway_dag_test.go#L718-L731](https://github.com/ipfs/gateway-conformance/blob/a95b2db5650b0b718cafcba9878e78714d49ea4d/tests/path_gateway_dag_test.go#L718-L731) makes no sense   conformance-wise, and should be removed from conformance suite itself, and remove any hacks we added to make it pass.

Lmk if  below missed the mark, but I think removing this edge case is the best way of handling it.

<details>
  <summary>💭 Click for (long) Rationale why</summary>

Making HTTP `Range` request with `Accept` that includes `text/html` make little sense.
The only meaningful `Range` tests for `dag-cbor` are ones that  fetch subset of  CBOR  bytes.
So if `Accept` is missing,  `*/*` or `application/vnd.ipld.dag-cbor` `application/vnd.ipld.cbor`.
The case where `Accept` is `text/html` is not meaningful.

This HTML is generated, getting subset of bytes of generated HTML preview page, while technically possible for HTTP client, is.. meh? Not useful at all, not related to receiving actual content-addressed data.

It would be a-okay regression test in Kubo, but should not be part of conformance.

I think the reason it ended up in conformance tests is that our approach to conformance was to play it safe and "better to test too much than too little". Which is good, and I still think was a good call.

When it comes to conformance, we have two opposite problems:

1. We test Kubo-specific behaviors. As we have implementations, we identify these, and over time relax, make generic, or remove.
2. We have no tests for implicit behaviors related to generic HTTP due to always using `net/http` from Golang
   - For example: for a long time we had HTTP Ranges working everywhere "for  free" because `net/http` from golang handled things like `Range` and `Content-Length` seamlessly.  But since `gateway/serve_http_content.go` was introduced in https://github.com/ipfs/boxo/pull/369 we do all these things by hand, and only where it brings any value. This created a gaps, for example, we are lacking some of basic Range request tests (https://github.com/ipfs/gateway-conformance/issues/154)

Hopefully this is useful context.
Over time, conformance tests will improve. The more we use them, the better.
</details>

hacdias:
@lidel great, I do agree. I opened a PR to remove the conformance test (https://github.com/ipfs/gateway-conformance/pull/202) and I will revert the change here that I made in order to pass such test. We'll need to make a release of the conformance tests once merged.

hacdias:
Besides removing the conformance test (PR [here](https://github.com/ipfs/gateway-conformance/pull/202)), this PR is ready to be reviewed. I just made a few cleanups, so here's a short summary of the most important:

- You can now override the `http.Client` used by remote fetchers where we were always using the default `newRemoteHTTPClient`.
- You can customize the `GetFetchTimeout` in the `CarBackend`
- I made it clear that the `CarBackend` is not strictly remote. Just like the `BlocksBackend`, we can use it with different `CarFetcher`s.
- I made the utility `NewRemoteCarBackend` that creates a `CarBackend` with a `RemoteCarFetcher`.
- I updated te `CarFetcher` interface to be more generic (path + CarParams) instead of the URL directly, making it more flexible.

hacdias:
Note that there is one test that is "expected" [by github actions] (`gateway-conformance`). This needs to be updated in the configuration of the repository after merging this PR, since we updated the name of the test.

lidel:
@hacdias thanks!

FYSA I've removed "gateway-conformance" workflow from required list for now.

Since `boxo/gateway` is de facto reference implementation of what is possible, `gateway-conformance.yml` workflow from this repo will be read by people who want to see prior art, learn how to integrate conformance tests into their projects. 
  
I took a stab at cleaning it up + adding the missing test for "remote block backend" (where blocks are fetched one by one via `?format=raw`) in https://github.com/ipfs/boxo/pull/587/commits/2ce7502f2b7b3af9eed59aa1c5958243a5d1d508. 

(I've run out of time, will try to finish reviewing this PR tomorrow)

gateway/blockstore.go
```diff
@@ -151,18 +158,23 @@ type remoteBlockstore struct {
...
-// specification for more details.
+// or more gateways that support [RAW block] requests. See the [Trustless Gateway]
+// specification for more details. You can optionally pass your own [http.Client].
 //
 // [Trustless Gateway]: https://specs.ipfs.tech/http-gateways/trustless-gateway/
-func NewRemoteBlockstore(gatewayURL []string) (blockstore.Blockstore, error) {
+// [RAW block]: https://www.iana.org/assignments/media-types/application/vnd.ipld.raw
+func NewRemoteBlockstore(gatewayURL []string, httpClient *http.Client) (blockstore.Blockstore, error) {
 	if len(gatewayURL) == 0 {
 		return nil, errors.New("missing gateway URLs to which to proxy")
```
lidel, line 168:
```suggestion
		return nil, errors.New("missing gateway URLs to which to proxy block (application/vnd.ipld.raw) requests")
```

gateway/errors.go
```diff
@@ -127,6 +129,37 @@ func (e *ErrorStatusCode) Unwrap() error {
...
+type CarResource struct {
+	Path   path.ImmutablePath
+	Params CarParams
+}
+
+func (epr ErrPartialResponse) Error() string {
+	if epr.error != nil {
+		return fmt.Sprintf("partial response: %s", epr.error.Error())
+	}
+	return "caboose received a partial response"
```
lidel, line 160:
```suggestion
	return "received a partial CAR response from the backend"
```

gateway/errors.go
```diff
@@ -127,6 +129,37 @@ func (e *ErrorStatusCode) Unwrap() error {
 	return e.Err
 }
 
+// ErrInvalidResponse can be returned from a [DataCallback] to indicate that the data provided for the
+// requested resource was explicitly 'incorrect' - that blocks not in the requested dag, or non-car-conforming
+// data was returned.
+type ErrInvalidResponse struct {
```
lidel, line 135:
```suggestion
// ErrInvalidResponse can be returned from a [DataCallback] to indicate that the data provided for the
// requested resource was explicitly 'incorrect', for example, when  received blocks did not belong to the requested dag, or non-car-conforming
// data was returned.
type ErrInvalidResponse struct {
```

go.mod
```diff
@@ -30,6 +31,7 @@ require (
 	github.com/ipfs/go-metrics-interface v0.0.1
 	github.com/ipfs/go-peertaskqueue v0.8.1
 	github.com/ipfs/go-unixfsnode v1.9.0
+	github.com/ipld/go-car v0.6.2
 	github.com/ipld/go-car/v2 v2.13.1
```
lidel, line 35: :thought_balloon: i guess we have to live with both now.

iiuc we need v1 for `WithErrorOnEmptyRoots` which was introduced in https://github.com/ipld/go-car/pull/461, and v2 does not support everything v1 does, and v3 mentioned in https://github.com/ipld/go-car/pull/461#issuecomment-1663273499 never happened.

examples/README.md
```diff
@@ -27,6 +27,7 @@ Once you have your example finished, do not forget to run `go mod tidy` and addi
 ## Examples and Tutorials
 
 - [Fetching a UnixFS file by CID](./unixfs-file-cid)
-- [Gateway backed by a CAR file](./gateway/car)
-- [Gateway backed by a remote blockstore and IPNS resolver](./gateway/proxy)
+- [Gateway backed by a CAR file](./gateway/car-file)
```
lidel, line 30:
```suggestion
- [Gateway backed by a local blockstore in form of a CAR file](./gateway/car-file)
```

examples/README.md
```diff
@@ -27,6 +27,7 @@ Once you have your example finished, do not forget to run `go mod tidy` and addi
...
+- [Gateway backed by a remote blockstore and IPNS resolver](./gateway/proxy-blocks)
+- [Gateway backed by a remote Trustless CAR Gateway](./gateway/proxy-car)
```
lidel, line 32:
```suggestion
- [Gateway backed by a remote CAR Gateway](./gateway/proxy-car)
```

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,47 @@
...
+```
+./gateway -g https://trustless-gateway.link -p 8040
+```
+
+### Subdomain gateway
+
+Now you can access the gateway in [localhost:8040](http://localhost:8040). It will
+behave like a regular [Subdomain IPFS Gateway](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#subdomain-gateway),
+except for the fact that it runs no libp2p, and has no local blockstore.
+All contents are provided by a remote  gateway and fetched as CAR files and IPNS Records, and verified locally.
```
lidel, line 28:
```suggestion
Now you can access the gateway at [`localhost:8040`](http://localhost:8040). It will
behave like a regular [Subdomain IPFS Gateway](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#subdomain-gateway),
except for the fact that it runs no libp2p, and has no local blockstore.
All data is provided by a remote trustless gateway, fetched as CAR files and IPNS Records, and verified locally.
```

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Path gateway
+
+If you don't need Origin isolation and only care about hosting flat files,
+a plain [path gateway](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#path-gateway) at [127.0.0.1:8040](http://127.0.0.1:8040)
```
lidel, line 33:
```suggestion
a plain [path gateway](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#path-gateway) at [127.0.0.1:8040](http://127.0.0.1:8040/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi)
```

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,47 @@
...
+First, you need a compliant gateway that supports both [CAR requests](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) and IPNS Record response
+types. Once you have it, run the proxy gateway with its address as the host parameter:
+
+```
+./gateway -g https://trustless-gateway.link -p 8040
+```
+
+### Subdomain gateway
+
+Now you can access the gateway in [localhost:8040](http://localhost:8040). It will
```
lidel, line 25:
```suggestion
Now you can access the gateway in [`localhost:8040`](http://localhost:8040/ipfs/bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze). It will
```

examples/gateway/proxy-car/README.md
```diff
@@ -0,0 +1,47 @@
+# Gateway as Proxy for Trustless CAR Remote Backend
+
+This is an example of building a Gateway that uses `application/vnd.ipld.car`
+responses from another gateway acting as a remote Trustless Gateway and IPNS resolver.
```
lidel, line 4:
```suggestion
This is an example of building a "verifying proxy" Gateway that has no blocktore, but instead uses `application/vnd.ipld.car` and `application/vnd.ipfs.ipns-record`
responses from a remote HTTP server that implements CAR support from [Trustless Gateway Specification](https://specs.ipfs.tech/http-gateways/trustless-gateway/).
```

lidel (approved):
Thank you for porting and cleaning this up @hacdias!

I think it is good enough: we have examples that show how to use it, and we dogfood both backends in conformance tests, which is a neat way to ensure they don't break without notice.


Lgtm, let's merge it. 


For anyone following this work: the next step will be to expose support for block and car proxy modes in rainbow (https://github.com/ipfs/rainbow/issues/88).

PR #592, opened 2024-03-22
feat: basic support for Ipfs-Path-Affinity from IPIP-462

lidel:
This PR aims to kick-off discussion how we would support [IPIP-462](https://github.com/ipfs/specs/pull/462) in `boxo/gateway`. 

For more info and header semantics, see https://github.com/ipfs/specs/pull/462

## proposed scope

What I want to do for now is to add minimal code to start leveraging `Ipfs-Path-Affinity` hints within existing `boxo/gateway` codebase, so we cna deploy it to our gateways and allow clients like [service-worker-gateway](https://github.com/ipfs-shipyard/helia-service-worker-gateway) or [ipfs-chromium](https://github.com/little-bear-labs/ipfs-chromium) to pass hint and retrieve content even when internal CID was not announced directly.

- [x] support percent-encoded values
- [x] limit to 3 hints per request 
- [x] cancel work if original request finished
- [x] skip work that would not produce any benefit in routing (if path from hint is already cached it won't trigger any additional provider lookups)

## future scope

We may formalize this by extending `IPFSBackend` with explicit place to pass this hint, and then wire it up into routing system of Kubo, but it feels more involved and would like to do that in follow-up, rather than block on it here.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -233,6 +233,8 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	i.handlePathAffinityHints(w, r, contentPath, logger)
+
 	// Detect when explicit Accept header or ?format parameter are present
 	responseFormat, formatParams, err := customResponseFormat(r)
 	if err != nil {
@@ -752,7 +754,7 @@ func (i *handler) handleWebRequestErrors(w http.ResponseWriter, r *http.Request,
 }
 
 // Detect 'Cache-Control: only-if-cached' in request and return data if it is already in the local datastore.
-// https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#cache-control-request-header
+// https://specs.ipfs.tech/http-gateways/path-gateway/#cache-control-request-header
 func (i *handler) handleOnlyIfCached(w http.ResponseWriter, r *http.Request, contentPath path.Path) bool {
 	if r.Header.Get("Cache-Control") == "only-if-cached" {
 		if !i.backend.IsCached(r.Context(), contentPath) {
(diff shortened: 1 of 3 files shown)


gateway/handler.go
```diff
@@ -233,6 +233,8 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	i.handlePathAffinityHints(w, r, contentPath, logger)
```
hacdias, line 236: What about only calling this if it's a request for `raw` blocks or CAR files as indicated [here](https://github.com/ipfs/specs/pull/462/files#diff-c18ce974c093e841cec3118920c4fa12613b711446890e5149cec768c71ea0b4R102-R104)? Or do you want to extend the scope?

hacdias (commented):
It seems like a good initial version, as it will trigger the search for providers for the affinity path, which are likely the same providers that contain the direct block.

I pushed a quick fix for the gateway sharness tests.

gateway/handler.go
```diff
@@ -233,6 +233,8 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
...
+
```
SgtPooki, line 237: As someone not familiar with go nor this codebase, these variable names need work 😳

gateway/handler.go
```diff
@@ -887,6 +889,103 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
...
+			continue
+		}
+
+		// Process hint in background without blocking response logic for contentPath
+		go func(contentPath path.Path, affinityPath path.Path, logger *zap.SugaredLogger) {
+			var immutableAffinityPath path.ImmutablePath
+			logger.Debugw("async processing of Ipfs-Path-Affinity hint", "affinity", affinityPath)
+			if affinityPath.Mutable() {
+				// Skip work if mutable affinity hint is a subset of mutable contentPath
+				if contentPath.Mutable() && strings.HasPrefix(contentPath.String(), affinityPath.String()) {
```
SgtPooki, line 952: Seems like hasPrefix check could be pulled out into a more semantic function since it's used at least twice.

gateway/handler.go
```diff
@@ -887,6 +889,103 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
 	return true
 }
 
+// Detect 'Ipfs-Path-Affinity' (IPIP-462) headers in request and use values as a content
+// routing hints if passed paths are not already in the local datastore.
+// These optional hints are mostly useful for trustless block requests.
+// See https://github.com/ipfs/specs/pull/462
+func (i *handler) handlePathAffinityHints(w http.ResponseWriter, r *http.Request, contentPath path.Path, logger *zap.SugaredLogger) {
+	headerName := "Ipfs-Path-Affinity"
```
gammazero, line 897:
```suggestion
	const headerName = "Ipfs-Path-Affinity"
```

gateway/handler.go
```diff
@@ -887,6 +889,103 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
...
+				// even when no other routing system had them announced. If
+				// original contentPath was received and returned to HTTP
+				// client before below get is done, the work is cancelled.
+
+				logger.Debugw("started async search for providers of Ipfs-Path-Affinity hint", "affinity", affinityPath)
+				_, _, err = i.backend.GetBlock(r.Context(), immutableAffinityPath)
+				logger.Debugw("ended async search for providers of Ipfs-Path-Affinity hint", "affinity", affinityPath, "error", err)
+			} else {
+				logger.Debugw("skipping Ipfs-Path-Affinity hint due to data being locally cached", "affinity", affinityPath)
+			}
```
gammazero, line 984: minor: Instead of `if .. else`, check if cached and return if it is.

```suggestion
			if i.backend.IsCached(r.Context(), immutableAffinityPath) {
				logger.Debugw("skipping Ipfs-Path-Affinity hint due to data being locally cached", "affinity", affinityPath)
				return
			}
			// The intention of below code is to asynchronously preconnect
			// gateway with providers of the affinityPath in
			// Ipfs-Path-Affinity hint. Once connected, these peers can be
			// asked directly (via mechanism like bitswap) for blocks
			// related to main request for contentPath, and retrieve them,
			// even when no other routing system had them announced. If
			// original contentPath was received and returned to HTTP
			// client before below get is done, the work is cancelled.

			logger.Debugw("started async search for providers of Ipfs-Path-Affinity hint", "affinity", affinityPath)
			_, _, err = i.backend.GetBlock(r.Context(), immutableAffinityPath)
			logger.Debugw("ended async search for providers of Ipfs-Path-Affinity hint", "affinity", affinityPath, "error", err)
```

gateway/handler.go
```diff
@@ -887,6 +889,103 @@ func (i *handler) handleSuperfluousNamespace(w http.ResponseWriter, r *http.Requ
...
+		// Confirm it is a valid content path
+		affinityPath, err := path.NewPath(headerValue)
+		if err != nil {
+			logger.Debugw("skipping invalid Ipfs-Path-Affinity hint", "error", err)
+			continue
+		}
+
+		// Skip duplicated work if immutable affinity hint is a subset of requested immutable contentPath
+		// (protect against broken clients that use affinity incorrectly)
+		if !contentPath.Mutable() && !affinityPath.Mutable() && strings.HasPrefix(contentPath.String(), affinityPath.String()) {
```
gammazero, line 941: The comment is a little confusing about which is a prefix of the other.
> if immutable affinity hint is a subset of requested immutable contentPath

It could mean that  the contentPath should be the prefix, since the subset is a more specific (longer) path:
```
contentPath = "/top-level"
affinityPath = "/top-level/sub-path"
```
or it could me that the affinity path is a _string subset_ of the longer content path:
```
contentPath = "/top-level/stuff/sub-stuff"
affinityPath = "/top-level/stuff"
```

gateway/handler.go
gammazero: @SgtPooki In Go HTTP server request handlers, `w` and `r` are pretty much convention for `http.ResponseWriter` and  `*http.Request` respectively. This is also seen the Go docs and [examples](https://pkg.go.dev/net/http#example-Handle).

PR #594, opened 2024-03-25
gateway: remote blocks backend

hacdias:
Partially addresses #576: moves all remote _blocks_ stuff to the gateway package, as well as the remote value store we used in the examples. It also includes the cached blockstore as well as the cached dns.

Moving the graph/CAR backend will be more complicated and involved. I will leave that for a separate PR.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 * `routing/http/server` now adds `Cache-Control` HTTP header to GET requests: 15 seconds for empty responses, or 5 minutes for responses with providers.
+* `gateway` now includes `NewRemoteBlocksBackend` which allows you to create a gateway backend that uses one or multiple other gateways as backend. These gateways must support RAW block requests, as well as IPNS Record requests. With this, we also introduced a `NewCacheBlockStore`, `NewCachedDNS`, `NewRemoteBlockstore` and `NewRemoteValueStore`.
```
```suggestion
* `gateway` now includes `NewRemoteBlocksBackend` which allows you to create a gateway backend that uses one or multiple other gateways as backend. These gateways must support RAW block requests (`application/vnd.ipld.raw`), as well as IPNS Record requests (`application/vnd.ipfs.ipns-record`). With this, we also introduced a `NewCacheBlockStore`, `NewCachedDNS`, `NewRemoteBlockstore` and `NewRemoteValueStore`. 
```

gateway/dns.go
lidel, line 1: :thought_balloon: iirc this was introduced ad a defensive mechanism in Rhea because of https://github.com/ipfs/bifrost-gateway/issues/34 + some things I can discuss 1:1 :see_no_evil: 

It feels like overkill in normal setups with generic DNS and more than one gateway backend on different DNS.

We can either:
- (A) simplify code, remove it from `NewRemoteBlockstore` and delete `CachedDNS` 
- (B) move it to namesys, and use for caching DNSLink in offline mode or when censored (https://github.com/ipfs/kubo/issues/8245) 

@hacdias  any preference? I think for simplicyty sake, we could go with (A), and once we pick up (B) we can revive useful parts of https://github.com/ipfs/boxo/blob/2036531ddab652fca0b8a06dd5ecf414bb0c32e7/gateway/dns.go, but no need to be distracted by it right now.

gateway/remote_blocks_backend.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net/http"
+	"time"
+
+	"github.com/ipfs/boxo/blockservice"
+	"github.com/ipfs/boxo/exchange/offline"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+const getBlockTimeout = time.Second * 60
```
lidel, line 13: No need to address this now, but worth leaving a TODO:

```suggestion
// TODO: make this configurable via BlocksBackendOption
const getBlockTimeout = time.Second * 60
```

lidel (approved):
sgtm, small nits below

PR #595, opened 2024-03-26
provider: prioritize roots and introduce NewPrioritizedProvider

hacdias:
See https://github.com/ipfs/kubo/pull/10376.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -118,7 +118,7 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
 			})
 		})
 		if err != nil {
-			logR.Errorf("reprovide direct pins: %s", err)
+			logR.Errorf("reprovide indirect pins: %s", err)
 			return
 		}
 	}()
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -71,43 +71,55 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
 }
 
 func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory, onlyRoots bool) (*cidutil.StreamingSet, error) {
-	// FIXME: Listing all pins code is duplicated thrice, twice in Kubo and here, maybe more.
-	// If this were a method of the [pin.Pinner] life would be easier.
 	set := cidutil.NewStreamingSet()
+	recursivePins := cidutil.NewSet()
 
 	go func() {
 		ctx, cancel := context.WithCancel(ctx)
 		defer cancel()
 		defer close(set.New)
 
-		for sc := range pinning.DirectKeys(ctx, false) {
-			if sc.Err != nil {
-				logR.Errorf("reprovide direct pins: %s", sc.Err)
-				return
-			}
-			set.Visitor(ctx)(sc.Pin.Key)
-		}
-
-		session := fetchConfig.NewSession(ctx)
+		// 1. Recursive keys
 		for sc := range pinning.RecursiveKeys(ctx, false) {
 			if sc.Err != nil {
 				logR.Errorf("reprovide recursive pins: %s", sc.Err)
 				return
 			}
-			set.Visitor(ctx)(sc.Pin.Key)
 			if !onlyRoots {
-				err := fetcherhelpers.BlockAll(ctx, session, cidlink.Link{Cid: sc.Pin.Key}, func(res fetcher.FetchResult) error {
-					clink, ok := res.LastBlockLink.(cidlink.Link)
-					if ok {
-						set.Visitor(ctx)(clink.Cid)
-					}
-					return nil
-				})
-				if err != nil {
-					logR.Errorf("reprovide indirect pins: %s", err)
-					return
-				}
+				// Save some bytes.
+				_ = recursivePins.Visit(sc.Pin.Key)
 			}
+			_ = set.Visitor(ctx)(sc.Pin.Key)
+		}
+
+		// 2. Direct pins
+		for sc := range pinning.DirectKeys(ctx, false) {
+			if sc.Err != nil {
+				logR.Errorf("reprovide direct pins: %s", sc.Err)
+				return
+			}
+			_ = set.Visitor(ctx)(sc.Pin.Key)
+		}
+
+		if onlyRoots {
+			return
+		}
+
+		// 3. Go through recursive pins to fetch remaining blocks if we want more
+		// than just roots.
+		session := fetchConfig.NewSession(ctx)
+		err := recursivePins.ForEach(func(c cid.Cid) error {
+			return fetcherhelpers.BlockAll(ctx, session, cidlink.Link{Cid: c}, func(res fetcher.FetchResult) error {
+				clink, ok := res.LastBlockLink.(cidlink.Link)
+				if ok {
+					_ = set.Visitor(ctx)(clink.Cid)
+				}
+				return nil
+			})
+		})
+		if err != nil {
+			logR.Errorf("reprovide direct pins: %s", err)
+			return
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


hacdias:
@aschmahmann what do you think? I can replace the map with a bloom filter. That can probably be better memory wise depending on how streams we're using.

provider/provider.go
```diff
@@ -113,3 +113,48 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
+							str := string(c.Bytes())
+							if _, ok := has[str]; ok {
+								continue
+							}
+
+							select {
+							case <-ctx.Done():
+								return
+							case outCh <- c:
+								has[str] = true
```
hacdias: I think that maybe we'll only be wanting to do this for the first stream, aka, the roots.  Otherwise, this can start becoming large. But then this function stops being as generic. Maybe rename to `NewPrioritizedRootsProvider(roots KeyChanFunc, rest KeyChanFunc)`.

provider/provider.go
```diff
@@ -113,3 +113,48 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
+		go func() {
+			defer close(outCh)
+			has := map[string]bool{}
+
+			for _, stream := range streams {
+				ch, err := stream(ctx)
+				if err != nil {
+					log.Warnf("error in prioritized strategy: %w", err)
+				} else {
+				L:
```
lidel: :thought_balloon: nit: can we simplify by removing  `L:`   and replacing `} else {` with `continue }` ?

provider/provider.go
```diff
@@ -113,3 +113,48 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
 	return set, nil
 }
+
+func NewPrioritizedProvider(streams ...KeyChanFunc) KeyChanFunc {
+	return func(ctx context.Context) (<-chan cid.Cid, error) {
+		outCh := make(chan cid.Cid)
+
+		go func() {
+			defer close(outCh)
+			has := map[string]bool{}
```
lidel: :thought_balloon: since we use it as set-like data structure, and care only about existence of key, and not value, we can avoid allocating bool values:
```suggestion
			has := map[string]struct{}
```


and then:

```diff
- has[str] = true
+ has[str] = struct{}{}
```

provider/provider.go
lidel: Yeah, we want to avoid memory footgun here, our users already run with extremely low memory setups.

Back-of-napkin memory penalty if we don't limit this to first pass:

- My PC node has ~1500 pins, but close to 3 million of blocks. That means pins make ~0.05% of my CIDs. Cluster nodes will likely be way bigger, but not multiple orders of magnitude bigger.
- If we have 10 million of CIDs, and each CID is v1 and ~60 bytes, that is close to >600MiB for this map alone. 
- Providing 50 million CIDs would require additional >3GiB of RAM.

So yes, if we want to enable this by default, as a smarter provider, we want to limit updating the map to the first pass, and then announce everything else.

I think going with explicit and basic API of `NewPrioritizedProvider(priorityCids KeyChanFunc, allCids KeyChanFunc)` makes sense.

For `Reprovider.Strategy` in Kubo, CID streams will be:
- `all`
  - first pass pin roots, then everything else
- `pinned`
  - `first pass pin roots, then pinned CIDs recursively (skipping already announced roots)
- `roots` 
  - no change, we announce pin roots only

lidel (changes requested):
I think we can ship this if we limit CID map to items from the first-pass (details inline).

provider/provider.go
hacdias: Adding the `continue` here is not going to help removing the `L` since the problem is the that you need to break out of the nested `select` to the `for`. There's no other way of doing this here AFAIK.

Anyways, it does not matter since this is going to be reworked to only work with two streams.

provider/provider.go
```diff
@@ -113,3 +113,60 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
 	return set, nil
 }
+
+func NewPrioritizedProvider(priorityCids KeyChanFunc, otherCids KeyChanFunc) KeyChanFunc {
+	return func(ctx context.Context) (<-chan cid.Cid, error) {
+		outCh := make(chan cid.Cid)
+
+		go func() {
+			defer close(outCh)
+			visited := map[string]struct{}{}
```
lidel: :thought_balloon: @aschmahmann  FYSA, any concerns? fwiw, even without bloom filter, since we populate this map only with roots of pins, this should be less  than what we already do in `pinSet` with `onlyRoots=false` (inside similar map that lives in `StreamingSet`)

lidel (approved):
Took it for a spin in https://github.com/ipfs/kubo/pull/10376 and seem to work as expected.

provider/provider.go
```diff
@@ -113,3 +113,60 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
+					select {
+					case <-ctx.Done():
+						return nil
+					case c, ok := <-ch:
+						if !ok {
+							return nil
+						}
+
+						// Likely faster than c.String()
+						str := string(c.Bytes())
```
aschmahmann: c.KeyString()? I'd hope the compiler makes the two the same thing 🤷, but wanted to flag that a function that gets you this already exists.

provider/provider.go
```diff
@@ -113,3 +113,60 @@ func pinSet(ctx context.Context, pinning pin.Pinner, fetchConfig fetcher.Factory
...
+				}
+			}
+
+			err := handleStream(priorityCids, true)
+			if err != nil {
+				log.Warnf("error in prioritized strategy while handling priority CIDs: %w", err)
+				return
+			}
+
+			err = handleStream(otherCids, false)
```
aschmahmann, line 174: Is the idea that each keyFunc should do its own internal deduplication and so setting this to false saves some memory? If so makes sense to me.

provider/reprovider_test.go
```diff
@@ -221,3 +223,93 @@ func TestOfflineRecordsThenOnlineRepublish(t *testing.T) {
...
+			expected: append(cids[3:], cids[:3]...),
+		},
+		{
+			name:     "no repeated",
+			priority: cids[3:],
+			all:      cids[3:],
+			expected: cids[3:],
+		},
+		{
+			name:     "no repeated intercalated",
```
aschmahmann: "intercalated" doesn't seem like the right word, but what's the test here "no repeated if duplicates in the prioritized channel"?

provider/provider.go
aschmahmann: Thinking a bit it might be more sensible to use the multihash bytes here rather than the CID bytes given that the major routing systems in use (Amino DHT, IPNI) are multihash based rather than CID based and so advertising 2 CIDs with the same multihash isn't really useful anyhow.

If people feel this abstraction should work on CIDs it's probably not too wasteful for most users, but I doubt changing will hurt anyone given routing systems that want more than just multihashes probably want more information than just CIDs (unless they're only differentiating full-DAG vs block)

provider/provider.go
aschmahmann: Yeah, this is obviously not ideal but may be fine at the moment (we should get people to test during RC and give feedback). See my comment https://github.com/ipfs/kubo/pull/10376/files#r1554294042 which might result in changes here

provider/provider.go
hacdias: No, the point was to avoid memory issues assuming that the priorityCids stream is much smaller than the otherCids. See more context here: https://github.com/ipfs/boxo/pull/595#discussion_r1539253269

hacdias: Well, it seems that each keyFunc does indeed do its own internal deduplication, so I think I can say: yes.

provider/provider.go
```diff
@@ -71,43 +71,55 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
...
+				clink, ok := res.LastBlockLink.(cidlink.Link)
+				if ok {
+					_ = set.Visitor(ctx)(clink.Cid)
 				}
-			}
+				return nil
+			})
+		})
+		if err != nil {
+			logR.Errorf("reprovide direct pins: %s", err)
```
lidel: nit: `err` here is for indirect (child of recursive pin)

lidel (approved):

PR #596, opened 2024-03-27
feat(routing/http/server): improve ux of /ipns

lidel:
- `Content-Disposition` with meaningful file extension
- default to application/vnd.ipfs.ipns-record when `Accept` is missing or includes `*/*` wildcard.

### DIFF
--- a/routing/http/server/server.go
+++ b/routing/http/server/server.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-multibase"
 
 	logging "github.com/ipfs/go-log/v2"
 )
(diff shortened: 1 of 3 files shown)


routing/http/server/server.go
```diff
@@ -370,8 +371,14 @@ func (s *server) findPeersNDJSON(w http.ResponseWriter, peersIter iter.ResultIte
...
 
 func (s *server) GetIPNS(w http.ResponseWriter, r *http.Request) {
-	if !strings.Contains(r.Header.Get("Accept"), mediaTypeIPNSRecord) {
-		writeErr(w, "GetIPNS", http.StatusNotAcceptable, errors.New("content type in 'Accept' header is missing or not supported"))
+	acceptHdrValue := r.Header.Get("Accept")
+	// When 'Accept' header is missing, default to 'application/vnd.ipfs.ipns-record'
+	// (improved UX, similar to how we default to JSON response for /providers and /peers)
+	if len(acceptHdrValue) == 0 {
+		acceptHdrValue = mediaTypeIPNSRecord
+	}
```
aschmahmann: Should we add a test for this behavior?

aschmahmann (commented):
LGTM, although @hacdias has been spending more time in this code than I have 😅

routing/http/server/server.go
lidel: Good catch: we had NO tests for implicit behavior in browser (when `Accept` is not provided).
Added tests for all endpoints in 3f251322

aschmahmann: @lidel might be off here (either in my understanding of the Go libraries or browsers), but doesn't the browser default to a set of accept headers rather than none https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values?

So if we want to default returning a specific type (e.g. `application/vnd.ipfs.ipns-record` for IPNS records) in a web browser (as opposed to say curl), then it seems like we'd need to do parsing of the Accept header(s?) and check for `application/vnd.ipfs.ipns-record` or something compatible like `*/*` or `application/*`.

Unfortunately Go seems to have no built-in tooling for this https://github.com/golang/go/issues/19307 🙃, but the linked issue points to a bunch of libraries that could help here if we decided to do something more than the basic manual checking (e.g. something like split on `,` and check for the 3 strings above and if we're being cautious also checking for repeated headers rather than using `header.Get`).

Also fair to push handling of `*/*` (and therefore web browser responses) to a separate issue and just leave this PR as is.

hacdias: Exactly, the browser will send `text/html` at least I think.

lidel: Good catch. Browsers send a version of  `Accept: text/plain,[bunch of image negotiations],*/*"`, curl sends `*/*`, other wildcards are not used in practice, we can do simple string match to keep code simple.

If present, do the same default response type we do when `Accept` is missing entirely.

Added fix + regression tests in 4c5650bb

hacdias (approved):

PR #597, opened 2024-04-08
bitswap: add missing client options to exchange

hacdias:


lidel (approved):
I guess we've missed/never needed them before?

PR #601, opened 2024-04-12
fix(routing): return 404 when no record is found

hacdias:
Addresses https://github.com/ipfs/someguy/issues/58.

Sets 404 if the router returns `routing.ErrNotFound` or if the list of providers is empty.

### DIFF
--- a/routing/http/server/server.go
+++ b/routing/http/server/server.go
@@ -546,6 +546,10 @@ func writeJSONResult(w http.ResponseWriter, method string, val interface{ Length
 }
 
 func writeErr(w http.ResponseWriter, method string, statusCode int, cause error) {
+	if errors.Is(cause, routing.ErrNotFound) {
+		setCacheControl(w, maxAgeWithoutResults, maxStale)
+	}
+
 	w.WriteHeader(statusCode)
 	causeStr := cause.Error()
 	if len(causeStr) > 1024 {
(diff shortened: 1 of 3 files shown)


routing/http/server/server.go
```diff
@@ -398,8 +409,13 @@ func (s *server) GetIPNS(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		writeErr(w, "GetIPNS", http.StatusInternalServerError, fmt.Errorf("delegate error: %w", err))
-		return
+		if errors.Is(err, routing.ErrNotFound) {
+			writeErr(w, "GetIPNS", http.StatusNotFound, fmt.Errorf("delegate error: %w", err))
+			return
+		} else {
+			writeErr(w, "GetIPNS", http.StatusInternalServerError, fmt.Errorf("delegate error: %w", err))
+			return
+		}
```
lidel, line 418: :thought_balloon: iiuc if we return here, we won't set meaningful cache-control on these 404s?

Probably not a big deal for IPNS responses, as their rate and failure raite will be lower, but could we refactor IPNS not found handling to include explicit `cache-control` header? 

Maybe reuse `setCacheControl(w, maxAgeWithoutResults, maxStale)` from other endpoints, and call it inside of `writeErr` if code is NotFound?

lidel (approved):
Lgtm, thanks for fixing this, 404s are cachable, 500 are not, so this will save resources across all deployments. 

Fine to ship, once small nit below is addressed.

PR #603, opened 2024-04-16
gateway: add Content-Location for non-default response formats

### LINKED ISSUE #606: gateway: fix car/block responses cached as deserialized ones when ?format=raw|car is missing
> Extracting context from internal https://github.com/orgs/ipshipyard/projects/1/views/1 to create a public log why `Content-Location` was added in https://github.com/ipfs/boxo/pull/603

### Problem
- `/ipfs/cid ` can return different bytes based on `Accept` header which caused bugs in the past (block response being returned instead of deserialized one etc)
  - We can't use `Vary: Accept` because we don't want cache key to be different for every blient/browser, we only want it to be different for block/car/ipns-record/tar/(dag-)json|cbor content types

### Solution

- Old bifrost-infra had custom code that uses custom cache key if response is block or car
  - Generic HTTP stack like LB from Cloudflare may not have that capability, and if it does, it requires custom config similar to one we did in Nginx.
  - Better to solve this at gateway level, using generic HTTP, with safe defaults that "just work" with stock HTTP tools and services, without requiring additional config on each deployment

### Ideas 

- (A) Ensure unique URL per response type, so it does not matter if HTTP middleware is not aware of our content-types
  - Return HTTP 302  to `?format=car|raw|foo` with `Cache-Control: no-cache` when alternative response format is requested
    - DOWNSIDE: bypasses http caches, hits backend every time (path needs to be resolved, blocks fetched/read)
- (B) 
  - Include [`Content-Location: [...]?format=foo`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) with non-default response 
  - Implement in boxo/gateway + update gateway specs
- (C) use `Vary: Accept` and pay the penalty of more cache misses

### Plan

- 2024-04-08 implement (B) for now, and test with popular LB/CDN provider like Cloudflare
- 2024-04-18 https://github.com/ipfs/boxo/pull/603 is ready

hacdias:
Addresses https://github.com/orgs/ipshipyard/projects/1/views/1?pane=issue&itemId=58406852
Close https://github.com/ipfs/boxo/issues/606

Specs PR: https://github.com/ipfs/specs/pull/471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The following emojis are used to highlight certain changes:
 * ✨ `gateway` has new backend possibilities:
   * `NewRemoteBlocksBackend` allows you to create a gateway backend that uses one or multiple other gateways as backend. These gateways must support RAW block requests (`application/vnd.ipld.raw`), as well as IPNS Record requests (`application/vnd.ipfs.ipns-record`). With this, we also introduced `NewCacheBlockStore`, `NewRemoteBlockstore` and `NewRemoteValueStore`.
   * `NewRemoteCarBackend` allows you to create a gateway backend that uses one or multiple Trustless Gateways as backend. These gateways must support CAR requests (`application/vnd.ipld.car`), as well as the extensions describe in [IPIP-402](https://specs.ipfs.tech/ipips/ipip-0402/). With this, we also introduced `NewCarBackend`, `NewRemoteCarFetcher` and `NewRetryCarFetcher`.
+* `gateway` now sets the [`Content-Location`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header for requests with non-default content format, as a result of content negotiation. This allows generic and misconfigured HTTP caches to store Deserialized, CAR and Block responses separately, under distinct cache keys.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The following emojis are used to highlight certain changes:
   * `NewRemoteBlocksBackend` allows you to create a gateway backend that uses one or multiple other gateways as backend. These gateways must support RAW block requests (`application/vnd.ipld.raw`), as well as IPNS Record requests (`application/vnd.ipfs.ipns-record`). With this, we also introduced `NewCacheBlockStore`, `NewRemoteBlockstore` and `NewRemoteValueStore`.
   * `NewRemoteCarBackend` allows you to create a gateway backend that uses one or multiple Trustless Gateways as backend. These gateways must support CAR requests (`application/vnd.ipld.car`), as well as the extensions describe in [IPIP-402](https://specs.ipfs.tech/ipips/ipip-0402/). With this, we also introduced `NewCarBackend`, `NewRemoteCarFetcher` and `NewRetryCarFetcher`.
 * `gateway` now sets the [`Content-Location`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header for requests with non-default content format, as a result of content negotiation. This allows generic and misconfigured HTTP caches to store Deserialized, CAR and Block responses separately, under distinct cache keys.
+* `gateway` now supports `car-dups`, `car-order` and `car-version` as query parameters in addition to the `application/vnd.ipld.car` parameters sent via `Accept` header. The parameters in the `Accept` header have always priority, but including them in URL simplifies HTTP caching and allows use in `Content-Location` header on CAR responses to maximize interoperability with wide array of HTTP caches.
 
 ### Changed
 
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -259,6 +259,8 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 		responseParams: formatParams,
 	}
 
+	addContentLocation(r, w, rq)
+
 	// IPNS Record response format can be handled now, since (1) it needs the
 	// non-resolved mutable path, and (2) has custom If-None-Match header handling
 	// due to custom ETag.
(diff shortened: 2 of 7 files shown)


hacdias:
@lidel before I open a PR for the specifications, we should decide what to do with CAR requests that contain "parameters" in the `Accept` header itself (e.g. `dups`, `order`).

These cannot be translated into query URL parameters, so the `Content-Location` header would not be fully correct. For now, I am not adding it. I wonder if we shouldn't instead make it possible to also set these parameters through the URL, or just keep is as it is.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The following emojis are used to highlight certain changes:
 * ✨ `gateway` has new backend possibilities:
   * `NewRemoteBlocksBackend` allows you to create a gateway backend that uses one or multiple other gateways as backend. These gateways must support RAW block requests (`application/vnd.ipld.raw`), as well as IPNS Record requests (`application/vnd.ipfs.ipns-record`). With this, we also introduced `NewCacheBlockStore`, `NewRemoteBlockstore` and `NewRemoteValueStore`.
   * `NewRemoteCarBackend` allows you to create a gateway backend that uses one or multiple Trustless Gateways as backend. These gateways must support CAR requests (`application/vnd.ipld.car`), as well as the extensions describe in [IPIP-402](https://specs.ipfs.tech/ipips/ipip-0402/). With this, we also introduced `NewCarBackend`, `NewRemoteCarFetcher` and `NewRetryCarFetcher`.
+* `gateway` now sets the [`Content-Location`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header for requests with non-default content format, as a result of content negotiation.
```
```suggestion
* `gateway` now sets the [`Content-Location`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header for requests with non-default content format, as a result of content negotiation. This allows generic and misconfigured HTTP caches to store Deserialized, CAR and Block responses separately, under distinct cache keys. 
```

gateway/handler.go
```diff
@@ -640,6 +648,39 @@ func customResponseFormat(r *http.Request) (mediaType string, params map[string]
...
+	}
+
+	// Copy known query parameters.
+	query := url.Values{}
+	for _, param := range []string{carRangeBytesKey, carTerminalElementTypeKey, "filename", "download"} {
+		if v := r.URL.Query().Get(param); v != "" {
+			query.Set(param, v)
+		}
+	}
+	query.Set("format", param)
```
lidel: nit: hardcoding list of safelisted params here will be hard to maintain.

Instead, copy/preserve all params from the original request URL. 

All we need is to add an explicit `format` on top of whatever was in the original request.

gateway/handler.go
```diff
@@ -640,6 +648,39 @@ func customResponseFormat(r *http.Request) (mediaType string, params map[string]
...
+
+	w.Header().Set("Content-Location", path+"?"+query.Encode())
```
lidel, line 688: @hacdias mind opening ipfs/spec PR to add `Content-Length` header to https://specs.ipfs.tech/http-gateways/path-gateway/ so we dont forget to document this? :pray: 

One paragraph description that it should be included in responses to request without `format` query parameter but with known `Accept`  type should be enough.

gateway/handler.go
```diff
@@ -640,6 +648,39 @@ func customResponseFormat(r *http.Request) (mediaType string, params map[string]
 	return "", nil, nil
 }
 
+// Add 'Content-Location' headers for non-default response formats. This allows
+// correct caching of such format requests when the format is passed via the
+// Accept header, for example.
+func addContentLocation(r *http.Request, w http.ResponseWriter, rq *requestData) {
+	if rq.responseFormat == "" {
+		return
+	}
```
lidel, line 659: I think we should also skip if `format` param is already present.

`Content-Location` should be returned only when `format` is missing.

```suggestion
  // Skip generating Content-Location if 
  // no explicit format was requested via Accept HTTP header,
  // or if it was requested via URL query parameter.
	if rq.responseFormat == "" || r.URL.Query().Get("format") != "" {
		return
	}
```

Rationale:
If we return `Content-Location` when `format` is already present in the original request, we will be creating a surface for cache-miss on subsequent requests due to `Content-Location` having query parameters in different order than the request made by non-golang client (that may have them in different order).

Cloudflare has "Enable Query String Sort" feature but it is only on Enterprise plan, so would not be something others can use.

gateway/handler.go
```diff
@@ -640,6 +648,39 @@ func customResponseFormat(r *http.Request) (mediaType string, params map[string]
...
+
+	// Response format parameters, such as 'dups' and 'order' for CAR requests
+	// cannot be translated into the URL. Therefore, we cannot add a 'Content-Location'
+	// header.
+	if len(rq.responseParams) != 0 {
+		return
+	}
```
lidel: :thought_balloon: Hm.. this is tricky.

The reason why we add `Content-Location` is to ensure CAR response triggered via `Accept` header  is not cached under `/ipfs/cid`, as that would be a bug when returned instead of default (deserialized) response type. 

If we skip here, we create a potential DoS hole, which allows clients to maliciously inject CAR with dups as response for non-CAR request.

I see two options here for the value in`Content-Location`:

- (A) update specs and conformance tests and allow for car content type params to be passed in URL query as `format=car&car-version=1&car-dups=n&car-order=dfs`.
- (B) avoid rabbit hole, keep change limited to `Content-Length` and its purpose.
  - Solve the problem at hand by including `format=car&something=hash(version+dups+order)` IFF any of car params in `Accept` header was different from the default.
  - This should be enough to ensure non-standard response body is not cached under generic `format=car`. 
  - We already have unique Etag per car params produced in `getCarEtag`, so you dont need to invent anything, maybe do `&etag=percent-encoded(getCarEtag(..))}`?

@hacdias maybe do (B) now, and fill issue for (A) in ipfs/specs?

lidel (changes requested):

gateway/handler.go
hacdias: There itis: https://github.com/ipfs/specs/pull/471

hacdias:
@lidel I've opened a PR in the specs repository: https://github.com/ipfs/specs/pull/471

gateway/handler.go
hacdias: @lidel (B) makes me feel uneasy due to the _meaning_ of the `Content-Location` header. If I saw the header response, I would expect that opening such URL would give me the same result as the current request. But that would not be the case here.

(A) likely takes a bit longer, but I think it should be prioritised. I don't think implementing it would take that long, or writing the specs, because:

1. We just ende to add three new query parameters: `car-version`, `car-dups` and `car-order`. Code wise this should be simple. The questions this may rise is: why are the other CAR parameters _not_ part of the `Accept` header (`entity-bytes` and `dag-entity`)?
2. `Accept` header parameter already has priority when both are included, so this is a unnecessary conversation.

hacdias: @lidel I implemented option (A), and additionally opened a PR in the specs:

- https://github.com/ipfs/boxo/pull/604
- https://github.com/ipfs/specs/pull/472

If this is good enough to merge, I think we should go ahead with it. Otherwise, I can also do (B).

lidel: Thanks! I think this is good enough, we did the due diligence and documented `Content-Location` story in specs, and it is better to have this solution for overwriting different CAR variant in cache, than not. 

To answer the question about `entity-bytes` and `dag-entity` not being parameters of `application/vnd.ipld.car`: these are not CAR specific, nor related to CAR format, but a way to limit the range of DAG being requested. Right now it is only meaningful for CARs, but we could have future format other than CAR that would also use them.

lidel (approved):
Thanks! I've reviewed ipfs/specs PRs,  merged #604   into this one and pushed fix/test for case where `format` is different than `Accept` (https://github.com/ipfs/boxo/pull/603/commits/91fdd909abdf5317a42f19e9627e84edd5fd372d). 

@hacdias If my changes look ok, feel free to merge, bubble up to rainbow, and bump our infra to rainbow image that includes this :pray:

PR #604, opened 2024-04-17
gateway: add support for car-* query parameters

hacdias:
Option (A) from https://github.com/ipfs/boxo/pull/603#discussion_r1567808439. Specifications: https://github.com/ipfs/specs/pull/472.


lidel (approved):
Thanks, let's merge into #604.

There is one case where I want to add test (`?format` being different than `Accept`), but will do that there.

PR #607, opened 2024-04-25
bitswap/server: allow overriding peer ledger with WithPeerLedger

hacdias:
Introduces `WithPeerLedger`, which allows to override the default peer ledger with anything else you want.

### DIFF
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -149,6 +149,8 @@ type Engine struct {
 
 	lock sync.RWMutex // protects the fields immediately below
 
+	noPeerLedger bool
+
 	// peerLedger saves which peers are waiting for a Cid
 	peerLedger *peerLedger
 
@@ -302,6 +304,15 @@ func WithSetSendDontHave(send bool) Option {
 func WithNotifyNewBlocks(notify bool) Option {
 	return func(e *Engine) {
 		e.notifyNewBlocks = notify
+	}
+}
+
+// WithNoPeerLedger disables the usage of a peer ledger to track want lists. This
+// means that this engine will not keep track of the CIDs that peers wanted in
+// the past.
+func WithNoPeerLedger() Option {
+	return func(e *Engine) {
+		e.noPeerLedger = true
 	}
 }
 
@@ -711,7 +722,10 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 			continue
 		}
 
-		e.peerLedger.Wants(p, entry.Entry)
+		if !e.noPeerLedger {
+			e.peerLedger.Wants(p, entry.Entry)
+		}
+
 		filteredWants = append(filteredWants, entry)
 	}
 	clear := wants[len(filteredWants):]
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -168,6 +168,8 @@ type Engine struct {
 
 	sendDontHaves bool
 
+	notifyNewBlocks bool
+
 	self peer.ID
 
 	// metrics gauge for total pending tasks across all workers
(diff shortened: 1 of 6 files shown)


bitswap/server/internal/decision/engine.go
```diff
@@ -699,7 +722,10 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 			continue
 		}
 
-		e.peerLedger.Wants(p, entry.Entry)
+		if !e.noPeerLedger {
+			e.peerLedger.Wants(p, entry.Entry)
+		}
+
```
hacdias: ℹ️ This should be enough as the only place to use `e.noPeerLedger`. All other operations will then return empty slices/0 or be no-ops.

bitswap/server/internal/decision/engine.go
```diff
@@ -294,6 +298,24 @@ func WithSetSendDontHave(send bool) Option {
...
+func WithNotifyNewBlocks(notify bool) Option {
+	return func(e *Engine) {
+		e.notifyNewBlocks = notify
+	}
+}
+
+// WithNoPeerLedger disables the usage of a peer ledger to track want lists. This
+// means that this engine will not keep track of the CIDs that peers wanted in
+// the past.
+func WithNoPeerLedger() Option {
```
aschmahmann: Haven't looked at this too much, but it might be helpful to describe the desired behavior more than the internal state. Perhaps something like `WithRequestResponseOnly` since the idea is for the server to only handle request-response interactions rather than subscription-like ones and therefore rely on rebroadcast if the data is added after a client has first asked for it (IIRC this is the behavior of some implementations like web3.storage's)

lidel: The `WithRequestResponseOnly` suggested by Adin feels a bit too high level, and too far away from the concept of wantlists and subscriptions. 

Our need at hand, is to disable per peer wantlists, and the reliable way to do it is to set noopPeerLedger. 

If we want high level option, perhaps replace `WithNoPeerLedger` and  `WithNotifyNewBlocks` with a single option: 

```go
WithPeerWantlists(false)
``` 

- `true` would be current default `peerLedger`, but allows implementers to explicitly flag they want wantlists
- `false` would force use of explicit `noopPeerLedger`,  this is what rainbow would use

Alternative is to not invent any new terms, and only allow passing custom peerLedger:

```go
WithPeerLedger(DefaultPeerLedger())
WithPeerLedger(NoopPeerLedger()) // no ledger → no wantlists → no peers to notify
```

This would allow people to implement better wantlists than default map-based we have.
:point_right:  fwiw, I would go with `WithPeerLedger(NoopPeerLedger())`  personally: solves our needs and makes library easier to customize in general.

bitswap/server/internal/decision/engine.go
```diff
@@ -294,6 +298,24 @@ func WithSetSendDontHave(send bool) Option {
 	}
 }
 
+// WithNotifyNewBlocks sets or not whether to notify peers when receiving a block.
+// By default, it is true. This can be useful if you want the server to only
+// reply if it has a block at the moment it receives a message, and not later.
+func WithNotifyNewBlocks(notify bool) Option {
+	return func(e *Engine) {
+		e.notifyNewBlocks = notify
```
lidel: nit: do we need this? having a separate `notifyNewBlocks` flag feels redundant in both scenarios:

- if we have working peerLedger,  we  want to send notifications
- if we have noop peerLedger, we dont track things, we won't send notifications

Perhaps we could remove it and only have  one flag that controls if we have noop peerLedger, or not? (see other comment)

lidel (changes requested):
I don't feel strongly how we name it, but
- feels like we could go away with adding just one option for this
- rainbow should use explicit `noopPeerLedger` that has all methods implemented as noop. this will be less prone to breakage / bugs during refactors.

details inline

bitswap/server/internal/decision/engine.go
hacdias: @lidel that was actually my original idea. I didn't do it at the time because it'll need some refactoring around the types. But I can take a better look tody.

hacdias:
@lidel I refactored to only have one new option (`WithPeerLedger`). I did **not** add a no-op peer ledger here. I will do this in Rainbow. My idea is that I didn't want to export custom peer ledgers here. We also do **not export** the default peer ledger, just like it has been done with the score ledger. Just trying to avoid inconsistencies here. If we want to use the no-op peer ledger in other places, we can then bring it to Boxo.

lidel (approved):
Lgtm, follows what we already had for `WithScoreLedger` and does not introduce anything new. 


---
 
Digression on future work:
We are still figuring out what appropriate settings for rainbow are, but once we do, it will be useful for others. If we want to add a higher abstraction similar to what @aschmahmann suggested (`WithRequestResponseOnly`),  I feel it needs to be more generic, e.g.: `WithProfile("minimal-request-response")` because it would adjust multiple settings, and work similar to Kubo profiles.

PR #615, opened 2024-06-07
feat(bitswap/client): add additional tracing

aschmahmann:
This is an attempt at adding some more visibility to the Bitswap tracing code path. This overall looks to be a good improvement, but I would expect the shape of the tracing (e.g. spans vs events, granularity and names of spans/events, etc.) to change over time here and don't think we want to commit to documenting each of those changes.

I'm labeling as skip/changelog, but happy to add some info if it makes sense.

Feedback definitely welcome 🙏

### DIFF
--- a/bitswap/client/internal/providerquerymanager/providerquerymanager.go
+++ b/bitswap/client/internal/providerquerymanager/providerquerymanager.go
@@ -2,13 +2,15 @@ package providerquerymanager
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
+	"github.com/ipfs/boxo/bitswap/client/internal"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	peer "github.com/libp2p/go-libp2p/core/peer"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var log = logging.Logger("bitswap")
@@ -39,7 +41,7 @@ type ProviderQueryNetwork interface {
 }
 
 type providerQueryMessage interface {
-	debugMessage() string
+	debugMessage()
 	handle(pqm *ProviderQueryManager)
 }
 
@@ -61,6 +63,7 @@ type newProvideQueryMessage struct {
 }
 
 type cancelRequestMessage struct {
+	ctx               context.Context
 	incomingProviders chan peer.ID
 	k                 cid.Cid
 }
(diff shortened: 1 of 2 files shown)


bitswap/client/internal/providerquerymanager/providerquerymanager.go
```diff
@@ -39,7 +41,7 @@ type ProviderQueryNetwork interface {
 }
 
 type providerQueryMessage interface {
-	debugMessage() string
+	debugMessage()
```
aschmahmann, line 44: I changed this because:
- It should be slightly more efficient by not needing to encode strings if no tracing or debugging is happening
- It enabled grouping tracing and debugging together when we were emitting roughly the same information

bitswap/client/internal/session/session.go
```diff
@@ -304,32 +305,44 @@ func (s *Session) run(ctx context.Context) {
 	for {
 		select {
 		case oper := <-s.incoming:
+			var opCtx context.Context
+			var span trace.Span
 			switch oper.op {
 			case opReceive:
 				// Received blocks
+				opCtx, span = internal.StartSpan(ctx, "Session.ReceiveOp")
```
aschmahmann: Most of these spans are frequently so small they could be events. I was trying to treat things fairly equally here since bugs should show up in any number of places. However, given this is an event loop and should be pretty tight I'd expect all the times to be almost 0.

I'm considering shrinking everything but the most obvious longer tasks (iirc broadcast and periodic search)

aschmahmann: I switched this to events so we didn't have to ignore `opCtx` for the linter and to shrink the number of spans

bitswap/client/internal/session/session.go
```diff
@@ -304,32 +305,44 @@ func (s *Session) run(ctx context.Context) {
 	for {
 		select {
 		case oper := <-s.incoming:
+			var opCtx context.Context
+			var span trace.Span
 			switch oper.op {
 			case opReceive:
 				// Received blocks
+				trace.SpanFromContext(ctx).AddEvent("Session.ReceiveOp")
 				s.handleReceive(oper.keys)
```
aschmahmann: It'd be nice to add the keys to the event here, although a little more logic is might be needed because stringifying every CID even when tracing is disabled (or on a minimal fraction of requests) would be a pain.

bitswap/client/internal/session/session.go
```diff
@@ -307,29 +308,39 @@ func (s *Session) run(ctx context.Context) {
 			switch oper.op {
 			case opReceive:
 				// Received blocks
+				trace.SpanFromContext(ctx).AddEvent("Session.ReceiveOp")
```
gammazero: Since this is in a loop, maybe better to call `span := trace.SpanFromContext(ctx)` outside loop, and then call `span.AddEvent()` inside the loop.

gammazero (commented):
minor comment, LGTM

gammazero (approved):

bitswap/client/internal/session/session.go
aschmahmann: Good call 🙏, previously these were all `internal.StartSpan` but didn't see the optimization here when I switched it.

bitswap/client/internal/session/session.go
aschmahmann: Going to skip this for now

PR #618, opened 2024-06-12
routing/http: fix goroutine/memory leak [skip changelog]

MichaelMure:
I've noticed some memory and goroutine buildup, linked to that function.

It seems like when the context get cancelled, the consumer could be gone, but that function would still attempt to write in the channel and block forever.


aschmahmann (approved):
LGTM, thanks 🙏

PR #621, opened 2024-06-18
feat(routing/none): remove nil router

### LINKED ISSUE #620: Duplicated Nil router
There are at least two implementations of a router the does nothing:

The nil router at https://github.com/ipfs/boxo/blob/980447ea44a13729f1d6d72df9ce278608021277/routing/none/none_client.go#L49 (originally from https://github.com/ipfs/go-ipfs-routing)

The Null router at https://github.com/libp2p/go-libp2p-routing-helpers/blob/77f4db3a49f600dc6f61ae7550d1f77450334c84/null.go#L12

The latter seems like the better implementation in that it properly returns errors.

Some options for what to do here:
1. Remove the boxo version and put some info in the changelog
2. Deprecate the boxo version and delete it later
3. Stub out the nil router in boxo to just instantiate the version from go-libp2p-routing-helpers and deprecate + delete later
4. Bring the go-libp2p-routing-helpers into boxo and then break / deprecate the older version. While it could certainly live independently in practice a lot of the work on in practice I tend to see PRs to go-libp2p-routing-helpers mostly come alongside discovery of bugs, missing features, etc. needed by boxo so it might make maintainers' lives easier.

I don't see a ton of users of `ConstructNilRouting` so the removal is probably not that big a deal.

aschmahmann:
Closes #620

This removes the nil router, if people would prefer something less harsh (e.g. deprecation followed by removal, stubbing followed by deprecation + removal) comment on the issue above


gammazero (approved):
LGTM after branch update.

Be aware of this code in boost:
- https://github.com/filecoin-project/boost/blob/main/cmd/booster-bitswap/client.go#L98-L102
- https://github.com/filecoin-project/boost/blob/main/cmd/booster-bitswap/server.go#L68-L72

PR #629, opened 2024-06-23
fix(bitswap): wantlist overflow handling

### LINKED ISSUE #527: bitswap/server: wantlist overflows fails in a toxic maner preventing any data transfer

gammazero:
Handle incoming wants that can not be added to the peer ledger without exceeding the peer want limit. These are handled by trying to make room for them by canceling existing wants for which there is no block. If this does not make sufficient room, then any lower priority wants that have blocks are canceled. This fix prevents the wantlist from filling up with CIDs that the server does not have.

Priority is also considered when truncating a wantlist that exceeds the size limit. Considering priority in wantlist truncation and overflow handling ensures that higher priority tasks handled when wantlist size needs to be limited. This will be more important in the future if/when priority is determined by a block's dag path (higher priority when closer to root).

Fixes #527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ The following emojis are used to highlight certain changes:
 
 - `routing/http`: the `FindPeer` now returns `routing.ErrNotFound` when no addresses are found
 - `routing/http`: the `FindProvidersAsync` no longer causes a goroutine buildup
-- bitswap wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have.
+- `bitswap`: wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have.
 
 ## [v0.20.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ The following emojis are used to highlight certain changes:
 ### Changed
 
 - `boxo/gateway` is now tested against [gateway-conformance v6](https://github.com/ipfs/gateway-conformance/releases/tag/v0.6.0)
-- `bitswap/client` supports additional tracing 
+- `bitswap/client` supports additional tracing
 
 ### Removed
 
@@ -41,6 +41,7 @@ The following emojis are used to highlight certain changes:
 
 - `routing/http`: the `FindPeer` now returns `routing.ErrNotFound` when no addresses are found
 - `routing/http`: the `FindProvidersAsync` no longer causes a goroutine buildup
+- bitswap wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have.
 
 ## [v0.20.0]
 
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -691,18 +691,9 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 		return true
 	}
 
-	// If there is a possibility of overflow, sort the wantlist to make sure
-	// the highest priority items are put into the free space.
-	freeSpace := int(e.maxQueuedWantlistEntriesPerPeer) - e.peerLedger.WantlistSizeForPeer(p)
-	if len(wants) > freeSpace {
-		// Sort incoming wants from most to least important.
-		slices.SortFunc(wants, func(a, b bsmsg.Entry) int {
-			return cmp.Compare(b.Entry.Priority, a.Entry.Priority)
-		})
-		// Do not take more wants that can be handled.
-		if len(wants) > int(e.maxQueuedWantlistEntriesPerPeer) {
-			wants = wants[:int(e.maxQueuedWantlistEntriesPerPeer)]
-		}
+	// Do not take more wants that can be handled.
+	if len(wants) > int(e.maxQueuedWantlistEntriesPerPeer) {
+		wants = wants[:int(e.maxQueuedWantlistEntriesPerPeer)]
 	}
 
 	// Get block sizes
@@ -739,7 +730,6 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 	}
 
 	if len(overflow) != 0 {
-		// Overflow is already sorted, so no need to do it here.
 		wants = e.handleOverflow(ctx, p, overflow, wants)
 	}
 
@@ -832,14 +822,15 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 // does not make sufficient room, then any lower priority wants that have
 // blocks are canceled.
 //
-// This assumes that overflow is already sorted from most to least important.
-//
 // Important: handleOverflwo must be called e.lock is locked.
 func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, overflow, wants []bsmsg.Entry) []bsmsg.Entry {
+	// Sort overflow from most to least important.
+	slices.SortFunc(overflow, func(a, b bsmsg.Entry) int {
+		return cmp.Compare(b.Entry.Priority, a.Entry.Priority)
+	})
+	// Sort existing wants from least to most important, to try to replace
+	// lowest priority items first.
 	existingWants := e.peerLedger.WantlistForPeer(p)
-
-	// Sort wl from least to most important, to try to replace lowest priority
-	// items first.
 	slices.SortFunc(existingWants, func(a, b wl.Entry) int {
 		return cmp.Compare(b.Priority, a.Priority)
 	})
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -316,8 +316,11 @@ func WithMaxOutstandingBytesPerPeer(count int) Option {
 	}
 }
 
-// WithMaxQueuedWantlistEntriesPerPeer limits how much individual entries each peer is allowed to send.
-// If a peer send us more than this we will truncate newest entries.
+// WithMaxQueuedWantlistEntriesPerPeer limits how many individual entries each
+// peer is allowed to send. If a peer sends more than this, then the lowest
+// priority entries are truncated to this limit. If there is insufficient space
+// to enqueue new entries, then older existing wants with no associated blocks,
+// and lower priority wants, are canceled to make room for the new wants.
 func WithMaxQueuedWantlistEntriesPerPeer(count uint) Option {
 	return func(e *Engine) {
 		e.maxQueuedWantlistEntriesPerPeer = count
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -133,7 +133,7 @@ type PeerEntry struct {
 // PeerLedger is an external ledger dealing with peers and their want lists.
 type PeerLedger interface {
 	// Wants informs the ledger that [peer.ID] wants [wl.Entry].
-	Wants(p peer.ID, e wl.Entry, limit int) bool
+	Wants(p peer.ID, e wl.Entry) bool
 
 	// CancelWant returns true if the [cid.Cid] was removed from the wantlist of [peer.ID].
 	CancelWant(p peer.ID, k cid.Cid) bool
@@ -406,7 +406,6 @@ func newEngine(
 		taskWorkerCount:                 defaults.BitswapEngineTaskWorkerCount,
 		sendDontHaves:                   true,
 		self:                            self,
-		peerLedger:                      NewDefaultPeerLedger(),
 		pendingGauge:                    bmetrics.PendingEngineGauge(ctx),
 		activeGauge:                     bmetrics.ActiveEngineGauge(ctx),
 		targetMessageSize:               defaultTargetMessageSize,
@@ -418,6 +417,11 @@ func newEngine(
 
 	for _, opt := range opts {
 		opt(e)
+	}
+
+	// If peerLedger was not set by option, then create a default instance.
+	if e.peerLedger == nil {
+		e.peerLedger = NewDefaultPeerLedger(e.maxQueuedWantlistEntriesPerPeer)
 	}
 
 	e.bsm = newBlockstoreManager(bs, e.bstoreWorkerCount, bmetrics.PendingBlocksGauge(ctx), bmetrics.ActiveBlocksGauge(ctx))
@@ -687,6 +691,20 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 		return true
 	}
 
+	// If there is a possibility of overflow, sort the wantlist to make sure
+	// the highest priority items are put into the free space.
+	freeSpace := int(e.maxQueuedWantlistEntriesPerPeer) - e.peerLedger.WantlistSizeForPeer(p)
+	if len(wants) > freeSpace {
+		// Sort incoming wants from most to least important.
+		slices.SortFunc(wants, func(a, b bsmsg.Entry) int {
+			return cmp.Compare(b.Entry.Priority, a.Entry.Priority)
+		})
+		// Do not take more wants that can be handled.
+		if len(wants) > int(e.maxQueuedWantlistEntriesPerPeer) {
+			wants = wants[:int(e.maxQueuedWantlistEntriesPerPeer)]
+		}
+	}
+
 	// Get block sizes
 	wantKs := cid.NewSet()
 	for _, entry := range wants {
@@ -708,7 +726,7 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 	if len(wants) != 0 {
 		filteredWants := wants[:0] // shift inplace
 		for _, entry := range wants {
-			if !e.peerLedger.Wants(p, entry.Entry, int(e.maxQueuedWantlistEntriesPerPeer)) {
+			if !e.peerLedger.Wants(p, entry.Entry) {
 				// Cannot add entry because it would exceed size limit.
 				overflow = append(overflow, entry)
 				continue
@@ -721,7 +739,8 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 	}
 
 	if len(overflow) != 0 {
-		wants = e.handleOverflow(ctx, p, wants, overflow)
+		// Overflow is already sorted, so no need to do it here.
+		wants = e.handleOverflow(ctx, p, overflow, wants)
 	}
 
 	for _, entry := range cancels {
@@ -812,14 +831,17 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 // make room by canceling existing wants for which there is no block. If this
 // does not make sufficient room, then any lower priority wants that have
 // blocks are canceled.
-func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, wants, overflow []bsmsg.Entry) []bsmsg.Entry {
+//
+// This assumes that overflow is already sorted from most to least important.
+//
+// Important: handleOverflwo must be called e.lock is locked.
+func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, overflow, wants []bsmsg.Entry) []bsmsg.Entry {
 	existingWants := e.peerLedger.WantlistForPeer(p)
-	// Sort wl and overflow from least to most important.
+
+	// Sort wl from least to most important, to try to replace lowest priority
+	// items first.
 	slices.SortFunc(existingWants, func(a, b wl.Entry) int {
-		return cmp.Compare(a.Priority, b.Priority)
-	})
-	slices.SortFunc(overflow, func(a, b bsmsg.Entry) int {
-		return cmp.Compare(a.Entry.Priority, b.Entry.Priority)
+		return cmp.Compare(b.Priority, a.Priority)
 	})
 
 	queuedWantKs := cid.NewSet()
@@ -841,19 +863,20 @@ func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, wants, overflow 
 				e.peerRequestQueue.Remove(w.Cid, p)
 			}
 			removed = append(removed, i)
-			// Add highest priority overflow.
-			lastOver := overflow[len(overflow)-1]
-			overflow = overflow[:len(overflow)-1]
-			e.peerLedger.Wants(p, lastOver.Entry, 0)
-			wants = append(wants, lastOver)
+			// Pop hoghest priority overflow.
+			firstOver := overflow[0]
+			overflow = overflow[1:]
+			// Add highest priority overflow to wants.
+			e.peerLedger.Wants(p, firstOver.Entry)
+			wants = append(wants, firstOver)
 			if len(overflow) == 0 {
-				break
+				return wants
 			}
 		}
 	}
 
-	// Not enough dont-haves removed. Replace existing entries, that are a
-	// lower priority, with overflow entries.
+	// Replace existing entries, that are a lower priority, with overflow
+	// entries.
 	var replace int
 	for _, overflowEnt := range overflow {
 		// Do not compare with removed existingWants entry.
@@ -861,17 +884,17 @@ func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, wants, overflow 
 			replace++
 			removed = removed[1:]
 		}
-		if overflowEnt.Entry.Priority <= existingWants[replace].Priority {
-			// Everything in existingWants is equal or more improtant, so this
-			// overflow entry cannot replace any existing wants.
-			continue
+		if overflowEnt.Entry.Priority < existingWants[replace].Priority {
+			// All overflow entries have too low of priority to replace any
+			// existing wants.
+			break
 		}
 		entCid := existingWants[replace].Cid
 		replace++
 		if e.peerLedger.CancelWant(p, entCid) {
 			e.peerRequestQueue.Remove(entCid, p)
 		}
-		e.peerLedger.Wants(p, overflowEnt.Entry, 0)
+		e.peerLedger.Wants(p, overflowEnt.Entry)
 		wants = append(wants, overflowEnt)
 	}
 
@@ -921,15 +944,6 @@ func (e *Engine) splitWantsCancelsDenials(p peer.ID, m bsmsg.BitSwapMessage) ([]
 
 	if len(wants) == 0 {
 		wants = nil
-	}
-
-	// Do not take more wants that can be handled.
-	if len(wants) > int(e.maxQueuedWantlistEntriesPerPeer) {
-		// Keep the highest priority wants.
-		slices.SortFunc(wants, func(a, b bsmsg.Entry) int {
-			return cmp.Compare(b.Entry.Priority, a.Entry.Priority)
-		})
-		wants = wants[:int(e.maxQueuedWantlistEntriesPerPeer)]
 	}
 
 	// Clear truncated entries.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ The following emojis are used to highlight certain changes:
 
 - `routing/http`: the `FindPeer` now returns `routing.ErrNotFound` when no addresses are found
 - `routing/http`: the `FindProvidersAsync` no longer causes a goroutine buildup
+- bitswap wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have.
```
```suggestion
- `bitswap`: wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have.
```

bitswap/server/internal/decision/engine.go
```diff
@@ -717,39 +717,72 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+		available := int(e.maxQueuedWantlistEntriesPerPeer) - s
+		if len(wants) > available {
+			needSpace := len(wants) - available
+			log.Debugw("wantlist overflow", "local", e.self, "remote", p, "would be", s+len(wants), "canceling", needSpace)
+			// Cancel any wants that are being requested again. This makes room
+			// for new wants and minimizes that existing wants to cancel that
+			// are not in the new request.
+			for _, entry := range wants {
+				if e.peerLedger.CancelWant(p, entry.Cid) {
+					e.peerRequestQueue.Remove(entry.Cid, p)
```
gammazero: It may be better to not remove the entry from `peerRequestQueue` here, since this one is being replaced with an identical want from the new request.

Wondertan:
[This option](https://github.com/ipfs/boxo/blob/dfd4a53ba828a368cec8d61c3fe12969ac6aa94c/bitswap/server/internal/decision/engine.go#L318-L320) will need a minor change

bitswap/server/internal/decision/engine.go
```diff
@@ -717,39 +717,72 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
-			log.Warnw("peer wants an identity CID", "local", e.self, "remote", p)
-			return true
+		wants = wants[len(filteredWants):]
+		for i := range wants {
+			wants[i] = bsmsg.Entry{} // early GC
 		}
-		if e.maxCidSize != 0 && uint(entry.Cid.ByteLen()) > e.maxCidSize {
-			// Ignore requests about CIDs that big.
-			continue
+		wants = filteredWants
```
aschmahmann, line 726: Not sure which line to leave this comment on, but adding some thoughts based on conversations with @gammazero and @Stebalien

It might be helpful if we logically had two "wantlists" here:
1. The set of blocks that the server is currently in the middle of processing responses for (i.e. request-response style semantics)
2. The set of blocks that the server doesn't currently have, but if it receives them it will notify / send them out to requesters (i.e. subscription semantics)

We currently have the wantlist and taskqueue which might be these two queues, but they also might not given we want to be able to do things like cancel tasks in the taskqueue without needing to re-enumerate the entire queue in case it's large. It might also be fine if the capacity for subscription-wants only really exists if the request-response wants have been satisfied already (e.g. if there are tons of new requests coming in flushing out all the old subscriptions is probably fine, but it seems ok to have more capacity for request-response which is supposed to be "moving" than for subscriptions that can stay and take up memory indefinitely).

Also, for others watching: I think it's fairly clear that some protocol changes to help with backpressure here would be great, but since they're protocol changes those will have to wait for another time 😅.

Wondertan: (slightly tangential)
We use Bitswap with a case where more than 1024(default max) CIDs are requested in a single `GetBlocks`. If we do that, the Bitswap hangs because of this limit and recovers in a minute after rebroadcasting. It would be great if Bitswap clients could handle that case and avoid sending more than protocol-wide constant by backpressure-ing the caller.

aschmahmann: > It would be great if Bitswap clients could handle that case and avoid sending more than protocol-wide constant by backpressure-ing the caller.

Could you clarify your suggestion? Reading it I see 3 options (but these might not even be correct 😅):
1. Having the server backpressure the client on the other side of the wire
   - Similar to my comment above, and AFAICT requires some protocol changes to accommodate
2. Have the bitswap client internally batch `GetBlocks` calls that are for larger than `maxWantlistSize` into batches before returning them in case the batch is fully returned before the rebroadcast interval?
   - Doable if it'll be helpful, although it's a bit gross since A) `maxWantlistSize` shouldn't really be a protocol/network-wide thing and be per-client B) this batching can be done outside of the bitswap client package.
3. The bitswap client backpressuring the caller (e.g. code that's walking a DAG)
   - My suspicion is this would best be served by having a streaming version of `GetBlocks` that you could block on which is a separate problem that seems like https://github.com/ipfs/boxo/issues/121. Which if you're interested in let's chat on the issue there.

Wondertan: Thank you for the elaborate response! 

My suggestion was some form of 2 and 3. The client should be able to deal with the server-side request rate-limiting; otherwise, the client gets stuck expecting to be served, while the server simply cuts off his wants. The rebroadcasting after one minute helps, but that's still one minute and request spamming overhead over the wire. 

Ideally, we would do a protocol change as described in 1, but as it's breaking, we may consider other less clean options, like or similar to 2. Setting protocol-wide `maxWantlistSize` is gross, I agree. Another option might be negotiating the limit between the client and the server so the client knows it should never exceed it.

The 3 is complimentary and provides a new, powerful way to interface with a client. However, I just realized that it is not necessary for our case if the client is smart enough. In our case, we have a flat structure, i.e. we don't traverse a DAG where we unpack IPLD nodes to get more CID links to fetch them and unpack again to get to the data. Essentially, we know all the CIDs in advance and could simply ask Bitswap to get all of them over `GetBlocks` as long the client is smart enough not to get into any limitations of its immediate peers, which we are currently facing with this issue.

Wondertan: Re protocol breaking 1. We are actually fine with this being a protocol-breaking change as we are building a new Bitswap-based protocol that hasn't been deployed yet and I believe there is more or less a clear way how to handle protocol version bumps in bitswap network component.

bitswap/server/internal/decision/engine.go
```diff
@@ -871,10 +923,19 @@ func (e *Engine) splitWantsCancelsDenials(p peer.ID, m bsmsg.BitSwapMessage) ([]
 		wants = nil
 	}
 
+	// Do not take more wants that can be handled.
+	if len(wants) > int(e.maxQueuedWantlistEntriesPerPeer) {
+		// Keep the highest priority wants.
+		slices.SortFunc(wants, func(a, b bsmsg.Entry) int {
+			return cmp.Compare(b.Entry.Priority, a.Entry.Priority)
+		})
+		wants = wants[:int(e.maxQueuedWantlistEntriesPerPeer)]
```
aschmahmann: This might be fine, however do we want to do this here given that there could be duplicates once we actually look at the wantlist?
 - Pro: We shed ram usage earlier
 - Con: There might be a better set of wants that we can honor if we wait

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+	})
+	slices.SortFunc(overflow, func(a, b bsmsg.Entry) int {
+		return cmp.Compare(a.Entry.Priority, b.Entry.Priority)
+	})
+
+	queuedWantKs := cid.NewSet()
+	for _, entry := range existingWants {
+		queuedWantKs.Add(entry.Cid)
+	}
+	queuedBlockSizes, err := e.bsm.getBlockSizes(ctx, queuedWantKs.Keys())
```
aschmahmann, line 840: I'm not sure I get what the idea is here and if this is necessary / if we can make this much cheaper
- Is this meant as "is there a DONT_HAVE response queued up that we should replace".
   - While I get this it might also be overkill, and it might be fine to respect the user priority in responding with DONT_HAVEs, HAVEs, and blocks in the same way.
- Is this meant as "I previously sent a DONT_HAVE and now this is sitting on my list as a subscription".
    - As discussed this definitely seems like something we should want to knock off our list if out of space

In either case it seems like we could add some extra data to the in-memory structs here rather than going to the blockstore to see if we have the data (and being at the mercy of whatever caching, bloom filters, etc. are used there)

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+	if err != nil {
+		log.Info("aborting overflow processing", err)
+		return wants
+	}
+
+	// Remove entries for blocks that are not present to make room for overflow.
```
aschmahmann, line 846: Related to above is this about blocks that aren't present or subscriptions?

Note: the reason I'm pushing on the difference is that from my perspective subscriptions are much more expensive by virtue of occupying memory for an indefinite amount of time rather than a transient "while I'm sending out a response". Not sure if that's enough to justify different lists, but it's how I'm thinking in my review here (but lmk if you disagree or think I'm missing the point).

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+	for _, overflowEnt := range overflow {
+		// Do not compare with removed existingWants entry.
+		for len(removed) != 0 && replace == removed[0] {
+			replace++
+			removed = removed[1:]
+		}
+		if overflowEnt.Entry.Priority <= existingWants[replace].Priority {
+			// Everything in existingWants is equal or more improtant, so this
+			// overflow entry cannot replace any existing wants.
+			continue
```
aschmahmann: IIUC the overflow and existingwants lists are already sorted by priority. If so, should this be a break out of the overflow iteration?

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+			e.peerLedger.Wants(p, lastOver.Entry, 0)
+			wants = append(wants, lastOver)
+			if len(overflow) == 0 {
+				break
+			}
+		}
+	}
+
+	// Not enough dont-haves removed. Replace existing entries, that are a
+	// lower priority, with overflow entries.
```
aschmahmann: In practice I expect this never to happen given that IIUC the boxo client (which is the most widely used one) just decreases the priority over time https://github.com/ipfs/boxo/blob/733fa556c1e6963b79f35be2baad10f2428a00d5/bitswap/client/internal/messagequeue/messagequeue.go#L305

Given that this is the case it'd be good to:
1. Make this shortcut fairly cheaply when looking at the overflow + existing lists (which have already been sorted earlier) we see that the lowest priority in the existingwants is higher than the highest in the overflow list (doesn't have to be an explicit check, but anything that allows this to be pretty cheap rather than linear)
2. Have it still work (and be tested) when people choose different priorities
3. Not being particularly expensive even in pathological cases

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+		if _, found := queuedBlockSizes[w.Cid]; !found {
+			// Cancel lowest priority dont-have.
+			if e.peerLedger.CancelWant(p, w.Cid) {
+				e.peerRequestQueue.Remove(w.Cid, p)
+			}
+			removed = append(removed, i)
+			// Add highest priority overflow.
+			lastOver := overflow[len(overflow)-1]
+			overflow = overflow[:len(overflow)-1]
+			e.peerLedger.Wants(p, lastOver.Entry, 0)
```
aschmahmann: Is this (and the time below) where we set the limit to 0 necessary? If we'd set the limit to whatever our size it it would be fine too, right?

Wondering whether we actually even need to bother passing limit in the interface or we can just put it into the peerledger struct since we don't seem to need to vary that parameter

bitswap/server/internal/decision/engine.go
```diff
@@ -827,19 +798,91 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+// blocks are canceled.
+func (e *Engine) handleOverflow(ctx context.Context, p peer.ID, wants, overflow []bsmsg.Entry) []bsmsg.Entry {
+	existingWants := e.peerLedger.WantlistForPeer(p)
+	// Sort wl and overflow from least to most important.
+	slices.SortFunc(existingWants, func(a, b wl.Entry) int {
+		return cmp.Compare(a.Priority, b.Priority)
+	})
+	slices.SortFunc(overflow, func(a, b bsmsg.Entry) int {
+		return cmp.Compare(a.Entry.Priority, b.Entry.Priority)
+	})
```
aschmahmann: Perhaps premature optimization (depends on how large people set their limits to and how frequently they're hit). Seems like we could spend a lot of time sorting here.

For the overflow list this is probably fine (shouldn't be that big anyway and it's related to message size), but if nodes spend significant time near the limit they'll be doing:
1. Copy the wantlist map into a list
2. Sort the list

For basically every message that comes in.

Can the PeerRequestQueue help us out here since it's already storing a prioritized queue of what needs to be done? It might not due to how the locking/concurrency works but could save a bunch of pain.

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1733,3 +1734,132 @@ func TestKillConnectionForInlineCid(t *testing.T) {
...
+	e := newEngineForTesting(ctx, bs, fpt, "localhost", 0, WithScoreLedger(NewTestScoreLedger(shortTerm, nil, clock.New())), WithBlockstoreWorkerCount(4), WithMaxQueuedWantlistEntriesPerPeer(limit))
+	e.StartWorkers(ctx, process.WithTeardown(func() error { return nil }))
+	warsaw := engineSet{
+		Peer:       peer.ID("warsaw"),
+		PeerTagger: fpt,
+		Blockstore: bs,
+		Engine:     e,
+	}
+
+	//warsaw := newTestEngine(ctx, "warsaw", WithMaxQueuedWantlistEntriesPerPeer(limit))
```
aschmahmann: debugging code?

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1733,3 +1734,132 @@ func TestKillConnectionForInlineCid(t *testing.T) {
...
+		}
+	}
+	// Check that high priority entries are all on wantlist.
+	for _, c := range highPrioCids {
+		if !findCid(c, wl) {
+			t.Fatal("expected high priority entry on wantlist")
+		}
+	}
+
+	// These 7 new wants should overflow and 5 ot them should replace existing
```
aschmahmann:
```suggestion
	// These 7 new wants should overflow and 5 of them should replace existing
```

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1733,3 +1734,132 @@ func TestKillConnectionForInlineCid(t *testing.T) {
 		t.Fatal("connection was not killed when receiving inline in cancel")
 	}
 }
+
+func TestWantlistOverflow(t *testing.T) {
```
aschmahmann, line 1823: This is good. Although I wonder if it'd also be good to have an integration test showing how things work (and when they slow down) as a result of basically not having backpressure.

It might make it easier for people to understand what happens if they run into issues here and for us to explain some motivations behind protocol change proposals.

Doesn't seem strictly necessary but would be good to have.

aschmahmann (changes requested):
Thanks for pushing on this 🙏

bitswap/server/internal/decision/engine.go
gammazero: I think it makes sense to truncate the list, but sorting can be avoided.

Truncation makes sense:
The incoming wants are already unique, so even if there are no existing wants, or all possible wants have duplicates, there is still no way any more than the limit can be used. If all those not added to the message queue are included in the overflow then they will end up getting dropped anyway because there will not be enough existing wants available to replace. Dropping them early does two things:
- Prevents trying to add excessive number of wants to peer ledger.
- Prevents a large number of wants from being sorted and incurring block size lookup in `handleOverflow`.

Truncation without sorting:
This will potentially lose higher priority wants, but avoids sorting an incoming wantlist of unknown size. In the usual case sorting is not needed anyway, so not sorting will avoid a performance hit.

> There might be a better set of wants that we can honor if we wait

True, but then it is necessary to examine all the wants (a possibly excessive amount) to see if they can be added directly to the message queue, or need to be handled as overflow.

bitswap/server/internal/decision/engine.go
gammazero: The overflow list is sorted in ascending priority. By continuing, this ignores a priority that is lower that anything on the wantlist and moves to the next item in the overflow list that might have a higher priority and therefore be usable to replace an item on the wantlist. Example, skipping 2, 3, 3.
```
wantlist: 1 4 5 5 5 5 5 
overflow: 2 2 3 3 6 7
            \_____^   continue
```

The code will be restructured slightly to resemble the example in the [following comment](https://github.com/ipfs/boxo/pull/629#discussion_r1667034384), which should clarify things.

bitswap/server/internal/decision/engine.go
gammazero: I also think it probably will be unlikely to happen given that the client decreases priority over time. The thinking here was looking more toward a future where priority may be set by path distance, where items closer to a DAG root have a higher priority. In that case it seems more likely that as new wants are requested that some do have a high priority because they are root or closer to root items.

1. Revised logic to break out as soon as lowest usable priority is hit on incoming wantlist (now ordered from most to least important).
2. Still works.
3. In the normal case, most of the work was already done in sorting the lists, so this compares list items until the priority in the ascending-sorted list is higher than the descending-sorted list. The pathological case is where all new incoming wants are at a higher priority. This can be detected, but there is not a better way to handle it since canceling the individual overflows need to be done. Clearing the peer's wantlist does the same thing, just for all CIDs, so that is not better.

bitswap/server/internal/decision/engine_test.go
gammazero: 😳

bitswap/server/internal/decision/engine.go
gammazero: The idea here is that there is a DONT_HAVE message queued for the peer, but it has not been sent yet and is blocking new messages from being queued for the peer. So, cancel the unsent DONT_HAVE and try to enqueue something possibly more important. Either a delayed HAVE message will replace the pending DONT_HAVE, or the peer can ask again later. This should keep messages moving, even if there is some backup sending DONT_HAVE messages to peers.

This also handles the case where a DONT_HAVE message has been sent, but is not removed from the queue. Once a [message is sent](https://github.com/ipfs/boxo/blob/main/bitswap/server/server.go#L306), the want is removed from the message queue and peer ledger only when [blocks](https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L960) have been sent or when [block presence](https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L966) has been sent. If a DONT_HAVE was sent the want remains on the queue and peer ledger as a place-holder should a block arrive later, and this is stopping new wants from being accepted. This is what the 5th bullet in #527 is referring to by:
> This is because the bitswap server never cleanup entries after sending DONT_HAVE

So, in short, it handles both cases.

> it seems like we could add some extra data to the in-memory structs here rather than going to the blockstore.

Yes, the wants for which block is found can be recorded in the peer ledger so that these can be ignored in overflow handling. However, that would need to be done in every call to `engine.MessageReceived`, and seems less preferable than doing something more expensive only during the exceptional case.

The task queue does already have this info, but this would require locking the tasqueue and the peer tracker for each overflow want CID to look at. Or, this would require a new taskqueue API to get a list of wants with `HaveBlock` set to `true` for a given peer. This last option might be less expensive than looking at the blockstore, but I was not comfortable with that amount of new plumbing for handling this bitswap exceptional case. WDYT?

bitswap/server/internal/decision/engine.go
gammazero: From the engine perspective, I do not think there is any need for distinction between subscription and request-response since that _I think_ only determines how long a peer is in the task queue/ledger.

Overall, it probably does make more sense to only do this overflow handling for subscriptions. I was thinking/hoping this would handle itself by subscriptions being the ones primarily affected in the first place and needing to do overflow handling. I think some real-world use is necessary to determine this. I will add logging that can be used to determine when overflow handling is happening.

bitswap/server/internal/decision/engine.go
gammazero: Overflow is now ordered most to least important, so it can break out as soon as a more important (higher priority) is seen in the existing want list.

bitswap/server/internal/decision/engine.go
gammazero: Now set in PeerLedger struct.

bitswap/server/internal/decision/engine.go
gammazero: The PeerRequestQueue cannot really help out, since that maintains a heap so getting an ordered list would require iterating the heap. I do not see a much less expensive alternative, other than just clearing the peer's want list at some point, like if overflow is happening too frequently.

Maybe if overflow happens 5 times in a row for a particular peer, then clear that peer's message queue? WDYT?

bitswap/server/internal/decision/engine_test.go
gammazero: Agree.

The whole problem stems from only removing wants when [blocks](https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L960) or [block presence](https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L966) has been sent. That, and the somewhat incorrect (IMO) [truncation of the incoming wants](https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L705-L714) based on the remaining space left in the message queue / peer ledger, before even updating existing wants. An integration test should only have to send a couple messages where the DONT_HAVE blocks add up to the queue size limit,  and before this PR bitswap should stop working, and after this PR it should continue working.

bitswap/server/internal/decision/engine.go
gammazero: @Wondertan With the latest changes in this PR (as of the date of this comment), the main problem should be solved. The problem was that wants that have DONT_HAVE responses remain on the queue, and cause new incoming messages to be dropped because there is no more free space on the queue (failure shown by `TestWantlistBlocked` when run against previous code). The same failure also prevented new wants from updating existing wants. This is fixed in this PR: Incoming wants are now applied to the queue, and any new ones that cannot be handled without increasing the queue over the size limit are handled specially as overflow, where their priority is considered for possibly replacing existing wants. Please take a look at the latest.

I still think that having the client limit its message sizes will be useful to avoid oversized wantlists from being truncated and allowing any overflow is handled. The client can also group batches of wants by priority if applicable. If the future, priority can be influenced by dag path (higher priority for items closer to dag root).

bitswap/server/internal/decision/engine_test.go
lidel, line 1: :+1:

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1733,3 +1734,235 @@ func TestKillConnectionForInlineCid(t *testing.T) {
...
+		t.Fatal("Sanity Check: Peers have same Key!")
+	}
+
+	warsaw.Engine.MessageReceived(ctx, riga.Peer, m)
+	// Check that the wantlist is at the size limit.
+	wl := warsaw.Engine.WantlistForPeer(riga.Peer)
+	if len(wl) != limit {
+		t.Fatal("wantlist size", len(wl), "does not match limit", limit)
+	}
+	t.Log("Senr message with", limit, "medium-priority wants and", limit-1, "have blocks present")
```
lidel:
```suggestion
	t.Log("Sent message with", limit, "medium-priority wants and", limit-1, "have blocks present")
```

bitswap/server/internal/decision/engine.go
```diff
@@ -132,9 +133,9 @@ type PeerEntry struct {
 // PeerLedger is an external ledger dealing with peers and their want lists.
 type PeerLedger interface {
 	// Wants informs the ledger that [peer.ID] wants [wl.Entry].
```
lidel, line 135:
```suggestion
	// Wants informs the ledger that [peer.ID] wants [wl.Entry].
	// If peer ledger exceed internal limit, then the entry is not added
	// and false is returned.
```

lidel (approved):
Smoke tested in https://github.com/ipfs/kubo/pull/10462 and lgtm.


Release / test plan: I'll apply cosmetics below and tag boxo release on Monday and bubble up to kubo 0.30.0-rc1 (https://github.com/ipfs/kubo/issues/10436), then deploy  rc1 to collab cluster and see if it impacts performance of its nodes.

bitswap/server/internal/decision/engine.go
gammazero: This conversation moved to issue #644

PR #630, opened 2024-06-24
fix(gateway): ipld.ErrNotFound should result in a 404

### LINKED ISSUE #593: gateway: ipld.ErrNotFound should result in a 404
It's a somewhat rare case, but it's possible for a gateway backend to return `ipld.ErrNotFound`:
- using an offline exchange, for example with `--offline` in kubo
- possibly https://github.com/ipfs/kubo/blob/master/docs/config.md#gatewaynofetch
- implementing some form of denylist or restriction at the storage/exchange level (which we do at Infura)

In those situations, kubo `v0.22.0` / boxo `v0.11.0` would return a 404 in the gateway, but the implementation rework changed that to return a 500.

See also https://github.com/ipfs/boxo/pull/440/files#r1297472794

I think 404 is the correct answer, as this is not a timeout while looking for content on the p2p network, or some missing link in an IPLD node (which would mean an incorrect request). It's a definitive answer that the content does not exist.

Optionally, another status code could be returned in some more specific situations (see https://github.com/ipfs/boxo/issues/591), but 404 seem to be the correct default to me.

https://github.com/MichaelMure/boxo/commit/d088e1ea7664f04bd90e20ee3ec8a9ba4147a178 trivially change that returned status, with no other impact as far as I can tell.

MichaelMure:
Fix https://github.com/ipfs/boxo/issues/593

In case of blockstore restriction like with --offline or similar restriction, returning a 500 is inapropriate.

It's also going back to a previous gateway behavior (at least in kubo v0.22, possibly later).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
-- `boxo/gateway` now returns 404 Status Not Found instead of 500 when the requested data cannot be found, without a fallback on bitswap or similar restriction.
+- `boxo/gateway` now correctly returns 404 Status Not Found instead of 500 when the requested content cannot be found due to offline exchange, gateway running in no-fetch (non-recursive) mode, or a similar restriction that only serves a specific set of CIDs.
 - `bitswap/client` fix memory leak in BlockPresenceManager due to unlimited map growth.
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


lidel (approved):
Thank you for fixing this @MichaelMure. Took bit longer to review as the semantic meaning of this ipld error type changed multiple times over the years, and we also had big refactor since v0.22, but it seems to do the trick with how things are today.

Tested in https://github.com/ipfs/kubo/pull/10466 and lgtm (we had NoFetch  tests that expected 500, changed them to 404).
Sharness tests failing here can be ignored (circular dep, will clear up once we ship releases in both places).

This will be included in Kubo 0.30 (https://github.com/ipfs/kubo/issues/10436).

PR #632, opened 2024-06-27
fix(bitswap): use unique logger name for MessageQueue

Wondertan:
`bitwap` is already taken by the root pkg and there is no way to access the MessageQueue logger

### DIFF
--- a/bitswap/client/internal/messagequeue/messagequeue.go
+++ b/bitswap/client/internal/messagequeue/messagequeue.go
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	log   = logging.Logger("bitswap")
+	log   = logging.Logger("bs:msgqueue")
 	sflog = log.Desugar()
 )
 


lidel (commented):
@Wondertan would you be up for unifying logger names in `boxo/bitswap`? 

We seem to have mix of `bs:foo` and [`bitswap_foo`](https://github.com/ipfs/boxo/blob/dfd4a53ba828a368cec8d61c3fe12969ac6aa94c/bitswap/network/ipfs_impl.go#L29).

Perhaps go with `bitswap:foo` everywhere?

gammazero:
@lidel I think we should use `bitswap/foo` as that seems to be the typical convention in other parts of boxo..

I went ahead and made a PR to do that: #637. If that looks good we can use it in place of this.

lidel:
Sgtm, let's continue in https://github.com/ipfs/boxo/pull/637
Thank you for prompting this cleanup @Wondertan  :+1:

Issue #574, opened 2024-02-01
bitswap/client: memory leak on BlockPresenceManager


Dreamacro:
Recently, I observed that when using bitswap for fetching data, there is a large amount of memory consumption and no release after downloading.

I use pprof and find most heap is occupied here.
https://github.com/ipfs/boxo/blob/89bceff34bf108a46795d26938e6aa8c2f876fc4/bitswap/client/internal/blockpresencemanager/blockpresencemanager.go#L40

I try to print stat after `bpm.presence[c] = make(map[peer.ID]bool)` and I found `bpm.presence` would delete on `BlockPresenceManager.RemoveKeys`, so I added some logs.

I found that there is a race condition problem here.

1. `BlockPresenceManager.RemoveKeys` has been called and remove the CidA
2. `BlockPresenceManager.ReceiveFrom` has been called and created `bpm.presence[CidA]`
3. `bpm.presence[CidA]` is no longer used or removed, and the map is getting bigger and bigger.

aschmahmann:
@Dreamacro can you attach or link to a pprof dump for analysis? Also, which versions of boxo and go-libp2p are you using?

Dreamacro:
@aschmahmann 

I first provide a few fragments of pprof/heap. I can confirm a memory leak here because the leakage disappears when I replace `map[cid.Cid]map[peer.ID]bool` with a cache with time. https://github.com/Dreamacro/boxo/commit/df0a587788c0e8ea70d22d051b11dae8580ad629



heap fragments: (Although boxo 0.15 is shown below, it has been upgraded to boxo 0.17)
```
258: 1387008 [2230: 11988480] @ 0xcd2266 0xcd3d10 0xcd7025 0x1c257ec 0x1c257f0 0x1c549fc 0x1c58ade 0x1c58ffa 0x1c6df45 0x1c39d31 0x1b0a0df 0x1b09108 0x1b32687 0xd32fe1
#	0x1c257eb	github.com/ipfs/boxo/bitswap/client/internal/blockpresencemanager.(*BlockPresenceManager).updateBlockPresence+0x22b	github.com/ipfs/boxo@v0.15.0/bitswap/client/internal/blockpresencemanager/blockpresencemanager.go:48
[...]

PR #636, opened 2024-07-08
fix(bitswap): memory leak on BlockPresenceManager

gammazero:
Limit BlockPresenceManager map growth be doing the following:

- Use nil map when BlockPresenceManager CID map is empty.
- Delete peer map when from CID map when peer map is empty.
- Remove peers from BlockPresenceManager when peers are pruned from session.

This allows GC to free memory when maps in BlockPresenceManager become empty.

Additional improvement: Do not look for lists of keys or peers in empty maps; return early instead.

Fix for issue #574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


lidel (approved):
Optimizations lgtm, even if not fully solving #574, they make sense on their own.
Let's land this and ship together with #629.

PR #637, opened 2024-07-09
refactor(bitswap): unify logger names

gammazero:
Use uniform pattern for logger naming: bitswap/pkgname

Slash delimiters are used to separate path segments used in logger names. This follows the convention used most commonly in other parts of boxo.


lidel (approved):
Lgtm, cleaning up inconsistency in naming here should make it easier to follow what is happening.

PR #638, opened 2024-07-13
test: use ipfs/go-test to consolidate code to generate test data

gammazero:
The go-test packages provides functionality for common types of test data generation. Use go-test for generating test data instead of internal utility packages.

The random test data can be generated deterministically by setting an initial seed if doing so is necessary to recreate certain test data.


util/util.go
```diff
@@ -53,20 +54,37 @@ func ExpandPathnames(paths []string) ([]string, error) {
...
+	return &randGen{
+		Rand: *rand.New(src),
+	}
+}
+
+func (r *randGen) Read(p []byte) (n int, err error) {
+	for i := 0; i < len(p); i++ {
+		p[i] = byte(r.Rand.Intn(255))
+	}
+	return len(p), nil
```
Jorropo:
```suggestion
	return rand.New(rand.NewSource(seed))
```

util/util.go
```diff
@@ -53,20 +54,37 @@ func ExpandPathnames(paths []string) ([]string, error) {
 	return out, nil
 }
 
+type randGen struct {
+	rand.Rand
+}
+
```
Jorropo:
```suggestion
```

util/util.go
```diff
@@ -53,20 +54,37 @@ func ExpandPathnames(paths []string) ([]string, error) {
...
 // NewTimeSeededRand returns a random bytes reader
 // which has been initialized with the current time.
 //
 // Deprecated: use github.com/ipfs/go-test/random instead.
 func NewTimeSeededRand() io.Reader {
-	return random.NewRand()
+	src := rand.NewSource(time.Now().UnixNano())
+	return &randGen{
+		Rand: *rand.New(src),
+	}
```
Jorropo:
```suggestion
	return rand.New(rand.NewSource(seed))
```

lidel (approved):
Thank you! (I'll try to resolve conflict and if CI remains green, merge later)

PR #639, opened 2024-07-25
fix(gateway): Cache-Control header for UnixFS directories

acejam:
Cache-Control headers are currently not being returned when viewing the directory index page of a UnixFS directory. This is because `rq.ttl` returns `0`.

For example, the following request path is missing a `Cache-Control` header: `/ipfs/<cid>/`

According to the Path Gateway Specification:

> Cache-Control: public, max-age=29030400, immutable must be returned for every immutable resource under /ipfs/ namespace.

https://specs.ipfs.tech/http-gateways/path-gateway/#cache-control-response-header

This PR sets the appropriate header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The following emojis are used to highlight certain changes:
 
 - `bitswap/server` minor memory use and performance improvements
 - `bitswap` unify logger names to use uniform format bitswap/path/pkgname
+- `gateway` now always returns meaningful cache-control headers for generated HTML listings of UnixFS directories
 
 ### Removed
 
--- a/gateway/handler_unixfs_dir.go
+++ b/gateway/handler_unixfs_dir.go
@@ -136,9 +136,14 @@ func (i *handler) serveDirectory(ctx context.Context, w http.ResponseWriter, r *
 	dirEtag := getDirListingEtag(resolvedPath.RootCid())
 	w.Header().Set("Etag", dirEtag)
 
-	// Add TTL if known.
+	// Set Cache-Control
 	if rq.ttl > 0 {
-		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", int(rq.ttl.Seconds())))
+		// Use known TTL from IPNS Record or DNSLink TXT Record
+		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d, stale-while-revalidate=2678400", int(rq.ttl.Seconds())))
+	} else {
+		// Cache for 1 week, serve stale cache for up to a month
+		// (style of generated HTML may change, should not be cached forever)
+		w.Header().Set("Cache-Control", "public, max-age=604800, stale-while-revalidate=2678400")
 	}
 
 	if r.Method == http.MethodHead {


gateway/handler_unixfs_dir.go
```diff
@@ -136,10 +136,8 @@ func (i *handler) serveDirectory(ctx context.Context, w http.ResponseWriter, r *
 	dirEtag := getDirListingEtag(resolvedPath.RootCid())
 	w.Header().Set("Etag", dirEtag)
 
-	// Add TTL if known.
-	if rq.ttl > 0 {
-		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", int(rq.ttl.Seconds())))
-	}
+	// Set Cache-Control
+	w.Header().Set("Cache-Control", immutableCacheControl)
```
lidel: @acejam mind switching to something like below? Your version broke TTL-based value we already correctly returned for `/ipns/{libp2p-key}` paths.

```suggestion
	// Set Cache-Control
	if rq.ttl > 0 { 
	  // Use known TTL from IPNS Record or DNSLink TXT Record
	  w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d, stale-while-revalidate=29030400", int(rq.ttl.Seconds())))
	} else {
	  // Cache for 1 week, serve stale cache for  up to a month
	  // (style of generated HTML may change, should not be cached forever)
          w.Header().Set("Cache-Control", "public, max-age=604800, stale-while-revalidate=29030400")
  }  
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The following emojis are used to highlight certain changes:
 
 - `bitswap/server` minor memory use and performance improvements
 - `bitswap` unify logger names to use uniform format bitswap/path/pkgname
+- `gateway` now returns immutable cache-control headers for unixfs directories
```
```suggestion
- `gateway` now always returns meaningful cache-control headers for generated HTML listings of UnixFS directories
```

lidel (changes requested):
Thank you for raising this. 

This is bit more tricky, because HTML listing is quality-of-life improvement for web browsers, and not the actual immutable data.

Generated HTML listing of UnixFS directory is NOT immutable and should not be cached forever.

- Proper fix for `/ipns/` is to implement TTL handling.
- Proper fix for `/ipfs/` may not exist. A fair compromise may be to cache for a week, and also allow for serving stale cache while check for update happens asynchronously. This ensures CDNs serve content fast, but we dont cache generated HTML forever. See proposed value inline.

Would you have time to submit related PRs to
- https://github.com/ipfs/specs/blob/main/src/http-gateways/path-gateway.md#cache-control-response-header
- https://github.com/ipfs/gateway-conformance 

?

We need to update conformance tests before this can be shipped. (CI is currently failing).

gateway/handler_unixfs_dir.go
acejam: Done

CHANGELOG.md
acejam: 👍

acejam:
@lidel `gateway-conformance` PR here: https://github.com/ipfs/gateway-conformance/pull/218

lidel:
Thanks! Had to push some test fixes but your fork does not allow it, so continuing in https://github.com/ipfs/boxo/pull/643

PR #641, opened 2024-07-26
Bitswap: move providing -> Exchange-layer, providerQueryManager -> routing

### LINKED ISSUE #640: Remove requirement on ProviderQueryManager from the bitswap client
## Background

This is the ProviderQueryManager that is in an internal package of the bitswap client

https://github.com/ipfs/boxo/blob/2816b715439a3472ccf7d78e04d5db3b22952728/bitswap/client/internal/providerquerymanager/providerquerymanager.go#L71-L78

Notably it does almost nothing that's specific to Bitswap, or even the concept of data transfer sessions. Instead it is an opinionated content routing wrapper that does things like:
1. Have max timeouts
2. Have predetermined max provider records to use
3. Does preconnects before returning the peers
4. Deduplicates ongoing queries
5. Rate limits requests

Unfortunately sometimes these opinions are too strong for consumers who'd like to experiment or use something else (e.g. higher rate limits, different timeouts, etc.

## Proposal

- Move the ProviderQueryManager package into the routing package
- Refactor it to match the `FindProvidersAsync` method from go-libp2p's content routing interface
- Continue using it by default in the bitswap client, but have an option to not use it
- Add some cleanup / options to the ProviderQueryManager for now and consider separating out the pieces later

Some advantages of this proposal are:
1. If these wrappers are needed/useful elsewhere they can be
2. While we can for now leave the ProviderQueryManager used by the Bitswap client by default we can add an option to disable it and just use the ContentRouting directly
3. While doing 2 we can allow the user to easily provide a modified ProviderQueryManager if it's useful to them by just passing it in as the ContentRouter.

## Alternatives Considered

- Do experiments in forks for the best configuration and upstream to everyone in bitswap/client -> Too cumbersome and inflexible
- Extract into `routing` but disable by default in the bitswap client -> could potentially cause problems for people updating without reading release notes, but would be much cleaner since then we could just leave this as a content routing wrapper
- Extract the ProviderQueryManager into multiple different components for each of the jobs it does -> Might be a good idea, but can also happen later. Particularly if we make the changes in the near future any breakages people experience won't be significant
- Extract into https://github.com/libp2p/go-libp2p-routing-helpers -> I sort of wish this package was already moved into boxo (maybe it should be) and at the moment I'd rather not add more things to go-libp2p-routing-helpers that could reasonably fit in boxo unless downstreams want it to be separate (if you're a downstream and you care, here's where to let me know I'm wrong 😉)

aschmahmann:
fixes #640

This is an attempt at splitting the ProviderQueryManager out of Bitswap so that we can configure it more in consumers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,52 @@ The following emojis are used to highlight certain changes:
 
 ## [Unreleased]
 
+- `bitswap`, `routing`, `exchange` ([#641](https://github.com/ipfs/boxo/pull/641)):
+  - ✨ Bitswap is no longer in charge of providing blocks to the newtork: providing functionality is now handled by a `exchange/providing.Exchange`, meant to be used with `provider.System` so that all provides follow the same rules (multiple parts of the code where handling provides) before.
+  - 🛠 `bitswap/client/internal/providerquerymanager` has been moved to `routing/providerquerymanager` where it belongs. In order to keep compatibility, Bitswap now receives a `routing.ContentDiscovery` parameter which implements `FindProvidersAsync(...)` and uses it to create a `providerquerymanager` with the default settings as before. Custom settings can be used by using a custom `providerquerymanager` to manually wrap a `ContentDiscovery` object and pass that in as `ContentDiscovery` on initialization while setting `bitswap.WithDefaultProviderQueryManager(false)` (to avoid re-wrapping it again).
+  - The renovated `providedQueryManager` will trigger lookups until it manages to connect to `MaxProviders`. Before it would lookup at most `MaxInProcessRequests*MaxProviders` and connection failures may have limited the actual number of providers found.
+  - 🛠 We have aligned our routing-related interfaces with the libp2p [`routing`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/routing#ContentRouting) ones, including in the `reprovider.System`.
+  - In order to obtain exactly the same behaviour as before (i.e. particularly ensuring that new blocks are still provided), what was done like:
+
+```go
+	bswapnet := network.NewFromIpfsHost(host, contentRouter)
+	bswap := bitswap.New(p.ctx, bswapnet, blockstore)
+	bserv = blockservice.New(blockstore, bswap)
+```
+  - becomes:
+
+```go
+	// Create network: no contentRouter anymore
+	bswapnet := network.NewFromIpfsHost(host)
+	// Create Bitswap: a new "discovery" parameter, usually the "contentRouter"
+	// which does both discovery and providing.
+	bswap := bitswap.New(p.ctx, bswapnet, discovery, blockstore)
+	// A provider system that handles concurrent provides etc. "contentProvider"
+	// is usually the "contentRouter" which does both discovery and providing.
+	// "contentProvider" could be used directly without wrapping, but it is recommended
+	// to do so to provide more efficiently.
+	provider := provider.New(datastore, provider.Online(contentProvider)
+	// A wrapped providing exchange using the previous exchange and the provider.
+	exch := providing.New(bswap, provider)
+
+	// Finally the blockservice
+	bserv := blockservice.New(blockstore, exch)
+	...
+```
+
+  - The above is only necessary if content routing is needed. Otherwise:
+
+```
+	// Create network: no contentRouter anymore
+	bswapnet := network.NewFromIpfsHost(host)
+	// Create Bitswap: a new "discovery" parameter set to nil (disable content discovery)
+	bswap := bitswap.New(p.ctx, bswapnet, nil, blockstore)
+	// Finally the blockservice
+	bserv := blockservice.New(blockstore, exch)
+```
+
+
+
 ### Added
 
 - `routing/http/server`: added built-in Prometheus instrumentation to http delegated `/routing/v1/` endpoints, with custom buckets for response size and duration to match real world data observed at [the `delegated-ipfs.dev` instance](https://docs.ipfs.tech/concepts/public-utilities/#delegated-routing). [#718](https://github.com/ipfs/boxo/pull/718) [#724](https://github.com/ipfs/boxo/pull/724)
@@ -117,6 +163,7 @@ The following emojis are used to highlight certain changes:
 - `bitswap/client` fix memory leak in BlockPresenceManager due to unlimited map growth. [#636](https://github.com/ipfs/boxo/pull/636)
 - `bitswap/network` fixed race condition when a timeout occurred before hole punching completed while establishing a first-time stream to a peer behind a NAT [#651](https://github.com/ipfs/boxo/pull/651)
 - `bitswap`: wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have. [#629](https://github.com/ipfs/boxo/pull/629)
+- 🛠 `bitswap` & `bitswap/server` no longer provide to content routers, instead you can use the `provider` package because it uses a datastore queue and batches calls to ProvideMany.
 
 ## [v0.21.0]
 
--- a/bitswap/client/bitswap_with_sessions_test.go
+++ b/bitswap/client/bitswap_with_sessions_test.go
@@ -26,16 +26,20 @@ const blockSize = 4
 func getVirtualNetwork() tn.Network {
 	// FIXME: the tests are really sensitive to the network delay. fix them to work
 	// well under varying conditions
-	return tn.VirtualNetwork(mockrouting.NewServer(), delay.Fixed(0))
+	return tn.VirtualNetwork(delay.Fixed(0))
 }
 
 func addBlock(t *testing.T, ctx context.Context, inst testinstance.Instance, blk blocks.Block) {
 	t.Helper()
-	err := inst.Blockstore().Put(ctx, blk)
+	err := inst.Blockstore.Put(ctx, blk)
 	if err != nil {
 		t.Fatal(err)
 	}
 	err = inst.Exchange.NotifyNewBlocks(ctx, blk)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = inst.Routing.Provide(ctx, blk.Cid(), true)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -46,7 +50,8 @@ func TestBasicSessions(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, nil)
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, nil)
 	defer ig.Close()
 
 	block := random.BlocksOfSize(1, blockSize)[0]
@@ -56,7 +61,7 @@ func TestBasicSessions(t *testing.T) {
 	b := inst[1]
 
 	// Add a block to Peer B
-	if err := b.Blockstore().Put(ctx, block); err != nil {
+	if err := b.Blockstore.Put(ctx, block); err != nil {
 		t.Fatal(err)
 	}
 
@@ -78,7 +83,7 @@ func TestBasicSessions(t *testing.T) {
 		t.Fatal("did not get tracable block")
 	}
 
-	if traceBlock.From != b.Peer {
+	if traceBlock.From != b.Identity.ID() {
 		t.Fatal("should have received block from peer B, did not")
 	}
 }
@@ -111,15 +116,16 @@ func TestSessionBetweenPeers(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	vnet := tn.VirtualNetwork(mockrouting.NewServer(), delay.Fixed(time.Millisecond))
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, []bitswap.Option{bitswap.SetSimulateDontHavesOnTimeout(false)})
+	vnet := tn.VirtualNetwork(delay.Fixed(time.Millisecond))
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, []bitswap.Option{bitswap.SetSimulateDontHavesOnTimeout(false)})
 	defer ig.Close()
 
 	inst := ig.Instances(10)
 
 	// Add 101 blocks to Peer A
 	blks := random.BlocksOfSize(101, blockSize)
-	if err := inst[0].Blockstore().PutMany(ctx, blks); err != nil {
+	if err := inst[0].Blockstore.PutMany(ctx, blks); err != nil {
 		t.Fatal(err)
 	}
 
@@ -147,7 +153,7 @@ func TestSessionBetweenPeers(t *testing.T) {
 		for b := range ch {
 			got = append(got, b)
 		}
-		if err := assertBlockListsFrom(inst[0].Peer, got, blks[i*10:(i+1)*10]); err != nil {
+		if err := assertBlockListsFrom(inst[0].Identity.ID(), got, blks[i*10:(i+1)*10]); err != nil {
 			t.Fatal(err)
 		}
 	}
@@ -171,7 +177,8 @@ func TestSessionSplitFetch(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, nil)
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, nil)
 	defer ig.Close()
 
 	inst := ig.Instances(11)
@@ -179,7 +186,7 @@ func TestSessionSplitFetch(t *testing.T) {
 	// Add 10 distinct blocks to each of 10 peers
 	blks := random.BlocksOfSize(100, blockSize)
 	for i := 0; i < 10; i++ {
-		if err := inst[i].Blockstore().PutMany(ctx, blks[i*10:(i+1)*10]); err != nil {
+		if err := inst[i].Blockstore.PutMany(ctx, blks[i*10:(i+1)*10]); err != nil {
 			t.Fatal(err)
 		}
 	}
@@ -203,7 +210,7 @@ func TestSessionSplitFetch(t *testing.T) {
 		for b := range ch {
 			got = append(got, b)
 		}
-		if err := assertBlockListsFrom(inst[i].Peer, got, blks[i*10:(i+1)*10]); err != nil {
+		if err := assertBlockListsFrom(inst[i].Identity.ID(), got, blks[i*10:(i+1)*10]); err != nil {
 			t.Fatal(err)
 		}
 	}
@@ -214,7 +221,8 @@ func TestFetchNotConnected(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, []bitswap.Option{bitswap.ProviderSearchDelay(10 * time.Millisecond)})
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, []bitswap.Option{bitswap.ProviderSearchDelay(10 * time.Millisecond)})
 	defer ig.Close()
 
 	other := ig.Next()
@@ -236,7 +244,6 @@ func TestFetchNotConnected(t *testing.T) {
 	thisNode := ig.Next()
 	ses := thisNode.Exchange.NewSession(ctx).(*session.Session)
 	ses.SetBaseTickDelay(time.Millisecond * 10)
-
 	ch, err := ses.GetBlocks(ctx, cids)
 	if err != nil {
 		t.Fatal(err)
@@ -246,7 +253,7 @@ func TestFetchNotConnected(t *testing.T) {
 	for b := range ch {
 		got = append(got, b)
 	}
-	if err := assertBlockListsFrom(other.Peer, got, blks); err != nil {
+	if err := assertBlockListsFrom(other.Identity.ID(), got, blks); err != nil {
 		t.Fatal(err)
 	}
 }
@@ -256,7 +263,8 @@ func TestFetchAfterDisconnect(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, []bitswap.Option{
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, []bitswap.Option{
 		bitswap.ProviderSearchDelay(10 * time.Millisecond),
 		bitswap.RebroadcastDelay(delay.Fixed(15 * time.Millisecond)),
 	})
@@ -294,12 +302,12 @@ func TestFetchAfterDisconnect(t *testing.T) {
 		got = append(got, b)
 	}
 
-	if err := assertBlockListsFrom(peerA.Peer, got, blks[:5]); err != nil {
+	if err := assertBlockListsFrom(peerA.Identity.ID(), got, blks[:5]); err != nil {
 		t.Fatal(err)
 	}
 
 	// Break connection
-	err = peerA.Adapter.DisconnectFrom(ctx, peerB.Peer)
+	err = peerA.Adapter.DisconnectFrom(ctx, peerB.Identity.ID())
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -323,7 +331,7 @@ func TestFetchAfterDisconnect(t *testing.T) {
 		}
 	}
 
-	if err := assertBlockListsFrom(peerA.Peer, got, blks); err != nil {
+	if err := assertBlockListsFrom(peerA.Identity.ID(), got, blks); err != nil {
 		t.Fatal(err)
 	}
 }
@@ -333,7 +341,8 @@ func TestInterestCacheOverflow(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, nil)
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, nil)
 	defer ig.Close()
 
 	blks := random.BlocksOfSize(2049, blockSize)
@@ -382,7 +391,8 @@ func TestPutAfterSessionCacheEvict(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, nil)
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, nil)
 	defer ig.Close()
 
 	blks := random.BlocksOfSize(2500, blockSize)
@@ -419,7 +429,8 @@ func TestMultipleSessions(t *testing.T) {
 	defer cancel()
 
 	vnet := getVirtualNetwork()
-	ig := testinstance.NewTestInstanceGenerator(vnet, nil, nil)
+	router := mockrouting.NewServer()
+	ig := testinstance.NewTestInstanceGenerator(vnet, router, nil, nil)
 	defer ig.Close()
 
 	blk := random.BlocksOfSize(1, blockSize)[0]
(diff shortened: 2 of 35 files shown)


bitswap/client/internal/session/session.go
```diff
@@ -78,6 +81,41 @@ type ProviderFinder interface {
 	FindProvidersAsync(ctx context.Context, k cid.Cid) <-chan peer.ID
 }
 
+type FindAllProviders struct {
+	Router bsnet.BitSwapNetwork
+}
```
aschmahmann: If we're going to keep the function here as not taking a `count` the name should get changed from `FindProvidersAsync` to one that doesn't collide with the more widely used function that does take a count so implementations can use both if they want.

bitswap/client/internal/session/session.go
```diff
@@ -78,6 +81,41 @@ type ProviderFinder interface {
...
+			case <-ctx.Done():
+				return
+			}
+		}(p)
+	}
+	go func() {
+		wg.Wait()
+		close(provCh)
+	}()
+	return provCh
```
aschmahmann: Note: this was copied out of the ProviderQueryManager because the internals (or at least the tests) seem to rely on having the connections made explicit here. It seems reasonable to extract this up a level out of the routing component though.

gammazero:
@aschmahmann Does this mean we still want the provider query manager, and need to close #536 which removes it.

bitswap/options.go
```diff
@@ -99,3 +99,11 @@ func WithTracer(tap tracer.Tracer) Option {
...
 	}
 }
+
+func WithClientOption(opt client.Option) Option {
+	return Option{opt}
+}
+
+func WithServerOption(opt server.Option) Option {
+	return Option{opt}
+}
```
hsanjuan, line 112: ?

bitswap/client/client.go
```diff
@@ -99,6 +99,15 @@ func WithoutDuplicatedBlockStats() Option {
 	}
 }
 
+// WithDefaultLookupManagement indicates whether a default manager for handling searching of new providers of data
+// should be used or whether the passed content router should be used directly.
+// See [defaults.WithDefaultLookupManagement] for the default.
+func WithDefaultLookupManagement(b bool) Option {
+	return func(bs *Client) {
+		bs.useDefaultLookupManagement = true
+	}
```
hsanjuan: In #535 , the Network is separated from the Router and that makes sense to me. As a user, if I want to use a custom Router, I am going to have to glue the normal Network (ConnectTo) and my custom Router (FindProvidersAsync) into a custom BitswapNetwork. I'm not sure if that requires much more refactoring though, given that #535 is based on other additional changes.

This would also allow to disable the ProviderQueryManager altogether when it is not passed in (here it's impossible).

hsanjuan (approved):
I guess this comes with the need not to replace the default provider itself, but to be able to create it by hand and turn the knobs (and guessing it's mostly about increasing the amount of workers).

It is similar to #535 although that includes some other things (FindProvidersAsync returns AddrInfos instead of PeerIDs, and Router and Network are separate things).

If we need it we can merge this, otherwise the direction of #535 seems better but I guess it's a more invasive change (haven't followed too much). Code-wise I don't see anything weird, so green light on that front.

bitswap/options.go
aschmahmann: This was an unrelated change and should be a separate PR, I'll raise it separately. TLDR though is that `bitswap.New` can take options for the client or the server, but right now the only way to construct them is if they're forwarded as regular Bitswap options, otherwise you have to construct the client and server yourself. This saves on duplication while being more explicit and was something I ran into here because a new client option was introduced.

bitswap/network/ipfs_impl.go
```diff
@@ -365,33 +364,16 @@ func (bsnet *impl) Stop() {
...
-			}
-			bsnet.host.Peerstore().AddAddrs(info.ID, info.Addrs, peerstore.TempAddrTTL)
-			select {
-			case <-ctx.Done():
-				return
-			case out <- info.ID:
-			}
-		}
-	}()
-	return out
```
hsanjuan, line 392: All of this is not needed because our query manager is doing `Connect(peerAddr)` instead of `ConnectTo(peerID)`. Addresses will be absorbed into the peerstore directly. At his point BSnet is only a convenience to have content router available where it is not passed explicitally.

routing/providerquerymanager/providerquerymanager.go
```diff
@@ -245,20 +292,20 @@ func (pqm *ProviderQueryManager) findProviderWorker() {
...
 			wg := &sync.WaitGroup{}
 			for p := range providers {
 				wg.Add(1)
-				go func(p peer.ID) {
+				go func(p peer.AddrInfo) {
 					defer wg.Done()
-					span.AddEvent("FoundProvider", trace.WithAttributes(attribute.Stringer("peer", p)))
-					err := pqm.network.ConnectTo(findProviderCtx, p)
+					span.AddEvent("FoundProvider", trace.WithAttributes(attribute.Stringer("peer", p.ID)))
+					err := pqm.dialer.Connect(findProviderCtx, p)
```
hsanjuan, line 341: The beauty is this can now be a libp2p.Host directly. No longer need to rely on the network having absorbed provider peer addresses into the peerstore prior to connecting because we have them here already.

bitswap/client/client.go
```diff
@@ -139,7 +167,16 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, bstore blockstore
 	sim := bssim.New()
 	bpm := bsbpm.New()
 	pm := bspm.New(ctx, peerQueueFactory, network.Self())
-	pqm := bspqm.New(ctx, network)
+	if bs.pqm == nil { // not set with the options
+		// network can do dialing and also content routing.
+		pqm, err := rpqm.New(ctx, network, network, rpqm.WithMaxProviders(10))
```
hsanjuan: If we don't want to pass-in a Host and a ContentRouter directly to the Client, we need to rely on Network, which has them, but it would be better to remove Content Routing capabilities from the Network because it is just wrapping the content Router and nothing else at this point.

hsanjuan (dismissed):

hsanjuan:
lol I meant to just comment, not approve my own changes

bitswap/client/client.go
```diff
@@ -99,6 +99,18 @@ func WithoutDuplicatedBlockStats() Option {
 	}
 }
 
+// WithCustomLookupManagement indicates we should use a custom
+// ProviderQueryManager, which provides a handling layer on top of the content
+// routing performed by the bitswap Network for finding providers for data.
+// The ProviderQueryManager provides bounded paralelism and limits for these
+// lookups. The ProviderQueryManager setup by default uses
+// maxInProcessRequests = 6 and maxProviders = 10.
+func WithCustomLookupManagement(pqm *rpqm.ProviderQueryManager) Option {
```
aschmahmann: Should this take a function / interface instead of a ProviderQueryManager struct? There are very few exposed functions on the struct (IIUC only FindProvidersAsync / `ProviderQueryRouter` is actually used)

routing/mock/centralized_client.go
```diff
@@ -51,7 +51,7 @@ func (c *client) FindProvidersAsync(ctx context.Context, k cid.Cid, max int) <-c
 	go func() {
 		defer close(out)
 		for i, p := range c.server.Providers(k) {
-			if max <= i {
```
aschmahmann, line 54: This is an unrelated fix due to this code not understanding that 0 means infinity, right?

bitswap/network/ipfs_impl.go
```diff
@@ -365,33 +364,16 @@ func (bsnet *impl) Stop() {
...
-// FindProvidersAsync returns a channel of providers for the given key.
-func (bsnet *impl) FindProvidersAsync(ctx context.Context, k cid.Cid, max int) <-chan peer.ID {
-	out := make(chan peer.ID, max)
-	go func() {
-		defer close(out)
-		providers := bsnet.routing.FindProvidersAsync(ctx, k, max)
-		for info := range providers {
-			if info.ID == bsnet.host.ID() {
-				continue // ignore self as provider
-			}
```
aschmahmann, line 383: Do we still need to filter this out, or is that not necessary?

aschmahmann (commented):
Thanks @hsanjuan for the changes. Looks reasonable to me, left a couple questions

bitswap/client/client.go
aschmahmann: If we make this a function that takes a bsnet and spits out a ProverQueryRouter would that reduce the likelihood of people making mistakes from having two routing systems (one unused in the bsnet and one here).

routing/mock/centralized_client.go
hsanjuan: I don't know, you did this change :man_shrugging:

hsanjuan:
Sync with @aschmahmann :

- ProviderQueryManager looks a lot like a ContentRouter itself.
- Hector proposes removing the ContentRouter (CR) part from BitSwapNetwork and passing it directly to Bitswap.New(). This is a breaking change (people need to pass CR to Bitswap.New() rather than to the bsnet.New().
- Adin wants to make ContentRouter completely optional in Bitswap Client, after removing from Server. That provides further justification to break bsnet and Bitswap.New() behaviour. 
- When ContentRouter is provided, it might already be a disguised ProviderQueryManager, so users could:
  - Use no ContentRouting
  - Use their custom ContentRouting
  - Use ProviderQueryManager
  - This means users would need to remember to manually create ProviderQueryManager on upgrade, to keep things working as before. That's shitty and they will likely forget.
  - Alternatively bitswap can:
    - Receive a ContentRouting
    - Also have an option about whether that content routing should be wrapped in a default ProviderQueryManager or not. Defaulting to yes.

I think this is the gist @aschmahmann . I think the last option would be ok (there's breakage but it both a) has a stronger reason b) is easy to fix.

routing/mock/centralized_client.go
aschmahmann: oh, lol. Yeah, that's why it's there. I was noticing some weirdness with testing.

aschmahmann:
> I think the last option would be ok (there's breakage but it both a) has a stronger reason b) is easy to fix.

Do you mean to do this?
1. Remove the Providing out of the Bitswap server (e.g. something like https://github.com/ipfs/boxo/pull/677)
2. Break the `bitswap.Network` by removing the ContentRouting
3. Add two options to bitswap.Client, 1 for adding a ProviderFinder and another for whether the default ProviderQueryManager should be used

Existing consumers would migrate by moving their ContentRouter from bitswap.Network into the bitswap.Client option

If so, that sounds reasonable to me. If you think pulling out the Providing is too much to do here we could also do something like:
1. Break the `bitswap.Network` by removing the ContentRouting
2. Add two options to bitswap.Bitswap, 1 for adding a ContentRouting and another (in bitswap.Client) for whether the default ProviderQueryManager should be used
...
In a follow up PR:
1. Remove the Providing out of the Bitswap server
2. Change the bitswap.Bitswap option to forward to a bitswap.Client option and pass a ProviderFinder instead of a ContentRouter (is just a subset so nothing should break)

hsanjuan:
> 1. Remove the Providing out of the Bitswap server (e.g. something like https://github.com/ipfs/boxo/pull/677)
> 2. Break the bitswap.Network by removing the ContentRouting
> 3. Add two options to bitswap.Client, 1 for adding a ProviderFinder and another for whether the default ProviderQueryManager should be used


ok, currently we are here... see my review for a couple of things

bitswap/client/client.go
```diff
@@ -97,6 +98,27 @@ func WithoutDuplicatedBlockStats() Option {
 	}
 }
 
+// WithContentRouting enables content.Routing lookups to find providers for
+// the content to be retrieved using Bitswap.
+func WithContentRouting(router routing.Routing) Option {
+	return func(bs *Client) {
+		bs.router = router
+	}
+}
```
hsanjuan: I find this problematic as it will silently disable providerQuery-ing if users do not remember to pass in the new option with a content router.

I would prefer to break New() and require a new router there (supporting that it is nil).

bitswap/client/client.go
```diff
@@ -97,6 +98,27 @@ func WithoutDuplicatedBlockStats() Option {
 	}
 }
 
+// WithContentRouting enables content.Routing lookups to find providers for
+// the content to be retrieved using Bitswap.
+func WithContentRouting(router routing.Routing) Option {
```
hsanjuan: We could have a `ProviderFinder` interface here instead of Routing. I think users may not necessarily figure out that a ProviderFinder is a subset of their already existing content routing though, so it is more self-documenting this way.

bitswap/testnet/virtual.go
```diff
@@ -254,20 +254,20 @@ func (nc *networkClient) Stats() bsnet.Stats {
 }
 
 // FindProvidersAsync returns a channel of providers for the given key.
-func (nc *networkClient) FindProvidersAsync(ctx context.Context, k cid.Cid, max int) <-chan peer.ID {
+func (nc *networkClient) FindProvidersAsync(ctx context.Context, k cid.Cid, max int) <-chan peer.AddrInfo {
```
hsanjuan: This can be removed now, but more important, we need to wire the mock content-router to the tests at a higher level (instance generator etc) as it is no longer in the network. currently the tests work without content routing.

bitswap/client/client.go
aschmahmann: > I would prefer to break New() and require a new router there (supporting that it is nil).

That's fine by me. In general I prefer options to passing nil's but for new consumers (as opposed to ones who will migrate during the breaking change to bsnet) I can see the value in making them think about whether they want to provide a content router or not.

bitswap/client/client.go
aschmahmann: I find the coupling of these unfortunate and non-Golike as it can end up resulting in wrapper layers and such. It's a bit like how the libp2p host is passed around everywhere even when only a fraction of its functionality is needed. Would just documenting this function and calling out that a `routing.ContentRouting` (https://github.com/libp2p/go-libp2p/blob/dc97b73e4d2970142e366e047a34853a5f54fd8c/core/routing/routing.go#L26) also suffices.

Note: even routing.Routing is way bigger than what we need. If breaking apart the interface in go-libp2p/core/routing would be helpful I'd be that'd be reasonable to the go-libp2p maintainers.

hsanjuan: Let's do that... I will open a PR on libp2p to split it and we can swap later.

hsanjuan: https://github.com/libp2p/go-libp2p/pull/3048

bitswap/network/ipfs_impl.go
hsanjuan: This is handled now... libp2p would return ErrDialToSelf so we need to ignore that explicitally as we don't know who we are and only have a general purpose dialer.

CHANGELOG.md
```diff
@@ -114,6 +114,7 @@ The following emojis are used to highlight certain changes:
 - `bitswap/client` fix memory leak in BlockPresenceManager due to unlimited map growth. [#636](https://github.com/ipfs/boxo/pull/636)
 - `bitswap/network` fixed race condition when a timeout occurred before hole punching completed while establishing a first-time stream to a peer behind a NAT [#651](https://github.com/ipfs/boxo/pull/651)
 - `bitswap`: wantlist overflow handling now cancels existing entries to make room for newer entries. This fix prevents the wantlist from filling up with CIDs that the server does not have. [#629](https://github.com/ipfs/boxo/pull/629)
+- 🛠 `bitswap` & `bitswap/server` no longer provide to content routers, instead you can use the `provider` package because it uses a datastore queue and batches calls to ProvideMany.
```
hsanjuan, line 166: Reminder to fix this before merge

provider/provider.go
```diff
@@ -18,7 +18,7 @@ var logR = logging.Logger("reprovider.simple")
 // Provider announces blocks to the network
 type Provider interface {
 	// Provide takes a cid and makes an attempt to announce it to the network
-	Provide(cid.Cid) error
+	Provide(context.Context, cid.Cid, bool) error
```
hsanjuan, line 21: I have updated this interface so that it matches the signature from ContentRouting, so that we can drop this interface altogether if they merge https://github.com/libp2p/go-libp2p/pull/3048

bitswap/testinstance/testinstance.go
```diff
@@ -8,18 +8,20 @@ import (
...
 	tnet "github.com/libp2p/go-libp2p-testing/net"
 	p2ptestutil "github.com/libp2p/go-libp2p-testing/netutil"
 	peer "github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/routing"
 )
 
 // NewTestInstanceGenerator generates a new InstanceGenerator for the given
 // testnet
-func NewTestInstanceGenerator(net tn.Network, netOptions []bsnet.NetOpt, bsOptions []bitswap.Option) InstanceGenerator {
+func NewTestInstanceGenerator(net tn.Network, routing mockrouting.Server, netOptions []bsnet.NetOpt, bsOptions []bitswap.Option) InstanceGenerator {
```
hsanjuan, line 24: We need to pass it in from tests because some tests do want to share a mock router among the different things that use it (Blockservice/Instance etc).

bitswap/testinstance/testinstance.go
```diff
@@ -84,10 +88,11 @@ func ConnectInstances(instances []Instance) {
 
 // Instance is a test instance of bitswap + dependencies for integration testing
 type Instance struct {
-	Peer            peer.ID
+	Identity        tnet.Identity
```
hsanjuan, line 91: There is a test that needs the identity to create a mock client provider on a different provider server to test that blocks provided on a separate network are not received etc etc. so this was necessary.

bitswap/client/bitswap_with_sessions_test.go
```diff
@@ -39,14 +39,19 @@ func addBlock(t *testing.T, ctx context.Context, inst testinstance.Instance, blk
 	if err != nil {
 		t.Fatal(err)
 	}
+	err = inst.Routing.Provide(ctx, blk.Cid(), true)
+	if err != nil {
+		t.Fatal(err)
+	}
```
hsanjuan, line 45: Since not done by network, it needs to be done here explicitally now.

bitswap/client/client.go
```diff
@@ -148,6 +198,14 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, bstore blockstore
...
 		self peer.ID,
 	) bssm.Session {
+		// careful when bs.pqm is nil. Since we are type-casting it
+		// into session.ProviderFinder when passing it, it will become
+		// not nil. Related:
+		// https://groups.google.com/g/golang-nuts/c/wnH302gBa4I?pli=1
+		var pqm bssession.ProviderFinder
+		if bs.pqm != nil {
+			pqm = bs.pqm
+		}
```
hsanjuan, line 208: Allowing to pass nil parameters is not awesome either...

bitswap/client/internal/session/session.go
```diff
@@ -75,7 +75,7 @@ type SessionPeerManager interface {
 // ProviderFinder is used to find providers for a given key
 type ProviderFinder interface {
 	// FindProvidersAsync searches for peers that provide the given CID
-	FindProvidersAsync(ctx context.Context, k cid.Cid) <-chan peer.ID
+	FindProvidersAsync(ctx context.Context, k cid.Cid) <-chan peer.AddrInfo
```
hsanjuan: To discuss: make this `FindProvidersAsync(ctx context.Context, k cid.Cid, max int)`. Consider this a `libp2p.ContentDiscovery` interface and not a custom one. Let the ProviderQueryManager implement that. Let `max` be a bitswap.Option and not a ProviderQueryManager one, or call it with `-1` or something to signal we want to use the "default".

blockservice/blockservice.go
```diff
@@ -474,47 +506,21 @@ func (s *Session) GetBlocks(ctx context.Context, ks []cid.Cid) <-chan blocks.Blo
 
 var _ BlockGetter = (*Session)(nil)
 
-// ContextWithSession is a helper which creates a context with an embded session,
-// future calls to [BlockGetter.GetBlock], [BlockGetter.GetBlocks] and [NewSession] with the same [BlockService]
-// will be redirected to this same session instead.
-// Sessions are lazily setup, this is cheap.
-// It wont make a new session if one exists already in the context.
-func ContextWithSession(ctx context.Context, bs BlockService) context.Context {
```
hsanjuan: Confirm we are definitely removing this

blockservice/providing_blockstore.go
```diff
@@ -0,0 +1,37 @@
...
+	if err := pbs.Blockstore.PutMany(ctx, b); err != nil {
+		return err // what are the semantics here, did some blocks were put ? assume PutMany is atomic
+	}
+
+	for _, b := range b {
+		if err := pbs.provider.Provide(ctx, b.Cid(), true); err != nil {
+			return err // this can only error if the whole provider is done for
+		}
+	}
+	return nil
```
hsanjuan: It was noted that this assumes the Provider does things async etc... is this the case?

blockservice/blockservice.go
hsanjuan: Decision: no, we don't want to remove this. Leave it

blockservice/providing_blockstore.go
hsanjuan: Decision: port the old logic. Make this part of an Exchange implementation (ProvidingExchange).

bitswap/client/internal/session/session.go
hsanjuan: @aschmahmann @gammazero

provider/provider.go
hsanjuan: It was merged, now wait for libp2p release

blockservice/blockservice.go
hsanjuan: These changes are now undone.

hsanjuan:
@aschmahmann @gammazero I have a question: I was going to port the "Provider" logic from bitswap server, with its workers and queues and what not. Then I saw that we have a provider.System (the reprovider) that does have queues, does batching and even stores pending provides to disk. Naively my thought is that, if such logic is already implemented in the reprovider, a providing.Exchange should not yet have its own way and we should just pass-in the reprovider when using it. Plus I suspect that we make use of things like Accelerated-DHT-providing, `ProvideMany`, so why going around...

What do you think?

blockservice/providing_blockstore.go
hsanjuan: See comment below: https://github.com/ipfs/boxo/pull/641#issuecomment-2489038266

aschmahmann:
>What do you think?

Makes sense to me to leave it out. It's unclear how useful the parallel queues were beforehand, but it seems like the kind of thing that's situational enough that people can choose to add whatever queues they need and we should start off keeping it simple and encouraging reuse of the existing queue.

bitswap/client/internal/session/session.go
aschmahmann: I think using the `libp2p.ContentDiscovery` interface (i.e. with the `max int`) makes sense.

> Let max be a bitswap.Option and not a ProviderQueryManager one, or call it with -1 or something to signal we want to use the "default".

Why not have bitswap call it with 0 (i.e. give me all of them) and have the ProviderQueryManager handle any "max" setting? The "max" on the FindProvidersAsync interface is about the max providers returned rather than the max providers searched for. The ProviderQueryManager could reasonably choose to search for 100 peers but only return the first 10 that it connected to and still be honoring `FindProvidersAsync(..., 10)`, even though right now it will search for 10 and return <= 10 even if there are 1000 providers that exist and 10 that are reachable.

blockservice/providing_blockstore.go
hsanjuan: Resolved by leaving a simple wrapper and we will try to use the Reprovider directly for this.

bitswap/client/internal/session/session.go
hsanjuan: If ProviderQueryManager implements the ContentDiscovery interface, then it should return an unlimited number of results when called with max == 0, per the requirements of that interface.

We could use <0 as the values that we use the MaxProviders option for (the routing interface does not specify what to do with negative values):

  * `pqm.FindProvidersAsync(ctx, cid, 10)`  -> returns at most 10 providers that we manage to connect to (it may search more)
  * `pqm.FindProvidersAsync(ctx, cid, 0)` -> returns unbounded number of providers that we manage to connect to.
  * `pqm.FindProvidersAsync(ctx, cid, -1)` -> returns at most pqm.maxProviders that we manage to connect to.

Thoughts?

routing/providerquerymanager/providerquerymanager.go
```diff
@@ -5,39 +5,45 @@ import (
...
 
-var log = logging.Logger("bitswap/client/provqrymgr")
+var log = logging.Logger("routing/provqrymgr")
 
 const (
-	maxProviders         = 10
-	maxInProcessRequests = 6
-	defaultTimeout       = 10 * time.Second
+	defaultMaxInProcessRequests = 6
+	defaultMaxProviders         = 0
```
hsanjuan, line 21: Are we sure we want 0 now?

bitswap/client/internal/session/session.go
```diff
@@ -403,14 +403,20 @@ func (s *Session) handlePeriodicSearch(ctx context.Context) {
...
+	if s.providerFinder == nil {
+		return
+	}
 	go func(k cid.Cid) {
 		ctx, span := internal.StartSpan(ctx, "Session.FindMorePeers")
 		defer span.End()
-		for p := range s.providerFinder.FindProvidersAsync(ctx, k) {
+		// Max is set to -1. This means "use the default limit" in the
+		// provider query manager.
+		for p := range s.providerFinder.FindProvidersAsync(ctx, k, -1) {
```
hsanjuan: It is a bit ugly that here it is assumed that the ContentDiscovery that we use understands what -1 means.

bitswap/client/internal/session/session.go
aschmahmann: We could do this, although is it worth the hassle given at the moment there's only one consumer that likely doesn't need this?

What about this?
- `pqm.FindProvidersAsync(ctx, cid, 10)` returns at most 10 providers that we manage to connect to (it will search for up to pqm.maxProviders which can be set to 0 and ignored)
- `pqm.FindProvidersAsync(ctx, cid, 0)` -> returns unbounded number of providers that we manage to connect to (it will search for up to pqm.maxProviders which can be set to 0 and ignored)
- default pqm.maxProviders is 0

In Bitswap we would default to `pqm.FindProvidersAsync(ctx, cid, 10)` and would set `pqm.maxProviders` to 10. In a followup PR (once we know this behaves well) we'd try setting pqm.maxProviders to 0 by default.

If in the future there's a desire to effectively set `pqm.maxProviders` per query as a parameter we can figure something out like:
1. Just add a new function `FindReachableProvidersAsync(ctx, cid, maxReturnedProviders, maxAttemptedProviders)`
2. Hide the `maxProviders` as a context settable option
3. Use negative numbers
   - `pqm.FindProvidersAsync(ctx, cid, -X)` returns at most X providers that we manage to connect to (it will search for up to X providers). We could then drop the ability to set `pqm.maxProviders` entirely and assert that setting the last parameter >= 0 is the same as the old `pqm.maxProviders=0` (i.e. the default)

hsanjuan: yeah ok, then we do your proposal

routing/providerquerymanager/providerquerymanager.go
hsanjuan: Apparently yes

bitswap/client/internal/session/session.go
hsanjuan: ok, now calling it with 0.

bitswap/client/internal/session/session.go
hsanjuan: fixed in latest commit

hsanjuan:
https://github.com/ipfs/kubo/pull/10595 is testing this.

blockservice/providing_blockstore.go
```diff
@@ -0,0 +1,37 @@
...
+}
```
hsanjuan: I need to remove this, the blockstore does not need to provide

gateway/backend_blocks.go
```diff
@@ -624,12 +624,6 @@ func (bb *BlocksBackend) IsCached(ctx context.Context, p path.Path) bool {
 	return has
 }
 
-var _ WithContextHint = (*BlocksBackend)(nil)
-
-func (bb *BlocksBackend) WrapContextForRequest(ctx context.Context) context.Context {
-	return blockservice.ContextWithSession(ctx, bb.blockService)
-}
-
```
hsanjuan: Need to re-add this

hsanjuan:
Squashed

gammazero (approved):

PR #643, opened 2024-07-29
fix(gateway): Cache-Control header for UnixFS directories on /ipfs

lidel:
Continuing from https://github.com/ipfs/boxo/pull/639

- [x] original fix from from https://github.com/ipfs/boxo/pull/639
- [x] limit it to immutable `/ipfs` to not over-cache DNSLink websites with default resolver (due to TTL gap described in https://github.com/ipfs/boxo/issues/329#issuecomment-1995236409
- [x] update tests 
- [x]  CI is green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The following emojis are used to highlight certain changes:
 
 - `bitswap/server` minor memory use and performance improvements
 - `bitswap` unify logger names to use uniform format bitswap/path/pkgname
+- `gateway` now always returns meaningful cache-control headers for generated HTML listings of UnixFS directories
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


lidel (commented):
Lgtm, we will likely refine cache-control across all response types, but this is useful fix of a hot path (dir listings on `/ipfs` should benefit from wider cache window).

Related spec clarification in https://github.com/ipfs/specs/pull/480

Digression:
- `max-age` based on TTL for `/ipns/{dnslink}` is unrelated to this fix, but we may want to fix it sooner than later, as https://github.com/ipfs/boxo/issues/329#issuecomment-1995236409 impacts directory listings, and that is why this PR is limited to dir listings on `/ipfs`.
- We are not married to the `stale-while-revalidate` value on IPNS, could adjust if needed (mutable conformance tests only care about `max-age`), but 1 month sounds more robust than a few days or a week, the intention is to avoid link rot, if http cache supports such wide window.

PR #645, opened 2024-07-29
feat(gateway): allow localhost http:// DoH resolvers

lidel:
This PR allows people to run own DNS over HTTP(S) resolver on the same box (localhost) without setting up unnecessary TLS certs.

Main use case is to simplify self-hosting of non-ICANN DNS systems such as ENS resolver (https://github.com/ethlimo/limo-web3-dns) by gateway operators who don't want to trust third-party DNS resolvers.

- [x] relax https://github.com/libp2p/go-doh-resolver/blob/f2e25860684789200e3a3a911dc9a74d01771d5b/resolver.go#L55
- ~[ ] relax https://github.com/multiformats/go-multiaddr-dns (if needed)~ (doesn't appear to be needed)
- [x] test end-to-end with localhost resolver

### DIFF
--- a/gateway/dns.go
+++ b/gateway/dns.go
@@ -15,8 +15,8 @@ var defaultResolvers = map[string]string{
 }
 
 func newResolver(url string, opts ...doh.Option) (madns.BasicResolver, error) {
-	if !strings.HasPrefix(url, "https://") {
-		return nil, fmt.Errorf("invalid resolver url: %s", url)
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "http://") {
+		return nil, fmt.Errorf("invalid DoH resolver URL: %s", url)
 	}
 
 	return doh.NewResolver(url, opts...)
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/ipld/go-codec-dagpb v1.6.0
 	github.com/ipld/go-ipld-prime v0.21.0
 	github.com/libp2p/go-buffer-pool v0.1.0
-	github.com/libp2p/go-doh-resolver v0.4.1-0.20241223184115-344e51cc8063
+	github.com/libp2p/go-doh-resolver v0.5.0
 	github.com/libp2p/go-libp2p v0.38.1
 	github.com/libp2p/go-libp2p-kad-dht v0.28.1
 	github.com/libp2p/go-libp2p-record v0.2.0
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/ipld/go-codec-dagpb v1.6.0
 	github.com/ipld/go-ipld-prime v0.21.0
 	github.com/libp2p/go-buffer-pool v0.1.0
-	github.com/libp2p/go-doh-resolver v0.4.0
+	github.com/libp2p/go-doh-resolver v0.4.1-0.20241223184115-344e51cc8063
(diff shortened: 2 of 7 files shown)


gateway/dns.go
```diff
@@ -15,8 +15,8 @@ var defaultResolvers = map[string]string{
 }
 
 func newResolver(url string, opts ...doh.Option) (madns.BasicResolver, error) {
-	if !strings.HasPrefix(url, "https://") {
-		return nil, fmt.Errorf("invalid resolver url: %s", url)
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "http://") {
```
aschmahmann, line 18: I don't think this is going to be enough:
https://github.com/libp2p/go-doh-resolver/blob/f2e25860684789200e3a3a911dc9a74d01771d5b/resolver.go#L55

Note: AFAICT there is no DNS-over-HTTP spec for some reason it's specified as being over HTTPS, however I definitely see your point and have definitely felt the pain over the self-signed cert dance just to run local resolvers

lidel: Yep, filled this PR because I am looking at https://github.com/libp2p/go-doh-resolver and https://github.com/multiformats/go-multiaddr-dns to wire up TTL (to fix https://github.com/ipfs/boxo/issues/329#issuecomment-1995236409 for DNSLinks), so likely will relax things there as well, and get back to this PR draft.

go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/ipld/go-codec-dagpb v1.6.0
 	github.com/ipld/go-ipld-prime v0.21.0
 	github.com/libp2p/go-buffer-pool v0.1.0
-	github.com/libp2p/go-doh-resolver v0.4.0
+	github.com/libp2p/go-doh-resolver v0.4.1-0.20241223184115-344e51cc8063
```
aschmahmann: TODO: Fix after https://github.com/libp2p/go-doh-resolver/pull/28 is released

aschmahmann:
> Main use case is to simplify self-hosting of ENS resolves by gateway operations etc.

This PR works and for tools like CoreDNS that support DNS over HTTP (and not just HTTPS) this might be enough. If there's not enough support here folks might end up asking us for DNS over UDP / TCP support, but we can tackle that as the need arises.

For what it's worth a nice thing about pushing on DNS-over-HTTP is that it makes it easier for folks who want to self-host while also using web-browsers since UDP/TCP are not available to javascript in browsers. For some cases like ENS this might not be a big deal since they can wrap the Ethereum JSON-RPC, but it might matter for other name systems.

aschmahmann:
Updated the comment above given that CoreDNS does support DNS-over-HTTP in addition to DNS-over-HTTPS.

FYI CoreDNS (and therefore the CoreDNS ENS resolver) supports DNS-over-HTTP, although the UX for this is that you configure listening on "https://" rather than "http://" and just not pass a certificate (which is what tripped me up initially 😅).

I think we should be good to go on a merge here, pending reviews.

lidel:
@aschmahmann thank you for adding tests and https://github.com/libp2p/go-doh-resolver/pull/28.  Lgtm.

I've reviewed https://github.com/libp2p/go-doh-resolver/pull/28, limited it  to `localhost`, merged, released, and switched this PR to go-doh-resolver [v0.5.0](https://github.com/libp2p/go-doh-resolver/releases/tag/v0.5.0).

Merging as it cleans up go.mod as bit as well.

Issue #647, opened 2024-07-30
Chunker Memory leak


Xib1uvXi:
When I add a folder with many small files to kubo, memory usage always spikes abnormally

- Total folder size: 400MiB
- Number of files: 10000
- Average size per file: 40KiB

I located the exception by analyzing the heap
![image](https://github.com/user-attachments/assets/56517ab7-9c3f-45a0-bb7e-7f3a1346bcb0)

```go

// NewSizeSplitter returns a new size-based Splitter with the given block size.
func NewSizeSplitter(r io.Reader, size int64) Splitter {
	return &sizeSplitterv2{
		r:    r,
		size: uint32(size),
	}
}

// NextBytes produces a new chunk.
func (ss *sizeSplitterv2) NextBytes() ([]byte, error) {
	if ss.err != nil {
		return nil, ss.err
	}

	full := pool.Get(int(ss.size))
	n, err := io.ReadFull(ss.r, full)
	switch err {
	case io.ErrUnexpectedEOF:
		ss.err = io.EOF
		small := make([]byte, n)
		copy(small, full) 
		pool.Put(full)  
		return small, nil
	case nil:
		return full, nil
	default:
		pool.Put(full)
		return nil, err
	}
}

// Reader returns the io.Reader associated to this Splitter.
func (ss *sizeSplitterv2) Reader() io.Reader {
	return ss.r
}
```

Q1: Will using a pool result in abnormal memory consumption when the file size is much smaller than the chunk size?

gammazero:
> Will using a pool result in abnormal memory consumption when the file size is much smaller than the chunk size?

The pool tries to allocate buffers sized to the power of 2 closest to the chunk size. A small file causes a new allocation, and the pool buffer is returned to the pool.
```go
		small := make([]byte, n)
		copy(small, full) 
		pool.Put(full)  
``` 
Returning the buffer to the pool does not necessarily free it and may keep it in the pool. A large number of files will cause these allocations for each when reading the last partial chunk. So I not think it has to do with small files, but the number of files.

It may be better to keep the pool-allocated buffer, and avoid the extra allocation and copy:
```go
		small := full[:n]
```
or use the pool to allocate the smaller buffer:
```go
		small := pool.Get(n)
		copy(small, full)
		pool.Put(full)
```
Having fewer different buffer sizes might help with GC. Will test to see if this makes any difference.

gammazero:
accidental close - repoened

Xib1uvXi:
[...]

PR #649, opened 2024-08-01
refactor(chunker): reduce overall memory use

gammazero:
When the chunker is not able to fill a chunk with data, it allocated a new buffer for the partial chunk of data. With many files this results in allocation of many small buffers of varying sizes, leading to heap fragmentation. This PR allocates a new buffer from the pool only if doing so would save space, otherwise it uses an partially filled (over-allocated) chunk. This makes all chunker allocation sizes be powers of 2.

Heap fragmentation is reduced at the cost of some temporary over-allocation. The advantage is that the overallocation is much shorter lived than the heap fragmentation.

Possible fix for issue #647


lidel (approved):
Smoke-tested in https://github.com/ipfs/kubo/pull/10485 as well, lgtm.
Let's see how it behaves in kubo 0.30.0-rc1

PR #651, opened 2024-08-05
fix(bitswap): increase timeout to ensure hole punching completes

2color:
## What's in this PR

This PR increases the Bitswap connection timeout from 5 seconds to 10 seconds. This is based on testing that shows that when one of the peers is behind NAT and holepunching is necessary, it can take a little bit over 5 seconds only for hole punching to succeed. 

Hole punching slowness will be handled independently in https://github.com/libp2p/go-libp2p/issues/2898. 

This issue aims to prevent bitswap connection timeouts that would otherwise succeed, if given a higher connection timeout. Increasing the timeout appears to be an effective workaround based on some testing.

Based on https://github.com/ipfs/go-bitswap/pull/477/files, it appears that the 5 second timeout was an arbitrary starting that was picked in 2021, prior to which there was no time out. [Here's the source of the suggestion of a 5 second timeout](https://github.com/ipfs/kubo/issues/7972#issuecomment-824327775).


bitswap/network/ipfs_impl.go
```diff
@@ -28,7 +28,7 @@ import (
 
 var log = logging.Logger("bitswap/network")
 
-var connectTimeout = time.Second * 5
+var connectTimeout = time.Second * 10
```
lidel, line 31: With https://github.com/libp2p/go-libp2p/pull/2907 being the default in [go-libp2p 0.36](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.36.0),  we probably could remove this timeout, but maybe aligning with stream timeout from that PR is safer. 

@aschmahmann @gammazero any concerns if we bump this to 15, to match modern timeouts in go-libp2p?

FYSA in latest go-libp2p, both Dial and NewStream have default timeouts of 15s:
https://github.com/libp2p/go-libp2p/blob/v0.36.0/p2p/net/swarm/swarm.go#L28-L36


```suggestion
var connectTimeout = time.Second * 15
```

lidel (changes requested):

bitswap/network/ipfs_impl.go
lidel: With https://github.com/libp2p/go-libp2p/pull/2907 being the default in [go-libp2p 0.36](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.36.0),  we probably could remove this timeout, but maybe aligning with stream timeout from that PR is safer. 


FYSA in latest go-libp2p, both Dial and NewStream have default timeouts of 15s:
https://github.com/libp2p/go-libp2p/blob/v0.36.0/p2p/net/swarm/swarm.go#L28-L36

lidel:
Triage note:
- decided to remove timeout and rely on default upstream ones introduced in https://github.com/libp2p/go-libp2p/pull/2907

lidel (approved):
After some spelunking we realized this timeout is no longer necessary, refactored this PR to reflect that. LGTM.

go-libp2p now has implicit ones for both Dial and NewStream:
https://github.com/libp2p/go-libp2p/blob/v0.36.1/p2p/net/swarm/swarm.go#L28-L36

(With https://github.com/libp2p/go-libp2p/pull/2907 being the default in [go-libp2p 0.36](https://github.com/ipfs/ipfs-desktop/releases/tag/v0.36.0), we have implicit timeout of 15s on stream as well)

PR #653, opened 2024-08-06
feat: support UnixFS 1.5 file mode and modification times

gammazero:
Replaces PR #34 written by @kstuart

Adds support for storing and retrieving file mode and last modification time.

Support added to:
- [X] Files
- [X] LinkFiles
- [X] Webfiles
- [X] Directories
- [X] Tar Archives
- [x] https://github.com/ipfs/boxo/pull/658

When the TAR archive (headers) include a file mode or modification time, the extractor will restore that metadata when supported for the underlying filesystem.

The Golang runtime currently does not support changing the times on a symlink, for Linux and some BSDs a custom solution has been implemented, for Darwin this is not the case so when copying a symlink to the filesystem the last modification time is not updated. Since for concrete files and directories stored modes and modification times are faithfully restored to the filesystem this should not be a breaking issue, a similar solution to that implemented for Linux/BSDs is likely implementable by a developer with access to a Darwin platform.

Replaces PRs:
- https://github.com/ipfs/go-ipfs-files/pull/31
- https://github.com/ipfs/tar-utils/pull/11
- #34

Relates to https://github.com/ipfs/kubo/issues/6920


ipld/unixfs/pb/unixfs.proto
```diff
@@ -21,8 +21,25 @@ message Data {
...
+	// Represents seconds of UTC time since Unix epoch
+	// 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to
+	// 9999-12-31T23:59:59Z inclusive.
+	required int64 seconds = 1;
+
+	// Non-negative fractions of a second at nanosecond resolution. Negative
+	// second values with fractions must still have non-negative nanos values
+	// that count forward in time. Must be from 0 to 999,999,999
+	// inclusive.
+	optional fixed32 nanos = 2;
```
lidel, line 44: I feel we should have some GO-JS interop tests for nanos somewhere.

Initially planned to do that in gateway-conformance after https://github.com/ipfs/boxo/pull/659 is done, but realized that the `Last-Modified` HTTP header in gateway responses will not include nanosecond precision. It only supports second precision with an optional millisecond component. 

@achingbrain  any suggestions? my initial idea is to add `mode` and `mtime` tests (live creation and with static dag-pb fixture) to https://github.com/ipfs/helia/tree/main/packages/interop#readme after we have kubo 0.30.0-rc1 ad updated [kubo-rpc-client](https://www.npmjs.com/package/kubo-rpc-client) that works with new RPC in JS, but lmk if there is a better place/way.

mfs/ops.go
```diff
@@ -243,3 +246,21 @@ func FlushPath(ctx context.Context, rt *Root, pth string) (ipld.Node, error) {
...
+}
+
+func Touch(rt *Root, pth string, ts time.Time) error {
+	nd, err := Lookup(rt, pth)
+	if err != nil {
+		return err
+	}
+
+	return nd.SetModTime(ts)
+}
```
lidel, line 266: @gammazero  found a bug:  `ipfs files touch` and `ipfs files chmod` fail when we execute them against `raw` blocks.

Repro steps with https://github.com/ipfs/kubo/pull/10478:

```console
$  echo -n "hello path gw" | ipfs add --cid-version 1 -Q  --raw-leaves --to-files /raw-test-1
bafkreiehuhfvxn7w4i5hexlhq4a3y7jf5kfogsvmm5zlfxg5poj4oqozyu

$ ipfs files touch /raw-test-1
Error: expected a ProtoNode as internal node

$ ipfs files chmod 0644 /raw-test-1
Error: expected a ProtoNode as internal node
```

lmk if you feel this is fixable in this PR before Wednesday, or should we move ahead with RC1, fill bug for this and fix later.


(hopefully we can add code that detects when node is a raw block and upgrade it to dag-pb before setting mode/time. otherwise, might cause a bigger refactor, because MFS hardcodes rawleaves=true for mutating files that use CIDv1)

lidel (commented):
Did bunch of testing with https://github.com/ipfs/kubo/pull/10478 and I think we are mostly good to go.

Need to decide if/how  we address below issues.

mfs/ops.go
lidel: To unblock RC testing, I've extracted it into https://github.com/ipfs/boxo/issues/660.

gammazero: @lidel I have fixed this in the commit [f4259d6](https://github.com/ipfs/boxo/pull/653/commits/f4259d68d2d12bbde204d1564b0c8cad7832cf55)

lidel: Thanks! Bumped and added end-to-end test in Kubo PR as well https://github.com/ipfs/kubo/pull/10478/commits/55dad60ae70eb5d6ebbba259bed1ad24e41e3fd0

lidel (approved):
Thank you @kstuart for initial version, and @gammazero for refactoring and implementing the missing parts and pushing over finish line.

This should be ready for shipping in kubo 0.30.0-rc1 and getting user feedback from folks that asked for `mode` and `mtime`. 

We have a bunch of tests here and in https://github.com/ipfs/kubo/pull/10478 (ignore sharness failing here, it passes in kubo PR), and these extra attributes have been around in JS for years, and are always opt-in, so relatively low risk.

PR #659, opened 2024-08-19
feat(gw): support UnixFS 1.5 on gateway responses as Last-Modified

lidel:
This is bit more involved, and I don't want to block https://github.com/ipfs/boxo/pull/653 on it, so filling this separate PR.

This PR builds on top of https://github.com/ipfs/boxo/pull/653 and exposes `mtime` from `dag-pb` as `Last-Modified` header on deserialized `/ipfs` and `/ipns` responses (ones that are used for web hosting, and not trustless responses).

## TODO

- [x] block backend
  - [x] return `Last-Modified`on GET responses
  - [x] return `Last-Modified`on HEAD responses
  - [x] 304 Not Modified on matching `If-Modified-Since` (UnixFS 1.5 file only)
  - [x] `Last-Modified` and `If-Modified-Since` tests with and without UnixFS 1.5

(Below    moved to https://github.com/ipfs/boxo/issues/664)
  
- [ ] car backend 
  - [ ] return `Last-Modified`on GET responses
  - [ ] return `Last-Modified`on HEAD responses
- [ ] add gateway-conformance test and release it as v7
  - [ ] confirm conformance passes with all backends (remote car uses different implementation of `Get` and `Head` and `ContentPathMetadata` needs to be updated there as well)


gateway/handler.go
```diff
@@ -745,6 +772,50 @@ func (i *handler) handleIfNoneMatch(w http.ResponseWriter, r *http.Request, rq *
 	return false
 }
 
+func (i *handler) handleIfModifiedSince(w http.ResponseWriter, r *http.Request, rq *requestData) bool {
```
lidel, line 775: :information_source:  this is based on what we already had in `handleIfNoneMatch` to support `If-None-Match`, which operates on Etags.

This one is executed only if `If-Modified-Since` is present in request, which usually is less popular than more deterministic `If-None-Match`. 

Still, we add is here as it costs us nothing to avoid returning payload and returning `HTTP 304 Not Modified` instead.

lidel (commented):
Discussed next steps here with @gammazero  during triage today and we decided to ship basic gateway support for `Last-Modified` / `If-Modified-Since` for UnixFS 1.5 files with `backend_blocks.go` (used by Rainbow and Kubo).

The remaining work can be done at a later time, if there is real world need for support in CAR backend. I've filled https://github.com/ipfs/boxo/issues/664 for that.

Merging this so we can include it in Kubo 0.30.0-rc2.

gateway/gateway_test.go
```diff
@@ -615,6 +549,112 @@ func TestHeaders(t *testing.T) {
 	})
 }
 
+// Testing a DAG with (optional) UnixFS1.5 modification time
+func TestHeadersUnixFSModeModTime(t *testing.T) {
+	t.Parallel()
+
+	ts, _, root := newTestServerAndNode(t, "unixfs-dir-with-mode-mtime.car")
```
lidel, line 556: :information_source:  created with Kubo 0.30.0-rc1

lidel:
Added more tests, CI in https://github.com/ipfs/kubo/pull/10494 is green, merging.

PR #667, opened 2024-09-05
fix: make timer operations work with golang 1.23

gammazero:
In golang 1.23 reading a timer's channel after calling timer.Stop will block forever. This is fixed by looking at the timer's channel capacity to determine if the channel may need to be read following a Stop.


Stebalien:
Note: it looks like go is fixing this upstream, so it might not be worth it to add this workaround.

gammazero:
Even so, clearing the channel is not necessary. The timeout of 1 hour immediately followed by `Stop` means the timer will never fire, so no need to read from the channel.

Stebalien:
> Even so, clearing the channel is not necessary. The timeout of 1 hour immediately followed by Stop means the timer will never fire, so no need to read from the channel.

Yeah, but I prefer to make code timing independent. Any reason not to just wait for 1.24, then bump the minimum go requirement to 1.23 to completely fix the race?

gammazero:
@Stebalien I made the code time independent. This PR is still needed since there are places where the timer channel is read following a Stop, and this breaks with go1.23.

Stebalien:
@gammazero https://github.com/golang/go/issues/69312? My understanding is that this is going to be fixed in a patch release very soon.

gammazero:
Will wait for go fix. Until then, the problem can be avoided by running with `GODEBUG="asynctimerchan=1"`

PR #671, opened 2024-09-12
feat: add routing filtering to delegated routing server IPIP-484

2color:
## What's in this PR

- Initial implementation of protocol filtering in the delegated routing server
- Implemented based on https://github.com/ipfs/specs/pull/484/
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- Added address and protocol filtering to the delegated routing server [IPIP-484](https://github.com/ipfs/specs/pull/484)
+
 ### Changed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
-- Added address and protocol filtering to the delegated routing server [IPIP-484](https://github.com/ipfs/specs/pull/484)
+- `routing/http`: added support for address and protocol filtering to the delegated routing server ([IPIP-484](https://github.com/ipfs/specs/pull/484)) [#671](https://github.com/ipfs/boxo/pull/671)
 
 ### Changed
 
--- a/routing/http/server/filters.go
+++ b/routing/http/server/filters.go
@@ -134,7 +134,9 @@ func applyProtocolFilter(peerProtocols []string, filterProtocols []string) bool 
 		}
 
 		for _, peerProtocol := range peerProtocols {
-			return peerProtocol == filterProtocol
+			if peerProtocol == filterProtocol {
+				return true
+			}
 		}
 	}
 	return false
--- a/routing/http/server/filters.go
+++ b/routing/http/server/filters.go
@@ -107,7 +107,7 @@ func applyFilters(provider *types.PeerRecord, filterAddrs, filterProtocols []str
 		return provider
 	}
 
-	if !applyProtocolFilter(provider.Protocols, filterProtocols) {
+	if !protocolsAllowed(provider.Protocols, filterProtocols) {
 		// If the provider doesn't match any of the passed protocols, the provider is omitted from the response.
 		return nil
 	}
@@ -177,7 +177,8 @@ func containsProtocol(protos []multiaddr.Protocol, proto multiaddr.Protocol) boo
 	return false
 }
 
-func applyProtocolFilter(peerProtocols []string, filterProtocols []string) bool {
+// protocolsAllowed returns true if the peerProtocols are allowed by the filter protocols.
+func protocolsAllowed(peerProtocols []string, filterProtocols []string) bool {
 	if len(filterProtocols) == 0 {
 		// If no filter is passed, do not filter
 		return true
--- a/routing/http/server/filters.go
+++ b/routing/http/server/filters.go
@@ -1,0 +1,143 @@
+package server
+
+import (
+	"reflect"
+	"slices"
+	"strings"
+
+	"github.com/ipfs/boxo/routing/http/types"
+	"github.com/multiformats/go-multiaddr"
+)
+
+// filters implements IPIP-0484
+
+func parseFilter(param string) []string {
+	if param == "" {
+		return nil
+	}
+	return strings.Split(strings.ToLower(param), ",")
+}
+
+func filterProviders(providers []types.Record, filterAddrs, filterProtocols []string) []types.Record {
+	if len(filterAddrs) == 0 && len(filterProtocols) == 0 {
+		return providers
+	}
+
+	filtered := make([]types.Record, 0, len(providers))
+
+	for _, provider := range providers {
+		if schema := provider.GetSchema(); schema == types.SchemaPeer {
+			peer, ok := provider.(*types.PeerRecord)
+			if !ok {
+				logger.Errorw("problem casting find providers result", "Schema", provider.GetSchema(), "Type", reflect.TypeOf(provider).String())
+				// if the type assertion fails, we exlude record from results
+				continue
+			}
+
+			record := applyFilters(peer, filterAddrs, filterProtocols)
+
+			if record != nil {
+				filtered = append(filtered, record)
+			}
+
+		} else {
+			// Will we ever encounter the SchemaBitswap type? Evidence seems to suggest that no longer
+			logger.Errorw("encountered unknown provider schema", "Schema", provider.GetSchema(), "Type", reflect.TypeOf(provider).String())
+		}
+	}
+	return filtered
+}
+
+// Applies the filters. Returns nil if the provider does not pass the protocols filter
+// The address filter is more complicated because it potentially modifies the Addrs slice.
+func applyFilters(provider *types.PeerRecord, filterAddrs, filterProtocols []string) *types.PeerRecord {
+	if !applyProtocolFilter(provider.Protocols, filterProtocols) {
+		// If the provider doesn't match any of the passed protocols, the provider is omitted from the response.
+		return nil
+	}
+
+	// return untouched if there's no filter or filterAddrsQuery contains "unknown" and provider has no addrs
+	if len(filterAddrs) == 0 || (len(provider.Addrs) == 0 && slices.Contains(filterAddrs, "unknown")) {
+		return provider
+	}
+
+	filteredAddrs := applyAddrFilter(provider.Addrs, filterAddrs)
+
+	// If filtering resulted in no addrs, omit the provider
+	if len(filteredAddrs) == 0 {
+		return nil
+	}
+
+	provider.Addrs = filteredAddrs
+	return provider
+}
+
+// If there are only negative filters, no addresses will be included in the result. The function will return an empty list.
+// For an address to be included, it must pass all negative filters AND match at least one positive filter.
+func applyAddrFilter(addrs []types.Multiaddr, filterAddrsQuery []string) []types.Multiaddr {
+	if len(filterAddrsQuery) == 0 {
+		return addrs
+	}
+
+	filteredAddrs := make([]types.Multiaddr, 0, len(addrs))
+
+	for _, addr := range addrs {
+		protocols := addr.Protocols()
+		includeAddr := true
+
+		// First, check all negative filters
+		for _, filter := range filterAddrsQuery {
+			if strings.HasPrefix(filter, "!") {
+				protocolStringFromFilter := strings.TrimPrefix(filter, "!")
+				protocolFromFilter := multiaddr.ProtocolWithName(protocolStringFromFilter)
+				if containsProtocol(protocols, protocolFromFilter) {
+					includeAddr = false
+					break
+				}
+			}
+		}
+
+		// If the address passed all negative filters, check positive filters
+		if includeAddr {
+			for _, filter := range filterAddrsQuery {
+				if !strings.HasPrefix(filter, "!") {
+					protocolFromFilter := multiaddr.ProtocolWithName(filter)
+					if containsProtocol(protocols, protocolFromFilter) {
+						filteredAddrs = append(filteredAddrs, addr)
+						break
+					}
+				}
+			}
+		}
+	}
+	return filteredAddrs
+}
+
+func containsProtocol(protos []multiaddr.Protocol, proto multiaddr.Protocol) bool {
+	for _, p := range protos {
+		if p.Code == proto.Code {
+			return true
+		}
+	}
+	return false
+}
+
+func applyProtocolFilter(peerProtocols []string, filterProtocols []string) bool {
+	if len(filterProtocols) == 0 {
+		// If no filter is passed, do not filter
+		return true
+	}
+
+	for _, filter := range filterProtocols {
+		filterProtocol := strings.TrimPrefix(filter, "!")
+
+		if filterProtocol == "unknown" && len(peerProtocols) == 0 {
+			return true
+		}
+
+		for _, peerProtocol := range peerProtocols {
+			return peerProtocol == filterProtocol
+		}
+	}
+	return false
+}
--- a/routing/http/server/filters.go
+++ b/routing/http/server/filters.go
@@ -39,8 +39,8 @@ func applyFiltersToIter(recordsIter iter.ResultIter[types.Record], filterAddrs, 
 			record, ok := v.Val.(*types.PeerRecord)
 			if !ok {
 				logger.Errorw("problem casting find providers record", "Schema", v.Val.GetSchema(), "Type", reflect.TypeOf(v).String())
-				// TODO: Do we want to let failed type assertions to pass through?
-				return v
+				// drop failed type assertion
+				return iter.Result[types.Record]{}
 			}
 
 			record = applyFilters(record, filterAddrs, filterProtocols)
@@ -55,8 +55,8 @@ func applyFiltersToIter(recordsIter iter.ResultIter[types.Record], filterAddrs, 
 			record, ok := v.Val.(*types.BitswapRecord)
 			if !ok {
 				logger.Errorw("problem casting find providers record", "Schema", v.Val.GetSchema(), "Type", reflect.TypeOf(v).String())
-				// TODO: Do we want to let failed type assertions to pass through?
-				return v
+				// drop failed type assertion
+				return iter.Result[types.Record]{}
 			}
 			peerRecord := types.FromBitswapRecord(record)
 			peerRecord = applyFilters(peerRecord, filterAddrs, filterProtocols)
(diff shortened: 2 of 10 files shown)


routing/http/server/server_test.go
```diff
@@ -136,7 +150,7 @@ func TestProviders(t *testing.T) {
 			limit = DefaultStreamingRecordsLimit
 		}
 		router.On("FindProviders", mock.Anything, cid, limit).Return(results, nil)
-		urlStr := serverAddr + "/routing/v1/providers/" + cidStr
+		urlStr := serverAddr + "/routing/v1/providers/" + cidStr + "?filter-addrs=" + filterAddrs + "&filter-protocols=" + filterProtocols
```
lidel: nit: append params only when  values are not empty

routing/http/server/server_test.go
```diff
@@ -174,29 +188,51 @@ func TestProviders(t *testing.T) {
...
 		accept := ""
-		runTest(t, accept, true, false, `{"Providers":null}`)
+		runTest(t, accept, "", "", true, false, `{"Providers":null}`)
 	})
 
 	t.Run("NDJSON Response", func(t *testing.T) {
-		runTest(t, mediaTypeNDJSON, false, true, `{"Addrs":[],"ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vn","Protocols":["transport-bitswap"],"Schema":"peer"}`+"\n"+`{"Schema":"bitswap","Protocol":"transport-bitswap","ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vz"}`+"\n")
+		runTest(t, mediaTypeNDJSON, "", "", false, true, `{"Addrs":["/ip4/127.0.0.1/tcp/4001","/ip4/127.0.0.1/udp/4001/quic-v1","/ip4/127.0.0.1/tcp/4001/ws","/ip4/102.101.1.1/tcp/4001/p2p/12D3KooWEjsGPUQJ4Ej3d1Jcg4VckWhFbhc6mkGunMm1faeSzZMu/p2p-circuit","/ip4/102.101.1.1/udp/4001/quic-v1/webtransport/p2p/12D3KooWEjsGPUQJ4Ej3d1Jcg4VckWhFbhc6mkGunMm1faeSzZMu/p2p-circuit","/ip4/8.8.8.8/udp/4001/quic-v1/webtransport"],"ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vn","Protocols":["transport-bitswap"],"Schema":"peer"}`+"\n"+`{"Schema":"bitswap","Protocol":"transport-bitswap","ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vz"}`+"\n")
```
lidel: @2color CI fails on this equals test ([log](https://github.com/ipfs/boxo/actions/runs/10834542823/job/30063981429?pr=671#step:13:324)), perhaps because  iteration order on maps is not specified in golang, so these tests worked just by luck, and once we added more logic, internal hashes changed and the order changed? 

Potential fix is to refactor `runTest` to do normalize json before comparing

routing/http/server/server.go
```diff
@@ -236,13 +241,78 @@ func (s *server) findProvidersJSON(w http.ResponseWriter, provIter iter.ResultIt
...
 	}
 
+	filteredProviders := filterProviders(providers, filterAddrs, filterProtocols)
+
 	writeJSONResult(w, "FindProviders", jsontypes.ProvidersResponse{
-		Providers: providers,
+		Providers: filteredProviders,
 	})
 }
+func (s *server) findProvidersNDJSON(w http.ResponseWriter, provIter iter.ResultIter[types.Record], filterAddrs, filterProtocols []string) {
```
lidel, line 249: @2color Any reason why we duplicate this code other than just having filtering on /providers and not /peers? 

Thoughts on adding filtering support to /peers and not having this duplicated handler? It feels useful for peer lookups too, at least filter-addrs one ("give me only /wss and /webrtc of this peer"), but supporting both for consistency allows us to be more DRY

lidel (changes requested):
CI is red, so only did a quick first-pass with some feedback inline

routing/http/server/server.go
2color: Yeah that's a good idea. One challenge was doing so with generics so I just focused on making it work first. 

I'll give it another look.

routing/http/server/server_test.go
2color: Actually the tests were wrong (I forgot to update the assertions after changing the test data. This was not a normalisation problem. 

The last commits should fix this.

routing/http/server/server.go
2color: I agree we should enable filtering to the `/peers`.

I had another look at this and haven't been able to find a good solution to reuse the code, due to the combination of generics and the different return types (types.Record and types.PeerRecord)

The main open question I have is why [`FindProviders` and `FindPeers`](https://github.com/ipfs/boxo/blob/171b0b72118578e2f42f7e5aa45b52ad33a04231/routing/http/server/server.go#L58-L67) have a different return type if they both return a list of `PeerRecord`:

```
FindProviders(ctx context.Context, cid cid.Cid, limit int) (iter.ResultIter[types.Record], error)
FindPeers(ctx context.Context, pid peer.ID, limit int) (iter.ResultIter[*types.PeerRecord], error)
``` 

Do you have an idea @lidel?

lidel: - `/routing/v1/peers` will always return results that use Peer schema.
- `/routing/v1/providers` may return different schemas, [Peer schema is not the only one](https://specs.ipfs.tech/routing/http-routing-v1/#known-schemas),  so a more generic type is used to allow pass-through in places like https://delegated-ipfs.dev/ which proxy results from other systems.
  - fysa `cid.contact/routing/v1` switched to Peer schema some time ago, making it now confusing why we dont use Peer schema everywhere, and why we waste space for Schema field in every response, but this is [probably both true] (a) a way of having open-ended API that can be evolved by community without breaking existing clients (b) reality of different teams working on different things with different priorities. 

Personally I see some value in (a), if we keep `FindProviders(ctx context.Context, cid cid.Cid, limit int) (iter.ResultIter[types.Record], error)` and return generic `Record` we  ensure   `Schema` field is not waste of space, and allow others to experiment without having to invent yet another API.

lidel: @2color  To address your task at hand, something we could do:
- keep interface as-is, write reusable things so they return `Record`, and in `FindPeers` have a wrapper that translates `Records` to `PeerRecord` for now (this way we dont break API)
  - this one is preferable, unless we have a good reason to break api (and we dont atm)
- make a breaking change and document in release notes: make `FindPeers` return generic `Record` as a way of deduplicating code boilerplate. Unlikely we will see different schema on `/peers`, but would not hurt to have that ability.

2color: I addressed your feedback taking the first approach and avoiding breaking changes. It felt a bit messy, but we haven't broken any public interfaces.

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+// If there are only negative filters, no addresses will be included in the result. The function will return an empty list.
+// For an address to be included, it must pass all negative filters AND match at least one positive filter.
+func applyAddrFilter(addrs []types.Multiaddr, filterAddrsQuery []string) []types.Multiaddr {
+	if len(filterAddrsQuery) == 0 {
+		return addrs
+	}
+
+	filteredAddrs := make([]types.Multiaddr, 0, len(addrs))
+
+	for _, addr := range addrs {
```
gammazero, line 141: Consider using [`multiaddr.FilterAddrs`](https://pkg.go.dev/github.com/multiformats/go-multiaddr#FilterAddrs) here.

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+	if len(filteredAddrs) == 0 {
+		return nil
+	}
+
+	provider.Addrs = filteredAddrs
+	return provider
+}
+
+// If there are only negative filters, no addresses will be included in the result. The function will return an empty list.
+// For an address to be included, it must pass all negative filters AND match at least one positive filter.
```
gammazero: This seems unexpected. If there are no positive filters, then I would expect that everything (that passed negative filters) is kept since there are no positive filters to exclude any more non-matches.

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+
+	for _, addr := range addrs {
+		protocols := addr.Protocols()
+		includeAddr := true
+
+		// First, check all negative filters
+		for _, filter := range filterAddrsQuery {
+			if strings.HasPrefix(filter, "!") {
+				protocolStringFromFilter := strings.TrimPrefix(filter, "!")
+				protocolFromFilter := multiaddr.ProtocolWithName(protocolStringFromFilter)
```
gammazero:
```suggestion
				protocolFromFilter := multiaddr.ProtocolWithName(filter[1:])
```

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+
+// If there are only negative filters, no addresses will be included in the result. The function will return an empty list.
+// For an address to be included, it must pass all negative filters AND match at least one positive filter.
+func applyAddrFilter(addrs []types.Multiaddr, filterAddrsQuery []string) []types.Multiaddr {
+	if len(filterAddrsQuery) == 0 {
+		return addrs
+	}
+
+	filteredAddrs := make([]types.Multiaddr, 0, len(addrs))
+
```
gammazero: In the case where there are filters, it is likely that many will be filtered out, so probably better not to pre-allocate.
```suggestion
	var filteredAddrs []types.Multiaddr

```

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+func containsProtocol(protos []multiaddr.Protocol, proto multiaddr.Protocol) bool {
+	for _, p := range protos {
+		if p.Code == proto.Code {
+			return true
+		}
+	}
+	return false
+}
+
+func applyProtocolFilter(peerProtocols []string, filterProtocols []string) bool {
```
gammazero: It is not immediately clear, without reading the oode, that returning true means the protocol is allowed. Perhaps rename this function to `protocolAllowed` or something similar.

routing/http/server/server.go
```diff
@@ -369,10 +380,33 @@ func (s *server) provide(w http.ResponseWriter, httpReq *http.Request) {
...
+
+		var record types.Record = v.Val
+		return iter.Result[types.Record]{Val: record}
+	})
+
+	filteredIter := applyFiltersToIter(mappedIter, filterAddrs, filterProtocols)
+
+	// Convert Record back to PeerRecord 🙃
+	finalIter := iter.Map(filteredIter, func(v iter.Result[types.Record]) iter.Result[*types.PeerRecord] {
+		if v.Err != nil || v.Val == nil {
```
gammazero, line 400: Please comment why it is OK to ignore the error if `v.Val` is not `nil`.

routing/http/server/server.go
```diff
@@ -383,8 +417,19 @@ func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[
 	})
 }
 
-func (s *server) findPeersNDJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord]) {
-	writeResultsIterNDJSON(w, peersIter)
+func (s *server) findPeersNDJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord], filterAddrs, filterProtocols []string) {
+	// Convert PeerRecord to Record so that we can reuse the filtering logic from findProviders
+	mappedIter := iter.Map(peersIter, func(v iter.Result[*types.PeerRecord]) iter.Result[types.Record] {
+		if v.Err != nil || v.Val == nil {
```
gammazero, line 423: Please comment why it is OK to ignore the error if `v.Val` is not `nil`.

routing/http/server/server.go
```diff
@@ -369,10 +380,33 @@ func (s *server) provide(w http.ResponseWriter, httpReq *http.Request) {
...
 }
 
-func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord]) {
+func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord], filterAddrs, filterProtocols []string) {
 	defer peersIter.Close()
 
-	peers, err := iter.ReadAllResults(peersIter)
+	// Convert PeerRecord to Record so that we can reuse the filtering logic from findProviders
+	mappedIter := iter.Map(peersIter, func(v iter.Result[*types.PeerRecord]) iter.Result[types.Record] {
+		if v.Err != nil || v.Val == nil {
```
gammazero, line 388: Do you mean to return the value and ignore the error when `v.Err != nil` and `v.Val` is not `nil`? If there is a use case for this, then a comment would be god here.

routing/http/types/iter/filter.go
```diff
@@ -0,0 +1,43 @@
...
+	return &FilterIter[T]{iter: iter, f: f}
+}
+
+type FilterIter[T any] struct {
+	iter Iter[T]
+	f    func(T) bool
+
+	done bool
+	val  T
+}
```
gammazero, line 14: NOTE: It would be nicer to use [go iterators](https://pkg.go.dev/iter) here to allow `range` to be used to iterate the filters. However, that requires go1.23 . If only we were requiring that.

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+	if param == "" {
+		return nil
+	}
+	return strings.Split(strings.ToLower(param), ",")
+}
+
+// applyFiltersToIter applies the filters to the given iterator and returns a new iterator.
+func applyFiltersToIter(recordsIter iter.ResultIter[types.Record], filterAddrs, filterProtocols []string) iter.ResultIter[types.Record] {
+	mappedIter := iter.Map(recordsIter, func(v iter.Result[types.Record]) iter.Result[types.Record] {
+		if v.Err != nil || v.Val == nil {
```
gammazero, line 33: Why is it OK to ignore the error if `v.Val != nil`? Please comment.

routing/http/server/filters.go
```diff
@@ -0,0 +1,201 @@
...
+			record, ok := v.Val.(*types.PeerRecord)
+			if !ok {
+				logger.Errorw("problem casting find providers record", "Schema", v.Val.GetSchema(), "Type", reflect.TypeOf(v).String())
+				// TODO: Do we want to let failed type assertions to pass through?
+				return v
+			}
+
+			record = applyFilters(record, filterAddrs, filterProtocols)
+			if record == nil {
+				return iter.Result[types.Record]{Err: errors.New("record is nil")}
```
gammazero: `applyFilters` returns nil when all providers are filtered out (no providers selected by filters), so this does not seem like it should be treated as an error. I think just returning an empty result, or use a defined `ErrNoProviders` error that can be differentiated for actual errors, would be preferable.

routing/http/server/filters.go
2color: If `v.Val != nil && v.Val != nil` , the result value will be untouched. The idea here was to avoid any processing if there's an error or a nil value, so that this function doesn't make any assumptions.  

Does that make sense?

routing/http/server/server.go
2color: But if there's an error, it won't be ignored, even if `v.Val != nil`. It will just return the error without any value.

routing/http/server/server.go
2color: See above. If v`.Val != nil` and there's an error, it will return a Result type with the error.

routing/http/server/server.go
2color: The idea was to return `nil` when there's an error **or** the value is nil.

routing/http/server/filters.go
2color: Good point.

routing/http/server/filters.go
2color: Good point. I'll return an empty result.

routing/http/server/filters.go
2color: I just had a look at this, however, because we're using the `types.Multiaddr` wrapper, this results in a bunch of conversions back and forth. So I've decided to skip this.

routing/http/server/filters.go
2color: Fixed in d8edbe6e9a1ce2388a8327842cc0dbb3a660380a

routing/http/server/filters.go
gammazero: yes

routing/http/server/server.go
gammazero: Sorry, for some reason I read it as `&&`.

gammazero (approved):
LGTM

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- Added address and protocol filtering to the delegated routing server [IPIP-484](https://github.com/ipfs/specs/pull/484)
```
```suggestion
- `routing/http`: added support for address and protocol filtering to the delegated routing server ([IPIP-484](https://github.com/ipfs/specs/pull/484)) [#671](https://github.com/ipfs/boxo/pull/671)
```

routing/http/server/filters.go
```diff
@@ -0,0 +1,229 @@
...
+			return v
+		}
+
+		switch v.Val.GetSchema() {
+		case types.SchemaPeer:
+			record, ok := v.Val.(*types.PeerRecord)
+			if !ok {
+				logger.Errorw("problem casting find providers record", "Schema", v.Val.GetSchema(), "Type", reflect.TypeOf(v).String())
+				// TODO: Do we want to let failed type assertions to pass through?
+				return v
```
lidel: I think if something went wrong here, its safer to skip result – may be malformed.

```suggestion
				// drop failed type assertion
				return iter.Result[types.Record]{}
```

routing/http/server/filters.go
```diff
@@ -0,0 +1,229 @@
...
+			v.Val = record
+
+		//lint:ignore SA1019 // ignore staticcheck
+		case types.SchemaBitswap:
+			//lint:ignore SA1019 // ignore staticcheck
+			record, ok := v.Val.(*types.BitswapRecord)
+			if !ok {
+				logger.Errorw("problem casting find providers record", "Schema", v.Val.GetSchema(), "Type", reflect.TypeOf(v).String())
+				// TODO: Do we want to let failed type assertions to pass through?
+				return v
```
lidel:
```suggestion
				// drop failed type assertion
				return iter.Result[types.Record]{}
```

routing/http/server/filters.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+
+		} else {
+			// Will we ever encounter the SchemaBitswap type? Evidence seems to suggest that no longer
+			logger.Errorw("encountered unknown provider schema", "Schema", record.GetSchema(), "Type", reflect.TypeOf(record).String())
+		}
+	}
+	return filtered
+}
+
```
lidel: Having inconsistency between `filterRecords` and `applyFiltersToIter` (one supports `SchemaBNitswap`, other dont)  feels weird.

`filterRecords` does not seem to be used anywhere, maybe just remove it and keep `applyFiltersToIter` ? (`//lint:ignore U1000 // ignore unused` silenced linter)

```suggestion
```

lidel (approved):
Tested in updated https://github.com/ipfs/someguy/pull/82 and LGTM, small suggestions inline, but I think its ok to merge once those are addressed.

lidel:
@2color we have pretty good changelog, so not the end of the world, but in future remember to squash PRs like this, to avoid noise in main branch :)  (in go repos we prefer squashes)

2color:
Sounds good! Sorry about that. Should we disable merge and rebase commits for this repo then?

lidel:
Afraid we need to keep it for the current release process: 
https://github.com/ipfs/boxo/blob/main/RELEASE.md#release-process

PR #672, opened 2024-09-13
feat: option to not read size of blocks for want-have requests

### LINKED ISSUE #657: bitswap/server: `WANT_HAVE` requests should use `Blockstore.Has`
Currently, `WANT_HAVE` requests [rely](https://github.com/ipfs/boxo/blob/635e1754f8eb888518f211d2e7253f384b0441d4/bitswap/server/internal/decision/engine.go#L701) on `GetSize` to check for presence of the requested block and the learnt [size gets discarded](https://github.com/ipfs/boxo/blob/635e1754f8eb888518f211d2e7253f384b0441d4/bitswap/server/internal/decision/engine.go#L793).

Basically, the Blockstore has to lookup data size that's unused. Depending on `Blockstore` implementation, this might have different consequences. Intuitively, looking up data size should be more expensive compared to checking data existence. In practice, this is what happens in our case.

Hence the proposal is to add new method to blockstore manager that performs only the Has check and refactor `ReceivedMessage` to decouple processing of `WANT_HAVE` and `WANT_BLOCK` messages.

gammazero:
When the replace-have-with-want-blocks is disabled, the block sizes for blocks corresponding to have-want requests are not read and instead the blockstore is checked only to see if the block is present.

The replace-have-with-want-blocks feature can be disabled using the new `WithReplaceHasWithBlockMaxSize(n)` option which can be used with `bitswap.New`. It sets the maximum size of a block in bytes up to which we will replace a want-have with a want-block. Setting a size of 0 disables want-have replacement and means that block sizes are not read for want-have requests.

Aims to fix #657

### DIFF
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -714,7 +714,6 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 			}
 			for blkCid := range hasBlocks {
 				blockSizes[blkCid] = 0
-				fmt.Println("   block cid:", blkCid)
 			}
 		}
 	}
--- a/bitswap/server/internal/decision/engine.go
+++ b/bitswap/server/internal/decision/engine.go
@@ -689,15 +689,35 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
 		return true
 	}
 
+	noReplace := e.maxBlockSizeReplaceHasWithBlock == 0
+
 	// Get block sizes for unique CIDs.
-	wantKs := cid.NewSet()
+	wantKs := make([]cid.Cid, 0, len(wants))
+	var haveKs []cid.Cid
 	for _, entry := range wants {
-		wantKs.Add(entry.Cid)
+		if noReplace && entry.WantType == pb.Message_Wantlist_Have {
+			haveKs = append(haveKs, entry.Cid)
+		} else {
+			wantKs = append(wantKs, entry.Cid)
+		}
 	}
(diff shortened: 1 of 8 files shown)


gammazero:
See configuration in kubo: https://github.com/ipfs/kubo/pull/10512

bitswap/server/internal/decision/engine.go
```diff
@@ -689,16 +685,39 @@ func (e *Engine) MessageReceived(ctx context.Context, p peer.ID, m bsmsg.BitSwap
...
+			log.Info("aborting message processing", err)
+			return false
+		}
+		if len(hasBlocks) != 0 {
+			if blockSizes == nil {
+				blockSizes = make(map[cid.Cid]int, len(hasBlocks))
+			}
+			for blkCid := range hasBlocks {
+				blockSizes[blkCid] = 0
+				fmt.Println("   block cid:", blkCid)
```
Wondertan: Relic. Let's remove it so we aren't flooded with those during testing

Wondertan (commented):
Our integration tests passed with the option applied. I will test this E2E once the debug log is removed

bitswap/server/internal/decision/engine.go
gammazero: Done

Wondertan:
I tested this E2E, and things have been going smoothly so far. Will see how it works over the weekend.

Wondertan:
The cluster performs correctly, and the nodes are still in sync. We haven't tested for performance improvements or degradations and I think that's an approve from our side.

Wondertan (approved):

lidel (approved):
lgtm, pushed cosmetic docs changes (godoc explaining feature bit more + ability to read implicit default via godoc)

@gammazero thanks again, if text looks good to you, free to squash and merge :rocket:

PR #678, opened 2024-10-01
feat: add protocol and address filtering to delegated routing api

2color:
## What's in this PR

- Adds client protocol and address filtering according to [IPIP-484](https://github.com/ipfs/specs/pull/484) to the HTTP Client

## Notes
- Since not all delegated routing servers support [IPIP-484](https://github.com/ipfs/specs/pull/484), e.g. IPNI, client side filtering is enabled by default, and needs to be disabled with `WithDisabledLocalFiltering`


routing/http/client/client.go
```diff
@@ -83,6 +90,37 @@ func WithIdentity(identity crypto.PrivKey) Option {
 	}
 }
 
+// WithDisabledLocalFiltering disables local filtering of the results.
+// This should be used for delegated routing servers that already implement filtering
+func WithDisabledLocalFiltering() Option {
```
gammazero: Maybe better to have this take a boolean, so it can be blindly passed a value from a config file instead of being called conditionally?

routing/http/client/client.go
```diff
@@ -184,7 +222,9 @@ func (c *Client) FindProviders(ctx context.Context, key cid.Cid) (providers iter
 	// TODO test measurements
 	m := newMeasurement("FindProviders")
 
-	url := c.baseURL + "/routing/v1/providers/" + key.String()
+	url := fmt.Sprintf("%s/routing/v1/providers/%s", c.baseURL, key.String())
```
gammazero: Consider using `path.Join` in case  the base or the key have unexpected separators.

gammazero (approved):
LGTM

gammazero (approved):

PR #688, opened 2024-10-10
fix(routing/http/client): avoid escaping comma query separator

lidel:
This PR makes delegated routing client make IPIP-484 queries with `,` instead of `%2C`.
This is backward-compatible cosmetic fix: routing/http/server works with both `,` and `%2C`, so clients that use boxo v0.24 will still work.

## Rationale

The comma is in the "sub-delims" set of characters that don't need to be encoded in most parts of a URL, including query parameters. Golang standard library percent-escapes it for consistency, but we prefer human-readable /routing/v1 URLs, and real comma is restored here to ensure human and machine requests hit the same HTTP cache keys.

cc @2color


2color (approved):
seems reasonable

PR #689, opened 2024-10-16
feat(routing/http/client): DefaultProtocolFilter for IPIP-484

lidel:
This PR makes `routing/http/client` initialize with implicit default protocol filter from [IPIP-484](https://github.com/ipfs/specs/pull/484) set to `unknown` + `transport-bitswap`, matching what we explicitly set in Kubo (https://github.com/ipfs/kubo/pull/10534) and Rainbow (https://github.com/ipfs/rainbow/pull/173/). 

## Rationale

 boxo users can change or disable default filter via `WithProtocolFilter` option, but the default should be filter that is optimized for golden path, which currently means bitswap. Processing peers that we know only support unsupported protocols is wasting resources and we should skip them.

Over time, we may add new protocols like http to boxo, and we will also adjust the default list accordingly.

## TODO

- [x] refactor `routing/http/client` to use `DefaultProtocolFilter`
- [x] refactor `examples/routing/delegated-routing-client` to reflect implicit defaults

### DIFF
--- a/examples/routing/delegated-routing-client/main_test.go
+++ b/examples/routing/delegated-routing-client/main_test.go
@@ -25,7 +25,7 @@ func TestFindProviders(t *testing.T) {
 		if r.URL.Path == "/routing/v1/providers/"+cidStr {
 			w.Header().Set("Content-Type", "application/x-ndjson")
 			w.Write([]byte(`{"Schema":"peer","ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vn","Addrs":["/ip4/111.222.222.111/tcp/5734"],"Protocols":["transport-bitswap"]}` + "\n"))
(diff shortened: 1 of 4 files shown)


examples/routing/delegated-routing-client/main_test.go
```diff
@@ -25,15 +25,15 @@ func TestFindProviders(t *testing.T) {
 		if r.URL.Path == "/routing/v1/providers/"+cidStr {
 			w.Header().Set("Content-Type", "application/x-ndjson")
 			w.Write([]byte(`{"Schema":"peer","ID":"12D3KooWM8sovaEGU1bmiWGWAzvs47DEcXKZZTuJnpQyVTkRs2Vn","Addrs":["/ip4/111.222.222.111/tcp/5734"],"Protocols":["transport-bitswap"]}` + "\n"))
-			w.Write([]byte(`{"Schema":"peer","ID":"12D3KooWB6RAWgcmHAP7TGEGK7utV2ZuqSzX1DNjRa97TtJ7139n","Addrs":["/ip4/127.0.0.1/tcp/5734"],"Protocols":["transport-horse"]}` + "\n"))
```
2color, line 28: Perhaps keep `"transport-horse"` and add an additional unknown one so that it's clear it's getting filtered

2color (approved):

PR #691, opened 2024-10-17
fix(bitswap/client/msgq): prevent duplicate requests

Wondertan:
Previously, in-progress requests could be re-requested again during periodic rebroadcast. The queue requests, and while awaiting a response, the rebroadcast event happens. Rebroadcast event changes previously sent WANTs to pending and sends them again in a new message, duplicating some WANT requests.

The solution here is to ensure WANT was in sent status for long enough before bringing it back to pending. This utilizes the existing `sendAt` map, which tracks when every CID was sent. Then, on every event, it compares if the message was around longer than `rebroadcastInterval`


bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -476,27 +476,43 @@ func (mq *MessageQueue) rebroadcastWantlist() {
...
-func (mq *MessageQueue) transferRebroadcastWants() bool {
+func (mq *MessageQueue) transferRebroadcastWants() int {
 	mq.wllock.Lock()
 	defer mq.wllock.Unlock()
 
-	// Check if there are any wants to rebroadcast
-	if mq.bcstWants.sent.Len() == 0 && mq.peerWants.sent.Len() == 0 {
-		return false
+	mq.rebroadcastIntervalLk.RLock()
+	rebroadcastInterval := mq.rebroadcastInterval
```
Wondertan, line 515: Alternatively, this could be a different new parameter/constant

Wondertan:
I tested this on a k8s cluster and with a local node connected to it. It works as expected, but I believe this would benefit a lot from a proper test. Unfortunately, I can't allocate time to writing one. It's not that straightforward.

Wondertan:
For context, I detect duplicates with a custom multihash that logs out when the same data is hashed again. This essentially uncovered #690, and this issue

bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -476,27 +476,43 @@ func (mq *MessageQueue) rebroadcastWantlist() {
...
-	mq.bcstWants.pending.Absorb(mq.bcstWants.sent)
-	mq.peerWants.pending.Absorb(mq.peerWants.sent)
+	for _, want := range mq.peerWants.sent.Entries() {
+		sentAt, ok := mq.peerWants.sentAt[want.Cid]
+		if ok && now.Sub(sentAt) >= rebroadcastInterval {
+			mq.peerWants.pending.Add(want.Cid, want.Priority, want.WantType)
+			mq.peerWants.sent.Remove(want.Cid)
+			toRebroadcast++
+		}
+	}
```
gammazero: This loop looks like a duplicate of the above.

Wondertan: Its peerWants vs broadcastWants tho

Wondertan: What I mean here is that it looks like the duplicate, but it works on different map

gammazero: The previous call to `mq.peerWants.pending.Absorb` also did:
```
// Invalidate the cache up-front to avoid doing any work trying to keep it up-to-date.                                 
w.cached = nil
```
Evan though all the send entries may not be added to pending in the new code, the cache still needs to be cleared. Otherwise, calling `mq.peerWants.pending.Entries()` will not return the newly added entries. Alternatively, the new entries can be to added to the wantlist's `cached` entries and sorted.

I suggest to keep the previous code:
```
mq.bcstWants.pending.Absorb(mq.bcstWants.sent)
mq.peerWants.pending.Absorb(mq.peerWants.sent)
```
And instead modify the `Absorb` function in `boxo/bitswap/client/wantlist/wantlist.go`. WDYT?

gammazero: What I meant is these look like they should call the same function defined for a `WantList`. Like `Absorb` in `boxo/bitswap/client/wantlist/wantlist.go`

bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -476,27 +476,43 @@ func (mq *MessageQueue) rebroadcastWantlist() {
...
 
 // Transfer wants from the rebroadcast lists into the pending lists.
-func (mq *MessageQueue) transferRebroadcastWants() bool {
+func (mq *MessageQueue) transferRebroadcastWants() int {
 	mq.wllock.Lock()
 	defer mq.wllock.Unlock()
 
-	// Check if there are any wants to rebroadcast
-	if mq.bcstWants.sent.Len() == 0 && mq.peerWants.sent.Len() == 0 {
-		return false
```
gammazero: This is probably good to leave since it avoids Lock/Unlock of `mq.rebroadcastIntervalLk` and `time.Now()`.
```go
	if mq.bcstWants.sent.Len() == 0 && mq.peerWants.sent.Len() == 0 {
		return 0
	}
```

gammazero:
triage note: This is a good candidate for testing in rainbow staging to observe performance differences.

bitswap/client/internal/messagequeue/messagequeue.go
Wondertan: This doesn't work because `pending *WantList` doesn't have access to `sentAt` map.

Wondertan: I can try to make a new method on `recallWantList` struct tho

bitswap/client/internal/messagequeue/messagequeue.go
Wondertan: The lock exists only for testing. The interval is never changed outside of the unit test. Thus, I don't see any contention zero length check could prevent.

bitswap/client/internal/messagequeue/messagequeue.go
Wondertan: Ok, check the new version

Wondertan: On the cache invalidation note. `WantList.Add` also invalidates the cache, so we don't have to it.

bitswap/client/wantlist/wantlist.go
```diff
@@ -130,13 +130,3 @@ func (w *Wantlist) Entries() []Entry {
 	w.cached = es
 	return es[0:len(es):len(es)]
 }
-
-// Absorb all the entries in other into this want list
-func (w *Wantlist) Absorb(other *Wantlist) {
```
Wondertan, line 135: Deleted as deadcode

bitswap/client/internal/messagequeue/messagequeue.go
hsanjuan: I think the comment is not about contention but about saving unnecessary lock/unlock calls, but if this only happens every 30 seconds, then it's probably not very important.

hsanjuan (approved):
The main thing to consider here is that:

  * before, a "want" would be re-broadcasted at most 30 seconds after it was sent (could be 0.1s)
  * after, a "want" would be re-broadcasted only after at least 30 seconds after it was sent (could be 59.9s).

In that respect the code looks good.

I am not sure how much of an improvement this is in practice (perhaps clients were lucky to hit a short rebroadcast period sometimes), but it makes clients more respectful at least and perf should not be based on "luck".

I think we can test on staging and discuss in the next triage if we accept the change.

gammazero:
Need to test on staging before merge.

gammazero:
This PR does make sure that the client does not resend wants to any peer before the rebroadcast interval has elapsed. In doing this is also makes some peers, that were just short of the interval, wait for another rebroadcast interval.

In summary it changes from "wait no more than X to resend wants" to "wait at least X, but no more than 2X, to resend wants".

If we want the PR, then we should consider calling `rebroadcastWantlist` at half (or less) of the rebroadcast interval. @hsanjuan WDYT?

Consider changing line 410 to
```go
const checksPerInterval = 2
mq.rebroadcastTimer = mq.clock.Timer(mq.rebroadcastInterval / checksPerInterval)
```

That will change the logic to "wait at least X, but no more than X+(X/checksPerInterval), to resend wants

PR #693, opened 2024-10-18
chore: update dependencies

gammazero:
- Use sync/atomic instead of uber/atomic


.github/workflows/gateway-sharness.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
```
lidel:
```suggestion
        uses: actions/setup-go@v5
```

.github/workflows/gateway-sharness.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         run: sudo apt install -y socat net-tools fish libxml2-utils
       - name: Replace boxo in Kubo go.mod
         run: |
+          go clean -cache
           go mod edit -replace=github.com/ipfs/boxo=../boxo
           make mod_tidy
```
lidel, line 36:
```suggestion
          make mod_tidy
          make clean
```

.github/workflows/gateway-sharness.yml
```diff
@@ -32,12 +32,14 @@ jobs:
...
-          cat go.mod
         working-directory: kubo
       - name: Install sharness dependencies
-        run: make test_sharness_deps
+        run: |
+          find . -name go.mod -execdir go mod tidy \;
```
lidel, line 40: @gammazero turns out running mod tidy second time here is what fixed the problem.

I've reproduced the CI error locally  by using `go mod edit -replace github.com/ipfs/boxo=/home/lidel/projects/ipfs/boxo` in my local Kubo and running `make mod_tidy` only once.  

Adding this extra mod tidy here applies 1 line change to `test/dependencies/go.mod` in such mod replace setup:

```diff
-       github.com/miekg/dns v1.1.61 // indirect
+       github.com/miekg/dns v1.1.62 // indirect
```

Once it is applied, build is green again.


This edge case is specific to `go mod edit -replace` case where version of one of dependencies of "replaced" boxo is higher than kubo, so imo ok to keep this here to avoid the problem coming back again in future.

.github/workflows/gateway-sharness.yml
```diff
@@ -32,12 +32,14 @@ jobs:
         run: sudo apt install -y socat net-tools fish libxml2-utils
       - name: Replace boxo in Kubo go.mod
         run: |
+          go clean -cache
```
lidel:
```suggestion
```

lidel (approved):

PR #700, opened 2024-10-23
Update go-test workflow to use ipdxco

gammazero:
Make go-test workflow use `ipdxco/unified-github-workflows/.github/workflows/go-test.yml@v1.0` and not `ipfs/uci/.github/workflows/go-test.yml@v1.0`

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -15,6 +15,6 @@ concurrency:
 
 jobs:
   go-test:
-    uses: ipfs/uci/.github/workflows/go-test.yml@v1.0
+    uses: ipdxco/unified-github-workflows/.github/workflows/go-test.yml@v1.0
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}


lidel:
Hm..  ubuntu tests fail due to [`Called workflows cannot be queued onto self-hosted runners across organizations/enterprises. Failed to queue this job. Labels: 'self-hosted , linux , x64 , 2xlarge'.`](https://github.com/ipfs/boxo/actions/runs/11488898887?pr=700)  

I think we need help from @ipfs/ipdx to resolve this.

galargh (changes requested):
`ipfs/uci` is a fork that is automatically updated by `ipdxco/unified-github-workflows` (https://github.com/ipdxco/unified-github-workflows/commit/2315cf0bd64fec262d8d976b0dd9ea8487fc3c1d - I just put it on schedule, not sure why it wasn't on it before). Unfortunately, we need this workaround because reusable workflows from outside the organization they're running in, do not have access to self-hosted runners.

lidel:
Thanks @galargh, in that case we can close this.

PR #703, opened 2024-10-24
bitswap(client/messageque): expose donthavetimeout config 

### LINKED ISSUE #701: bitswap(messagequeue): Expose DONT_HAVE timeout knobs
### Context
Don't have timeout is an important security feature that prevents malicious peers from never responding with DONT_HAVE. If a peer we sent WANT_BLOCK never responds, a Bitswap session stalls and does not re-request other peers. This mechanism has a [bunch of configuration parameters ](https://github.com/ipfs/boxo/blob/392493dbca6031051da7558ca8d1f0a3db972856/bitswap/client/internal/messagequeue/donthavetimeoutmgr.go#L13-L39).

### Request
Expose those parameters for protocol-specific fine tuning. The defaults timeout too quickly, sometimes causing:
* Duplclicate requests
  * A asked B for Cid -> A timed out waiting for Cid and asked C -> B kept processing requests, and with C now -> A received two responses.
* Peers eviction from session. Once 64 DONT_HAVE received, the peer is dropped, even tho the peer is healthy.

Wondertan:
Closes #701

### DIFF
--- a/bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
+++ b/bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
@@ -10,33 +10,41 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
 )
 
-const (
-	// dontHaveTimeout is used to simulate a DONT_HAVE when communicating with
+func defaultDontHaveTimeoutConfig() *DontHaveTimeoutConfig {
+	cfg := &DontHaveTimeoutConfig{
+		DontHaveTimeout:            5 * time.Second,
+		MaxExpectedWantProcessTime: 2 * time.Second,
+		PingLatencyMultiplier:      3,
+		MessageLatencyAlpha:        0.5,
+		MessageLatencyMultiplier:   2,
+	}
+
+	cfg.MaxTimeout = cfg.DontHaveTimeout + cfg.MaxExpectedWantProcessTime
+	return cfg
+}
+
+type DontHaveTimeoutConfig struct {
+	// DontHaveTimeout is used to simulate a DONT_HAVE when communicating with
 	// a peer whose Bitswap client doesn't support the DONT_HAVE response,
 	// or when the peer takes too long to respond.
 	// If the peer doesn't respond to a want-block within the timeout, the
 	// local node assumes that the peer doesn't have the block.
-	dontHaveTimeout = 5 * time.Second
-
-	// maxExpectedWantProcessTime is the maximum amount of time we expect a
+	DontHaveTimeout time.Duration
+	// MaxExpectedWantProcessTime is the maximum amount of time we expect a
 	// peer takes to process a want and initiate sending a response to us
-	maxExpectedWantProcessTime = 2 * time.Second
-
-	// maxTimeout is the maximum allowed timeout, regardless of latency
-	maxTimeout = dontHaveTimeout + maxExpectedWantProcessTime
-
-	// pingLatencyMultiplier is multiplied by the average ping time to
+	MaxExpectedWantProcessTime time.Duration
+	// MaxTimeout is the maximum allowed timeout, regardless of latency
+	MaxTimeout time.Duration
+	// PingLatencyMultiplier is multiplied by the average ping time to
 	// get an upper bound on how long we expect to wait for a peer's response
 	// to arrive
-	pingLatencyMultiplier = 3
-
-	// messageLatencyAlpha is the alpha supplied to the message latency EWMA
-	messageLatencyAlpha = 0.5
-
+	PingLatencyMultiplier int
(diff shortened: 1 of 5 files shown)


bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
```diff
@@ -92,39 +96,33 @@ type dontHaveTimeoutMgr struct {
...
-		onDontHaveTimeout:          onDontHaveTimeout,
-		timeoutsTriggered:          timeoutsTriggered,
+		clock:             clock,
+		ctx:               ctx,
+		shutdown:          shutdown,
+		peerConn:          pc,
+		activeWants:       make(map[cid.Cid]*pendingWant),
+		timeout:           cfg.DontHaveTimeout,
+		messageLatency:    &latencyEwma{alpha: cfg.MessageLatencyAlpha},
+		onDontHaveTimeout: onDontHaveTimeout,
```
gammazero, line 123: The `onDontHaveTimeout` function seems to accompany the `config` is many/most places. So, `onDontHaveTimeout` callback may be a candidate for moving into the `DontHaveTimeoutConfig` type.

bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
```diff
@@ -10,33 +10,41 @@ import (
...
+	cfg := &DontHaveTimeoutConfig{
+		DontHaveTimeout:            5 * time.Second,
+		MaxExpectedWantProcessTime: 2 * time.Second,
+		PingLatencyMultiplier:      3,
+		MessageLatencyAlpha:        0.5,
+		MessageLatencyMultiplier:   2,
+	}
+
+	cfg.MaxTimeout = cfg.DontHaveTimeout + cfg.MaxExpectedWantProcessTime
+	return cfg
```
lidel, line 23: Two concerns with this approach:
1. We expose a way for overriding the defaults without exposing the implicit default values themselves
   -  people will keep hardcoding current values instead of referencing const that may change in the future
2. We use a struct which forces override fo all values
   - This makes it tedious to override a single value
   - This means if we add a new knob, people who already override the struct will override default with implicit zero value for that type (likely causing bugs)
  
My mental model for makign things like this configurable is to 
- have `WithFoo` configuration options which allow overriding each knob individually 
- and expose `DefaultFoo` to allow people to programmatically refer to implicit default, rather than hardcoding them in their own code.

This may be more work in this PR upfront, but will save us from headache in the future. Thoughts?

lidel (changes requested):

bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
Wondertan: I suggest we expose the default struct for people to override necessary knobs this way:
```go
cfg := DefaultDontHaveTimeoutConfig()
// example of granular customization
cfg.MaxExpectedWantProcessTime = 5 * time.Second
// use the global option.
WithDontHaveTimeoutConfig(cfg)
```

This solves your concern without creating "options hell". This is equivalent to having multi-level options, yet it's cleaner. Also, it doesn't require forwarding all the options from `client` to `bitswap` in case you want to add more options horizontally, which I tried initially, and got lost.

Wondertan: Basically, the compromise I propose it expose the default configuration solving your concern while keeping things straightforward.

bitswap/client/internal/messagequeue/donthavetimeoutmgr.go
Wondertan: I guess the question is: "Do we want users to register a custom handler there?". I would be opposed to that unless there is a prominent use case.

Wondertan:
@gammazero, @lidel, one thing I realized about this component is that it's actually never disabled. The option to disable donthavetimeout handling exists in bitswap. However, it simply makes the handler no-op, while the component is still active and running, doing work that is costly (measuring latency, firing timeouts in routines, etc). And disabled donthavetimeout is the default behaviour in Kubo.

lidel:
Triage notes:
- exposing defaults sgtm
  - nit: make it public?  `defaultDontHaveTimeoutConfig → `DefaultDontHaveTimeoutConfig` 
- @Wondertan  what do you mean by "And disabled donthavetimeout is the default behaviour in Kubo."? Mind linking to relevant code/config?

gammazero:
I am planning on changing this to enable and configure onDontHaveTime using an option param. This will likely be done in a different PR.

gammazero:
Replaced by #750

In #750, the option to disable donthavetimeout handling is respected and the dontHaveTimeoutMgr is disabled if this option is false.

Wondertan:
Feel free to close this one then

Wondertan:
Nvm, I'll just close it myself

PR #710, opened 2024-10-30
Remove dependency on goprocess

### LINKED ISSUE #130: [ipfs/go-bitswap] Cleanup usage of processes and contexts
We should never pass around _both_ a context _and_ a goprocess.

### LINKED ISSUE #709: Remove dependency on goprocess
The dependency on `github.com/jbenet/goprocess` is not needed by boxo. Removing it removes the need to support it in code dependent on boxo, making boxo simpler to use.

gammazero:
The dependency on goprocess is not needed by boxo, and removing it removes the need to support it in code dependent on boxo.

Closes #709
Closes #130

### DIFF
--- a/bitswap/server/server.go
+++ b/bitswap/server/server.go
@@ -299,12 +299,7 @@ func (bs *Server) startWorkers(ctx context.Context) {
 	if bs.provideEnabled {
 		bs.waitWorkers.Add(1)
 		go bs.provideCollector(ctx)
-
-		// Spawn up multiple workers to handle incoming blocks
-		// consider increasing number if providing blocks bottlenecks
-		// file transfers
-		bs.waitWorkers.Add(1)
-		go bs.provideWorker(ctx)
+		bs.startProvideWorkers(ctx)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


bitswap/server/server.go
```diff
@@ -508,18 +501,16 @@ func (bs *Server) provideCollector(ctx context.Context) {
...
+func (bs *Server) provideWorker(ctx context.Context) {
 	limit := make(chan struct{}, provideWorkerMax)
+	defer func() {
+		// Wait until all limitGoProvide goroutines are done before declaring
+		// this worker as done.
+		for i := 0; i < provideWorkerMax; i++ {
+			limit <- struct{}{}
+		}
+		bs.waitWorkers.Done()
+	}()
```
hsanjuan: I reviewed the provideWorker() closely and I think it works as before. I'm just not very sure why we have 1 provideWorker which then launches short-lived provideWorkerMax-goroutines all the time, vs having provideWorkerMax workers that are long-lived and read from the provideKeys channel and on closing we wait for them to finish with a waitGroup.

Not worth refactoring here though.

hsanjuan (approved):
Super happy to see goprocess no more. I believe this is good, I hope I'm not missing something. We should pay attention to goroutine leakage when rolling out RCs in case something is not getting closed as before, but it should be fine.

bitswap/server/server.go
gammazero: The only reason this was done in the PR was to work exactly as before, with only the removal of goprocess.  Now that you have seen it too, and have the same question I did, I would like to change this to use a set of long-lived goroutines.

hsanjuan: Do it in a different PR? This one fits its purpose already.

gammazero: I made the change in this PR, see [diff](https://github.com/ipfs/boxo/pull/710/commits/d3697effad16de9c4e24f6def4ca3e8a88fcdc31).

The only reason for the original implementation (if any) was to have a different `wid` value in the debug log for each time a provide was done. That is `runCount` is added.

hsanjuan: LGTM

hsanjuan:
Lastest commit LGTM too

PR #712, opened 2024-11-04
chore: update deps

aschmahmann:
Updates directly imported dependencies (left the indirect dependencies alone)


gammazero (approved):

lidel:
https://github.com/ipfs/kubo/pull/10572 is green, will release this as a patch

PR #715, opened 2024-11-11
bitswap: SessionInterestManager can consider unwanted blocks as interesting

hsanjuan:
FilterSessionInterested should only return "wanted" blocks. Currently it returns blocks that have been marked as unwanted (flipped entry in map to false but remained in the map).

I need a second pair of eyes... so in `RemoveSessionWants()` we mark the interest of a session in a block as `false`:

https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go#L81-L86

We however do not remove the entry from the map as done in `RemoveSessionInterested`. But later in `FilterSessionInterested` we only check whether the session entry exists, and not whether it is set to `true`:

https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go#L132-L136

It would seem as if there was some distinction between "session interest", and "session want", but this is not reflected in how things are used. i.e. `Session.ReceiveFrom()` uses `FilterSessionInterest` and `Session.handleReceive()` uses `RemoveSessionWants()`, so I think we might be requesting blocks that we already have because `FilterSessionInterest` is going to still return removed wants.

### DIFF
--- a/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
+++ b/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
@@ -131,7 +131,7 @@ func (sim *SessionInterestManager) FilterSessionInterested(ses uint64, ksets ...
 		// For each key in the list
 		for _, c := range ks {
 			// If there is a session that's interested, add the key to the set
-			if _, ok := sim.wants[c][ses]; ok {
+			if wanted, ok := sim.wants[c][ses]; ok && wanted {
 				has = append(has, c)
 			}
 		}
--- a/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager_test.go
+++ b/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager_test.go
@@ -83,6 +83,20 @@ func TestInterestedSessions(t *testing.T) {
 	}
 }
 
+func TestRemoveSessionWants(t *testing.T) {
+	sim := New()
+
+	const ses = 1
+	cids := random.Cids(3)
+	sim.RecordSessionInterest(ses, cids[0:2])
+	sim.RemoveSessionWants(ses, cids)
+	res := sim.FilterSessionInterested(ses, cids)
+
+	if len(res) > 0 {
+		t.Fatal("wants have not been removed")
+	}
+}
+
 func TestRemoveSession(t *testing.T) {
 	sim := New()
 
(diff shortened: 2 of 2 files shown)


hsanjuan:
A test confirms that RemoveSessionWants() have no effect on FilterSessionInterested().

hsanjuan:
If confirmed that there is indeed a problem, I would change altogether `wants: make(map[cid.Cid]map[uint64]bool),` into `wants: make(map[cid.Cid]map[uint64]struct{},`

lidel:
triage note:
- pre-review before we invest time into refactoring failing test

bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -131,7 +131,7 @@ func (sim *SessionInterestManager) FilterSessionInterested(ses uint64, ksets ...
 		// For each key in the list
 		for _, c := range ks {
 			// If there is a session that's interested, add the key to the set
-			if _, ok := sim.wants[c][ses]; ok {
+			if wanted, ok := sim.wants[c][ses]; ok && wanted {
```
gammazero, line 134: I think the original code was intentional.
```
// The boolean indicates whether the session still wants the block                                                                                                                                                                                    
// or is just interested in receiving messages about it. 
```
Also, see comments here:
https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go#L153-L156

I think the original code is correct, because it adds the key to the `has` set if the session is interested in the blocks (`wanted==true`) ot just the messages (`wanted==false`). This function is asking if the session has an interest - not specifically if it wants the blocks.

gammazero (changes requested):
I beleive the original code is correct. See comments.

bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
hsanjuan: Thanks so much!

For my own notes: `FilterSessionInterested` is only called from https://github.com/ipfs/boxo/blob/73057e2145e4f105cf75ce6894ad07c0a37b6c73/bitswap/client/internal/session/session.go#L194 which then calls `sws.Update`. 

sws.Update will then removeWant() for the keys that we were "interested in" when it processes the update as it contains the cids of the blocks received. iirc this also means that the wants for a peer a both tracked in their own ways in sessionInterestManager and SessionWantSender.

PR #716, opened 2024-11-12
Provider query manager improvements

gammazero:
- Use atomic value to protect variable timeout.
- Remove `inProgressRequestStatuses` map when empty.
- Do not append to slice forever. Use type that reuses items removed from front of list (e.g. [channelqueue](https://pkg.go.dev/github.com/gammazero/channelqueue), [deque](https://pkg.go.dev/github.com/gammazero/deque)
- Add configurable bitswap options:
  - `WithMaxConcurrentFinds`
  - `WithMaxProvidersPerFind`
- Make bitswap options configurable through client.


hsanjuan:
@gammazero heads up that I'm not looking into this one until #641  is through, at which point it will need to be redone...

PR #717, opened 2024-11-12
feat: add WithMiddleware option to the delegated routing server

2color:
Because the router is defined inside the server here, consumers of the server have no clean way to instrument each delegated routing endpoint independently. 

This PR adds a new option that allows passing in middleware similar to the https://github.com/slok/go-http-metrics/blob/master/examples/gorilla/main.go example.


2color:
Closing this, as this won't help with setting a different handler ID per route.

PR #721, opened 2024-11-14
fix(bitswap/client/providerquerymanager): use non-timed out context for tracing

aschmahmann:
@gammazero going to test this out, but this seems like a relatively simple fix to the issue where we cancel the context which stops the tracing. Instead we pass the uncancelled context from the parent.

Note: I skipped the changelog since it seems like a fairly minor change, but can add if you think it makes sense


gammazero (commented):
I still see one missing item in the trace, but this is better.

gammazero (approved):

PR #718, opened 2024-11-13
feat(routing/http/server): expose prometheus metrics 

2color:
## What's in this PR

- **feat: add instrumentation with handler label to delegated routing endpoints**
- **feat: allow passing in custom prometheus registry**

## Why

Until now, we relied on instrumentation in consumers of the delegated routing server, e.g. [in someguy](https://github.com/ipfs/someguy/blob/49038325fb109bc04e070053f03a76f8642b225d/server.go#L126). The problem is that you cannot get endpoint/handler specific metrics. 

The duration buckets were chosen based on [probelab data](https://probelab.io/ipfs/amino/#performance) and production someguy metrics. In many cases, requests take over 10 seconds.

PR #722, opened 2024-11-18
refactor: default to prometheus.DefaultRegisterer

lidel:
This PR contains changes on top of https://github.com/ipfs/boxo/pull/718 

It replaces `prometheus.NewRegistry()` calls with `prometheus.DefaultRegisterer` to ensure that by default boxo users who do not specify an explicit registry are not missing some of the metrics.

This convention is what go-libp2p does already, and I feel boxo should follow, to remove confusion why some metrics are present, and some are missing.

## TODO

- [x] merge https://github.com/ipfs/boxo/pull/718 and rebase this on main branch
- [x] sharness tests will fail until fix from https://github.com/ipfs/kubo/pull/10567  lands in Kubo master
- [x] refactor so boxo modules won't panic if `prometheus.DefaultRegisterer` is used and tests are run in parallel
   - [x] log error where possible
   - [x] use unique prefix for metric name where not possible to avoid `registry.MustRegister`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ The following emojis are used to highlight certain changes:
 
   - The above is only necessary if content routing is needed. Otherwise:
 
-```
+```go
 	// Create network: no contentRouter anymore
 	bswapnet := network.NewFromIpfsHost(host)
 	// Create Bitswap: a new "discovery" parameter set to nil (disable content discovery)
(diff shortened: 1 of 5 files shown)


lidel:
I think the approach of special-casing tests is too tedious, especially if people elaready use boxo modules in their tests, we will be breaking  them forcing people to add similar wrappers.

We have a different approach in [boxo/gateway/metrics](https://github.com/ipfs/boxo/blob/main/gateway/metrics.go#L299) where we catch "duplicate registration" error  and log error instead. This way all tests can run in parallel without extra wrapper. 


So next steps here:

- [x] undo changes to tests and refactor to use approach from gateway

lidel:
Pushed refactored version. 

Boxo will no longer panic when `prometheus.DefaultRegisterer` is used, and someone runs ours (or theirs) tests in parallel. 

This way
- things work: boxo users are never missing metrics when they run with defaults
- ux remains good: existing boxo users will not have to refactor their existing tests due to vague panic occuring after boxo update

hsanjuan (approved):

PR #723, opened 2024-11-19
fix(bitswap/server): pass context to server engine to register metrics

gammazero:
Fixes invalid metrics errors logged by kubo at daemon startup.

Fixes the problem from  https://github.com/ipfs/kubo/pull/10567#pullrequestreview-2443616123


hsanjuan (approved):

lidel (approved):
Thanks! I think this is good enough fix problem at hand (metrics regression due to goprocess cleanup that landed with https://github.com/ipfs/boxo/pull/710).


In future we may want to rework the way bitswap metrics are gathered, but that would be a separate PR (while working on https://github.com/ipfs/boxo/pull/722 i've identified few places where we do different things, eg. boxo/gateway only uses default registry and ignores duplicate registration errors)

PR #725, opened 2024-11-20
fix(bitswap/client/providerquerymanager): don't end trace span until …

gammazero:
…all providers are returned

cherry picked from #641


gammazero:
We can merge this now or wait for #641 to merge

aschmahmann (approved):
LGTM, this way we don't have to wait on the other PR to get better tracing results

PR #729, opened 2024-11-27
Add debug logging for deduplicated queries

gammazero:
Also, added comment about why request context is not used in `inProgressRequestStatus`.

Would deduplicated requests be worth tracking with a prometheus gauge?

### DIFF
--- a/routing/providerquerymanager/providerquerymanager.go
+++ b/routing/providerquerymanager/providerquerymanager.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap/zapcore"
 )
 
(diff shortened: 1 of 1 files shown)


hsanjuan (approved):
Sure why not

routing/providerquerymanager/providerquerymanager.go
```diff
@@ -504,6 +507,7 @@ func (npqm *newProvideQueryMessage) handle(pqm *ProviderQueryManager) {
 		}
 	} else {
 		trace.SpanFromContext(npqm.ctx).AddEvent("JoinQuery", trace.WithAttributes(attribute.Stringer("cid", npqm.k)))
+		log.Debugf("Joined existing query for cid %s which now has %d queries in progress", npqm.k, len(requestStatus.listeners)+1)
```
lidel: not sure if meaningful, but maybe wrap this with `if log.Level().Enabled(zapcore.DebugLevel) {` so len is not calculated needlessly?

PR #732, opened 2024-11-28
[skip changelog] bitswap/client: fix wiring when passing custom providerFinder

hsanjuan:
When passing in a custom provider finder, we need to pass it to the session. We were not doing that, so lookups were effectively disabled in that case.


hsanjuan:
I will add a test tomorrow

gammazero (approved):

PR #734, opened 2024-12-01
chore: no lifecycle context to shutdown ProviderQueryManager

gammazero:
Call `Close` instead of cancelling long-lived context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ The following emojis are used to highlight certain changes:
 - `gateway`: `NewCacheBlockStore` and `NewCarBackend` will use `prometheus.DefaultRegisterer` when a custom one is not specified via `WithPrometheusRegistry` [#722](https://github.com/ipfs/boxo/pull/722)
 - `filestore`: added opt-in `WithMMapReader` option to `FileManager` to enable memory-mapped file reads [#665](https://github.com/ipfs/boxo/pull/665)
 - `bitswap/routing` `ProviderQueryManager` does not require calling `Startup` separate from `New`. [#741](https://github.com/ipfs/boxo/pull/741)
+- `bitswap/routing` ProviderQueryManager does not use liftcycle context.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


hsanjuan:
I understand the part about auto-starting it... but removing the context? It is nice that things die on context cancellation by themselves.

gammazero:
@hsanjuan IMO Contexts are best for request-scoped operations. For globally shared resources, traditional patterns may be better. Passing a context info `New` means the caller needs to understand whether the context is used to cancel the call to `New`, or whether it is used in place of a `Close` function for the object being created.

Also, passing the context into `New` generally requires storing the context in the struct that defines the object. According to the [documentation](https://golang.org/pkg/context/#:~:text=Do%20not%20store%20Contexts%20inside), contexts should never be stored in struct type, but rather it should be passed through as an argument to a function. There are a few exceptions; context.CancelFunc is one such exception that can be put in a structure.

hsanjuan:
> @hsanjuan IMO Contexts are best for request-scoped operations. For globally shared resources, traditional patterns may be better. Passing a context info `New` means the caller needs to understand whether the context is used to cancel the call to `New`, or whether it is used in place of a `Close` function for the object being created.
> 
> Also, passing the context into `New` generally requires storing the context in the struct that defines the object. According to the [documentation](https://golang.org/pkg/context/#:~:text=Do%20not%20store%20Contexts%20inside), contexts should never be stored in struct type, but rather it should be passed through as an argument to a function. There are a few exceptions; context.CancelFunc is one such exception that can be put in a structure.

Interesting! I've somehow always considered that contexts could be used for objects' lifecycle management because deriving a context and cancelling the mother of all contexts facilitated cascading cleanup of things, rather than remembering to Close() here and there. Thanks for the pointer! I will not discuss The Go Way :)

gammazero:
> I've somehow always considered that contexts could be used for objects' lifecycle management

That is still a valid use for context, I just have a particular opinion. I think we need a consistent design paradigm that we can all mostly agree on. So, I pose it as a question to the maintainers here: **Do we generally want to limit the scope of a context to that of a request or session, or do we also want to use contexts to for objects' lifecycle management?**

hsanjuan:
(We can merge and discuss this later, I think this is an improvement and shouldn't stay behind waiting on discussions)

PR #737, opened 2024-12-03
Bitswap default ProviderQueryManager uses explicit options

gammazero:
Set explicit options for bitswap default PQM, to make it more conservative and isolate it from a PQM without options.

### DIFF
--- a/bitswap/client/client.go
+++ b/bitswap/client/client.go
@@ -97,13 +97,18 @@ func WithoutDuplicatedBlockStats() Option {
 	}
 }
 
-// WithDefaultProviderQueryManager indicates wether we should use a the
-// default ProviderQueryManager, a wrapper of the content Router which
-// provides bounded paralelism and limits for these lookups. The
-// ProviderQueryManager setup by default uses maxInProcessRequests = 6 and
-// maxProviders = 10.  To use a custom ProviderQueryManager, set to false and
-// wrap directly the content router provided with the WithContentRouting()
-// option. Only takes effect if WithContentRouting is set.
+// WithDefaultProviderQueryManager indicates whether to use the default
+// ProviderQueryManager as a wrapper of the content Router. The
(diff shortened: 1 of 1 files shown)


bitswap/client/client.go
```diff
@@ -176,7 +181,7 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder Pr
 
 	if bs.providerFinder != nil && bs.defaultProviderQueryManager {
 		// network can do dialing.
-		pqm, err := rpqm.New(ctx, network, bs.providerFinder, rpqm.WithMaxProviders(10))
+		pqm, err := rpqm.New(ctx, network, bs.providerFinder)
```
aschmahmann: AFAIK we don't yet have any evidence that this is a better default to set, I wouldn't change the defaults for people unless we have done some tests demonstrating that this is the case.

Note: setting this number to be unbounded might be wasteful since if one request has many providers we might connect to lots of them even though we only need a few, it might be fine in practice though (especially given we have a 5s timeout for the entire query). If we wanted to set the number to be higher / unbounded but were noticing issues related to having too many connections there are probably some more sophisticated approaches we could take instead but better to leave that until it looks necessary 😅

gammazero: The point of this PR is not to change the defaults, but that the default PQM should use its own defaults. So if I understand your comment, then what this PR should also do then is to set `defaultMaxProviders = 10` in `providerquerymanager.go`. Right? 

Or, are you saying that this PQM is the _client's_ default which should have 10 max providers, even though a PQM has a default of unlimited providers?

hsanjuan: Yeah, we made the PQM-defaults not be the Bitswap-Default-PQM defaults because those were more conservative. I am fine with aligning. Increasing what we had a little bit probably doesn't hurt.

PR #743, opened 2024-12-05
fix(bitswap): autostart network on new client/server

### LINKED ISSUE #68: [ipfs/go-bitswap] Bitswap client usage unintuitive

fabiobozzo:
**Add NewWithAutoStart constructors for bitswap client and server**

Fixes [#68](https://github.com/ipfs/boxo/issues/68)

This PR adds `NewWithAutoStart` constructors to both client and server packages that automatically call `network.Start()`. This provides a more intuitive API when using standalone client or server, avoiding the common footgun where users need to manually call `network.Start()`.

The existing `New()` constructors are preserved for backwards compatibility and now include documentation about the `Start()` requirement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ The following emojis are used to highlight certain changes:
 - `gateway`: `NewCacheBlockStore` and `NewCarBackend` will use `prometheus.DefaultRegisterer` when a custom one is not specified via `WithPrometheusRegistry` [#722](https://github.com/ipfs/boxo/pull/722)
 - `filestore`: added opt-in `WithMMapReader` option to `FileManager` to enable memory-mapped file reads [#665](https://github.com/ipfs/boxo/pull/665)
 - `bitswap/routing` `ProviderQueryManager` does not require calling `Startup` separate from `New`. [#741](https://github.com/ipfs/boxo/pull/741)
+- `bitswap/client`, `bitswap/server`: Added `NewWithAutoStart` constructors that automatically call `network.Start()`, providing a more intuitive API when using standalone client or server. 
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


gammazero:
Thank you for your interest in contributing to this project. However, we are choosing to decline this PR in the interest of having fewer / simpler APIs.

By simpler, this means fewer expectations. For example:
- If a client or server is created using `NewWithAutoStart`, then is it expected that the client or server also calls `network.Stop()` or should that still be done outside the client or server?
- Should there also be a new version of `client.Close` named `client.CloseWithAutoStop`
- What is the best way to start multiple test clients using the same network? Seems that 'New` is still needed to register all the clients and then start the network, so it is not just for backward compatibility in this case.

In other words, the preference is to have only one way to do things, even at the expense of having to remember to call `network.Start`

What would be preferable IMO is to have the network started when the network is created, so there is no need to explicitly start it as a separate step. Then each client or server can register itself separately, and that can be done within the client's or server's `New`. This means that the `BitSwapNetwork` interface will need to change to remove the `Start` function and instead have  a `Register(...Receiver)` function. There may be additional changes to make sure that registering a new receiver works if the network is already started. All this is potentially much more work without any real functional gain, only cleaner style, over the current API so we have not done this.

PR #750, opened 2024-12-12
bitswap/client/messagequeue: expose dontHaveTimeoutMgr configuration

### LINKED ISSUE #701: bitswap(messagequeue): Expose DONT_HAVE timeout knobs

gammazero:
- Create an exported `DontHaveTimeoutConfig` to hold configuration values for `dontHaveTimeoutMrg`
- Provide function `DefaultDontHaveTimeoutConfig` to return a `DontHaveTimeoutConfig` populated with default values
- Bitswap client `New` has new option for caller to supply a `DontHaveTimeoutConfig`
- If the required `onHaveTimeout` function is `nil`, then disable the `dontaHaveTimeoutMgr`
- The option to disable donthavetimeout handling is respected and the `dontHaveTimeoutMgr` is disabled if this option is `false`.
 
This PR replaces #703 

Closes #701 

Credit to @Wondertan for the original PR #703 that this one is based on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,14 +17,17 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 - `bitswap/client`: Improved timeout configuration for block requests
-  - Exposed `DontHaveTimeoutConfig` to hold configuration values for `dontHaveTimeoutMgr` which controls how long to wait for requested block, and if timeout should emit a synthetic DontHave response
+  - Exposed `DontHaveTimeoutConfig` to hold configuration values for `dontHaveTimeoutMgr` which controls how long to wait for requested block before emitting a synthetic DontHave response
   - Added `DefaultDontHaveTimeoutConfig()` to return a `DontHaveTimeoutConfig` populated with default values
   - Added optional `WithDontHaveTimeoutConfig` to allow passing a custom `DontHaveTimeoutConfig`
-  - If `DontHaveTimeoutConfig.onDontHaveTimeout` is `nil`, then `dontaHaveTimeoutMgr` is disabled and synthetic DontHave won't be emitted on timeout
+  - Setting `SetSendDontHaves(false)` works the same as before. Behind the scenes, it will disable `dontHaveTimeoutMgr` by passing a `nil` `onDontHaveTimeout` to `newDontHaveTimeoutMgr`.
+
 
 ### Changed
 
 ### Removed
+
+
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `bitswap/client`: Improved timeout configuration for block requests
+  - Exposed `DontHaveTimeoutConfig` to hold configuration values for `dontHaveTimeoutMgr` which controls how long to wait for requested block, and if timeout should emit a synthetic DontHave response
+  - Added `DefaultDontHaveTimeoutConfig()` to return a `DontHaveTimeoutConfig` populated with default values
+  - Added optional `WithDontHaveTimeoutConfig` to allow passing a custom `DontHaveTimeoutConfig`
+  - If `DontHaveTimeoutConfig.onDontHaveTimeout` is `nil`, then `dontaHaveTimeoutMgr` is disabled and synthetic DontHave won't be emitted on timeout
```
gammazero: @lidel From a user's perspective this last point did not change. The only change was internal - to not run the timeout manager, instead of running it but calling a no-op function.

lidel: @gammazero  Indeed, I was very confused what is the point of `DontHaveTimeoutConfig.onDontHaveTimeout`, after closer inspection it is not used for anything(?)

The one passed to `newDontHaveTimeoutMgr` as a separate argument is used instead, and we create it in bitswap client.

Removed `DontHaveTimeoutConfig.onDontHaveTimeout` in 068cd7b42e95b50e048f3a7a27e2e8dc8b0bed5f, now the public struct is just timeouts. 

Updated this line in changelog too. 
If there is not more to it, lgtm.

lidel (approved):

gammazero:
Thank you for removing the unused value from config, I had forgotten to do that. Everything looks good at this point, so will merge.

PR #751, opened 2024-12-12
mfs: clean cache on sync

hsanjuan:
There have been multiple reports of people complaining that MFS slows down to a crawl and triggers OOM events when adding many items to a folder.

One underlying issue is that the MFS Directory has a cache which can grown unbounded. All items added to a directory are cached here even though they are already written to the underlying Unixfs folder on mfs writes.

The cache serves Child() requests only. The purpose is the a Directory can then give updated information about its children without being "up to date" on its own parent: when a new write happens to a folder, the parent folder needs to be notified about the new unixfs node of their child. With this cache, this notification can be delayed until a Flush() is requested, one of the children triggers a bubbling notification, or the parent folder is traversed.

When the parent folder is traversed, it will write cached entries of its childen to disk, including that of the current folder. In order to write the current folder it has triggered a GetNode() on it, which in turn has written all the cached nodes  in the current folder to disk. This cascades. In principle, leafs do not have children and should not need to be rewritten to unixfs, but without caching the TestConcurrentWrites tests start failing. Caching provides a way for concurrent writes to access a single reference to a unixfs node, somehow making things work.

Apart from cosmetic changes, this commits allows resetting the cache when it has been synced to disk. Other approaches have been tested, like fully removing the cache, or resetting it when it reaches certain size. All were insatisfactory in that MFS breaks in one way or other (i.e. when setting maxCacheSize to 1). Setting size to ~50 allows tests to pass, but just because no tests are testing with 50+1 items.

This change does not break any tests, but after what I've seen, I can assume that concurrent writes during a Flush event probably result in broken expectations, and I wouldn't be surprised if there is a way to concatenate writes, lookups and flushes on references to multiple folders on the tree and trigger errors where things just work now.

The main problem remains: the current setup essentially keeps all mfs in memory. This commit could allevaite the fact that reading the directory object triggers as many unixfs.AddChild() as nodes in the directory. After this, AddChild() would be triggered only on nodes cached since the last time.

I don't find a reliable way of fixing MFS and I have discovered multiple gotchas into what was going to be a quick fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+* `mfs`: directory cache is now cleared every time the directory node is read, somewhat limiting unbounded growth and time to sync it to the underlying unixfs.
+
 ### Security
 
 ## [v0.25.0]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
-* `mfs`: directory cache is now cleared every time the directory node is read, somewhat limiting unbounded growth and time to sync it to the underlying unixfs.
+* `mfs`: directory cache is now cleared on Flush(), liberating the memory used by the otherwise ever-growing cache. References to directories and sub-directories should be renewed after flushing.
 
 ### Security
 
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -99,10 +99,11 @@ func (d *Directory) localUpdate(c child) (*dag.ProtoNode, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
-	err := d.updateChild(c)
+	err := d.unixfsDir.AddChild(d.ctx, c.Name, c.Node)
 	if err != nil {
 		return nil, err
 	}
+
 	// TODO: Clearly define how are we propagating changes to lower layers
 	// like UnixFS.
 
@@ -125,29 +126,8 @@ func (d *Directory) localUpdate(c child) (*dag.ProtoNode, error) {
 	// TODO: Why do we need a copy?
 }
 
-// Update child entry in the underlying UnixFS directory.
-func (d *Directory) updateChild(c child) error {
-	err := d.unixfsDir.AddChild(d.ctx, c.Name, c.Node)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
-
 func (d *Directory) Type() NodeType {
 	return TDir
-}
-
-// childNode returns a FSNode under this directory by the given name if it exists.
-// it does *not* check the cached dirs and files
-func (d *Directory) childNode(name string) (FSNode, error) {
-	nd, err := d.childFromDag(name)
-	if err != nil {
-		return nil, err
-	}
-
-	return d.cacheNode(name, nd)
 }
 
 // cacheNode caches a node into d.childDirs or d.files and returns the FSNode.
@@ -219,7 +199,12 @@ func (d *Directory) childUnsync(name string) (FSNode, error) {
 		return entry, nil
 	}
 
-	return d.childNode(name)
+	nd, err := d.childFromDag(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.cacheNode(name, nd)
 }
 
 type NodeListing struct {
@@ -361,29 +346,24 @@ func (d *Directory) AddChild(name string, nd ipld.Node) error {
 		return err
 	}
 
-	err = d.unixfsDir.AddChild(d.ctx, name, nd)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return d.unixfsDir.AddChild(d.ctx, name, nd)
 }
 
-func (d *Directory) sync() error {
+func (d *Directory) syncCache(clean bool) error {
 	for name, entry := range d.entriesCache {
 		nd, err := entry.GetNode()
 		if err != nil {
 			return err
 		}
 
-		err = d.updateChild(child{name, nd})
+		err = d.unixfsDir.AddChild(d.ctx, name, nd)
 		if err != nil {
 			return err
 		}
 	}
-
-	// TODO: Should we clean the cache here?
-
+	if clean {
+		d.entriesCache = make(map[string]FSNode)
+	}
 	return nil
 }
 
@@ -408,7 +388,7 @@ func (d *Directory) GetNode() (ipld.Node, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
-	err := d.sync()
+	err := d.syncCache(true)
 	if err != nil {
 		return nil, err
 	}
--- a/mfs/dir.go
+++ b/mfs/dir.go
@@ -323,7 +323,7 @@ func (d *Directory) Unlink(name string) error {
 }
 
 func (d *Directory) Flush() error {
-	nd, err := d.GetNode()
+	nd, err := d.getNode(true)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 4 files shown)


mfs/dir.go
```diff
@@ -99,10 +99,11 @@ func (d *Directory) localUpdate(c child) (*dag.ProtoNode, error) {
 	d.lock.Lock()
 	defer d.lock.Unlock()
 
-	err := d.updateChild(c)
+	err := d.unixfsDir.AddChild(d.ctx, c.Name, c.Node)
```
hsanjuan, line 102: These are just subtituting one-liner functions for the actual call.

mfs/dir.go
```diff
@@ -125,31 +126,10 @@ func (d *Directory) localUpdate(c child) (*dag.ProtoNode, error) {
...
-// it does *not* check the cached dirs and files
-func (d *Directory) childNode(name string) (FSNode, error) {
-	nd, err := d.childFromDag(name)
-	if err != nil {
-		return nil, err
-	}
-
-	return d.cacheNode(name, nd)
-}
-
```
hsanjuan, line 152: one-liners removed.

mfs/dir.go
```diff
@@ -361,29 +346,24 @@ func (d *Directory) AddChild(name string, nd ipld.Node) error {
...
 		if err != nil {
 			return err
 		}
 	}
-
-	// TODO: Should we clean the cache here?
-
+	if clean {
+		d.entriesCache = make(map[string]FSNode)
+	}
```
hsanjuan, line 366: The only relevant change.

mfs/dir.go
```diff
@@ -361,29 +346,24 @@ func (d *Directory) AddChild(name string, nd ipld.Node) error {
...
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return d.unixfsDir.AddChild(d.ctx, name, nd)
 }
 
-func (d *Directory) sync() error {
+func (d *Directory) syncCache(clean bool) error {
```
gammazero: Is there any situation where we would not want `clean = true`? Seems like we should always clear the cache.

gammazero (approved):
LGMT. One question.

mfs/dir.go
hsanjuan: Originally I wanted to not clean on GetNode() while cleaning on reaching maxCacheSize.

I am not sure this PR is good in general... probably there should be some refcounting about which objects can be pushed out of the cache. My feeling based on how tests misbehave when removing things is at least that files that are open for writing should probably stay...

hsanjuan: And there's also the issue that this still allows the cache to grow forever until sync. And such sync may take minutes or hours...

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+* `mfs`: directory cache is now cleared every time the directory node is read, somewhat limiting unbounded growth and time to sync it to the underlying unixfs.
```
lidel: I'm scared to replace known MFS problems with new unknown MFS problems, but also not familiar with codebase well enough to have informed opinion :(

I know folks have pipelines that do a lot of concurrent writes to MFS via `ipfs files` CLI/RPC and likely figured out what works and what does not.

Are we sure new version removes or at least decreases surface for data loss during concurrent writes?

Perhaps useful context: in kubo, we have --flush flag that (if true) calls `mfs.FlushPath` after each MFS operation.

Intention behind disabling flush is to allow user to avoid sync during many small operations and flush to disk until "final" version of file or directory is ready, but that comes at consistency risk if there is more than one thread doing mutation:

Most of the subcommands of `ipfs files` accept the `--flush` flag. It defaults
to true. Use caution when setting this flag to false. It will improve
performance for large numbers of file operations, but it does so at the cost
of consistency guarantees.

So effectively, patching a lot of things should be faster with flush=false but you need to do things sequentially. This is the mental model Kubo users have.

Are edge cases mentioned in  https://github.com/ipfs/boxo/pull/751/files avoidable  with  `--flush=true` ? Or are we introducing racy risk there, where it was not before?

If we make "safe" default less "safe", could/should we tie cache cleanup to happen only in  mfs.FlushPath instead (explicit `--flush=true`)? (call `syncCache` with `true` equivalent only if in the context of explicit `mfs.FlushPath`?)

hsanjuan: That might be an option... 

I did not know we flushed by default... that also means this commit will essentially kill the cache on every operation. I cannot say with certainty but this probably brings the new "unknown MFS problems" as it will be equivalent to disabling the cache.

Let's do something, let me spend another while trying to figure out why exactly concurrent writes to a single file break with things like cache-size=1 and if something else is broken that way. Maybe I can find a workaround to feel better about it.

gammazero:
This helps/solves a known problem, so this change does have positive value. Can we run this is cluster or somewhere  to gain more confidence that there are not any negative behaviors with this change?

hsanjuan:
Ok,

I have been checking... as said, the cache solves dealing with writes that have not been bubbled to mfs root and references which have not been renewed. The idea is that your cache has a reference that was used to make changes, so even if those changes aren't reflected in your MFS dir, since you looked-up a node using the cached version, things work.

If the cache wasn't there, it would be more cumbersome to work with MFS (do something, bubble up, re-obtain up-to-date references). At the same time, the whole thing works because we are holding all the MFS structure in memory. Only the `root` has a method to clear up cache entries in the root dir (`FlushMemFree`) with big warnings around it.

The whole thing is very hairy, as said, things are written to unixfs, dag-service... persisted, but changes are not really distributed along the whole tree all the time. One could think that MFS-changes could all be in memory until Flush(), at which point the unixfs nodes are updated and they are committed to the DAGService etc. but this is not the case.

Anyways, what I've done now is to clear the directory cache on Flush(). I think this is relatively safe in that, flushing hopefully indicates that you are generally done with what you were doing. This does not break any tests, but also note that there is no general "flushing" of the MFS tree.

Flushing of a file/folder notifies the parents so that they update references, but parents don't "flush" themselves. I have tried clearing the cache on all parents but this again introduces issues with the concurrent writes test which is writing on multiple goroutines and therefore contains references that become outdated unless cached up to the top of the tree.

In Kubo, we call FlushPath which calls Flush() on a specific mfs node and not on the parent folder, so we have to discuss if we somehow introduce a general MFS free-mem flush.

lidel:
The usual pattern of Kubo's MFS user is to mutate a directory with `files cp|rm|mv`, and then do `ipfs files stat` on it to learn new CID, so flushing to disk / cleaning cache of only current dir and not its parents should be fine.  

@hsanjuan mind  opening a PR draft in Kubo with boxo from this PR to test if legacy MFS tests are all passing there as well, just a precaution (iirc `test/sharness/t0250-files-api.sh`, `test/sharness/t0040-add-and-cat.sh` are not run in this repo).

hsanjuan:
> @hsanjuan mind opening a PR draft in Kubo with boxo from this PR to test if legacy MFS tests are all passing there as well, just a precaution (iirc `test/sharness/t0250-files-api.sh`, `test/sharness/t0040-add-and-cat.sh` are not run in this repo).

https://github.com/ipfs/kubo/pull/10628 seems we're good on that front.

I still have to test if my computer blows up when adding files. Will report back.

hsanjuan:
>  so flushing to disk / cleaning cache of only current dir and not its parents should be fine.

What I mean, is that we should flush the parent dir. i.e. `ipfs files cp /ipfs/Qmxyz /mfs/myfile` flushes `/mfs/myfile` by default, but this won't clear the cache of the `/mfs` folder.

I have verified that with the changes in this PR only Kubo still leaks memory when running ~6000 `ipfs files cp` operations and then starts freezing. However, if calling `Flush` on the parent folder, the memory consumption is greatly reduced and things no longer explode (there's a slowdown which I think I can attribute to the HAMT growing, making inserts more and more expensive). So perhaps we can do that? Move ahead with this and then ensure `files cp/mv/write` Flush the parent folder?

PR #754, opened 2024-12-13
improve mfs republisher

gammazero:
- Get updated values while retrying to publish
- Prefer Close to lifecycle context
- Do not require call to Start separate from New

### DIFF
--- a/mfs/repub.go
+++ b/mfs/repub.go
@@ -111,6 +111,11 @@ func (rp *Republisher) Update(c cid.Cid) {
 // duration. The `quick` timer allows publishing sooner if there are no more
 // updates available.
 //
+// In other words, the quick timeout means there are no more values to put into
+// the "batch", so do update. The long timeout means there are that the "batch"
+// is full, so do update, even though there are still values (no quick timeout
+// yet) arriving.
+//
 // If a publish fails, retry repeatedly every `longer` timeout.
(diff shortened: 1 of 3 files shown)


mfs/repub.go
```diff
@@ -123,19 +141,20 @@ func (rp *Republisher) Run(lastPublished cid.Cid) {
 				break
 			}
 
-			// If we aren't already waiting to publish something,
-			// reset the long timeout.
+			// If mot already waiting to publish something, reset the long
```
hsanjuan: mot?

hsanjuan (approved):
part of me wonders if something similar cannot be achieved with a single timer that is reset for long/short intervals depending how the publishing went.

gammazero:
> part of me wonders if something similar cannot be achieved with a single timer that is reset for long/short intervals depending how the publishing went.

As far as I can see both timers are needed. It like when batching by count: you need to know both the batch size and how many values are in the batch so far. This is batching, but with time.

The quick timer (waiting for additional results) keeps getting reset when another result arrives. If new results are coming in continuously, resetting quick timer each time, there needs to be another timer running that will trigger an update at some point even though new values keep arriving.

In other words, the quick timeout means there are no more values to put into the "batch", so do update. The long timeout means there are that the "batch" is full, so do update, even though there are still values (no quick timeout yet) arriving.

PR #756, opened 2024-12-16
Fix leak due to cid queue never getting cleaned up

gammazero:
This removes entries from the queue portion of CidQueue that were removed from the map portion.

Credit to @Wondertan for finding and proposing fix.


bitswap/client/internal/session/cidqueue.go
```diff
@@ -31,14 +31,7 @@ func (cq *cidQueue) Pop() cid.Cid {
...
 	// Lazily delete from the list any cids that were removed from the set
-	if cq.elems.Len() > cq.eset.Len() {
-		for i := 0; i < cq.elems.Len(); i++ {
-			c := cq.elems.PopFront()
-			if cq.eset.Has(c) {
-				cq.elems.PushBack(c)
-			}
-		}
-	}
+	cq.GC()
```
hsanjuan: I don't think this function is called from anywhere. We could delete it.

hsanjuan: And if it is not called from anywhere, do we need the set?

What we could have is a set of CIDs that need to be skipped/removed on Pop().

So when we `Pop()`, we check if the CID is in the set. If so, we remove it from the set and Pop again (the inverse of what happens now on that regard). We can then save on GC calls, I think?

hsanjuan (changes requested):

bitswap/client/internal/session/cidqueue.go
gammazero: I think the reason to do the GC separate from the pop is because a number of CIDs are Popped in a loop. That would require traversing the CID list each Pop. Doing GC after the loop of Pops means only traversing the list once.

gammazero:
@hsanjuan I think my latest change satisfies what you were asking for. I will merge this PR in a bit unless you have any objections.

PR #759, opened 2024-12-16
Remove want-block sent tracking from sessionWantSender

gammazero:
Tracking want-blocks that were sent to a peer is not necessary to do in sessionwantsender. This is already handled in peermanager/peerwantmanager.

Removing this tracking reduces unnecessary memory use.


bitswap/client/internal/session/sessionwantsender.go
```diff
@@ -577,11 +570,6 @@ func (sws *sessionWantSender) sendNextWants(newlyAvailable []peer.ID) {
 func (sws *sessionWantSender) sendWants(sends allWants) {
 	// For each peer we're sending a request to
 	for p, snd := range sends {
-		// Piggyback some other want-haves onto the request to the peer
-		for _, c := range sws.getPiggybackWantHaves(p, snd.wantBlocks) {
-			snd.wantHaves.Add(c)
-		}
-
```
hsanjuan, line 584: Are we sure removing this has no effect?

swbt was introduced here: https://github.com/ipfs/go-bitswap/commit/b3a47bcf5080c734346fc39400802c35aedd6428#diff-92a0b168d045013a05e87675741bb3cfa356a674c6f1535fbe88fc8e89d31e53

It is not explained what exactly motivated it, but it sounds like it is somehow keeping track of some wants that would otherwise not be sent?

guillaumemichel (approved):
It seems that the only place where the use of `sentWantBlocksTracker` is questionable is [`getPiggybackWantHaves`](https://github.com/ipfs/boxo/blob/bcf475cfd7fb78762d595f8c4fad9dbaa0d3ff6a/bitswap/client/internal/session/sessionwantsender.go#L606).

```go
// sendNextWants sends wants to peers according to the latest information
// about which peers have / dont have blocks
func (sws *sessionWantSender) sendNextWants(newlyAvailable []peer.ID) {
	toSend := make(allWants)


	for c, wi := range sws.wants {
		// Ensure we send want-haves to any newly available peers
		for _, p := range newlyAvailable {
			toSend.forPeer(p).wantHaves.Add(c)
		}


		// We already sent a want-block to a peer and haven't yet received a
		// response yet
		if wi.sentTo != "" {
			continue
		}


		// All the peers have indicated that they don't have the block
		// corresponding to this want, so we must wait to discover more peers
		if wi.bestPeer == "" {
			// TODO: work this out in real time instead of using bestP?
			continue
		}


		// Send a want-block to the chosen peer
		toSend.forPeer(wi.bestPeer).wantBlocks.Add(c)


		// Send a want-have to each other peer
		for _, op := range sws.spm.Peers() {
			if op != wi.bestPeer {
				toSend.forPeer(op).wantHaves.Add(c)
			}
		}
	}


	// Send any wants we've collected
	sws.sendWants(toSend)
	...
}
```

`getPiggybackWantHaves` seems to cover a specific corner case (I doubt it was done voluntarily though).
1. A new `WANT(k)` is added to the client for cid `k`
2. A `WANT_BLOCK` is sent to `wi.bestPeer`, and `WANT_HAVE`s are sent to all other peers in `sws.spm.Peers()`
3. `wi.bestPeer` becomes unavailable, triggering `wi.sentTo` to be reset. (this is the only place where `wi.sentTo` can be reset).
4. At the next call of `sendNextWants`, `WANT_HAVE(k)` will be sent to all peers in `sws.spm.Peers()` again (as long as not peers have return a `DONT_HAVE`)
5. `getPiggybackWantHaves` should add `WANT_HAVE(kk)` for all `kk` in `sws.wants` where `kk != k` and `WANT_HAVE(kk)` wasn't sent to that peer before, for peers that are part of `sws.spm.Peers()` but NOT `newlyAvailable`

However I would say that this set of peers will always be empty.
1. If a peer was part of `sws.spm.Peers()` at the time the new `WANT(k)` has been added, then it would already have received a `WANT_HAVE`. 
2. If it wasn't part of `sws.spm.Peers()` at the time the new `WANT(k)` has been added , then it means that it would have been part of `newlyAvailable` at some point, and hence would have received the `WANT_HAVE`.

So I would say that `getPiggybackWantHaves` and the whole `sentWantBlocksTracker` logic can be removed safely, without impacting Bitswap behaviour.

gammazero:
@guillaumemichel Thank you for this careful review and analysis of the logic.

PR #760, opened 2024-12-16
[skip changelog] fix: Drop stream references on Close/Reset

MarcoPolo:
Lets connections get GC'd sooner (possibly before the streamMessageSender is GC'd)

There's a small extension to this pr in https://github.com/ipfs/boxo/tree/marco/with-AfterFunc that uses `context.AfterFunc`. But even without that, from my tests, this seems to help as well.

### DIFF
--- a/bitswap/network/ipfs_impl.go
+++ b/bitswap/network/ipfs_impl.go
@@ -82,9 +82,32 @@ type impl struct {
 	receivers []Receiver
 }
 
+// interfaceWrapper is concrete type that wraps an interface. Necessary because
+// atomic.Value needs the same type and can not Store(nil). This indirection
+// allows us to store nil.
+type interfaceWrapper[T any] struct {
+	t T
+}
+type atomicInterface[T any] struct {
+	iface atomic.Value
+}
+
+func (a *atomicInterface[T]) Load() T {
+	var v T
+	x := a.iface.Load()
+	if x != nil {
+		return x.(interfaceWrapper[T]).t
+	}
+	return v
+}
+
+func (a *atomicInterface[T]) Store(v T) {
+	a.iface.Store(interfaceWrapper[T]{v})
+}
+
 type streamMessageSender struct {
 	to     peer.ID
-	stream network.Stream
+	stream atomicInterface[network.Stream]
 	bsnet  *impl
 	opts   *MessageSenderOpts
 }
@@ -95,7 +118,7 @@ type HasContext interface {
 
 // Open a stream to the remote peer
 func (s *streamMessageSender) Connect(ctx context.Context) (network.Stream, error) {
-	stream := s.stream
+	stream := s.stream.Load()
 	if stream != nil {
 		return stream, nil
 	}
@@ -111,17 +134,22 @@ func (s *streamMessageSender) Connect(ctx context.Context) (network.Stream, erro
 	if err != nil {
 		return nil, err
 	}
+	if withCtx, ok := stream.Conn().(HasContext); ok {
+		context.AfterFunc(withCtx.Context(), func() {
+			s.stream.Store(nil)
+		})
+	}
 
-	s.stream = stream
+	s.stream.Store(stream)
 	return stream, nil
 }
 
 // Reset the stream
 func (s *streamMessageSender) Reset() error {
-	stream := s.stream
+	stream := s.stream.Load()
 	if stream != nil {
 		err := stream.Reset()
-		s.stream = nil
+		s.stream.Store(nil)
 		return err
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


bitswap/network/ipfs_impl.go
```diff
@@ -108,29 +135,39 @@ func (s *streamMessageSender) Connect(ctx context.Context) (network.Stream, erro
...
+		err := stream.Reset()
+		s.stream.Store(nil)
 		return err
 	}
 	return nil
 }
 
 // Close the stream
 func (s *streamMessageSender) Close() error {
-	return s.stream.Close()
```
hsanjuan, line 128: wouldn't nil-ifying `s.stream` here also free the stream reference to be GC'ed ?

hsanjuan: Also, I don't think `Close()` is used. And Reset() only on queue shutdown. Even or read/write errors the stream isn't closed nor reset, it seems.

hsanjuan (commented):
What is the motivation for this? I'm not sure streams are Closed/Reset at all much, so maybe that's the real problem.

I am not sure the code needs to be concurrency-safe either. Does it? The messagequeue seems to SendMessage sequentially.

bitswap/network/ipfs_impl.go
gammazero: Maybe we should add
```go
    defer sender.Close()
```
at this [location](https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/messagequeue/messagequeue.go#L558)?

I am guessing that the reason for not doing this is to keep the connection open, and instead rely on resource manager/connection manager to limit the number of open connections.

MarcoPolo:
> I am not sure the code needs to be concurrency-safe either. Does it? The messagequeue seems to SendMessage sequentially.

Correct. It's necessary if we use the AfterFunc as in this change (which was a small diff on this): https://github.com/ipfs/boxo/pull/763/files. Let me change this to remove that and move the atomic to that PR.

gammazero:
@MarcoPolo Thank you. IMO we will want both PRs anyway since it seems like we would want to nil the connection if the context was canceled anyway. That way we have a better chance of reconnecting _before_ using the dead connection instead of getting an error while using it.

bitswap/network/ipfs_impl.go
hsanjuan: (the location doesn't make sense?)

I guess there's nothing wrong with re-using a stream and only cleaning up on peer disconnect?

MarcoPolo: > wouldn't nil-ifying s.stream here also free the stream reference to be GC'ed ?

Are you suggesting we don't close the stream here and drop our reference? That would be wrong. We should clean up our resources when we are done with them.

Ideally we'd use one stream per message and close the stream once we're done as @gammazero suggests. I believe the reason we don't do that is that in Yamux the initial send buffer is [constant](https://github.com/hashicorp/yamux/blob/060257c97bcf6317b24df54b9f8abe8e451d4606/const.go#L125) per stream (rather than [configurable](https://datatracker.ietf.org/doc/html/rfc9000#section-4) both within a stream and per connection connection as in QUIC). This means that if our initial message is larger than that initial send capacity, then we have to wait for a round trip to send the whole message. By reusing the stream, the peer can increase our send buffer size in a later message.

It may be worth changing this in Yamux.

MarcoPolo: > I guess there's nothing wrong with re-using a stream and only cleaning up on peer disconnect?

The real issue in this code is that the cleanup of peer disconnects is slow. A lot of time is spent on acquiring the lock by many different goroutines (ping me on slack and I'll point you to the pprofs). This PR is merely a small refactor in opportunistically dropping the stream reference if we know it's done.

The benefit of the one stream per message approach is that you avoid the whole issue of maintaining a reference and knowing if a connection is even alive or not.

MarcoPolo:
Makes sense. I just tidied up this PR by moving the atomic things to the next PR.

bitswap/network/ipfs_impl.go
gammazero: I really like the idea of having one stream per message. Reusing the stream is like double-muxing just to avoid a buffer sizing issue. Being able to have different messages on the same connection is point of muxing and streams is the mechanism for doing that.

The initial stream buffer size is 256K. If most of our messages are smaller, then that is fine. I suspect we do need a larger buffer, particularly for block transfer. Seems like a change to yamux is worth it, unless the occasional extra round trip is not a huge problem.

gammazero: Discussion moved to issue: https://github.com/ipfs/boxo/issues/765

gammazero (approved):

PR #763, opened 2024-12-17
Marco/with after func

gammazero:
refactor: if the conn has a context, use `context.AfterFunc` for cleanup


MarcoPolo:
Note that go-libp2p doesn't yet support the conn context. I'm thinking it'll come in the next release. I just didn't want to rush it in here.

PR #781, opened 2025-01-07
fix(gateway): update DoH resolver for .eth DNSLink

### LINKED ISSUE #771: bug: resolver.cloudflare-eth.com is down; change default ENS resolver?
# Problem

DNSLinks that use ENS (websites on `.eth` TLD) are broken for users that run `boxo/gateway` with default settings (incl. [Kubo](https://github.com/ipfs/kubo#readme), [Rainbow](https://github.com/ipfs/rainbow#readme), [IPFS Desktop](https://github.com/ipfs/ipfs-desktop#readme)).

Cloudflare resolver at `https://resolver.cloudflare-eth.com/dns-query` is currently  broken and results no DNSLink results (`"Answer":[]` at the end):

```console
$ curl -s -H "accept: application/dns-json" "https://resolver.cloudflare-eth.com/dns-query?name=_dnslink.vitalik.eth&type=TXT"
{"AD":true,"CD":false,"RA":true,"RD":true,"TC":false,"Status":3,"Question":[{"name":"_dnslink.vitalik.eth.","type":16}],"Answer":[]
```

We've reported outage to Cloudflare, but if it does not get fixed until January when the team is back from holidays,  we should consider removing .eth support from implicit defaults in Boxo (and Kubo), or switch implicit default to a different DoH resolver.

# Solution

We could make things more robust by supporting fallbacks (https://github.com/ipfs/kubo/issues/8173) but for that we need more than one, and it seems ENS has only one stable resolver atm. 

So the options are:

## Solution (A): do nothing, wait for `resolver.cloudflare-eth.com` to be fixed

This happens once a year on average. Acceptable? :shrug:  

## Solution (B): remove default resolver for `.eth` 

This would break all ENS websites, and require all end users to choose or set up their own resolver.
Probably not what we want given ENS+IPFS use in the wider ecosystem, but writing this down here for completeness.

## Solution (C): switch URL to `https://dns.eth.limo/dns-query`

The DoH at `https://dns.eth.limo/dns-query` is a good candidate, seems to be well maintained and provided non-empty DNSLink in `"Answer"`:

```console
$ curl -s -H "accept: application/dns-json" "https://dns.eth.limo/dns-query?name=_dnslink.vitalik.eth&type=TXT"
{"Status":"0","TC":false,"Question":[{"name":"_dnslink.vitalik.eth","type":16}],"Answer":[{"name":"_dnslink.vitalik.eth","data":"dnslink=/ipfs/bafybeifvusbh4iunpvwjlowu47sxnt4hjlebx46kxi4yz5zdsoecfpkkei","type":16,"ttl":300}]}
```

(this is not final decision, consider this commit as a way fo kicking-off conversation what we should do in 2025 to minimize issues with proxied naming systems like ENS)

## Solution (D): ?

Other ideas welcome.

lidel:
> Ref. https://developers.cloudflare.com/fundamentals/api/reference/deprecations/#2025-07-01:
> image

Cloudflare decommissioned their resolver and started redirecting (HTTP 308) to https://dns.eth.link/dns-query which is an alias for eth.limo.

We change default to https://dns.eth.limo/dns-query to avoid redirects and cache splitting.

Closes #771 by implementing option (C)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+- `gateway` The default DNSLink resolver for `.eth` TLD changed to `https://dns.eth.limo/dns-query` [#781](https://github.com/ipfs/boxo/pull/781)
+
 ### Removed
 
 ### Fixed
--- a/gateway/dns.go
+++ b/gateway/dns.go
@@ -10,7 +10,7 @@ import (
 )
 
 var defaultResolvers = map[string]string{
-	"eth.":    "https://resolver.cloudflare-eth.com/dns-query",
+	"eth.":    "https://dns.eth.limo/dns-query",
 	"crypto.": "https://resolver.cloudflare-eth.com/dns-query",
 }
 


aschmahmann (approved):
LGTM

lidel (commented):
CI green, merging.

Issue #772, opened 2024-12-23
bug: all "Unstoppable Domains" (.crypto)  DoH resolvers are down


lidel:
This is similar to https://github.com/ipfs/boxo/issues/771 but at the time of writing this _all_ UD DoH resolvers seem to be down, breaking DNSLink websites loaded via `boxo/gateway` if the domain name uses the `.crypto` TLD.

Current implicit default is Cloudflare at `https://resolver.cloudflare-eth.com/dns-query` and it stopped returning DNSLink results earlier today (filled support ticket to CF).

The "official" DoH at `https://resolver.unstoppable.io/dns-query` just hangs and timeouts.

I guess we wait for Cloudflare fix, but if that does not start working we need to consider altenatives (are there any? removal?)

Update: see https://github.com/ipfs/boxo/issues/772#issuecomment-2578597255

lidel:
Friendly ping to https://github.com/orgs/unstoppabledomains/people: @nickshatilo @bonsaiben

How does DoH story look like for UD? Should we keep defaulting to Cloudflare (assuming its fixed soon)?
We would prefer using `https://resolver.unstoppable.io/dns-query` directly, but it seems to be offline.

I'm testing with:

```console
$ curl 'https://resolver.unstoppable.io/dns-query?name=_dnslink.test.crypto&type=TXT' -H 'accept: application/dns-json' -v
* Host resolver.unstoppable.io:443 was resolved.
* IPv6: (none)
* IPv4: 3.33.139.32
*   Trying 3.33.139.32:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
```

and just hangs.

Team is on holiday, but we would like to resolve this in January (either fix `.crypto`, or remove default  `resolver.cloudflare-eth.com` if not fixed by then)

lidel:
As noted in https://github.com/ipfs/boxo/issues/771#issuecomment-2573189494 now redirects to `https://dns.eth.link/dns-query`, which does not support UD (`.crypto`) domains. 

We are planning to replace default resolver with `https://resolver.unstoppable.io/dns-query` giving UD community a chance to get their act together and fix the DNS bridge at the domain name they control:

- https://github.com/ipfs/boxo/pull/782

If this does not happen and the DoH resolver at `https://resolver.unstoppable.io/dns-query` remains broken in Q1, we will remove the implicit resolver for `.crypto`.

lidel:
[...]

PR #782, opened 2025-01-07
fix(gateway): update DoH resolver for .crypto DNSLink

lidel:
> Ref. https://developers.cloudflare.com/fundamentals/api/reference/deprecations/#2025-07-01:
> image

Cloudflare decomissionned their resolver and started redirecting (HTTP 308) to https://dns.eth.link/dns-query which is an alias for eth.limo.

That endpoint does not support resolving DNSLink of `.crypto` TLD and always returns no results, effectively breaking those DNSLinks for everyone using them with IPFS systems.

This change switches the default implicit DoH resolver of `.crypto` TLS to one provided by UD project, allowing their community to fix issue without involving a third-party.

Allows UD community to fix resolution themselves – see #772

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The following emojis are used to highlight certain changes:
 ### Fixed
 
 - `gateway` Fix redirect URLs for subdirectories with characters that need escaping. [#779](https://github.com/ipfs/boxo/pull/779)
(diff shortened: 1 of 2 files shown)


lidel (commented):
CI is green, merging to make the source of UD issues more obvious to users trying to resolve `.crypto`.

This is provisional fix on our end to allow UD community to fix resolution themselves.
We will keep this status quo for 3-6 months – the future of implicit resolver will be discussed in https://github.com/ipfs/boxo/issues/772#issuecomment-2578597255

PR #784, opened 2025-01-08
do not send cancel message to peer that sent block

### LINKED ISSUE #694: bitswap: Don't send CANCEL to peer we got block from
### Overview
Currently, we [send CANCEL to every peer in a session](https://github.com/ipfs/boxo/blob/19bcc75077891427aba3e74cec467dde4d5a1af3/bitswap/client/internal/sessionmanager/sessionmanager.go#L175-L176) once we got a block from a peer. However, the serving peer can and [cleans the client's wantlint](https://github.com/ipfs/boxo/blob/19bcc75077891427aba3e74cec467dde4d5a1af3/bitswap/server/internal/decision/engine.go#L1046-L1049) after serving the block. This suggests that sending CANCEL to the serving peer is redundant.

### Proposal
Add peer parameter to [`PeerManager.SendCancels`](https://github.com/ipfs/boxo/blob/19bcc75077891427aba3e74cec467dde4d5a1af3/bitswap/client/internal/peermanager/peermanager.go#L157) and exclude it [when broadcasting cancels](https://github.com/ipfs/boxo/blob/19bcc75077891427aba3e74cec467dde4d5a1af3/bitswap/client/internal/peermanager/peerwantmanager.go#L259-L260).

gammazero:
The serving peer cleans the client's wantlint after serving the block, making sending CANCEL to the serving peer redundant. So, exclude the serving peer when sending cancels after receiving a block.

Closes #694


Wondertan (approved):

hsanjuan (approved):
So we are sure that we have not requested the same CIDs to the same peer via some other session etc. right?

My main worry would be we start leaking something.

Wondertan:
PeerManager also takes care of deduplication if multiple session want the same data

PR #787, opened 2025-01-10
bitswap: messagequeue: lock only needed sections

hsanjuan:
A couple of locks here seem to lock also while doing things like sending things on channels. This seems prone to deadlocks and is wasteful.


gammazero (approved):
The last unlock may be important if this ever blocks:
```
mq.events <- messageFinishedSending
```

PR #789, opened 2025-01-10
Define a `go_package` for protobuf, rename to a more unique `ipns-record.proto`

### LINKED ISSUE #788: record.proto is too generic and results in conflict with other packages
## Background
I work on the [k3s project](https://github.com/k3s-io/k3s), which has a large dependency trail. Recently we attempted to move to a new version of [spegel](https://github.com/spegel-org/spegel), which pulled in new versions of [libp2p/go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht/compare/v0.25.2..v0.28.1) which bumped boxo from from [v0.10.0 -> v0.24.3](https://github.com/ipfs/boxo/compare/v0.10.0...v0.24.3)

## Issue
All these dependencies thrown together has resulted in conflicts around `record.proto` registration:
```
panic: proto: file "record.proto" is already registered
	See https://protobuf.dev/reference/go/faq#namespace-conflict
```	

Looking thru our vendor list you can see that there are a multitude of dependencies that use a very generic `record.proto` for some protobuf.

```
grep -rF 'record.proto' vendor/
vendor/go.etcd.io/etcd/server/v3/wal/walpb/record.pb.go:// source: record.proto
vendor/go.etcd.io/etcd/server/v3/wal/walpb/record.pb.go:func init() { proto.RegisterFile("record.proto", fileDescriptor_bf94fd919e302a1d) }
vendor/github.com/libp2p/go-libp2p-record/pb/record.pb.go:// source: record.proto
vendor/github.com/libp2p/go-libp2p-record/pb/record.pb.go:func init() { proto.RegisterFile("record.proto", fileDescriptor_bf94fd919e302a1d) }
vendor/github.com/libp2p/go-libp2p-record/pb/record.proto:// `protoc --gogo_out=. record.proto`
vendor/github.com/libp2p/go-libp2p-kad-dht/pb/dht.proto:import "github.com/libp2p/go-libp2p-record/pb/record.proto";
vendor/github.com/libp2p/go-libp2p/p2p/protocol/identify/pb/identify.proto:  // github.com/libp2p/go-libp2p/core/peer/pb/peer_record.proto for message definitions.
vendor/github.com/libp2p/go-libp2p/p2p/protocol/identify/pb/identify.pb.go:	// github.com/libp2p/go-libp2p/core/peer/pb/peer_record.proto for message definitions.
vendor/github.com/libp2p/go-libp2p/core/peer/pb/peer_record.pb.go:// source: core/peer/pb/peer_record.proto
vendor/github.com/ipfs/boxo/ipns/pb/record.pb.go:// source: record.proto
vendor/github.com/ipfs/boxo/ipns/record.go://go:generate protoc -I=pb --go_out=pb --go_opt=module=github.com/ipfs/boxo/ipns/pb pb/record.proto
```
Looking at the FAQ around how to resolve this that https://protobuf.dev/reference/go/faq#namespace-conflict, it seems that boxo does not define a  unique enough `option go_package` for its protobuf. Additionally, it is likely a good idea to just change the name entirely from something so generic as `record`.

dereknola:
This is an attempt to address https://github.com/ipfs/boxo/issues/788

Note in the first commit that I first attempted to follow the online recommendations from:
 - https://protobuf.dev/reference/go/go-generated/#package
 - https://protobuf.dev/reference/go/faq/
 - https://rkiselenko.dev/blog/grpc-in-golang/
And only define a unique `option go_package`. This failed to resolve the issue as the go compile still seems to complain and is very upset that the file itself is still called `record.proto`. 

Thus I also renamed to file itself to something less generic. Note that this should not affect external use, as the package is still `pb` and since https://github.com/ipfs/boxo/pull/339, external usage of the protobuf is around the `ipns` package, not the `ipns/pb` package directly.

If someone with more knowledge around protobuf generation can see a better way to address the issue or improve this PR, please let me know.


ipns/pb/ipns-record.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package github.com.boxo.ipns.pb;
 
-option go_package = ".;pb";
```
dereknola, line 5: Based off my reading around, the usage of a `.` for the package name is essentially a no-op and is not recommended. https://github.com/golang/protobuf/issues/1249#issuecomment-736859185

guillaumemichel: IIUC `.` tells protoc to generate the file in the same directory, and `pb` indicates the go package name

guillaumemichel:
@dereknola does the updated protobuf actually fix your issues?

I suspect that it might be coming from other protobufs where the package name might be too generic such as [go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht/blob/dbb8d9efab982decb474995e7f91d0101ef7233a/pb/dht.proto) and [go-libp2p-record](https://github.com/libp2p/go-libp2p-record/blob/d72bab7b487a62123614a944a0d99801b8d22cf2/pb/record.proto) for example.

I already have opened PRs for these two repos, I'll update the package name, and push to get them merged.
* https://github.com/libp2p/go-libp2p-kad-dht/pull/975
* https://github.com/libp2p/go-libp2p-record/pull/60

dereknola:
@guillaumemichel Yes, the PR I've submitted does fix our issue in K3s. I've got a branch [here](https://github.com/dereknola/k3s/commits/boxo_fork/) where the boxo dependecy is replaced with my own version (i.e. this current PR commit). Here is the record grep on that branch
```
derek@degion:~/rancher/k3s$ grep -rF 'record.proto' vendor/
vendor/go.etcd.io/etcd/server/v3/wal/walpb/record.pb.go:// source: record.proto
vendor/go.etcd.io/etcd/server/v3/wal/walpb/record.pb.go:func init() { proto.RegisterFile("record.proto", fileDescriptor_bf94fd919e302a1d) }
vendor/github.com/libp2p/go-libp2p-record/pb/record.pb.go:// source: record.proto
vendor/github.com/libp2p/go-libp2p-record/pb/record.pb.go:func init() { proto.RegisterFile("record.proto", fileDescriptor_bf94fd919e302a1d) }
vendor/github.com/libp2p/go-libp2p-record/pb/record.proto:// `protoc --gogo_out=. record.proto`
vendor/github.com/libp2p/go-libp2p-kad-dht/pb/dht.proto:import "github.com/libp2p/go-libp2p-record/pb/record.proto";
vendor/github.com/libp2p/go-libp2p/p2p/protocol/identify/pb/identify.proto:  // github.com/libp2p/go-libp2p/core/peer/pb/peer_record.proto for message definitions.
vendor/github.com/libp2p/go-libp2p/p2p/protocol/identify/pb/identify.pb.go:	// github.com/libp2p/go-libp2p/core/peer/pb/peer_record.proto for message definitions.
vendor/github.com/libp2p/go-libp2p/core/peer/pb/peer_record.pb.go:// source: core/peer/pb/peer_record.proto
vendor/github.com/ipfs/boxo/ipns/pb/ipns-record.pb.go:// source: ipns-record.proto
vendor/github.com/ipfs/boxo/ipns/record.go://go:generate protoc -I=pb --go_out=paths=source_relative:pb pb/ipns-record.proto
```

I appreciate you opening the other two PRs. Its possible that if those got merged it would also fix our issue, I haven't tried replacing those.

guillaumemichel (approved):
Thanks @dereknola for your contribution!

Looks good to me! What seems to matter is the `go_package` option, filename shouldn't create conflict, though renaming it to `ipns-record.proto` adds clarity 👍🏻 

TIL: https://protobuf.dev/programming-guides/proto3/#packages

> In Go, the `package` directive is ignored, and the generated `.pb.go` file is in the package named after the corresponding `go_proto_library` Bazel rule. For open source projects, you must provide either a `go_package` option or set the Bazel `-M` flag.

PR #792, opened 2025-01-17
Do not return unused values from wantlists

gammazero:
- Remove unnecessary setting empty dict to `nil`
- Do not return unused value from `Remove` or from `Contains`
- Do not export functions unnecessarily.

Skip changelog because there are not functional changes here.


guillaumemichel (approved):
LGTM

The change is breaking `Wantlist` compatibility (`Remove`, `Contains`), but I guess there aren't many users anyway

gammazero:
@guillaumemichel I think it is safe to assume that no one is using the `bitswap/client/wantlist` package outside of this code.

PR #793, opened 2025-01-17
reduce default number of routing in-process requests

gammazero:
Reduce the routing ProviderQueryManager default MaxInProcessRequests from 16 to 8. This prevents a situation where goroutines can be created faster than they can be completed, leading to OOM.

### DIFF
--- a/bitswap/client/client.go
+++ b/bitswap/client/client.go
@@ -109,9 +109,7 @@ func WithoutDuplicatedBlockStats() Option {
 // lookups. The bitswap default ProviderQueryManager uses these options, which
 // may be more conservative than the ProviderQueryManager defaults:
 //
-//   - WithMaxInProcessRequests(8)
-//   - WithMaxProviders(10)
-//   - WithMaxTimeout(10 *time.Second)
+//   - WithMaxProviders(defaults.BitswapClientDefaultMaxProviders)
 //
 // To use a custom ProviderQueryManager, set to false and wrap directly the
 // content router provided with the WithContentRouting() option. Only takes
@@ -195,7 +193,8 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder Pr
 
 	if bs.providerFinder != nil && bs.defaultProviderQueryManager {
 		// network can do dialing.
-		pqm, err := rpqm.New(network, bs.providerFinder, rpqm.WithMaxProviders(10))
+		pqm, err := rpqm.New(network, bs.providerFinder,
+			rpqm.WithMaxProviders(defaults.BitswapClientDefaultMaxProviders))
 		if err != nil {
 			// Should not be possible to hit this
 			panic(err)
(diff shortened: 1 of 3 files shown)


routing/providerquerymanager/providerquerymanager.go
```diff
@@ -114,7 +114,7 @@ func WithMaxTimeout(timeout time.Duration) Option {
 
 // WithMaxInProcessRequests is the maximum number of requests that can be
 // processed in parallel. If this is 0, then the number is unlimited. Default
-// is defaultMaxInProcessRequests (16).
```
hsanjuan, line 117: Perhaps we should make those defaults public, because reading the docs and finding a default is defined in a private const and you have to go read the code for that is not fantastic.

hsanjuan (approved):

routing/providerquerymanager/providerquerymanager.go
gammazero: Should these consts be exported, or should there be a different set of exported consts for setting up the default provider query manager?

bitswap/client/client.go
```diff
@@ -196,7 +196,7 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder Pr
 	if bs.providerFinder != nil && bs.defaultProviderQueryManager {
 		// network can do dialing.
 		pqm, err := rpqm.New(network, bs.providerFinder,
-			rpqm.WithMaxInProcessRequests(16),
+			rpqm.WithMaxInProcessRequests(8),
```
gammazero: @hsanjuan Do we want exported consts used here?

I think we only set these options if they differ from providerquerymanager default.

routing/providerquerymanager/providerquerymanager.go
lidel: We have prior art in https://github.com/ipfs/boxo/blob/main/bitswap/internal/defaults/defaults.go – a separate package called defaults – maybe add there, or something similar?

gammazero: consts are exported now

gammazero: @lidel Done. Values for default bitswap client come from `bitswap/internal/defaults/defaults.go`

routing/providerquerymanager/providerquerymanager.go
```diff
@@ -20,9 +20,15 @@ import (
...
-	defaultTimeout              = 10 * time.Second
+	// DefaultMaxInProcessRequests is the default maximum number of requests
+	// that are processed concurrently. A value of 0 means unlimited.
+	DefaultMaxInProcessRequests = 8
+	// DefaultMaxProviders is the default maximum number of providers that are
+	// looked up per find request. 0 value means unlimited.
+	DefaultMaxProviders = 0
+	// DefaultTimeout is the limit on the amount of time to spend waiting for
+	// the maximum number of providers from a find request.
+	DefaultTimeout = 10 * time.Second
```
gammazero, line 31: Do we still want to export these?

lidel: I believe so – as user i would expect a const with default for every thing that we also allow being tweaked via `WithFoo` funcs, if possible.

I guess if user wants to adjust this, they will provide their own manager?

gammazero: That is correct. We can add config values to configure the default pqm in a separate PR.

PR #794, opened 2025-01-17
fix: ipns protobuf namespace conflict

guillaumemichel:
## Context

To solve protobuf namespace conflicts, https://github.com/ipfs/boxo/pull/789 renamed `record.proto` to `ipns-record.proto` breaking protobuf dependency import since the filename has now changed. I think there is a better way to solve https://github.com/ipfs/boxo/issues/788.

## Proposed solution

While the `go_package` option set in the `.proto` file isn't included in the registered protobuf name, we can use a custom `Makefile` to register the full package name. This seems like a cleaner approach, since it registers the protobuf with the full package path, instead of just `ipns-record.proto`, and it doesn't break existing `record.proto` imports.

Protobuf package source [before](https://github.com/ipfs/boxo/blob/b62b60e0ecb8353d1c3e32621b687737a72a30d0/ipns/pb/ipns-record.pb.go#L5) and [after](https://github.com/ipfs/boxo/blob/70a5a10155b9a2986d33653bba0ef45ae20ae7f4/ipns/pb/record.pb.go#L5).

Also see https://github.com/libp2p/go-libp2p-record/pull/62

Would that work for you @dereknola?


dereknola (approved):
That works for K3s, I'm not seeing any conflicts when I inject your PR commit.

ipns/pb/Makefile
```diff
@@ -0,0 +1,31 @@
...
+	mkdir -p ${PKG_PATH}
+	cp ${PROTOFILE} ${PKG_PATH} 
+	protoc --go_out=. ${PKG_PATH}/${PROTOFILE}
+	mv ${PKG_PATH}/${PROTONAME}.pb.go .
+	rm -rf $(firstword $(subst /, ,${PKG_PATH}))
+
+clean:
+	rm -f *.pb.go
+
+.PHONY: all build clean
```
gammazero: I think it would be preferable to do that as a go script instead of as a makefile. That would ensure compatibility across all platforms that have a go compiler. The go script would then be called using `go generate`, with this in `ipns/record.go`:
```
//go:generate go run build_proto.go

```
Another option, instead of a go script, the commands could also be called as a series of `//go:generate` directives.

@lidel Are there other examples of similar things in the IPFS ecosystem that we want emulate for consistency?

ipns/pb/Makefile
```diff
@@ -0,0 +1,31 @@
...
+PROTOFILE = ${PROTONAME}.proto
+
+all: clean build
+
+build:
+	mkdir -p ${PKG_PATH}
+	cp ${PROTOFILE} ${PKG_PATH} 
+	protoc --go_out=. ${PKG_PATH}/${PROTOFILE}
+	mv ${PKG_PATH}/${PROTONAME}.pb.go .
+	rm -rf $(firstword $(subst /, ,${PKG_PATH}))
```
gammazero: This may be specific to `gmake` which is generally fine, but could be a problem if building on a non-gnu-based system. I think better to avoid makefile in general, see comment below.

gammazero (commented):
Might be better to avoid makefile. See comments.

gammazero (approved):
Lets merge this before the next boxo version and kubo RC2. We can decide between `Makefile` and `go:generate` in a later PR.

gammazero (approved):

gammazero:
Using go:generate also means it can be run from a top-level directory, `go generate ./...`,  without having to know where a makefile is located

PR #797, opened 2025-01-19
style: cleanup code by golangci-lint

Dreamacro:
As described in #775, `boxo` requires a better static code analysis tool since the currently used `ipdxco/unified-github-workflows` has limitation capabilities.

Therefore, I am attempting to use `golangci-lint` here (while retaining the original ipdxco/unified-github-workflows). I have disabled all `golangci-lint` default configurations and enabled some checks that I find particularly valuable.

golangci-lint config
```yaml
linters:
  disable-all: true
  enable:
    - gci
    - gofumpt
    - gosimple
    - govet
    - ineffassign
    - misspell
    - staticcheck
    - unconvert
    - unused
    - usestdlibvars

issue:
  exclude-files:
    - ".*\\.pb\\.go"

```

After performing the lint fixes, `boxo` generally has the following issues:
1. Some files have imported package names divided into groups without any discernible pattern (if there is one, please let me know), and `gci` can fix this.
for example:
https://github.com/ipfs/boxo/blob/8ca0ca28612a37df47deb72bc9d7e188e5c2b96d/ipld/unixfs/io/dagreader_test.go#L3-L15
2. many unnecessary type conversions can be detected and fixed using `unconvert`.
3. Other minor issues, such as unnecessary blank lines, octal notation, [statements that will never be executed](https://github.com/ipfs/boxo/blob/8ca0ca28612a37df47deb72bc9d7e188e5c2b96d/gateway/backend_blocks.go#L517), etc.

There might be some other valuable plugins, such as `unused`, and I would like to hear your opinions on them.


lidel (changes requested):
Triage notes 
- thank you @Dreamacro  – parking for now as we are unsure if we want to manually add this to every go repo
- @galargh thoughts if this could/should be part of Unified CI for GO instead?

hsanjuan:
We could merge the fixes without the CI changes for now?

Dreamacro:
@hsanjuan I'll revert the CI changes later.

Dreamacro:
done

galargh:
`go-check` from uci uses `govet` and `staticcheck`. I think the addition of extra linters listed here to the common workflow should be evaluated individually:
```
- gci
- gofumpt
- gosimple
- ineffassign
- misspell
- unconvert
- unused
- usestdlibvars
```

It might be also worth considering making golangci-lint workflow part of the go-check, and exposing the linter selection to the workflow users.

One thing that I think we should keep in mind during that process is the quality of the local development iteration. We definitely want some level of static checking of the code, but we also don't want the checks to negatively impact the contribution experience.

gammazero (commented):
Everything _looks_ good

lidel (approved):
I think its fine to land it here for now, given how prvalent boxo is, and then evaluate doing everywhere in  `go-check` later

PR #798, opened 2025-01-20
fix potential crash in unixfs directory

### LINKED ISSUE #675: Crash in sizeBelowThreshold
I found a bug that can cause a crash. In the directory [sizeBelowThreshold](https://github.com/ipfs/boxo/blob/main/ipld/unixfs/io/directory.go#L492) function, we call the EnumLinksAsync function, then call parallelShardWalk, and the sizeBelowThreshold function may be end if we have already fetched enough shards to assert we are above the threshold. At this time, parallelShardWalk may still running the [nextShard.walkChildren](https://github.com/ipfs/boxo/blob/main/ipld/unixfs/hamt/hamt.go#L508).
Next, because sizeBelowThreshold has been completed, we can continue to modify direcotry. At this time, walkChildren will cause a crash because the hamt ds.childer has been modified.

An example is:
Call direcotry.unlink(), and then immediately call direcotry.addChild(). If run many times, it may trigger a crash.

gammazero:
Fix potential crash in sizeBelowThreshold when modifying children in a directory.

Closes #675


gammazero:
Needs test

lidel:
Triage notes:
- ok to merge as long there is Kubo PR using this PR and the CI there is green (there are autosharding tests in Kubo iirc that should catch any unexpected regression)

PR #801, opened 2025-01-21
fix: replace fmt.Printf with fmt.Println for empty line output

0xwitty:
I noticed we're using `fmt.Printf` to print two empty lines, but since we're not utilizing any formatting, it's more straightforward to use `fmt.Println()`.

This change simplifies the code and achieves the same result, as `fmt.Println` automatically adds a newline.

### DIFF
--- a/cmd/boxo-migrate/boxomigrate.go
+++ b/cmd/boxo-migrate/boxomigrate.go
@@ -71,7 +71,7 @@ func main() {
 						return err
 					}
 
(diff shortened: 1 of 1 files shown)


guillaumemichel:
`fmt.Println()` isn't equivalent to `fmt.Printf("\n\n")`, since the former prints a single newline and the second prints 2 newlines.

`fmt.Printf("\n\n")` is also preferred to `fmt.Printfln("\n")`

PR #805, opened 2025-01-22
fix sending cancels when excluding peer

### LINKED ISSUE #694: bitswap: Don't send CANCEL to peer we got block from

gammazero:
When sending cancels for want-block and want-have requests, a peer may be excluded because it was the sender of a block and that peer will cancel the want itself without having to receive a cancel message. The would-be sender of the cancel message still needs to do everything, such as adjusting want counts, as if the cancel were being sent and then not send out the cancel message.

Closes #694

### DIFF
--- a/bitswap/client/internal/peermanager/peermanager_test.go
+++ b/bitswap/client/internal/peermanager/peermanager_test.go
@@ -281,7 +281,7 @@ func TestSendCancelsExclude(t *testing.T) {
 	// Clear messages
 	collectMessages(msgs, 2*time.Millisecond)
 
-	// Send cancels for 1 want-block and 1 want-have
+	// Send cancels for 1 want-block and 1 want-have, excluding peer1.
 	peerManager.SendCancels(ctx, []cid.Cid{cids[0], cids[2]}, peer1)
 	collected := collectMessages(msgs, 2*time.Millisecond)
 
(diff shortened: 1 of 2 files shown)


lidel (approved):
Not familiar with wantlist internals, but staging test suggests metric is fixed:

staging 01 – 0.32.1
staging 02 - kubo master / this PR

## Before this PR
We had metric go negative on box 02:

> image



## With this PR

Metric seems to be fixed:

> image

hsanjuan:
LGTM

hsanjuan:
lol it's bad anyway

PR #809, opened 2025-01-22
Revert peer exclude cancel

gammazero:
This reverts both PR #805 and #784.

Reverting the ~fix~ attempt to prevent sending cancels to a peers that sent blocks. Logic needs to be implemented that has to go all the way to modifying any already queued outgoing messages to that peer, as that message may contain other things that are not being canceled.

This issue can be addressed in a future rework of bitswap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,6 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


lidel (commented):
Thank you for catching this, It seems that reverting to version from boxo v0.26.0 is the right play for now. 

If things look on on staging, feel free to release boxo patch.
We will include this in Kubo 0.33.0-rc3.

Issue #679, opened 2024-10-01
gateway: ability to set response write timeout


lidel:
## Problem

At [Shipyard](https://ipshipyard.com/) we've run some A/B tests at [public gateway](https://docs.ipfs.tech/concepts/public-utilities/#public-ipfs-gateways) and lowering nginx timeout from 5m to 30s.
This produced better UX and also raised the number of 200s while lowering 504s.

Right now, the `boxo/gateway` library does not have any timeout, aside from [this failsafe 1h one](https://github.com/ipfs/boxo/blob/19a402b7dc34491a735696840a67a5343def4dd4/gateway/handler.go#L131), so we set timeout at .nginx is sitting in front of rainbow. 

This is extra step that most of people running gateways does not do, thus wasting resources while looking for content that is not provided correctly to certain degree (nginx default timeout is 60s, while it could be lowered).

What is really unfortunate is that [IPFS Desktop](https://docs.ipfs.tech/install/ipfs-desktop/) users hit gateway directly, and they never hit any timeout, unless it is their user agent (browser).

## Proposed feature

We should introduce feature similar to nginx's [`proxy_read_timeout`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) directly in the `boxo/gateway` library, make it configurable, but also set it to some implicit default (e.g. 30s). 

It should not depend on any internal gateway logic, but solely count the time between two successful writes from server to the client.

This way everyone using boxo will save resources, and Desktop users will get meaningful error page sooner, and we will not regress.

### Implementation ideas

Details tbd, but broad strokes idea for the `boxo/gateway` library will be to wrap existing handler in a generic response writer timeout handler:

```go
func main() {
    gwHandler := // current boxo/gateway handler
    timeoutHandler := WithResponseWriteTimeout(gwHandler, 30*time.Second) // future handler will act like this
    http.ListenAndServe(":8080", timeoutHandler)
}
```

- The `WithResponseWriteTimeout` middleware creates a timeoutResponseWriter and starts a timer.
  -   "timeoutResponseWriter"  wraps the original `ResponseWriter` and tracks the last successful write.
  - Every time data is written successfully, the timer is reset.
  - If no data is written for the specified duration, the timer expires, and a 504 Gateway Timeout status is sent to the client.

[...]

PR #812, opened 2025-01-23
feat(gateway): add configurable response write timeout

gitsrc:
This commit introduces a configurable response write timeout for the IPFS gateway. 
The timeout can be set via the `ResponseWriteTimeout` field in the gateway configuration. 
If not set, a default timeout of 30 seconds is applied.

The implementation includes:
- A new `timeoutResponseWriter` struct that wraps the standard `http.ResponseWriter` 
  and enforces the timeout.
- A middleware function `WithResponseWriteTimeout` that applies the timeout logic 
  to the HTTP handler chain.
- Comprehensive unit tests to verify the timeout behavior under various scenarios.

The timeout ensures that slow or unresponsive clients do not indefinitely hold 
server resources, improving the overall reliability and stability of the gateway.

This change also attempts to address the issue described in https://github.com/ipfs/boxo/issues/679 
by providing a mechanism to handle slow or stuck HTTP responses more gracefully.


gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+
+	if !w.done {
+		w.done = true
+		w.cancel() // Propagate context cancellation
+		w.ResponseWriter.WriteHeader(http.StatusGatewayTimeout)
+	}
+}
+
+// WithResponseWriteTimeout creates middleware for response write timeout handling
+func WithResponseWriteTimeout(next http.Handler, timeout time.Duration) http.Handler {
```
gammazero, line 122: This does not appear to need to be exported.
```suggestion
// withResponseWriteTimeout creates middleware for response write timeout handling
func withResponseWriteTimeout(next http.Handler, timeout time.Duration) http.Handler {
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
 //
 // [IPFS HTTP Gateway]: https://specs.ipfs.tech/http-gateways/
 func NewHandler(c Config, backend IPFSBackend) http.Handler {
-	return newHandlerWithMetrics(&c, backend)
+	handler := newHandlerWithMetrics(&c, backend)
+
+	// Use the configured timeout or fall back to a default value
+	timeout := c.ResponseWriteTimeout
+	if timeout == 0 {
+		timeout = 30 * time.Second // Default timeout of 30 seconds
```
gammazero, line 77: Define a `const` for the default value.
```suggestion
		timeout = defaultResponseTimeout
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+	}
+
+	// Apply the timeout middleware
+	return WithResponseWriteTimeout(handler, timeout)
```
gammazero, line 81:
```suggestion
	return withResponseWriteTimeout(handler, timeout)
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+}
+
+// timeoutResponseWriter implements http.ResponseWriter with timeout control
+type timeoutResponseWriter struct {
+	http.ResponseWriter                    // Embedded standard response writer
+	timer               *time.Timer        // Timeout tracking mechanism
+	timeout             time.Duration      // Configured timeout duration
+	mu                  sync.Mutex         // Mutex for concurrent access protection
+	done                bool               // Completion state flag
+	requestCtx          context.Context    // Original request context
```
gammazero, line 91: This can be removed as it is never referenced in the struct.
```suggestion
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+
+// cancelRequest handles timeout termination sequence
+func (w *timeoutResponseWriter) cancelRequest() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if !w.done {
+		w.done = true
+		w.cancel() // Propagate context cancellation
+		w.ResponseWriter.WriteHeader(http.StatusGatewayTimeout)
```
gammazero, line 117: If this were a timeout waiting to read a response then returning `http.StatusGatewayTimeout` is appropriate. Since this is a write timeout, I think `http.StatusServiceUnavailable` is preferable.
```suggestion
		w.ResponseWriter.WriteHeader(http.StatusServiceUnavailable)
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+	return http.HandlerFunc(func(origWriter http.ResponseWriter, r *http.Request) {
+		// Create derived context with cancellation capability
+		ctx, cancel := context.WithCancel(r.Context())
+
+		// Initialize enhanced response writer
+		tw := &timeoutResponseWriter{
+			ResponseWriter: origWriter,
+			timeout:        timeout,
+			timer:          time.NewTimer(timeout),
+			requestCtx:     ctx,
```
gammazero, line 132: not used here
```suggestion
```

gateway/handler.go
```diff
@@ -68,7 +69,83 @@ type handler struct {
...
+		w.ResponseWriter.WriteHeader(http.StatusGatewayTimeout)
+	}
+}
+
+// WithResponseWriteTimeout creates middleware for response write timeout handling
+func WithResponseWriteTimeout(next http.Handler, timeout time.Duration) http.Handler {
+	return http.HandlerFunc(func(origWriter http.ResponseWriter, r *http.Request) {
+		// Create derived context with cancellation capability
+		ctx, cancel := context.WithCancel(r.Context())
+
```
gammazero, line 126: Terminate goroutine when handler returns
```suggestion
defer cancel()
```

gammazero (commented):
This addresses what was asked for in #679 but I think that if can live without the ask, "Every time data is written successfully, the timer is reset" then things can be much simpler. For example: https://github.com/ipfs/boxo/pull/818

gitsrc:
OK, I understand, it would be cleaner to set a uniform timeout.

> This addresses what was asked for in #679 but I think that if can live without the ask, "Every time data is written successfully, the timer is reset" then things can be much simpler. For example: #818

lidel:
Triage note:
- likely we want to focus on "timeout between byte flushes to client" rather than generic timeout (https://github.com/ipfs/boxo/pull/818)
- we will look into this before 0.34

gateway/gateway.go
```diff
@@ -48,6 +48,11 @@ type Config struct {
 	// directory listings, DAG previews and errors. These will be displayed to the
 	// right of "About IPFS" and "Install IPFS".
 	Menu []assets.MenuItem
+
+	// ResponseWriteTimeout is the maximum duration the gateway will wait for a response
+	// to be written before timing out and returning a 504 Gateway Timeout error.
+	// If not set, a default timeout will be used.
```
lidel, line 54:
```suggestion
	// ResponseWriteTimeout is the maximum duration the gateway will wait for
	// new bytes to be retrieved from [gateway.IPFSBackend]. This timeout is
	// reset on every [http.ResponseWriter] Write, which is protecting both
	// client and server from wasting resources when parts of requested file or
	// DAG have no providers and would hang forever.
	// Setting to 0 disables this timeout.
```

gateway/handler_test.go
```diff
@@ -28,3 +32,90 @@ func TestEtagMatch(t *testing.T) {
...
+						w.Write([]byte("chunk\n")) // Resets timer on each write
+					case <-r.Context().Done():
+						return
+					}
+				}
+			},
+			timeout:        300 * time.Millisecond,
+			expectStatus:   http.StatusOK,
+			expectedChunks: 3,
+		},
```
lidel, line 87: Missing test like this, but that writes two chunks and timeouts on third.

We need to see what happens when Headers were already sent with status 200, but payload truncated due to timeout.

Perhaps it would be useful to `w.Write([]byte("\n[TRUNCATED DUE TO gateway.IPFSBackend TIMEOUT]"))` to make it easier for clients to understand what happened if they debug things?

gateway/handler.go
lidel, line 81: Hm.. this makes disabling timeout impossible. 

Let's change this to no timeout by default (Kubo and Rainbow will set the default themselves), and skip entire things if no timeout.

```suggestion
	if c.ResponseWriteTimeout != 0 {
		handler = WithResponseWriteTimeout(handler, timeout)
	}

	// Apply the timeout middleware
	return handler
```

gateway/handler.go
lidel: +1, why we need requestCtx @gitsrc, does not seem to be used?

gateway/handler_test.go
lidel, line 1: nit: run gofmt on this file

lidel (changes requested):
I'd like to run this on kubo staging before merging. Pushing this to Kubo 0.35.

, in the meantime, needs additional  test and some cleanup – details inline.

gateway/handler.go
gitsrc: You are right, this variable is not used and can be deleted.

lidel:
Triage notes: moving to ~0.36~ 0.37

lidel:
I will be looking at this as part of v0.37, likely cherry-pick useful commits from this PR and https://github.com/ipfs/boxo/pull/887 and introduce both settings in a single PR to avoid duplicated orchestration.

As for this PR, quick notes for self (things to check):
- Race condition: Timer reset in Write() method creates concurrency hazards?
- Double response risk: 504 status may conflict with handler's already-written response
- Goroutine leak: Missing(?) cleanup mechanism for timeout monitor goroutine

lidel:
Let's go with https://github.com/ipfs/boxo/pull/994 which consolidates both types of limits + adds metrics and more tests.

PR #816, opened 2025-01-27
go1.23

hsanjuan:


hsanjuan:
ok, so running with Go latest version - 1 is a feature...

aschmahmann:
> ok, so running with Go latest version - 1 is a feature...

Yeah, we try and maintain compatibility with the current + previous versions of Go so that people don't have to be on the latest, but we also don't have to support ancient versions of the language.

PR #817, opened 2025-01-27
fix: bitswap lock contention under high load

gammazero:
### Summary

Fix runaway goroutine creation under high load. Under high load conditions, goroutines are created faster than they can complete and the more goroutines creates the slower them complete. This creates a positive feedback cycle that ends in OOM. The fix dynamically adjusts message send scheduling to avoid the runaway condition.

### Description of Lock Contention under High Load

The peermanager acquires the peermanager mutex, does peermanager stuff, and then acquires the messagequeue mutex for each peer to put wants/cancels on that peer's message queue. Nothing is blocked indefinitely, but all session goroutines wait on the peermanager mutex.

The messagequeue event loop for each peer is always running in a separate goroutine, waking up every time new data is added to the message queue. The messagequeue acquires the messagequeue mutex to check the amount of pending work and send a message if there is enough work.

The frequent lock/unlock of each messagequeue mutex delays each session goroutine from adding items to messagequeues, as they wait to acquire each peer's messagequeue mutex to enqueue a message. These delays cause the peermanager mutex to be held longer by each goroutine. When there are a sufficient number of peers and want requests, goroutines end up waiting on the peermanager mutex for a longer time, on average, that it takes for an additional request to arrive and start another goroutine. This leads to a positive feedback loop where the number of goroutines increases until their number alone is sufficient to cause OOM.

### How this PR Fixes this

This PR avoids waking up the messagequeue event loop on every item added to the message queue, thus avoiding the high-frequency messagequeue mutex lock/unlock. Instead, the event loop wakes up after a delay, sends the accumulated work, then goes back to sleep for another delay. During the delay, wants and cancels are accumulated. This allows the session goroutines to add items to message queues without contending with the messagequeue event loop for the messagequeue mtuex.

The delay dynamically adjusts, between 20ms and 1 second, based on the number peers. The delay per peer is configurable, with a default of 1/8 millisecond (125us).


bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -565,38 +542,48 @@ func (mq *MessageQueue) sendMessage() {
...
+		// send more if the the workcount is above the cutoff. Otherwise,
+		// schedule sending the rest of the wants in the next iteration of the
+		// event loop.
+		pendingWork := mq.pendingWorkCount()
+		if pendingWork < sendMessageCutoff {
+			if pendingWork > 0 {
+				mq.signalWorkReady()
+			}
+			return
+		}
```
lidel, line 610: @gammazero it feels we changed the role of  `sendMessageCutoff` a bit here?


Previously,  `signalWorkReady()`  was guarded with `pendingWork > sendMessageCutoff`, now its called only if `pendingWork < sendMessageCutoff && pendingWork > 0`


Is it possible to have a spike in workCount in-between calls of `sendMessage()` that workCount it goes beyond `sendMessageCutoff` (256) and we forget to call `signalWorkReady()` here?

bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -42,12 +42,9 @@ const (
...
 	// when we reach sendMessageCutoff wants/cancels, we'll send the message immediately.
 	sendMessageCutoff = 256
-	// sendMessageDebounce is the debounce duration when calling sendMessage()
-	sendMessageDebounce = time.Millisecond
-	// when we debounce for more than sendMessageMaxDelay, we'll send the
-	// message immediately.
-	sendMessageMaxDelay = 20 * time.Millisecond
-	sendTimeout         = 30 * time.Second
+	// wait this long before sending next message
+	sendMessageDelay = time.Second
```
lidel: @gammazero is my understanding correct this PR in current form effectively raises delay in sending bitswap wants from <20ms to 1 second? 

This feels pretty hardcore change for websites, where a page can have multiple assets, and each may incur this penalty.

Did quick anecdotal test locally and even simple pages like wikipedia seem to incur increased load times on empty repo:

- http://en-wikipedia--on--ipfs-org.ipns.localhost:8080/wiki/Peace_of_Westphalia
  - kubo master → ~4s
    > image
  - kubo master + this PR → >8s (twice as long)
    > image
- http://en-wikipedia--on--ipfs-org.ipns.localhost:8080/wiki/Peace_treaty
  - kubo master → ~3s
    > image
  - kubo master + this pr  → ~9s (three times as long)
    > image

It feels like a dramatic change that may impact other latency-sensitive users (e.g. chains that care about sync time). 

Will sleep on this and we can sync tomorrow, but I wonder, would it be possible to bring it back to 1-20ms, and then have some sort of exponential backoff up to 1s once we hit high load?

lidel (commented):
For posterity, staging box looks really promising, the window of time when this was deployed to box 02 was significantly in better shape than kubo 0.32.1 (01 box):
> image
> image

HTTP success rate is higher too:

> image

EOD for me, but I'll do more test tomorrow morning and see if any questions arise. Some quick ones inline.

bitswap/client/internal/messagequeue/messagequeue.go
gammazero: Yes, you understand correctly. This is experimental and was originally to show/confirm the cause of the problem, and a 1s delay is much too course for a releasable fix.

It does demonstrate that a fix to the problem must reduce lock contention at sufficient load to cause the problem. Reducing lock contention is done here by a very course manner, by introducing a big fixed delay. A delay may not be the only or best way to accomplish this.

### Other Approaches

#### Use a dedicated queue
Instead of having each session goroutine acquire the messagequeue mutex to update the lists of CIDs in a peer's message queue, write an update to an actual queue - a channel read by a dedicated goroutine. This will effectively separate enqueuing new messages from the peer's messagequeue being busy.

#### Dynamic Delay
Adjust the send message delay based on sampled metrics (number of peers, wantlist size, etc.)

gammazero: ### Experiments with queue
After experimentation with a separate queue, this works better but does not completely fix the problem. Even if there is a dedicated goroutine reading from the channel that updates are written to, writing these into a queue, and writing these to another channel. Copies of CID lists (wants and cancels) must be made to deliver to the asynchronous queue, instead of updating the messagequeue datastructures directly.. Under high load this copying, and the need to run the queue handling goroutine at high frequency, can cause writers to back up waiting to write to the queue.

### Adjusting timing
Reducing the sendMessage delay to 200ms seems to work just as well as having a 1 second delay at the loads tested. With lighter workloads the delay could be reduced farther.

A way to dynamically adjust this would be to increase the delay proportional to the pending updates coming into the message queue.

bitswap/client/internal/messagequeue/messagequeue.go
gammazero: Nothing is missed. It always calls `signalWorkReady()` when exiting the loop/function. If `pendingWork` is larger than the cutoff, then the loop iterates again and does another send immediately.

This allows sending large messages in back-to-back chunks without having to wait for the next send, and prevents getting behind when there are many updates.

bitswap/client/client.go
```diff
@@ -178,7 +187,9 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder ro
 		}
 	}
 	peerQueueFactory := func(ctx context.Context, p peer.ID) bspm.PeerQueue {
-		return bsmq.New(ctx, p, network, onDontHaveTimeout, bsmq.WithDontHaveTimeoutConfig(bs.dontHaveTimeoutConfig))
+		return bsmq.New(ctx, p, network, onDontHaveTimeout,
+			bsmq.WithDontHaveTimeoutConfig(bs.dontHaveTimeoutConfig),
+			bsmq.WithPerPeerSendDelay(bs.perPeerWantSendDelay))
```
lidel: nit: bit odd that we end up  setting more broad "PerPeerSendDelay" to kinda narrowly-named "PerPeer**Want**SendDelay". Someone may think it only applies to Wants, but it will also aply to other things, no?

Could we rename `[With]PeerWantSendDelay` to  `[With]PerPeerSendDelay` so its named the same everywhere and there is no confusion?

lidel (approved):
Lgtm, this is such improvement we should ship this as a patch release next week.


For posterity:  based on our ([Shipyard](https://ipshipyard.com/)) staging tests the impact on high load providers is significant.

Below is a sample from HTTP gateway  processing ~80 requests per second (mirrored organic cache-miss from ipfs.io). "01" is latest Kubo (0.33.0) without this fix, and "02" is with this fix (0.33.1):
> kubo-with-boxo-goroutine-fix
> kubo-rsp-rate-bs-fix

bitswap/client/client.go
gammazero: Good point. Done.

PR #818, opened 2025-01-28
gateway: add configurable response write timeout

### LINKED ISSUE #679: gateway: ability to set response write timeout

gammazero:
Closes #679
Replacement for #812

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `gateway`: Configurable write timeout [#818](https://github.com/ipfs/boxo/pull/818)
+
 ### Changed
 
 ### Removed
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -48,6 +48,11 @@ type Config struct {
 	// directory listings, DAG previews and errors. These will be displayed to the
 	// right of "About IPFS" and "Install IPFS".
 	Menu []assets.MenuItem
+
+	// ResponseWriteTimeout is the maximum duration the gateway will wait for a
+	// response to be written before returning a 503 Service Unavailable error.
+	// An unset or zero value uses the default timeout. -1 means no timeout.
+	ResponseWriteTimeout time.Duration
 }
 
 // PublicGateway is the specification of an IPFS Public Gateway.
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -32,6 +32,8 @@ import (
 var log = logging.Logger("boxo/gateway")
 
 const (
+	defaultResponseTimeout = 30 * time.Second
+
 	ipfsPathPrefix        = "/ipfs/"
 	ipnsPathPrefix        = ipns.NamespacePrefix
 	immutableCacheControl = "public, max-age=29030400, immutable"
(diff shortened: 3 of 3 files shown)


hsanjuan:
This doesn't act as nginx's `proxy_read_timeout` though?... i.e. a roundtrip should not be killed just because it took 30 seconds... the problem is 30 seconds without writing any data on the response body.

Instead of #812, which goes in that direction, perhaps we can get away with it by setting `ResponseHeaderTimeout` in the client configuration. Normally if a request lasts more than 30 seconds without any data coming back, the reason is that we are still looking for the content, and I doubt we have written any response headers by then.

lidel:
Yeah, we want `proxy_read_timeout`-like behavior. 
Sadly we have no means to control HTTP clients

Let's go back to https://github.com/ipfs/boxo/pull/812  and see if we could simplify it.

PR #819, opened 2025-01-29
bitswap/client: fewer wantlist iterations in sendCancels

gammazero:
Combine loops to remove redundant wantlist iteration.


guillaumemichel (approved):
That was much needed, thanks @gammazero!

Issue #752, opened 2024-12-12
bitswap: long-running session leaks


Wondertan:
We've been running long-running Bitswap sessions and encountered a series of memory leaks that we've successfully identified and fixed in our fork. I plan to start upstreaming them eventually, but I am currently botted with millions of things. Anyway, I want to make you aware of their existence already in case you have the capacity to look into those.

The full branch is located [here](https://github.com/ipfs/boxo/compare/main...celestiaorg:boxo:fixes). Some of them were merged, some of them aren't. The most notable are:

* [fix leak ensuring SessionInterestManager is cleaned up](https://github.com/ipfs/boxo/commit/6cd1ed917fece931595fe65ffc9450a0737d4c06)
* [fix leak in block presence manager](https://github.com/ipfs/boxo/commit/7adce5f5ee7afb69684a817bf6578d0ee46762d5) along [with](https://github.com/ipfs/boxo/commit/70a650316c3b9e4734df0bcb79ea8e530d86aae8) 
* [cleanup toFetch cidqueue](https://github.com/ipfs/boxo/commit/b0a7e7e3f8b60ac2baca056be7429bcb6a0f1fd6)
* [cleanup peerresponsetracker](https://github.com/ipfs/boxo/commit/ea61bc62e8f90e975fc8cf37950edae2d418e7bd)
* [rebroadcast only when there are live CIDs](https://github.com/ipfs/boxo/commit/2690bfaf80ae011dbe46d6b07955c731ddf5e1cd) - not a leak, resolves issue with timer continue ticking even when there are not CIDs to rebroadcast further

All of those fixes have been tested on hundreds of nodes in production, and things have been running smoothly and steadily for 3 weeks. I am happy to answer if you start working on this sooner than when I get to opening PRs myself.

cc @hsanjuan, @gammazero, @lidel

gammazero:
I will create separate PR for each of these and review and merge these into our code base, unless you would rather do this from your fork. We are also resolving memory consumption issues with connections, so it would be great to get all the known memory issues handled in at the same time. So let me know soon.

Thank you.

gammazero:
The first two fixes appear to be incorrect according to the [comment](https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go#L25-L27) in `SessionInterestManager`:
```
		// Note that once the block is received the session no longer wants
		// the block, but still wants to receive messages from peers who have
[...]

PR #821, opened 2025-01-29
fix(bitswap): simplify SessionInterestManager

guillaumemichel:
Follow up to https://github.com/ipfs/boxo/issues/752

After investigating around https://github.com/ipfs/boxo/commit/6cd1ed917fece931595fe65ffc9450a0737d4c06, I found that `SessionInterestManager` keeping track of `Interested` session, doesn't have a single use case, as @Wondertan suggested.

`SessionInterestManager` is recording a session's `wants` (blocks that we want, and that we didn't get yet), as well as `interests` (block that we want, and we have already). `SessionsInterestManager` isn't doing anything about the `interests`, hence I removed `interests` tracking altogether.

cc: @gammazero


bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -153,7 +130,7 @@ func (sim *SessionInterestManager) SplitWantedUnwanted(blks []blocks.Block) ([]b
 		// For each session that is interested in the key
 		for ses := range sim.wants[c] {
 			// If the session wants the key (rather than just being interested)
```
gammazero:
```suggestion
```

bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -153,7 +130,7 @@ func (sim *SessionInterestManager) SplitWantedUnwanted(blks []blocks.Block) ([]b
 		// For each session that is interested in the key
```
gammazero:
```suggestion
		// For each session that wants the key.
```

gammazero (approved):
Thanky for following up in this. Couple of comments to fix, now that there is no distinction between wants and interests.

PR #822, opened 2025-01-29
fix(bitswap): filter interests from received messages

guillaumemichel:
Fixresource leak by adding a CID to BlockResponseManager only if there a session interested in it.

Follow up to https://github.com/ipfs/boxo/issues/752
Based on https://github.com/ipfs/boxo/pull/821

@Wondertan does that address the leak that you are referring to in https://github.com/ipfs/boxo/commit/70a650316c3b9e4734df0bcb79ea8e530d86aae8?

cc: @gammazero

### DIFF
--- a/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
+++ b/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
@@ -181,11 +181,11 @@ func (sim *SessionInterestManager) FilterInterests(keySets ...[]cid.Cid) [][]cid
 	sim.lk.RLock()
 	defer sim.lk.RUnlock()
 
-	result := make([][]cid.Cid, len(keySets))
+	result := keySets[:0]
 	// For each set of keys
 	for i, ks := range keySets {
 		// The set of keys that at least one session is interested in
-		wanted := make([]cid.Cid, 0, len(ks))
+		wanted := ks[:0]
 
 		// For each key in the set
 		for _, c := range ks {
@@ -194,7 +194,7 @@ func (sim *SessionInterestManager) FilterInterests(keySets ...[]cid.Cid) [][]cid
 				wanted = append(wanted, c)
 			}
 		}
-		result[i] = wanted
+		result = appent(result, wanted)
 	}
 	return result
 }
--- a/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
+++ b/bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
@@ -183,18 +183,18 @@ func (sim *SessionInterestManager) FilterInterests(keySets ...[]cid.Cid) [][]cid
 
 	result := keySets[:0]
 	// For each set of keys
-	for i, ks := range keySets {
-		// The set of keys that at least one session is interested in
+	for _, ks := range keySets {
+		// The set of keys wanted by at least one session
 		wanted := ks[:0]
 
 		// For each key in the set
 		for _, c := range ks {
-			// If there are any sessions interested in this key
+			// If any session wants the key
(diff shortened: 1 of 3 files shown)


bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -177,3 +177,26 @@ func (sim *SessionInterestManager) InterestedSessions(keySets ...[]cid.Cid) []ui
 	}
 	return ses
 }
+
+// Filters only the keys that are wanted by at least one session
+func (sim *SessionInterestManager) FilterInterests(keySets ...[]cid.Cid) [][]cid.Cid {
+	sim.lk.RLock()
+	defer sim.lk.RUnlock()
+
+	result := make([][]cid.Cid, len(keySets))
```
gammazero: Instead of creating a new slice, modify in place.
```suggestion
	result := keySets[:0]
```

bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -177,3 +177,26 @@ func (sim *SessionInterestManager) InterestedSessions(keySets ...[]cid.Cid) []ui
...
+	// For each set of keys
+	for i, ks := range keySets {
+		// The set of keys that at least one session is interested in
+		wanted := make([]cid.Cid, 0, len(ks))
```
gammazero: Instead of creating new slices, modified in place.
```suggestion
		wanted := ks[:0]
```

gammazero (commented):
See suggested changes to avoid allocation.

bitswap/client/internal/sessioninterestmanager/sessioninterestmanager.go
```diff
@@ -177,3 +177,26 @@ func (sim *SessionInterestManager) InterestedSessions(keySets ...[]cid.Cid) []ui
...
+
+		// For each key in the set
+		for _, c := range ks {
+			// If there are any sessions interested in this key
+			if _, ok := sim.wants[c]; ok {
+				wanted = append(wanted, c)
+			}
+		}
+		result[i] = wanted
```
gammazero: If modifying in place:
```suggestion
		result = appent(result, wanted)
```

gammazero (commented):
Additionl change

gammazero (approved):

bitswap/client/internal/sessionmanager/sessionmanager.go
```diff
@@ -153,6 +153,11 @@ func (sm *SessionManager) GetNextSessionID() uint64 {
 
 // ReceiveFrom is called when a new message is received
 func (sm *SessionManager) ReceiveFrom(ctx context.Context, p peer.ID, blks []cid.Cid, haves []cid.Cid, dontHaves []cid.Cid) {
+	// Keep only the keys that at least one session wants
+	keys := sm.sessionInterestManager.FilterInterests(blks, haves, dontHaves)
+	blks = keys[0]
+	haves = keys[1]
+	dontHaves = keys[2]
```
gammazero, line 160: Is this already being done redundantly in the call to `sess.ReceiveFrom` below (line 175)?

See: https://github.com/ipfs/boxo/blob/main/bitswap/client/internal/session/session.go#L189

Wondertan:
@guillaumemichel, I think it does and should also fix the DOS vector. 
In my solution, I combined filtering with `InterestedSessions` to loop less over map in `sim`, but that's up to you

P.S. This was really an annoying one to figure out.

Wondertan:
After this PR, https://github.com/ipfs/boxo/commit/7adce5f5ee7afb69684a817bf6578d0ee46762d5 should be the last leak to fix

PR #829, opened 2025-02-03
Always send cancels even if peer has no interest

gammazero:
Possible fix for memory leak in wantlists.

It appears necessary to send cancels even if the receiving peer has no interest in the CIDs according to the session interest manager. The handling of cancels cleans up the CIDs that remain on the peerwantmanager wantlists.

### DIFF
--- a/bitswap/client/internal/sessionmanager/sessionmanager.go
+++ b/bitswap/client/internal/sessionmanager/sessionmanager.go
@@ -157,7 +157,9 @@ func (sm *SessionManager) GetNextSessionID() uint64 {
 // their contents. If the caller needs to preserve a copy of the lists it
 // should make a copy before calling ReceiveFrom.
 func (sm *SessionManager) ReceiveFrom(ctx context.Context, p peer.ID, blks []cid.Cid, haves []cid.Cid, dontHaves []cid.Cid) {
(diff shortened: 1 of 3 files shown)


bitswap/client/internal/sessionmanager/sessionmanager.go
```diff
@@ -157,6 +157,9 @@ func (sm *SessionManager) GetNextSessionID() uint64 {
 // their contents. If the caller needs to preserve a copy of the lists it
 // should make a copy before calling ReceiveFrom.
 func (sm *SessionManager) ReceiveFrom(ctx context.Context, p peer.ID, blks []cid.Cid, haves []cid.Cid, dontHaves []cid.Cid) {
+	// Send CANCEL to all peers with want-have / want-block
+	sm.peerManager.SendCancels(ctx, blks)
```
hsanjuan: Perhaps explain why this sits here now, as I can imagine a developer undoing this in a couple of years when trying to optimize.

hsanjuan (approved):
Minor nitpick

bitswap/client/internal/sessionmanager/sessionmanager.go
gammazero: Done

PR #830, opened 2025-02-03
fix(ipns): reading records with raw []byte Value

lidel:
This PR aims to improve / maximize interop with legacy/alternative IPNS clients by restoring support for `[]byte` CID in `Value` field and gracefully handling empty `Value` in scenarios where IPNS is used for its `Data` field, and not IPFS paths.

- [x] `record.Value()` will convert a binary CID to a valid `path.Path`. 
- [x] Empty `Value()` will produce `NoopPath` (`/ipfs/bafkqaaa`) to avoid breaking existing code that expects a valid record to always produce a valid content path.
- [x] tests to ensure we test both + error if `Value()` can't produce a meaningful path

Thank you @ianopolous and @aschmahmann for catching this.

If / once this lands, I'll update phrasing in https://specs.ipfs.tech/ipns/ipns-record/ to make it clear binary CID and empty Value are both allowed.

Aiming to include this in Kubo 0.34.

### DIFF
--- a/ipns/record.go
+++ b/ipns/record.go
@@ -12,6 +12,7 @@ import (
 	ipns_pb "github.com/ipfs/boxo/ipns/pb"
 	"github.com/ipfs/boxo/path"
 	"github.com/ipfs/boxo/util"
+	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/dagcbor"
@@ -30,6 +31,9 @@ type ValidityType int64
 // ValidityEOL means "this record is valid until {Validity}". This is currently
 // the only supported Validity type.
 const ValidityEOL ValidityType = 0
+
+// NoopValue is an identity CID that points at zero bytes.
+const NoopValue = "/ipfs/bafkqaaa"
 
 // Record represents an [IPNS Record].
 //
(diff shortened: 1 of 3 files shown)


ipns/record.go
```diff
@@ -89,10 +93,29 @@ func (rec *Record) Value() (path.Path, error) {
...
-	p, err := path.NewPath(string(value))
+	if len(value) == 0 {
+		// To maximize interop across implementations, turn empty value into zero-length identity CID.
+		// This is a convenience placeholder used when the value in record is empty or does
+		// not matter because IPNS is used for custom CBOR in the Data field.
+		value = []byte(NoopValue)
+	}
+
+	// parse as a string with content path
+	if len(value) > 0 && value[0] == '/' {
```
gammazero: The test of `len(value) > 0` is unnecessary because of the `if len(value) == 0` above, and `len` will not return a negative value.

ipns/record.go
```diff
@@ -89,10 +93,29 @@ func (rec *Record) Value() (path.Path, error) {
...
+		p, err := path.NewPath(string(value))
+		if err != nil {
+			return nil, multierr.Combine(ErrInvalidPath, err)
```
gammazero, line 109: Having both `ipns.ErrInvalidPath` and [`path.ErrInvalidPath`](https://github.com/ipfs/boxo/blob/main/path/error.go#L14) is redundant. Maybe just use `path.ErrInvalidPath` and sinply return `err` here?

ipns/record.go
```diff
@@ -89,10 +93,29 @@ func (rec *Record) Value() (path.Path, error) {
...
+		}
+		// done, finish fast
+		return p, nil
+	}
+
+	// fallback: for legacy/optimization reasons, the value could be a valid CID in byte form
+	binaryCid, err := cid.Cast(value)
 	if err != nil {
 		return nil, multierr.Combine(ErrInvalidPath, err)
```
gammazero, line 118: Maybe use `path.ErrInvalidPath`?

lidel: this is on purpose, to have IPNS type that works with `errors.Is` and `errors.As` but also human-readable information that error occurred in CID fallback (this user-friendly error will show up in `ipfs` CLI output etc)

debugging IPNS is hard, this makes it bit easier for end user

ipns/record.go
lidel: I'd keep it for consistency with https://github.com/ipfs/boxo/pull/830#discussion_r1941342309 and the rest of record.go

gammazero (approved):

PR #835, opened 2025-02-06
bitswap/client/internal/messagequeue: run tests in parallel

Jorropo:
On my computer, make the tests take 12s rather than 65s. Skimming through them I didn't found any dependency or shared resources in them.


gammazero (approved):
Nice, thank you!

PR #838, opened 2025-02-07
Move from gogo protobuf

gammazero:


ipld/unixfs/pb/unixfs.pb.go
```diff
@@ -1,28 +1,25 @@
-// Code generated by protoc-gen-gogo. DO NOT EDIT.
-// source: unixfs.proto
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.36.3
+// 	protoc        v5.29.3
+// source: github.com/ipfs/boxo/ipld/unixfs/pb/unixfs.proto
 
-package unixfs_pb
+package pb
```
aschmahmann, line 7: @gammazero if possible I would:
1. Separate switching everything but UnixFS from the UnixFS switch
2. If / when switching UnixFS be very careful in checking that the generated protobufs should be byte-wise identical (includes validating in kubo, etc. that data imports result in the same CIDs, etc.)

gammazero: Done in #839.  Will add UnixFS and `ipld.merkledag` in separate PRs.

gammazero:
Replaced by #839

PR #845, opened 2025-02-14
fix: gateway/blocks-backend: GetBlock should not perform IPLD decoding

### LINKED ISSUE #673: gateway/blocks-backend:  GetBlock should not perform IPLD decoding
The gateway BlocksBackend's `GetBlock` function under the hood uses the dagservice's Get function which means it attempts to do an IPLD decoding of the block. This is not good because it means the gateway needs to understand the blocks codec in order to return it even when it doesn't really need to.

https://github.com/ipfs/boxo/blob/171b0b72118578e2f42f7e5aa45b52ad33a04231/gateway/backend_blocks.go#L213

https://github.com/ipfs/boxo/blob/171b0b72118578e2f42f7e5aa45b52ad33a04231/gateway/backend_blocks.go#L553

As an example: rainbow does not support dag-jose and so sending a `format=raw` request to a rainbow gateway will fail even though it could pass.

A fix here is to use the blockservice directly instead of the dagservice for the GetBlock function's retrieval purposes after doing any path traversals required to get to the final block.

gsergey418:
Fixes #673.


gammazero (approved):
Tested in kubo CI also. Looks good.

@gsergey418alt Thank you for this.

PR #851, opened 2025-02-20
feat(examples): add UnixFS builder example with HAMT sharding

### LINKED ISSUE #663: example: turn directory into UnixFS DAG
## Problem

We have many unixfs implementations, and it is extremely confusing for new developers to understand what they should use and why. (cc @2color)

## Proposed improvement

The [./exmaples](https://github.com/ipfs/boxo/tree/main/examples) should have an example of `ipfs add + ipfs dag export` equivalent for reading a file or a directory from local filesystem, and turning it into a CAR with UnixFS DAG that was properly chunked / HAMT-sharded. 

The example should be a CLI tool that accepts same flags as `ipfs add`, demonstrating how to adjust chunker, hash function, or decide if mode/mtime should be preserved.

PsychoPunkSage:
Closes #663

This PR adds a new example demonstrating how to programmatically create UnixFS DAGs with features similar to `ipfs add`. The example shows how to:

- Create UnixFS structures from files and directories
- Handle large directories using HAMT sharding
- Preserve file metadata (mtime, mode)
- Customize chunking strategies
- Display progress with human-readable sizes

The example includes proper error handling, tests, and documentation for common use cases. This helps developers understand how to work with UnixFS programmatically using boxo's lower-level APIs.

### DIFF
--- a/examples/unixfs-builder/README.md
+++ b/examples/unixfs-builder/README.md
@@ -1,0 +1,35 @@
+# UnixFS Builder Example
+
+This example demonstrates how to programmatically create UnixFS data structures that are compatible with IPFS. It shows how to:
+
+- Add single files to IPFS with custom chunking and hashing
+- Create directories with multiple files
+- Handle large directories using HAMT sharding
+- Preserve file metadata (mode, mtime)
+- Export the resulting data as a CAR file
+
+## Usage
+
+```bash
+# Add a single file
+go run . add myfile.txt
+
+# Add a directory
+go run . add -r mydirectory/
+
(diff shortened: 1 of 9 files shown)


PsychoPunkSage:
Hi @lidel
Can you have a quick look here as soon as you find some time?

gammazero:
@PsychoPunkSage Thank you for your interest in this. However, we generally recommend doing this using MFS and would prefer a published example to show that. If other techniques are necessary, then it indicates a need for revised APIs instead of extensive examples.

PR #855, opened 2025-02-22
feat(gateway): introduce typed errors for blocked content

### LINKED ISSUE #591: gateway: dedicated error type for denylists
Right now we don't use proper error types to convey that some content requested by a gateway was blocked. Instead we have this string based checking tied to nopfs

https://github.com/ipfs/boxo/blob/b101ba0858150cd6a34b6cb68eefd28c411bd58c/gateway/errors.go#L211-L216

@MichaelMure recently let me know that:
1. Infura also does some string based error checking for their errors
2. They return 451 when rather than just 410

So it seems to me like this should be bundled into boxo tooling.
1. Have an error type indicating that it's a blocked error (which other tools like nopfs or others can wrap)
2. Given that 451 and 410 both seem plausible either that should be configurable by the error tooling, or at the gateway level
   - At the moment I don't know anyone mixing 410s + 451s but that may not always be the case

@MichaelMure let me know if there's any context I missed.

cc @lidel

PsychoPunkSage:
Previously, blocked content detection relied on string matching which is fragile and implementation-specific. This PR:

- Adds `ErrorContentBlocked` type to properly handle blocked content errors
- Allows configurable HTTP status codes (410 Gone or 451 Unavailable For Legal Reasons) 
- Makes error handling consistent across implementations (nopfs, Infura etc.)
- Eliminates string matching in favor of proper error type checking

Related: closes #591

### DIFF
--- a/gateway/errors.go
+++ b/gateway/errors.go
@@ -36,6 +36,12 @@ type ErrorRetryAfter struct {
 	RetryAfter time.Duration
 }
 
+// ErrorContentBlocked represents an error when content is blocked due to filtering
+type ErrorContentBlocked struct {
+	Err        error
+	StatusCode int
+}
+
 func NewErrorRetryAfter(err error, retryAfter time.Duration) *ErrorRetryAfter {
 	if err == nil {
 		err = ErrServiceUnavailable
@@ -46,6 +52,18 @@ func NewErrorRetryAfter(err error, retryAfter time.Duration) *ErrorRetryAfter {
 	return &ErrorRetryAfter{
 		RetryAfter: retryAfter,
 		Err:        err,
+	}
+}
+
+func NewErrorContentBlocked(err error, statusCode int) *ErrorContentBlocked {
+	// Will default to Error:451 if no status code is provided
+	if statusCode != http.StatusGone && statusCode != http.StatusUnavailableForLegalReasons {
+		statusCode = http.StatusUnavailableForLegalReasons // 451 by default
+	}
+
+	return &ErrorContentBlocked{
+		Err:        err,
+		StatusCode: statusCode,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gateway/errors.go
```diff
@@ -49,6 +55,18 @@ func NewErrorRetryAfter(err error, retryAfter time.Duration) *ErrorRetryAfter {
 	}
 }
 
+func NewErrorContentBlocked(err error, statusCode int) *ErrorContentBlocked {
+	// Will default to Error:451 if no status code is provided
+	if statusCode != http.StatusGone && statusCode != http.StatusUnavailableForLegalReasons {
+		statusCode = http.StatusUnavailableForLegalReasons // 451 by default
+	}
```
lidel, line 62: nit: legal one should not be the default, switch to 410

gateway/errors.go
```diff
@@ -255,5 +298,6 @@ func isErrNotFound(err error) bool {
 func isErrContentBlocked(err error) bool {
 	// TODO: we match error message to avoid pulling nopfs as a dependency
 	// Ref. https://github.com/ipfs-shipyard/nopfs/blob/cde3b5ba964c13e977f4a95f3bd8ca7d7710fbda/status.go#L87-L89
-	return strings.Contains(err.Error(), "blocked and cannot be provided")
+	var blocked *ErrorContentBlocked
+	return errors.As(err, &blocked)
```
lidel, line 302: Right now nothing will return `ErrorContentBlocked` type, so `isErrContentBlocked` will always return false and break content blocking response code in projects that use `boxo/gateway`, like Kubo.

What is the plan for making this work? 

We don't want to depend on https://github.com/ipfs-shipyard/nopfs as noted  in the comment here. Perhaps we want to keep the `string.Contains` and check it in addition to the type check here?

lidel (changes requested):
There is bit more to do here, this PR in current form does not solve #591 – see comment inline. 


@PsychoPunkSage Mind opening a companion PR against https://github.com/ipfs/kubo  that uses boxo from this PR? We have content blocking tests there in https://github.com/ipfs/kubo/blob/master/test/cli/content_blocking_test.go that need to pass before we are comfortable merging this.

Also cc'ing @MichaelMure  for feedback from his end: does this look like useful direction?

gateway/errors.go
PsychoPunkSage: Done

gateway/errors.go
PsychoPunkSage: done

lidel:
Triage notes:
- marking requested changes as "resolved" and "done"  but not addressing them
- closing due to the lack of engagement and suspicion this was generated with AI
  - if this was not the case, and you simply were afk for prolonged amount of time, feel free to open new PR (once feedback is addressed)

PR #865, opened 2025-03-03
feat(bitswap/client): MinTimeout for DontHaveTimeoutConfig

Wondertan:
In low latency environments, the latency estimation becomes very sensitive to sudden spikes in serving time for a Block.
It may happen that Block serving takes more than expected for a ordinary reason, yet triggering don't have timeout to fire a new often duplicate request. The min timeout configuration allows us to set the baseline timeout, beyond which a new request should be fired, substantially minimizing amount of duplicates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The following emojis are used to highlight certain changes:
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


guillaumemichel (approved):
Seems fair.

@Wondertan could you add a changelog entry?

Wondertan:
@guillaumemichel, done

PR #870, opened 2025-03-05
provider: add a buffered KeyChanFunc.

hsanjuan:
This provider helper allows to buffer all the results from a keyChanFunc in memory.

The purpose is to fix issues with slow re-providing. In the case of https://github.com/ipfs/kubo/issues/10596, the slow re-providing process causes starvation of any operations trying to read/write to the pinset.

With the new buffered KeyChanFunc, we can read everything we need to announce into memory first, and free any locks as soon as possible.

Given the compact size of CIDs (<50bytes), I don't think any complexer approach (batch reading and lock/unlock handling) is justified. People with pinsets of 20 million items that want to announce everything can well spare an extra GB of RAM. For smaller repos the impact becomes negligible.

The test targets precisely the use-case and ensures we don't starve operations while interacting with dspinner.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -9,7 +9,7 @@ package provider
 import (
 	"context"
 
-	"github.com/gammazero/deque"
+	"github.com/gammazero/chanqueue"
 	blocks "github.com/ipfs/boxo/blockstore"
 	"github.com/ipfs/boxo/fetcher"
 	fetcherhelpers "github.com/ipfs/boxo/fetcher/helpers"
@@ -79,10 +79,10 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
 }
 
 // NewBufferedProvider returns a KeyChanFunc supplying keys from a given
-// KeyChanFunction, but buffering all the keys in memory first until they are
-// read.  This allows the underlying KeyChanFunc to finish listing pins as
-// soon as possible releasing any resources, locks, at the expense of memory
-// usage.
+// KeyChanFunction, but buffering keys in memory if we can read them faster
+// they are consumed.  This allows the underlying KeyChanFunc to finish
+// listing pins as soon as possible releasing any resources, locks, at the
+// expense of memory usage.
 func NewBufferedProvider(pinsF KeyChanFunc) KeyChanFunc {
 	return func(ctx context.Context) (<-chan cid.Cid, error) {
 		pins, err := pinsF(ctx)
@@ -90,35 +90,17 @@ func NewBufferedProvider(pinsF KeyChanFunc) KeyChanFunc {
 			return nil, err
 		}
 
-		// Use very efficient queue implementation for storing the cids.
-		queue := new(deque.Deque[cid.Cid])
-		queue.SetBaseCap(1024)
-
-		totalCids := 0
-		for c := range pins {
-			queue.PushBack(c)
-			totalCids++
-		}
-
-		outCh := make(chan cid.Cid) // channel for the consumer.
+		queue := chanqueue.New[cid.Cid]()
 
 		go func() {
-			defer close(outCh)
-			for {
-				if totalCids == 0 {
-					return
-				}
-				c := queue.PopFront()
-				select {
-				case <-ctx.Done():
-					return
-				case outCh <- c:
-					totalCids--
-				}
+			in := queue.In()
+			defer close(in)
+			for c := range pins {
+				in <- c
 			}
 		}()
 
-		return outCh, nil
+		return queue.Out(), nil
 	}
 }
 
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -90,16 +90,7 @@ func NewBufferedProvider(pinsF KeyChanFunc) KeyChanFunc {
 			return nil, err
 		}
 
-		queue := chanqueue.New[cid.Cid]()
-
-		go func() {
-			in := queue.In()
-			defer close(in)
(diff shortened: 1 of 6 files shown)


guillaumemichel (approved):
The reasoning and the code make sense. The windows CI doesn't seem happy with the tests though

hsanjuan:
yeah @gammazero is this related to dequed "power of two" grow allocations?

gammazero:
>  is this related to dequed "power of two" grow allocations?

I do not think so. It does not look like your test creates such a huge number of pins that it should cause OOM. Besides, the grow by power of two is what a slice does anyway. I will test this locally and see what it eating up all the memory.

The `deque` is very memory efficient when inflow and outflow are somewhat balanced, i.e. when size changes do not incur many queue resizes. However, if the use pattern is to enqueue a large (relative to the base capacity) amount of data and then read all of the data, then a better deque implementation is this: https://github.com/edwingeng/deque. It really depends on the use pattern which is more efficient.

provider/provider.go
```diff
@@ -70,6 +78,50 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
...
+				case <-ctx.Done():
+					return
+				case outCh <- c:
+					totalCids--
+				}
+			}
+		}()
+
+		return outCh, nil
+	}
```
gammazero, line 95: You may consider using [`chanqueue`](https://github.com/gammazero/chanqueue?tab=readme-ov-file#chanqueue) instead of `NewBufferedProvider`, as it basically does what the code above does. A chanqueue can supply its own in and out channels or use channels provided to it for input and output.

provider/provider_test.go
```diff
@@ -0,0 +1,86 @@
...
+	root1count := 0
+	root2count := 0
+	for range root2pins {
+		root2count++
+	}
+	for range root1pins {
+		root1count++
+	}
+	require.Equal(t, root1count, 2048, "first pin should have provided 2048 cids")
+	require.Equal(t, root2count, 4096+2048, "second pin should have provided 4096 cids")
```
gammazero: For [`require.Equal`](https://pkg.go.dev/github.com/stretchr/testify/require#Equal) the expected value needs to be first.
```suggestion
	require.Equal(t, 2048, root1count, "first pin should have provided 2048 cids")
	require.Equal(t, 4096+2048, root2count, "second pin should have provided 4096 cids")
```

provider/provider.go
```diff
@@ -70,6 +78,50 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
...
+		}
+
+		outCh := make(chan cid.Cid) // channel for the consumer.
+
+		go func() {
+			defer close(outCh)
+			for {
+				if totalCids == 0 {
+					return
+				}
```
gammazero:
```suggestion
			for queue.Len() != 0 {
```
This way you can get rid of `totalCids`.

provider/provider.go
```diff
@@ -70,6 +78,50 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
...
+
+		// Use very efficient queue implementation for storing the cids.
+		queue := new(deque.Deque[cid.Cid])
+		queue.SetBaseCap(1024)
+
+		totalCids := 0
+		for c := range pins {
+			queue.PushBack(c)
+			totalCids++
+		}
```
gammazero: If this section was run in a goroutine, and was writing to the queue, then the cids could be read from the queue without having wait for all the pins to be enqueued first.  Doing this would mean protecting the queue with synchronization. If there were something that looked like an infinite-length channel, that would work. See [`chanqueue`](https://github.com/gammazero/chanqueue?tab=readme-ov-file#chanqueue)

If you did this, then your tests would need to change to read all the CIDs from the `keyChanF` returned by the first `NewBufferedProvider`, before pinning the next root.

hsanjuan: I should have seen `chanqueue` before. Let's switch to it.

hsanjuan:
The test failure was due to recursion when ipld-prime traverses a 4092-deep graph.

provider/provider.go
```diff
@@ -70,6 +78,32 @@ func NewPinnedProvider(onlyRoots bool, pinning pin.Pinner, fetchConfig fetcher.F
...
+
+		go func() {
+			in := queue.In()
+			defer close(in)
+			for c := range pins {
+				in <- c
+			}
+		}()
+
+		return queue.Out(), nil
```
gammazero: If you update to chanqueue [v1.1.0](https://github.com/gammazero/chanqueue/releases/tag/v1.1.0), then  the code can be simplified to:
```suggestion
		queue := chanqueue.New(chanqueue.WithInputRdOnly[cid.Cid](pins))
		return queue.Out(), nil
```

gammazero: It will operate the same, but without the additional goroutine. `queue.Out()` will be closed when `pins` in closed and items are read.

The only difference from your original code is that canceling the context does not cause the output channel to immediately close, requiring all the output to be read first. If handling the closed context immediately is a concern, then it can be done by:
```go
context.AfterFunc(ctx, func() {
	queue.Shutdown()
})
```
... as Shutdown drains the Chanqueue.

gammazero (approved):
See comment.

gammazero (approved):

provider/provider.go
hsanjuan: Perfect thanks!

PR #871, opened 2025-03-05
fix(provider): call reprovider throughput callback only if reprovide is enabled

guillaumemichel:
Addresses https://github.com/ipfs/kubo/issues/10714

`shouldReprovide` doesn't test whether the `reprovideInterval` is `0`, and instructs to reprovide even when it shouldn't.

### DIFF
--- a/provider/reprovider.go
+++ b/provider/reprovider.go
@@ -44,6 +44,8 @@ type reprovider struct {
 	closewg sync.WaitGroup
 	mu      sync.Mutex
 
+	// reprovideInterval is the time between 2 reprovides. A value of 0 means
+	// that no automatic reprovide will be performed.
 	reprovideInterval        time.Duration
 	initalReprovideDelay     time.Duration
 	initialReprovideDelaySet bool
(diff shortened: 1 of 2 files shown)


provider/reprovider.go
```diff
@@ -383,7 +383,7 @@ func (s *reprovider) run() {
 
 			s.throughputDurationSum += dur
 			s.throughputProvideCurrentCount += uint(len(keys))
-			if s.throughputCallback != nil && s.throughputProvideCurrentCount >= s.throughputMinimumProvides {
+			if s.throughputCallback != nil && s.reprovideInterval > 0 && s.throughputProvideCurrentCount >= s.throughputMinimumProvides {
```
lidel: @guillaumemichel hm.. but this code would execute only if  `performedReprovide` is `true`, so we are not fixing underlying problem, just hiding error message, or did I misread whats happening? 

any idea why `performedReprovide` would be `true` if reprovide interval is `0`?

guillaumemichel: Found the cause and addressed it in https://github.com/ipfs/boxo/pull/871/commits/01710006ac0c1991039532d5309156ec4df1eb38

gammazero (approved):

PR #872, opened 2025-03-06
fix(bitswap/client): dont set nil for DontHaveTimeoutConfig

Wondertan:
I thought I was going nuts... That's not even funny.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- `bitswap/client`: Fix unintentional ignoring `DontHaveTimeoutConfig` [#872](https://github.com/ipfs/boxo/pull/872)
+
 ### Security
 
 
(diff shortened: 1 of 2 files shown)


guillaumemichel (approved):
Thanks @Wondertan!

guillaumemichel (changes requested):
Actually, since `bs.dontHaveTimeoutConfig` is only used to define `peerQueueFactory`, setting it to nil should have no effect on the closure previously defined. This looks like a memory optimization.

Any opinions @gammazero?

Wondertan:
@guillaumemichel, closure takes value from Client struct tho, so this nil sets the config entirely to zero even if its set.

The DontHaveTimeout component then makes a default value silently if nil is given, so in the end DontHaveTimeout component works but with default config even when user passes a modified config.

This is the reason why I was going "nuts". I spent too much time debugging the config and my understanding of Bitswap code there, while this was an issue...

gammazero:
> Actually, since `bs.dontHaveTimeoutConfig` is only used to define `peerQueueFactory`, setting it to nil should have no effect on the closure previously defined. This looks like a memory optimization.
> 
> Any opinions @gammazero?

The values inside in the function are not evaluated until the function is run, so when `peerQueueFactory` is called, `bs.perPeerSendDelay` will appear to be `nil`. This is nothing more than an attempt to dispose of memory that is not needed, but it affects what it inside the closure.

gammazero (approved):

Wondertan:
should we make it skipchangelog?

gammazero:
> should we make it skipchangelog?

I think it would be good make a note that we no longer ignore DontHaveTimeoutConfig. If there is a change in functionality or in the way something is used, then it should be noted in change log. I think this qualifies.

PR #875, opened 2025-03-07
Fix PeerManager mutex bottleneck

gammazero:
Do not try to acquire a MessageQueue mutex while holding the PeerManager mutex. A MessageQueue mutex may already be held for message processing and waiting for these will delay releasing the global PeerManager mutex. Instead, enqueue calls to MessageQueue and let an asynchronous goroutine execute the calls so that the PeerManager lock can be released sooner, without having to wait for the MessageQueue operations.

In short, this decouples the PeerManager from the per-peer MessageQueue mutexes.

Kubo PR: https://github.com/ipfs/kubo/pull/10749

### DIFF
--- a/bitswap/client/internal/messagequeue/messagequeue.go
+++ b/bitswap/client/internal/messagequeue/messagequeue.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/filecoin-project/go-clock"
+	"github.com/gammazero/chanqueue"
 	bswl "github.com/ipfs/boxo/bitswap/client/wantlist"
 	bsmsg "github.com/ipfs/boxo/bitswap/message"
 	pb "github.com/ipfs/boxo/bitswap/message/pb"
@@ -107,6 +108,8 @@ type MessageQueue struct {
 	events chan<- messageEvent
 
 	perPeerDelay time.Duration
+
+	requests *chanqueue.ChanQueue[func()]
 }
 
 // recallWantlist keeps a list of pending wants and a list of sent wants
@@ -325,7 +328,17 @@ func newMessageQueue(
 		msg:    bsmsg.New(false),
 		clock:  clk,
 		events: events,
+
+		requests: chanqueue.New[func()](),
 	}
+}
+
+func (mq *MessageQueue) Sync() <-chan struct{} {
+	done := make(chan struct{})
+	mq.requests.In() <- func() {
+		close(done)
+	}
+	return done
 }
 
 // Add want-haves that are part of a broadcast to all connected peers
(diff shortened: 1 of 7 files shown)


bitswap/client/internal/messagequeue/messagequeue.go
```diff
@@ -325,7 +328,17 @@ func newMessageQueue(
 		msg:    bsmsg.New(false),
 		clock:  clk,
 		events: events,
+
+		requests: chanqueue.New[func()](),
+	}
+}
+
+func (mq *MessageQueue) Sync() <-chan struct{} {
```
guillaumemichel: `Sync()` seems unused

bitswap/client/internal/peermanager/peermanager.go
```diff
@@ -96,18 +95,16 @@ func (pm *PeerManager) Connected(p peer.ID) {
 // Disconnected is called to remove a peer from the pool.
 func (pm *PeerManager) Disconnected(p peer.ID) {
 	pm.pqLk.Lock()
+	defer pm.pqLk.Unlock()
```
guillaumemichel: Does it help to hold the lock longer here?

bitswap/client/internal/peermanager/peermanager.go
```diff
@@ -119,9 +116,9 @@ func (pm *PeerManager) Disconnected(p peer.ID) {
 // Note that this is just used to calculate latency.
 func (pm *PeerManager) ResponseReceived(p peer.ID, ks []cid.Cid) {
 	pm.pqLk.RLock()
-	pq, ok := pm.peerQueues[p]
-	pm.pqLk.RUnlock()
+	defer pm.pqLk.RUnlock()
```
guillaumemichel, line 121: Does it help to hold the lock longer here?

bitswap/client/internal/peermanager/peerwantmanager_test.go
```diff
@@ -213,6 +223,7 @@ func TestPWMSendWants(t *testing.T) {
 }
 
 func TestPWMSendCancels(t *testing.T) {
+	t.Skip()
```
guillaumemichel: Why?

guillaumemichel (commented):
Do we have a way to quantify performance gains of linearizing the execution using `chanqueue`?

bitswap/client/internal/messagequeue/messagequeue.go
gammazero: Yes, was there for previous testing, no longer used. Removed.

bitswap/client/internal/peermanager/peermanager.go
gammazero: It is necessary to call `pq.ResponseReceived` while holding the lock. Otherwise another goroutine could shutdown the  MessageQueue, between `RUnlock` and `pq.ResponseReceived`,  and that would cause use of a shutdown message queue that is no longer reading channels.

bitswap/client/internal/peermanager/peerwantmanager_test.go
gammazero: left by mistake during debug

bitswap/client/internal/peermanager/peermanager.go
gammazero: no, changed back.

gammazero:
> Do we have a way to quantify performance gains of linearizing the execution using `chanqueue`?

Watch staging metrics for now.

There is not any "linearizing the execution" since the code passed to the chanqueue was already serialized by the MessageQueue `wllock` mutex, and there is a separate chanqueue for each peer (one per MessageQueue).

My bigger concern is that now there are 2 more goroutines per peer, so if this does not reduce waiting goroutines then it actually can make things worse.

gammazero:
PR is not significantly effective

PR #876, opened 2025-03-08
refactor: use t.TempDir() instead of os.MkdirTemp

MarkDaveny:
`TempDir()` is a method introduced in Go 1.15 for `testing.T`. It automatically creates a temporary directory and cleans it up after the test is complete, eliminating the hassle of manually handling errors and cleanup.  More detail about TempDir  https://pkg.go.dev/testing#B.TempDir

### DIFF
--- a/examples/gateway/car-file/README.md
+++ b/examples/gateway/car-file/README.md
@@ -2,7 +2,7 @@
 
 This is an example that shows how to build a Gateway backed by the contents of
 a CAR file. A [CAR file](https://ipld.io/specs/transport/car/) is a Content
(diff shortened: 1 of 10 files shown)


hsanjuan:
Yeah but why? what is the improvement here other that merits the hassle of opening and reviewing a PR?

gammazero:
Looks like this PR missed other places this same change could have been made:
- `files/filter_test.go`
- `filestore/filestore_test.go`
- `tar/extractor_test.go`

This PR is not complete and what is here will be combined with other nonfunctional changes.

PR #877, opened 2025-03-10
bitswap/httpnet: Add WithDenylist option

### LINKED ISSUE #861: bitswap/httpnet: add WithDenylist option
We have ability to enable HTTP retrieval for specific hostnames (only do HTTP with hosts from allowlist) via `httpnet.WithAllowlist`

To prepare for the future where we enable this by default in places like Kubo, there should be a companion, opposite flag that allows us to pass explicit denylist (with known hosts that should NOT be used for HTTP retrieval) – `httpnet.WithDenylist`

The denylist should apply to both original URL and any HTTP redirects.

The main reason here is to avoid situation where  defunct gateways start returning redirects to `ipfs.io` and `dweb.link`, effectively acting as amplification vector (note: there may  be a better fix in https://github.com/ipfs/boxo/issues/862).

cc @hsanjuan

hsanjuan:
Fixes #861


hsanjuan:
Perhaps this PR is a good opportunity to solve an additional problem:

- If a peer has no good http addresses to connect to, perhaps we should fail-over to bitswap. For this we need special handling in the proxy.

And an additional issue:

- If we manage to add HTTP addresses  from a peer to the peerstore without having called Connect() (i.e. somehow the other peer connected to us etc.), the HTTP codepaths will trigger and the MessageSender hits http addresses bypassing allowlists. I have seen this a couple of times.

hsanjuan:
> If a peer has no good http addresses to connect to, perhaps we should fail-over to bitswap. For this we need special handling in the proxy.

This was already the case...

bitswap/network/httpnet/httpnet.go
```diff
@@ -366,97 +376,103 @@ func (ht *Network) Self() peer.ID {
...
-		ht.connEvtMgr.Connected(p.ID)
-		ht.pinger.startPinging(p.ID)
-		log.Debugf("connect success to %s (supports HEAD: %t)", p.ID, supportsHead)
-		// We "connected"
-		return nil
+	// if we have 0 new, valid urls (or not space for them) and not
+	// connected already, error if we are connected, ignore.
+	if len(workingAddrs) == 0 {
+		err := fmt.Errorf("connect failure to %s: %w", p, ErrNoSuccess)
+		log.Debug(err)
```
gammazero: Since the error is returned, would it be better to let the caller log the error, since the caller may have more context to include in a log message?

bitswap/network/router.go
```diff
@@ -146,14 +146,19 @@ func (rt *router) Stats() Stats {
...
-	// IF we did not manage to connect to any HTTP address beforehand, we
-	// should not have them in the peerstore.
-	pi := rt.Peerstore.PeerInfo(p)
-	htaddrs, _ := SplitHTTPAddrs(pi)
-	if len(htaddrs.Addrs) > 0 {
-		return rt.HTTP.NewMessageSender(ctx, p, opts)
+	// There seem to be cases when we have HTTP addresses in the peerstore
+	// even though we have not connected to the peer via HTTP before. This
+	// can then bypass allowlists etc.  HTTP.NewMessageSender() checks the
+	// the peer is known to HTTP or errors.
```
gammazero:
```suggestion
	// can then bypass allowlists etc.  HTTP.NewMessageSender() checks if
	// the peer is known to HTTP or errors.
```

gammazero (approved):

bitswap/network/httpnet/httpnet.go
hsanjuan: In this case the caller is the providerQueryManager which has no knowledge on how connection happens (bitswap/http).

As a general rule I guess you are right, here it is useful to have this as part of `httpnet` debug facility so that it is easy to know and debug 1) who you are connecting to via http, 2) why it isn't working.

hsanjuan: Reading again, we are actually loging the failures in more detail in the connectToURL function so we can skip this one. I have updated the line and the comment above which was also semi-leftover.

Issue #862, opened 2025-02-28
bitswap/httpnet: request with CDN-Loop  (RFC 8586)


lidel:
We need to find a solution to situation where defunct gateways start returning redirects back to `ipfs.io` and `dweb.link`, effectively acting as amplification vector.

There is some prior art solution in form of `CDN-Loop` header which helps in avoiding unintended HTTP loops like that:

> 
> This document defines the CDN-Loop request header field for HTTP.
> CDN-Loop addresses an operational need that occurs when an HTTP
> request is intentionally forwarded between Content Delivery Networks
> (CDNs), but is then accidentally or maliciously re-routed back into
> the original CDN causing a non-terminating loop.  The new header
> field can be used to identify the error and terminate the loop.

Ref. 
- https://datatracker.ietf.org/doc/html/rfc8586
- https://blog.cloudflare.com/preventing-request-loops-using-cdn-loop/
- https://github.com/ipfs/specs/issues/426

I think in case of Rainbow, we could leverage SEED and have all nodes set `CDN-Loop` to the same deterministic hash value, and refuse to process requests with matching one. This way HTTP traffic looped/proxied back to our infra is ignored.

cc @hsanjuan for feedback, maybe there a better way?

hsanjuan:
@lidel we could simply not support redirects... Our urls come from content provider records, which, iirc, have expirations and need to be renewed regularly. If an endpoint changes, the action to take is to update the provider record, rather than setting a redirect.

It's also not even about loops, its also about redirect hops (1 hop means opening a new connection somewhere else which is already expensive by itself in our context).

lidel:
@hsanjuan good point. not following redirects sounds ok to me. 

interpreting 3XX as 404 will force people to update announcements rather than keeping dead domains alive and then breaking months/years later. 
take a look at proposed spec clarification in https://github.com/ipfs/specs/pull/494


still, the utility of `CDN-Loop` goes beyond redirects. we've seen people setting up dumb Nginx-based proxies that blindly forwarded (not redirected) requests to `ipfs.io` with all original headers (*cough* rogue Rhea/Saturn nodes). 

perhaps for now, disable following redirects, and keep this issue about `CDN-Loop` as low priority (P3) until we actually need it?

PR #878, opened 2025-03-10
bitswap/httpnet: do not follow redirects

hsanjuan:
Avoid following redirects. We treat them as an incorrect provider record.  We don't want to be coerced into opening new connections to new hosts and we prefer provider records to be up to date rather than them relying in redirects being followed. There are also problems with potential loops etc.

https://github.com/ipfs/boxo/issues/862

### DIFF
--- a/bitswap/network/httpnet/httpnet.go
+++ b/bitswap/network/httpnet/httpnet.go
@@ -267,6 +267,12 @@ func New(host host.Host, opts ...Option) network.BitSwapNetwork {
 
 	c := &http.Client{
 		Transport: t,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			// we do not follow redirects. Providers should keep
+			// announcements up to
+			// date. https://github.com/boxo/issues/862.
+			return http.ErrUseLastResponse
+		},
 	}
 	htnet.client = c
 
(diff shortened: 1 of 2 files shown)


bitswap/network/httpnet/httpnet.go
```diff
@@ -267,6 +267,12 @@ func New(host host.Host, opts ...Option) network.BitSwapNetwork {
 
 	c := &http.Client{
 		Transport: t,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			// we do not follow redirects. Providers should keep
+			// announcements up to
+			// date. https://github.com/boxo/issues/862.
+			return http.ErrUseLastResponse
```
lidel, line 274: https://pkg.go.dev/net/http states 
> ErrUseLastResponse can be returned by Client.CheckRedirect hooks to control how redirects are processed. If returned, the next request is not sent and the most recent response is returned with its body unclosed.

@hsanjuan  does this mean we need to manually close body in client code that called `client.Do`?

hsanjuan: Whops cafa7e2 now closes the body after reading.

hsanjuan:
Actually closing the body only matters for HTTP/1. I deployed this and I don't see any perf improvement, if anything perf is worse...

gammazero (approved):

hsanjuan:
Note: I deployed on the other staging box and perf remained the same, meaning no perf impact, differences can be attributed to the local machines.

PR #880, opened 2025-03-11
Refactor(hostname): Skip DNSLink for local IP addresses to avoid DNS queries

### LINKED ISSUE #879: Slow responses to requests on 127.0.0.1 when DNSLink enabled
When DNSLink is enabled in kubo (by default), queries to kubo on `127.0.0.1` take longer to respond to than the same queries to `localhost`.

The code in the gateway hostname handler, [here](https://github.com/ipfs/boxo/blob/main/gateway/hostname.go#L24), recognizes `localhost` in the `Host` header, but it does not recognize `127.0.0.1`. This leads it all the way [here](https://github.com/ipfs/boxo/blob/main/gateway/hostname.go#L215) after checking for DNSLink record.

The hostname handler code should recognize IP addresses (both IPv4 and IPv6) and avoid DNSLink lookup for these.

gitsrc:
This PR refactors the hostname handling logic to skip DNSLink lookups for IP addresses.

The hasDNSLinkRecord function now explicitly checks if the provided hostname is an IP address. If it is an IP address, the function immediately returns false, skipping the DNSLink lookup.

This change avoids unnecessary DNS queries when the hostname is an IP address. This improves efficiency and reduces potential delays, especially in local development environments where DNS lookups for IPs are not needed and can sometimes cause issues.

 Closes #879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,7 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
-- `gateway`: Skip DNSLink lookup for local IP addresses to avoid unnecessary DNS queries
-
+- `gateway`: Skip DNSLink lookup for IP addresses to avoid unnecessary DNS queries [#880](https://github.com/ipfs/boxo/pull/880)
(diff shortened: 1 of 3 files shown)


gammazero:
Turns out that we need to igore app IP addresses, not just localhost.

gammazero (approved):

PR #883, opened 2025-03-12
Limit bitswap sessions

gammazero:
Limit number of bitswap client sessions to prevent goroutine runaway


gammazero:
Closing as this does not really have the desired effect of reducing want list buildup. It needs to also include some sort of wantlist GC.

PR #886, opened 2025-03-18
fix(gateway): preserve query parameters on _redirects

SgtPooki:
Preserve query parameters for _redirects file redirects

Still needs a test.. and potentially CHANGELOG.md update. 

related https://github.com/ipfs/service-worker-gateway/issues/628 + https://github.com/ipfs/specs/pull/498

cc @lidel


lidel (approved):
This won't help with Cloudflare-hosted sw gateway, but we should fix the bug where params are not preserved ( we preserve them in all other cases such as subdomain redirects).

- [x] Added support for merging static and dynamic query parameters, prioritizing user values
- [x] Added tests and
- [x] CHANGELOG.md updated
- [x] Opened https://github.com/ipfs/specs/pull/498 to fill the spec gap and clarify how query parameters should be handled.


@SgtPooki in spare time take a look, if specs/tests look fine, I think it is fine to merge this fix (query params should be preserved).

SgtPooki (commented):
tests look good to me

Issue #881, opened 2025-03-11
gateway: limit the number of requests in flight


lidel:
> Filling issue based on triage discussion today, cc @gammazero  if below prior art context is useful

Currently, we have no "requests in flight" cap in `boxo/gateway` and require reverse proxy in front of go binary to enforce the limit.

## Proposed improvement

There should be a built-in mechanism to limit the number of ongoing HTTP gateway requests in flight, and return HTTP 503 Service Unavailable with [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) when limit is reached.

Make it a config option, `boxo/gateway` should ship with some implicit default, protecting people who expose it blindly to the internet 

### Prior art 

- At public gateway, 
  - the current ceiling exist at nginx and for our rainbow prod seems to be really high: [`worker_connections`](https://nginx.org/en/docs/ngx_core_module.html#worker_connections) is set to `10240` (per each instance). For reference, nginx default is 512, raising to 1024 seems to be a sensible place to start as default, protecting from traffic spikes.
  - there is a timeout of 30 seconds (nginx's `proxy_read_timeout`)  – we dont have built-in config yet, but its tracked in https://github.com/ipfs/boxo/issues/679

PR #887, opened 2025-03-19
feat(gateway): implement request concurrency limit

gitsrc:
This commit introduces a mechanism to limit the number of concurrent in-flight HTTP requests to the gateway.

- Implements a configurable limit for concurrent requests.
- Returns HTTP 503 Service Unavailable with Retry-After header when the limit is reached.
- Includes a default limit to protect against traffic spikes when exposed without a reverse proxy.

This change enhances gateway's resilience and prevents overload by limiting concurrent requests.

#881

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


hsanjuan:
Triage notes: this will be triaged as part Kubo v0.35 (@lidel )

lidel:
Triage notes: moving to ~0.36~ 0.37

hsanjuan:
Triage notes: @lidel is still on it, has not forgotten about, will be dealt with in due time.

lidel:
This endeavor needs more metrics and tests and integration with timeouts.

The atomic.AddInt32 proposed in this PR is ok, however feels like not the best tool for the job. Let's go with more idiomatic semaphore approach from https://github.com/ipfs/boxo/pull/994  which has lower risk of busy-waiting/checking/mutating shared int value in a tight loop.

PR #888, opened 2025-03-19
Improve performance of data onboarding

gammazero:
Adding a large directory of data to IPFS can take a long time. A significant amount of this time is spent writing to and reading from the persisted provider queue, depending on the underlying datastore. The time spent doing this can be greatly reduced by writing queued items to the datastore in batches. In addition to batched writes, reading queued items can be read directly from the input buffer if there are no queued items in the datastore. This allows in-memory only operation when the number of queued items is below the batch size, and when the queue is not idle for too long.

kubo PR: https://github.com/ipfs/kubo/pull/10758
part of: https://github.com/ipshipyard/roadmaps/issues/6 / https://github.com/ipshipyard/roadmaps/issues/7

## Time Comparison
Using freshly initialized kubo each time. All files in test data were 1Mib and filled with random data.

### Before PR (No Buffer)

```
> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-100M > /dev/null
 100.00 MiB / 100.00 MiB [=====================================================================] 100.00%
real	0m6.464s

> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-1G > /dev/null
 1000.00 MiB / 1000.00 MiB [===================================================================] 100.00%
real	1m10.542s

> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-10G > /dev/null
 10.00 GiB / 10.00 GiB [=======================================================================] 100.00%
real	24m5.744s
```

### After PR (With Buffer)
```
> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-100M > /dev/null
 100.00 MiB / 100.00 MiB [=====================================================================] 100.00%
real	0m0.323s

> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-1G > /dev/null
 1.00 GiB / 1.00 GiB [=========================================================================] 100.00%
real	0m2.721s

> time kubo/cmd/ipfs/ipfs add -r /tmp/testfiles-10G > /dev/null
 10.00 GiB / 10.00 GiB [=======================================================================] 100.00%
real	0m35.042s
```

### Summary

| Data     | Before PR | After PR | 
| -------- | --------- | ----------- |
| 100M     | 6.5s    | 0.32s      | 
| 1G       | 70.5s | 2.7s      | 
| 10G      | 1446s  | 35s    |

### DIFF
--- a/provider/internal/queue/queue.go
+++ b/provider/internal/queue/queue.go
@@ -22,7 +22,7 @@ const (
 	batchCommitInterval = 5 * time.Second
 
 	// Number of input CIDs to buffer without blocking.
-	inputBufferSize = 1024 * 256
+	inputBufferSize = 1024 * 64
 	// Time for Close to wait to finish writing CIDs to datastore.
 	shutdownTimeout = 30 * time.Second
 )
@@ -41,7 +41,6 @@ type Queue struct {
 	// e.g. provider vs reprovider
 	ds        datastore.Batching
 	dequeue   chan cid.Cid
-	enqueue   chan cid.Cid
 	inBuf     *chanqueue.ChanQueue[cid.Cid] // in-memory queue to buffer input
 	close     context.CancelFunc
 	closed    chan struct{}
@@ -53,23 +52,17 @@ type Queue struct {
 
 // NewQueue creates a queue for cids
 func NewQueue(ds datastore.Batching) *Queue {
-	dequeue := make(chan cid.Cid)
-	enqueue := make(chan cid.Cid)
-
 	ctx, cancel := context.WithCancel(context.Background())
+
 	q := &Queue{
 		close:    cancel,
 		closed:   make(chan struct{}),
 		ds:       namespace.Wrap(ds, datastore.NewKey("/queue")),
-		dequeue:  dequeue,
-		enqueue:  enqueue,
+		dequeue:  make(chan cid.Cid),
+		inBuf:    chanqueue.New(chanqueue.WithCapacity[cid.Cid](inputBufferSize)),
 		syncDone: make(chan struct{}, 1),
 	}
 
-	q.inBuf = chanqueue.New(
-		chanqueue.WithInput(enqueue),
-		chanqueue.WithCapacity[cid.Cid](inputBufferSize),
-	)
 	go q.worker(ctx)
 
 	return q
@@ -111,7 +104,7 @@ func (q *Queue) Enqueue(cid cid.Cid) (err error) {
 			err = errors.New("failed to enqueue CID: shutting down")
 		}
 	}()
-	q.enqueue <- cid
+	q.inBuf.In() <- cid
 	return
 }
 
@@ -126,7 +119,7 @@ func (q *Queue) worker(ctx context.Context) {
 	defer q.inBuf.Shutdown()
 
 	var (
-		c       cid.Cid = cid.Undef
+		c       cid.Cid
 		counter uint64
 		cstr    string
 		k       datastore.Key = datastore.Key{}
@@ -152,8 +145,9 @@ func (q *Queue) worker(ctx context.Context) {
 	batchTicker := time.NewTicker(batchCommitInterval)
 	defer batchTicker.Stop()
 
+	var lastCid cid.Cid
+
 	for {
-		//fmt.Println("---> inbuf len:", q.inBuf.Len(), "batch count:", batchCount)
 		if c == cid.Undef {
 			head, err := q.getQueueHead(ctx)
 			if err != nil {
@@ -188,12 +182,15 @@ func (q *Queue) worker(ctx context.Context) {
 				return
 			}
 			if toQueue == cid.Undef {
-				if batchCount != 0 {
-					commit = true
-				}
+				commit = true
 				needSync = true
 				break
 			}
+			if toQueue == lastCid {
+				continue
+			}
+			lastCid = toQueue
+
 			// Add suffix to key path to allow multiple entries with same sequence.
 			nextKey := datastore.NewKey(fmt.Sprintf("%020d/%s", counter, cstr))
 			counter++
@@ -215,10 +212,22 @@ func (q *Queue) worker(ctx context.Context) {
 				commit = true
 			}
 		case <-batchTicker.C:
-			if batchCount != 0 && q.inBuf.Len() == 0 {
-				commit = true
-			}
+			commit = q.inBuf.Len() == 0
 		case dequeue <- c:
+			// CID may still be in uncommitted batch, so commit current batch first.
+			if batchCount != 0 {
+				if err = b.Commit(ctx); err != nil {
+					log.Errorf("Failed to write cid batch, stopping provider: %s", err)
+					return
+				}
+				b, err = q.ds.Batch(ctx)
+				if err != nil {
+					log.Errorf("Failed to create batch, stopping provider: %s", err)
+					return
+				}
+				batchCount = 0
+			}
+
 			// Do not batch delete. Delete must be committed immediately, otherwise the same head cid will be read from the datastore.
 			err := q.ds.Delete(ctx, k)
 			if err != nil {
@@ -231,25 +240,27 @@ func (q *Queue) worker(ctx context.Context) {
 		}
 
 		if commit {
-			if err = b.Commit(ctx); err != nil {
-				log.Errorf("Failed to write cid batch, stopping provider: %s", err)
-				return
+			if batchCount != 0 {
+				if err = b.Commit(ctx); err != nil {
+					log.Errorf("Failed to write cid batch, stopping provider: %s", err)
+					return
+				}
+				b, err = q.ds.Batch(ctx)
+				if err != nil {
+					log.Errorf("Failed to create batch, stopping provider: %s", err)
+					return
+				}
+				batchCount = 0
 			}
-			b, err = q.ds.Batch(ctx)
-			if err != nil {
-				log.Errorf("Failed to create batch, stopping provider: %s", err)
-				return
-			}
-			batchCount = 0
 			commit = false
-		}
 
-		if needSync {
-			needSync = false
-			select {
-			case q.syncDone <- struct{}{}:
-			case <-ctx.Done():
-				return
+			if needSync {
+				needSync = false
+				select {
+				case q.syncDone <- struct{}{}:
+				case <-ctx.Done():
+					return
+				}
 			}
 		}
 	}
--- a/provider/internal/queue/queue.go
+++ b/provider/internal/queue/queue.go
@@ -37,7 +37,6 @@ const (
 type Queue struct {
 	// used to differentiate queues in datastore
 	// e.g. provider vs reprovider
-	ctx       context.Context
 	ds        datastore.Datastore // Must be threadsafe
 	dequeue   chan cid.Cid
 	enqueue   *chanqueue.ChanQueue[cid.Cid] // in-memory queue to buffer input
--- a/provider/internal/queue/queue.go
+++ b/provider/internal/queue/queue.go
@@ -18,7 +18,7 @@ import (
 var log = logging.Logger("provider.queue")
 
 const (
-	// Number of input CIDs to buffer without blocking. If <= 1, then use channel.
+	// Number of input CIDs to buffer without blocking.
 	inputBufferSize = 65536
 	// Time for Close to wait to finish writing CIDs to datastore.
 	shutdownTimeout = 5 * time.Second
@@ -47,24 +47,32 @@ type Queue struct {
 
 // NewQueue creates a queue for cids
 func NewQueue(ds datastore.Datastore) *Queue {
-	ctx, cancel := context.WithCancel(context.Background())
+	dequeue := make(chan cid.Cid)
+	enqueue := make(chan cid.Cid)
+
 	q := &Queue{
-		ds:      namespace.Wrap(ds, datastore.NewKey("/queue")),
-		dequeue: make(chan cid.Cid),
-		close:   cancel,
-		closed:  make(chan struct{}),
+		dequeue: dequeue,
+		enqueue: enqueue,
 	}
-	if inputBufferSize > 1 {
-		q.enqueue = make(chan cid.Cid)
+
+	if ds == nil {
 		q.inBuf = chanqueue.New(
-			chanqueue.WithInput(q.enqueue),
+			chanqueue.WithInput(enqueue),
+			chanqueue.WithOutput(dequeue),
 			chanqueue.WithCapacity[cid.Cid](inputBufferSize),
 		)
 	} else {
-		q.enqueue = make(chan cid.Cid, inputBufferSize)
+		ctx, cancel := context.WithCancel(context.Background())
+		q.close = cancel
+		q.inBuf = chanqueue.New(
+			chanqueue.WithInput(enqueue),
+			chanqueue.WithCapacity[cid.Cid](inputBufferSize),
+		)
+		q.ds = namespace.Wrap(ds, datastore.NewKey("/queue"))
+		q.closed = make(chan struct{})
+		go q.worker(ctx)
 	}
 
-	go q.worker(ctx)
 	return q
 }
 
@@ -73,15 +81,19 @@ func (q *Queue) Close() error {
 	var err error
 	q.closeOnce.Do(func() {
 		// Close input queue and wait for worker to finish reading it.
-		close(q.enqueue)
-		select {
-		case <-q.closed:
-		case <-time.After(shutdownTimeout):
-			q.close() // force immediate shutdown
-			<-q.closed
-			err = fmt.Errorf("provider queue: %d cids not written to datastore", q.inBuf.Len())
+		if q.ds == nil {
+			q.inBuf.Shutdown()
+		} else {
+			q.inBuf.Close()
+			select {
+			case <-q.closed:
+			case <-time.After(shutdownTimeout):
+				q.close() // force immediate shutdown
+				<-q.closed
+				err = fmt.Errorf("provider queue: %d cids not written to datastore", q.inBuf.Len())
+			}
+			close(q.dequeue) // no more output from this queue
 		}
-		close(q.dequeue) // no more output from this queue
 	})
 	return err
 }
@@ -112,11 +124,7 @@ func (q *Queue) worker(ctx context.Context) {
 	var counter uint64
 
 	var readInBuf <-chan cid.Cid
-	if q.inBuf != nil {
-		readInBuf = q.inBuf.Out()
-	} else {
-		readInBuf = q.enqueue
-	}
+	readInBuf = q.inBuf.Out()
 
 	for {
 		if c == cid.Undef {
--- a/provider/internal/queue/queue.go
+++ b/provider/internal/queue/queue.go
@@ -21,10 +21,11 @@ const (
 	batchSize           = 16384
 	batchCommitInterval = 5 * time.Second
 
-	// Number of input CIDs to buffer without blocking.
+	// Number of input CIDs to buffer without blocking. This capacity is only
+	// used when writing batches to the datastore takes some time.
 	inputBufferSize = 1024 * 64
 	// Time for Close to wait to finish writing CIDs to datastore.
-	shutdownTimeout = 30 * time.Second
+	shutdownTimeout = 20 * time.Second
 )
 
 // Queue provides a FIFO interface to the datastore for storing cids.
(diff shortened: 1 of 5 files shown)


provider/reprovider.go
```diff
@@ -424,7 +424,7 @@ func (s *reprovider) run() {
 				return
 			}
 
-			err := s.reprovide(s.ctx, false)
+			err := s.reprovide(context.Background(), false)
```
gammazero: This is done because `s.reprovide` is already selecting over `s.ctx` so there is no additional context to watch when calling `s.reprovide`

provider/internal/queue/queue_test.go
```diff
@@ -5,73 +5,65 @@ import (
...
-
 	ds := sync.MutexWrap(datastore.NewMapDatastore())
 	queue := NewQueue(ds)
 	defer queue.Close()
 
-	cids := makeCids(10)
-
+	cids := random.Cids(10)
 	for _, c := range cids {
+		//t.Log(c.String())
```
guillaumemichel:
```suggestion
```

guillaumemichel (approved):

provider/internal/queue/queue.go
```diff
@@ -15,96 +17,127 @@ import (
...
+// Cids in the process of being provided when a crash or shutdown occurs may be
+// in the queue when the node is brought back online depending on whether they
+// were fully written to the underlying datastore.
 //
-// Best-effort durability just means that cids in the process of being provided when a
-// crash or shutdown occurs may be in the queue when the node is brought back online
-// depending on whether the underlying datastore has synchronous or asynchronous writes.
+// Input to the queue is buffered in memory, up to inputBufferSize, to increase
+// the speed of data onboarding. This input buffer behaves as a channel with a
+// very large capacity.
```
hsanjuan: Note: If the queue exists so that we don't forget to provide in case of crash, I would expect things are persisted asap. If things are buffered, we partially lose the purpose of the queue?

provider/internal/queue/queue.go
```diff
@@ -15,96 +17,127 @@ import (
 
 var log = logging.Logger("provider.queue")
 
-// Queue provides a best-effort durability, FIFO interface to the datastore for storing cids
+const (
+	// Number of input CIDs to buffer without blocking. If <= 1, then use channel.
+	inputBufferSize = 65536
+	// Time for Close to wait to finish writing CIDs to datastore.
+	shutdownTimeout = 5 * time.Second
```
hsanjuan: This should be enough, but I wonder what if it isn't. Should we wait more? Do we just drop things?

provider/internal/queue/queue.go
```diff
@@ -15,96 +17,127 @@ import (
...
+	var err error
+	q.closeOnce.Do(func() {
+		// Close input queue and wait for worker to finish reading it.
+		close(q.enqueue)
+		select {
+		case <-q.closed:
+		case <-time.After(shutdownTimeout):
+			q.close() // force immediate shutdown
+			<-q.closed
+			err = errors.New("provider queue: some cids not written to datastore")
```
hsanjuan: Can we log how many?

provider/internal/queue/queue.go
```diff
@@ -114,37 +147,48 @@ func (q *Queue) worker() {
...
 
 			if c == cid.Undef {
 				// fast path, skip rereading the datastore if we don't have anything in hand yet
 				c = toQueue
 				k = nextKey
+				cstr = c.String()
 			}
 
-			if err := q.ds.Put(q.ctx, nextKey, toQueue.Bytes()); err != nil {
+			if err := q.ds.Put(ctx, nextKey, toQueue.Bytes()); err != nil {
```
hsanjuan: If this is a source of slowness, would opening and batch and committing as a batch help? We must limit the batch size but maybe batching could just solve the whole problem of buffering and we don't need the special queue.

hsanjuan (changes requested):
I'm lukewarm about the approach:

- We should probably batch datastore writes when we have a Batching datastore and see if that solves most of the problem, as it can improve write perf and at the same time works as buffering and introduces back-pressure like here.
- I would log a warning message on Close saying "We are storing x CIDs before shutting down, waiting for 5 seconds at most". Helps surface that things are not going well on that front even if it does finish before 5 seconds.

provider/internal/queue/queue.go
gammazero: Maybe better to not have a timeout and either it will finish or the process will be killed by impatient user? The reason for the timeout was to be able to generate error (and log it) if items were not persisted to datastore.

Do we really care if they are not persisted, since periodic reprovide will announce them eventually? Another option is to not wait at all, which would be the same as before this PR.

provider/internal/queue/queue.go
lidel, line 165: @hsanjuan  I think there is more to it, especially if datastore is on slow hdd. 

some notes from colo discussion:
- provide queue being persisted in the same hard drive as pins and blocks is probably making things worse for setups like our collab cluster (with slow hdds):
  - CIDs are read from datastore, and then written back to it, and then read and deleted by reprovide task – on slow HDDs, this likely contributes to overall IO wait
- perhaps for stable, long-running nodes like our cluster, there should be an option to avoid hitting the hard drive on reprovide, trading increased memory usage for extra performance?
  - could we add an opt-in configuration option to provider system that avoids hitting disk (datastore) for persisting the queue entirely? 
  - alternative: use in-memory datastore for the provide queue?
    - if we make the queue keys land under deterministic prefix, we could define in-memory datastore only for `/foo` prefix, similar to how we already do that for [`/blocks` in Kubo](https://github.com/ipfs/kubo/blob/b2efaa992f979f4053eee8decbb6d918da3fa247/config/init.go#L170) (where `/blocks` is flatfs, mounted on top of global `/` handled by leveldb)
    - - downside: the `getQueueHead` will still incur cost due to `query.OrderByKey`, but maybe in "performance mode" we dont need this to be sorted?

lidel: @hsanjuan  I think there is more to it, especially if datastore is on slow hdd. 

some notes from colo discussion:
- right now, by default in kubo, the provider queue is being persisted in leveldb datastore
- provide queue being persisted in the same hard drive as pins and blocks is probably making things worse for setups like our collab cluster (with slow hdds):
  - CIDs are read from datastore, and then written back to it, and then read and deleted by reprovide task – on slow HDDs, this likely contributes to overall IO wait
- perhaps for stable, long-running nodes like our cluster, there should be an option to avoid hitting the hard drive on reprovide, trading increased memory usage for extra performance?
  - could we add an opt-in configuration option to provider system that avoids hitting disk (datastore) for persisting the queue entirely? 
  - alternative: use in-memory datastore for the provide queue?
    - if we make the queue keys land under deterministic prefix, we could define in-memory datastore only for `/foo` prefix, similar to how we already do that for [`/blocks` in Kubo](https://github.com/ipfs/kubo/blob/b2efaa992f979f4053eee8decbb6d918da3fa247/config/init.go#L170) (where `/blocks` is flatfs, mounted on top of global `/` handled by leveldb)
    - downside: the `getQueueHead` will still incur cost due to `query.OrderByKey`, but maybe in "performance mode" we dont need this to be sorted?

provider/internal/queue/queue.go
gammazero: Buffering the input does not change the result. CIDs are being persisted just as fast as without buffering, and buffering is done concurrently with persisting the CIDs. ing buffered. The only difference is that the caller does not have to wait for queue to be persisted. A crash that happens in half the time it takes to persist all the CIDs still results in hald the CIDs not getting persisted.

I am wondering about the value of persisting the queue at all. If there is a crash/restart then the periodic reprovide should announce the CIDs after some time. So, the only value of the persisted queue is that these items can be kept on disk instead of in memory,.

provider/internal/queue/queue.go
gammazero: Yes, this is the source of the slowness. Yes, opening a batch does help somewhat, depending on the datastore used. This changes the behavior because with batching nothing is written to the datastore until the batch is full or flushed after some time. This means
- cids spend longer in memroy before being persisted to datastore
- takes longer for a concurrent consumer to read cids from the persisted queue/datastore

Buffering, on the other hand, does not delay cids from being written to datastore because taking data from the input buffer and writing it to datastore is done concurrently with new data being put into the input buffer.

I think it may still be a useful benchmark to see how much difference batching alone makes.

provider/internal/queue/queue.go
hsanjuan: > I am wondering about the value of persisting the queue at all

Right. Isn't the main thing that a reprovide operation can continue where it left off on node restart? (rather than starting from the beginning?).

The side-effects is that content that was just added should be provided asap, even after a restart, so that it makes user experience better (ability to find and fetch content from other node).

A second side-effect is that, when not using the default provide strategy, we are probably announcing blocks that otherwise will not be announced again. i.e. `roots`  + `ipfs add --pin=false` will announce things that should not be announced per the configuration (iirc).

provider/internal/queue/queue.go
hsanjuan: > Buffering, on the other hand, does not delay cids from being written to datastore

Good points, yeah it all depends on how much perf you gain with batching. We could have both too I guess (with something like a per-second batch commit).

hsanjuan: > the provider queue is being persisted in leveldb datastore

We do probably need a way, in very large nodes, to not re-do a reprovide from the beginning on node restart right? If a reprovide sweep is interrupted, it should continue from where it left off rather than starting from scratch. Moving everything to memory is an approach, but we should attempt persistance on shutdown.

Also, perhaps the datastore is not the best place to persist a queue. Perhaps it should be a file that can be read or mmaped.

gammazero: @lidel 
> alternative: use in-memory datastore for the provide queue?

There is no need for a datastore at all if storing in memory, There are no queries to do. All that is needed is an ordered list (a slice is totally sufficient).  The `ChanQueue` that serves as the input buffer will work perfectly.

Even for persisting to disk, using a datastore is not necessary since all that is being done is to keep an ordered list. A alternative could be to append to a file (yes, over-simplified).

provider/internal/queue/queue.go
```diff
@@ -15,144 +17,141 @@ import (
 
 var log = logging.Logger("provider.queue")
 
-// Queue provides a best-effort durability, FIFO interface to the datastore for storing cids
+const (
+	// Number of input CIDs the queue will hold..
+	defaultQueueCapacity = 65536
```
hsanjuan: Seems rather low to me... I would do something like 1M ?

provider/internal/queue/queue.go
```diff
@@ -15,144 +17,141 @@ import (
...
+	// Time for Close to wait to finish writing CIDs to datastore.
+	shutdownTimeout = 5 * time.Second
```
hsanjuan:
```suggestion
	shutdownTimeout = 20 * time.Second
```

? Systemd waits ups to 1m30s when something doesn't want to stop, so I'm guessing we are still way more polite to the user.

provider/internal/queue/queue.go
```diff
@@ -15,144 +17,141 @@ import (
...
-				log.Errorf("Failed to delete queued cid %s with key %s: %s", c, k, err)
-				continue
+			key := datastore.NewKey(fmt.Sprintf("%020d/%s", counter, c.String()))
+			if err := q.ds.Put(ctx, key, c.Bytes()); err != nil {
+				return fmt.Errorf("failed to write provider queue: %w", err)
 			}
-			c = cid.Undef
-		case <-q.ctx.Done():
-			return
+			counter++
```
hsanjuan: Why not make a batch and commit here? We want to shutdown asap. Specially if we take my proposal to increase capacity by x20.

provider/internal/queue/queue.go
```diff
@@ -15,144 +17,141 @@ import (
...
-func (q *Queue) Enqueue(cid cid.Cid) error {
-	select {
-	case q.enqueue <- cid:
-		return nil
-	case <-q.ctx.Done():
-		return errors.New("failed to enqueue CID: shutting down")
-	}
+// Cap returns the queue's capacity.
+func (q *Queue) Cap() int {
+	return q.q.Cap()
```
hsanjuan: Are these exported for a reason? They seem internal so far but maybe there are plans to use them? (just making sure we don't export more than necessary).

provider/internal/queue/queue.go
```diff
@@ -15,144 +17,141 @@ import (
...
+			continue
 		}
+		qIn <- c
 
-		// If c != cid.Undef set dequeue and attempt write, otherwise wait for enqueue
-		var dequeue chan cid.Cid
-		if c != cid.Undef {
-			dequeue = q.dequeue
+		if err = q.ds.Delete(ctx, k); err != nil {
+			log.Errorw("Failed to delete queued cid", "cid", c, "key", k, "err", err)
```
hsanjuan: I would also do the deletions as a batch..

hsanjuan (changes requested):
I'm good with the approach. My main comment is to batch-commit and batch-delete, which can probably make a difference with large queue capacity.

provider/internal/queue/queue.go
gammazero: For testing, but since test already in same package these are not needed. Removed.

provider/internal/queue/queue.go
gammazero: Done

provider/internal/queue/queue.go
gammazero: removed

provider/internal/queue/queue.go
gammazero: Cannot do that du to the way the next queued item is read from the datastore.

provider/internal/queue/queue.go
gammazero: This is probably higher than it ever needs to be. The capacity will only be used if writing batches becomes slow.

provider/internal/queue/queue.go
gammazero: Done

provider/internal/queue/queue.go
```diff
@@ -15,136 +17,254 @@ import (
...
+					if !errors.Is(err, context.Canceled) {
+						log.Errorf("%w, stopping provider", err)
+					}
+					return
+				}
+			}
+
+			// Do not batch delete. Delete must be committed immediately,
+			// otherwise the same head cid will be read from the datastore.
+			if err = q.ds.Delete(ctx, k); err != nil {
```
hsanjuan: This part makes me wonder some things. 

I would expect that there is a path were things get buffered, but:

- If they are read fast enough, nothing is written.
- If they are read slow, the batch fills up and gets commited eventually, otherwise it may be reduced in size if we start dequeuing faster.

But it seems (sorry if I am wrong) here that:
  - every time we enqueue, something is written to the batch.
  - every time we dequeue, we commit the batch, and then delete the element that we batched.
  
As long as I'm not wrong, this means we write to the datastore on dequeue. If we have fast-queues and slow dequeues, we can assume we write larger batches. But if have slow queues and fast-dequeues, do we still write-and-delete a single or a few elements (based on the random selection between `readInBuf` and `dequeue`?

My expectation would be that instead we call `b.Delete` and delete the element directly from the batch, without having committed it at all (assuming that is possible). Unless I'm missing something here and the code does not trigger unnecessary writes.

hsanjuan (commented):
Complexity has increased a bit!

provider/internal/queue/queue.go
gammazero: You are correct about what ie seems (last two bullets).

> If we have fast-queues and slow dequeues, we can assume we write larger batches. 

yes

> if have slow queues and fast-dequeues, do we still write-and-delete a single or a few elements

Yes, a few elements are written (whatever is in the pending batch) and one is deleted (the one just removed from the provider queue). While this may seem inefficient, it was not the cause of the majority of the previous time spent. 

> instead we call b.Delete and delete the element directly from the batch, without having committed it 

This would be my preference as well.That also means that the batch needs to be committed in order for the delete to complete before trying to read the next head from the datastore. OK, that's fine.  The main issue is that I was not sure if it was safe to assume that doing a batch delete would delete the item from the batch and that the delete operation will come after the add operation. If that is safe then it should work. Will retest.

Seems to be ok.  After thinking about it more, it would be a broken batching implementation if the above was not guaranteed.

Done.

provider/internal/queue/queue.go
```diff
@@ -15,136 +17,244 @@ import (
...
+			if err = b.Delete(ctx, k); err != nil {
 				log.Errorf("Failed to delete queued cid %s with key %s: %s", c, k, err)
 				continue
 			}
 			c = cid.Undef
-		case <-q.ctx.Done():
+			batchCount++
+			// Delete (in batch) must be committed immediately, otherwise the
+			// same head cid will be read from the datastore.
+			commit = true
```
hsanjuan: shouldn't this  `batchCount--` (making sure it doesn't go negative) and `commit = false` (delete the line) ?

- Since no error, we deleted an element from the batch, so size has been reduced.
- If we had reached batch-size, we are now below, so we don't have to commit

hsanjuan: There's also the case where the batch has been committed due to getting over the batchCount, so now it is empty, and when we dequeue we do need to make a direct-delete from db and I don't see this happening?

hsanjuan (commented):
Additional comment:

- We enqueue 1 element
- We write to batch (but not commit yet)
- `getQueueHead()` returns nothing because nothing was written (uncommited batch)

This means we need to wait for the batchCommitInterval to trigger a write, so if we provide something below `16*1024` CIDs, we will only start providing after 5 seconds, right?

This is ok, I think, just making sure I understand the flow..

provider/internal/queue/queue.go
gammazero: > shouldn't this `batchCount--` (making sure it doesn't go negative) and `commit = false` (delete the line) ?
> 
> * Since no error, we deleted an element from the batch, so size has been reduced.
> * If we had reached batch-size, we are now below, so we don't have to commit

We deleted an element, but we do not know whether or not it was in the current batch without searching the batch. So we cannot decrement the count.

That head element that was just dequeued and deleted may have been the first one received when the queue was empty, and has not yet been committed, in which case it is in the pending batch.  That head element that was just dequeued and deleted may have been committed 5 batches ago as was read from the datastore, in which case it is not in the batch.

The count can be seen as a counter of batch "events" to commit, not how many items are in the batch. The counter only counts up. If we have n Puts and m Deletes, and do know how many are for the same items, then it seems we need to count batch events, not batch contents. When the commit happens the counter is reset to 0.

 
> There's also the case where the batch has been committed due to getting over the batchCount, so now it is empty, and when we dequeue we do need to make a direct-delete from db and I don't see this happening?

A direct delete is still needed. Consider this sequence.
1. datastore contains A, batch contains XYZ, read head CID A (making it next to dequeue)
2. receive new CIDs X Y Z
3. commit batch, datastore contains AXYZ, 
4. Dequeue head CID A, but skip delete. Datastore contains AXYZ.
5. Read next head CID from datastore, get A (making it next to dequeue).

The immediate delete between steps 4 and 5 is still needed. That's what I was trying to say in the comment:
```go
// Delete (in batch) must be committed immediately, otherwise the
// same head cid will be read from the datastore.
```

hsanjuan: > We deleted an element, but we do not know whether or not it was in the current batch without searching the batch. So we cannot decrement the count.

I was thinking the batch would return Not Found when the element deleted from it is not in it, but I see the other issues now.

Wouldn't it be simpler then to have a batch-to-add and a batch-to-delete in the forms of cid.Set then?

- Enqueue -> add to batch-to-add set. Remove from batch-to-delete-set.
- Dequeue -> read-from-db, check if in batch-to-delete. If yes, skip. If no, delete from batch to add, add to batch to delete and send on dequeue channel.

Then commit adds+deletes in a single batch based on time, size as now.

gammazero: I had considered using a `cid.Set` exactly this way, but chose not to as that would lose the ordering that the queue is making such an effort to preserve. I think the only reason the ordering is done is to ensure that items do not sit in the queue for longer than newer ones due to chance selection.

Keeping ordering means that a sequence or timestamp needs to be part of the key or value. I think we can get rid of ordering if the provide queue is alway consumed until empty, eventually. If that is not guaranteed, and there are always some CIDs in the queue, then ordering is needed. 

If a sequence number is kept in the key, as it is currently, then the batch-to-delete set cannot be a `cid.Set`, since it is necessary to know what key delete. Instead it will need to be a map of cid to key-string, where the key-string is the key that was read from the data store.

> Dequeue -> read-from-db, check if in batch-to-delete. If yes, skip. If no, delete from batch to add, add to batch to delete and send on dequeue channel.

Batch deletes do not work with maintaining ordering. When the next CID is read from the datastore, it must be immediately deleted and not batch deleted. Otherwise, we will keep reading the same next CID until the batch-to-delete is committed.

gammazero: I cannot find any good reason for ordering of provides, and the fact that CIDs are getting stored in a map after being dequeued, [here](https://github.com/ipfs/boxo/blob/main/provider/reprovider.go#L290), indicates that ordering does not really matter. This means that a `cid.Set` can be used, and the queue does not need to write a sequence into the key used to store a CID in the datastore.

gammazero: I will change to use different batches for pending-adds and pending-deletes, which will each be `cid.Set`.  The will only be turned into a datastore Batch at commit time.

As a secondary improvement, if we want:  The key with sequence number is no longer needed, and instead just the CID can be stored in the datastore (key only with no value). Will need to delete or convert old provide queue data in datastore.

hsanjuan (approved):

gammazero:
Revised with improved solution:

Input to the queue is buffered in memory. The contents of the buffer are written to the datastore when the input buffer contains batchSize items, or when the queue has been idel (no input or output) for a long enough time. CIDs to dequeue are read, in order, from the input buffer when there are none in the datastore. Otherwise they are read from the datastore.

Reading queued items from the input buffer means items can be dequeued without having to first be written to the datastore. This way CIDs can remain in memory, since the queue is not idle and the input buffer size is reduced by reading items from it. When the queue is closed, any remaining items in memory are written to the datastore.

This gives us in-memory behavior until there are too many CIDs in memory, or until items have been sitting idle in memory too long.

PR #889, opened 2025-03-20
Rename and move bitswap -> blockexchange

hsanjuan:
This does the following:

* The bulk of bitswap/bitswap bitswap/client bitswap/server is moved to `exchange/blockexchange`
* `bitswap/network` becomes `swap`. `bsnet` -> `swap/bitswap`. `httpnet` -> `swap/httpswap`.
* `bitswap/message` -> becomes `swap/message`.

Types are renamed, comments/readmes are updated.

~~TODO: compatiblity later in old `bitswap`.~~
TODO: renamings in test packages.
TODO: internal variable renamings `bsnet` -> `bitswap`. aliases etc.


hsanjuan:
Kubo and rainbow are tested to compile without any changes.

gammazero:
Maybe add what is written [here](https://github.com/ipfs/boxo/pull/889/files#diff-d355dc0639449f5ee011da5eb46ee9ec1e8149952e428c7e27ae3e2e9086adefR17) to the changelog as well?

exchange/blockexchange/README.md
```diff
@@ -0,0 +1,30 @@
+BlockExchange
+=============
+
+> An implementation of the Exchange interface
+
+# Background
+
+The BlockExchange is an implementation of the Exchange interface aimed to
```
gammazero:
```suggestion
The BlockExchange is an implementation of the Exchange interface aimed at
```

exchange/blockexchange/README.md
```diff
@@ -0,0 +1,30 @@
...
+# Formery known as Bitswap
+
+The bulk of this code was previously named "bitswap" and was part of the
+"bitswap protocol implementation". This was before HTTP block retrieval was
+introduced. The reality is that the code here was mostly guided to support the
+exchange of blocks using the bitswap multistream protocol and implement the
+Bitswap spec, but since this is no longer the only way swapping blocks and the
+Bitswap spec itself is unrelated to how we manage queues, sessions etc and all
+the work to organize the retrieval of blocks, things have been re-named
+accordingly.
```
gammazero:
```suggestion
Bitswap spec. Since this is no longer the only way of swapping blocks and since the
Bitswap spec itself is unrelated to how we manage queues, sessions, and organizing
the retrieval of blocks, things have been re-named accordingly.
```

exchange/blockexchange/client/docs/how-bitswap-works.md
```diff
@@ -1,33 +1,33 @@
-How Bitswap Works
+How BlockExchange Works
```
gammazero: Should the name of this file also change to `how-blockexchange-works.md`

gammazero (approved):

guillaumemichel (approved):
nit: we could also mark bitswap as deprecated in the [bitswap readme](https://github.com/ipfs/boxo/blob/blockexchange/bitswap/README.md) and point to `exchange/blockexchange`

hsanjuan:
Per discussion yesterday, let's give 24h for additional possible improvements in naming. We have things like:

- `blockexchange.Blockexchange`
- `blockexchange/client.Client` and `blockexchange/server.Server`
- `swap.Network`
- `swap/message.Wantlist` (pb)
- `swap/httpswap`, `swap/bitswap`

I'm generally ok. We could do:

- `blockexchange.Exchange` (or `blockexchange.FullExchange`) + `server.BlockServer` + `client.Exchange` perhaps.

I'm not 100% sure of what happens to `blockexchange/server` in general, currently it does bitswap and not whether we should plug-in http endpoint handlers in there somehow (even if we re-use the gateway's ones).

hsanjuan:
> TODO: renamings in test packages.
> TODO: internal variable renamings `bsnet` -> `bitswap`. aliases etc.

This will come as follow-up PR, as it's not something that interlaces a lot with any other work that can happen on these modules and I don't want to hold this longer than necessary.

lidel:
@hsanjuan before we make decision, mind updating https://github.com/ipfs/boxo/tree/main/examples/bitswap-transfer to show how the new names look in real user code? 

I'd like us see how this looks for end users before we give this a go (cc @2color). 

This PR means most of `bitswap/` users will stop seeing "Bitswap" and will start seeing "BlockExchange"  

I have slight worry that 
- splitting `bitswap/` into `exchange/blockexchange/` and separate top-level `swap/bitswap|httpswap` is confusing – any reason `swap` ended up on top level and not in `exchange/`?
- new term "swap" NEXT to "exchange"
  - wouldn't keeping "network" and going with something like `exchange/network`  be more intuitive?
    - OR go the other way, make "swap" a thing and use "blockswap" instead of "blockexchange"
    - replacing "bitswap" with "blockexchange" as the top interface that end user interacts with makes new name bit awkward. How about `blockswap`?

hsanjuan:
> splitting bitswap/ into exchange/blockexchange/ and separate top-level swap/bitswap|httpswap is confusing – any reason swap ended up on top level and not in exchange/?

Seems like a block-swapping network deserved it's own space, as it is not necessarily related to an exchange interface... but the reason is not too strong... (also maintaining something "swap" as top-level, since no more bitswap).

hsanjuan:
> * replacing "bitswap" with "blockexchange" as the top interface that end user interacts with makes new name bit awkward. How about `blockswap`

`blockswap.Exchange` sounds nice. In that case:

- `exchange/blockswap/message.Wanlist`
- `exchange/blockswap/network/bitswap` and `blockswap/network/httpswap`

?

lidel:
Yeah, I think `exchange/blockswap/network/bitswap` looks decent (to me, a non-native speaker), kinda keeps things together, and allows us to add other exchange interfaces (e.g. ones that operate on graphs in the future).

@gammazero thoughts?

bitswap/README.md
```diff
@@ -1,120 +1,12 @@
-go-bitswap
-==================
+Bitswap (moved to `blockexchage` and `swap`)
```
lidel, line 1: :thought_balloon:  This readme will be what current bitswap users see, so good opportunity to point people at examples that use new API:
- https://github.com/ipfs/boxo/tree/main/examples/bitswap-transfer 
- https://github.com/ipfs/rainbow/blob/main/setup_bitswap.go (once we refactor both)

(sometimes seeing a working example makes refactor easier than reading godocs)

aschmahmann:
A brief thought on the language of exchange/swap

To me exchange/swap sounds like it draws from bittorrent where it's an intentional part of the system that clients can download data faster if they're also servers (i.e. an exchange / swap of data).

However, nothing here implements this and it's not how it's being branded. Instead the code here is setup more like an http server which can setup whatever logic it wants on rate-limiting, fairness, etc. and so what's happening is more like we have a data transfer module that's doing block-based data retrieval. Would some other language such as having a path like `transfer/blocks/network/bitswap` be better?

Could also be a good opportunity to keep the clients and servers more separate (and glued together only as needed for protocols like bitswap)

hsanjuan:
Good points from @aschmahmann . While I don't find the latest proposal incorrect in itself, after sleeping on it, moving `bitswap` inside `blockexchange` and renaming it but leaving the core tree "as it is now" is perhaps too much trouble for what we get (just moving things around).

I think my original feeling was that there is a part of "bitswap" that is block-exchange-implementation related (i.e. how many providers we look for, or how we handle exchange sessions), and there is a part which is "bitswap-protocol" related (the network), and this is why I opted for making `swap` a top-level thing, perhaps under the feeling that you could use a `swap/*` without necessarily putting a full exchange on top, and thus giving it top-level module status.

To ellaborate on this idea: The `network` interface in itself is not related to the exchange interface. The exchange interface deals with getting blocks by finding where they are etc., where the `network` interface deals with getting Wantlists from a known place, with bitswap being one protocol to transfer wantlists.

So under this light, packing `network/*` inside exchange seems a bit too much. It may be that in the future we decide to have a CARExchange that is implemented with an expanded `swap/bitswap` as well, or that we expand the current `exchange` but add new `carswap` protocols.

Anyways, just trying to justify where the original choice comes from.

To build on top of Adin's, we could also simplify and do:

- `transfer/bitswap`
- `transfer/httpswap`

Or otherwise:

- `transfer/wantlist/bitswap`
- `transfer/wantlist/httpswap` -> Since the wantlist (and not the block) is the common skeleton of the protocols.

And:

- `exchange/wantlist` -> where we have a `wantlist.NewExchange(... n bitswap.Transfer...)`, which implements a blockexchange using Wantlist transfers.

Last, I want to say I'm also ok dropping this completely. This should be a meaningful improvement made to last. If we are unsure better to leave things alone. Maybe when we think about CAR transfers we get better ideas.

hsanjuan:
Latest naming idea:

- `session.Exchange` / `blockSwap.Exchange`
- `transfer/wantlist`
- `transfer/bitswap`
- `transfer/httpswap` 


Discussion:

- We don't want to break bitswap forks
- We want to re-organize things with better naming
- We will start new modules with better naming at some point
- It will wrap the existing code, but fixing the initialization and public methods to improved versions.
- At some point then we will move the bulk of the existing code.

(DO NOT DELETE BRANCH)

gammazero:
I prefer the latest naming. Seems like perhaps `transfer/wantlist` should be `exchange/wantlist`. "transfer" being ways stuff is moved around, and a "wantlist" is thing that gets moved around, or exchanged between peers.

PR #893, opened 2025-03-23
feat(gateway):dedicated error type for denylists

shobit000:
This pull request addresses [issue #591](https://github.com/ipfs/boxo/issues/591) by introducing a dedicated error type for denylists in the IPFS gateway. The changes ensure that blocked content errors are properly conveyed using a specific error type instead of relying on string-based error checking.

## Changes Made

1. Introduced BlockedError type
2. Updated isErrContentBlocked function
3. Updated webError function

### DIFF
--- a/gateway/errors.go
+++ b/gateway/errors.go
@@ -18,6 +18,9 @@ import (
 	"github.com/ipld/go-ipld-prime/schema"
 )
 
+// Configuration for HTTP status code (451 vs 410)
+var use451StatusCode = true // This should be set based on your configuration
+
 var (
 	ErrInternalServerError = NewErrorStatusCodeFromStatus(http.StatusInternalServerError)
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
@shobit000  thank you for opening this, but was this generated with AI?

Same concerns as in 
-  https://github.com/ipfs/boxo/pull/855#discussion_r1972583915 – who will produce this error? 
- https://github.com/ipfs/boxo/pull/855#discussion_r1972568949 default to 410, no need for extra flag, nobody will default to "legal reasons".

shobit000:
yeag , used  ai but will definately update you on my work on this pr

gammazero:
The information on the reason for being blocked needs to come from nopfs as it gives more context for the appropriate error to return. We need a solution that takes this into account. Closing, as this and not the way we want to solve this.

PR #897, opened 2025-03-25
ipld/unixfs/io: MaxLinks and MaxHAMTFanout for directories

hsanjuan:
This PR paves the way to support controlling and limiting the max number of links a directory can have by supporting option-based limitations on the unixfs directory implementation.

The result is that we can set a maximum number of links in Dynamic folders that will trigger conversion to HAMT, which has in turn it's own Fanout option.

This PR has a follow up at #906 .

### DIFF
--- a/ipld/unixfs/io/directory.go
+++ b/ipld/unixfs/io/directory.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"os"
+	"time"
 
 	"github.com/alecthomas/units"
 	mdag "github.com/ipfs/boxo/ipld/merkledag"
@@ -69,6 +70,57 @@ type Directory interface {
 
 	// GetCidBuilder returns the CID Builder used.
 	GetCidBuilder() cid.Builder
+
+	// setMaxLinks sets the max width of the Directory.
+	setMaxLinks(n int)
+}
+
+type directoryWithOptions interface {
+	Directory
+	// setMaxLinks sets the max width of the Directory.
+	setMaxLinks(n int)
+	setStats(os.FileMode, time.Time)
+}
+
+type DirectoryOption func(directoryWithOptions)
+
+// WithMaxLinks stablishes the max number of links allowed for a directory:
+//
+// - On Dynamic directories using a BasicDirectory, it can trigger conversion
+// to HAMT when set and exceeded. The subsequent HAMT nodes will use MaxLinks
+// as ShardWidth. Conversion can happen too based on HAMTShardingSize.
+//
+// - On Dynamic directories using a HAMTDirectory, it can trigger conversion
+// to BasicDirectory when the number of links is below MaxLinks (and
+// HAMTShardingSize allows).
+//
+// - On Basic directories, it causes an error when adding more than MaxLinks
+// children.
+//
+// - On HAMT directories, it sets the ShardWidth, otherwise DefaultShardWidth
+// is used.
+//
+// HAMT directories require this value to be a power of 2 and a multiple of 8.
+// If it is not the case, it will not be used and DefaultHAMTWidth be used
+// instead.
+func WithMaxLinks(n int) DirectoryOption {
+	return func(d directoryWithOptions) {
+		d.setMaxLinks(n)
+	}
+}
+
+// WithCidBuilder sets the CidBuilder for new Directories.
+func WithCidBuilder(cb cid.Builder) DirectoryOption {
+	return func(d directoryWithOptions) {
+		d.SetCidBuilder(cb)
+	}
+}
+
+// WithStat can be used to set the empty directory node permissions.
+func WithStat(mode os.FileMode, mtime time.Time) DirectoryOption {
+	return func(d directoryWithOptions) {
+		d.setStats(mode, mtime)
+	}
 }
 
 // TODO: Evaluate removing `dserv` from this layer and providing it in MFS.
@@ -84,6 +136,12 @@ func init() {
 	linksize.LinkSizeFunction = productionLinkSize
 }
 
+var (
+	_ directoryWithOptions = (*DynamicDirectory)(nil)
+	_ directoryWithOptions = (*BasicDirectory)(nil)
+	_ directoryWithOptions = (*HAMTDirectory)(nil)
+)
+
 // BasicDirectory is the basic implementation of `Directory`. All the entries
 // are stored in a single node.
 type BasicDirectory struct {
@@ -96,6 +154,14 @@ type BasicDirectory struct {
 	// (We maintain this value up to date even if the HAMTShardingSize is off
 	// since potentially the option could be activated on the fly.)
 	estimatedSize int
+	totalLinks    int
+
+	// opts
+	// maxNumberOfLinks. If set, can trigger conversion to HAMT directory.
+	maxLinks   int
+	cidBuilder cid.Builder
+	mode       os.FileMode
+	mtime      time.Time
 }
 
 // HAMTDirectory is the HAMT implementation of `Directory`.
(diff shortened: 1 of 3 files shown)


hsanjuan:
(This is a supporting PR to being able to set dag-width in Kubo).

ipld/unixfs/io/directory.go
```diff
@@ -104,30 +170,107 @@ type HAMTDirectory struct {
...
+
+	return dir, nil
+}
+
+// NewHAMTDirectoryFromNode creates a HAMT directory from the given node,
+// which must correspond to an existing HAMT.
+func NewHAMTDirectoryFromNode(dserv ipld.DAGService, node ipld.Node) (*HAMTDirectory, error) {
+	dir := new(HAMTDirectory)
+	dir.dserv = dserv
+	dir.sizeChange = 0
```
gammazero: Unnecessary initialization

ipld/unixfs/io/directory.go
```diff
@@ -104,30 +170,107 @@ type HAMTDirectory struct {
...
-// NewDirectory returns a Directory implemented by DynamicDirectory
-// containing a BasicDirectory that can be converted to a HAMTDirectory.
-func NewDirectory(dserv ipld.DAGService) Directory {
-	return &DynamicDirectory{newEmptyBasicDirectory(dserv)}
+// NewHAMTDirectory creates an empty HAMT directory with the given options.
+func NewHAMTDirectory(dserv ipld.DAGService, sizeChange int, opts ...DirectoryOption) (*HAMTDirectory, error) {
+	dir := new(HAMTDirectory)
+	dir.dserv = dserv
+	dir.sizeChange = sizeChange
+	dir.maxLinks = 0
```
gammazero: `dir.maxLinks` should already be initialized to 0 when `dir` was created with `new`.
```suggestion
	dir := &HAMTDirectory{
		dserv: dserv,
		sizeChange: sizeChange,
	}
```

ipld/unixfs/io/directory.go
```diff
@@ -148,23 +291,33 @@ func NewDirectoryFromNode(dserv ipld.DAGService, node ipld.Node) (Directory, err
...
-		return &DynamicDirectory{&HAMTDirectory{shard, dserv, 0}}, nil
+		return &DynamicDirectory{hamtDir}, nil
 	}
 
 	return nil, ErrNotADir
 }
 
-func (d *BasicDirectory) computeEstimatedSize() {
+func (d *BasicDirectory) setMaxLinks(n int) {
+	d.maxLinks = n
```
gammazero, line 326: Any bounds checks needed here, or in caller?

ipld/unixfs/io/directory.go
```diff
@@ -231,12 +395,20 @@ func (d *BasicDirectory) addLinkChild(ctx context.Context, name string, link *ip
 	if err != nil && err != os.ErrNotExist {
 		return err
 	}
+	if err == nil { // existed
+		d.totalLinks--
+	}
```
gammazero:
```suggestion
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			return err
		}
	} else { // existed
		d.totalLinks--
	}
```

ipld/unixfs/io/directory.go
```diff
@@ -472,14 +658,32 @@ func (d *HAMTDirectory) needsToSwitchToBasicDir(ctx context.Context, name string
...
-	// HAMTShardingSize threshold to trigger a switch.
-	return d.sizeBelowThreshold(ctx, operationSizeChange)
+	canSwitchMaxLinks := true
+	if d.maxLinks > 0 {
+		total := d.totalLinks
+		if nodeToAdd != nil && entryToRemove == nil {
+			total++
+		} else if nodeToAdd == nil && entryToRemove != nil {
+			total--
+		}
```
gammazero: Are the extra checks worth it?
```suggestion
		if nodeToAdd != nil {
			total++
		}
		if entryToRemove != nil {
			total--
		}
```

ipld/unixfs/io/directory.go
```diff
@@ -554,49 +758,58 @@ func (d *DynamicDirectory) AddChild(ctx context.Context, name string, nd ipld.No
...
 		return basicDir.AddChild(ctx, name, nd)
 	}
-	hamtDir, err = basicDir.switchToSharding(ctx)
+
+	maxLinks := DefaultShardWidth
+	// Verify that our maxLinks is usuable for ShardWidth (power of 2, multiple of 8)
+	if validShardWidth(basicDir.maxLinks) {
+		maxLinks = basicDir.maxLinks
+	} else if basicDir.maxLinks > 0 {
+		log.Warning("using DefaultShardWidth for HAMT directories. MaxLinks cannot be used as it is not a power of 2 and multiple of 8.")
```
gammazero: I think `log.Warn` is what is normally used and `log.Warning` may actually be deprecated, but I am not sure. 
```suggestion
		log.Warn("using DefaultShardWidth for HAMT directories. MaxLinks cannot be used as it is not a power of 2 and multiple of 8.")
```

ipld/unixfs/io/directory.go
```diff
@@ -608,14 +821,23 @@ func (d *DynamicDirectory) RemoveChild(ctx context.Context, name string) error {
...
+
+	basicDir.setMaxLinks(hamtDir.maxLinks - 1)
+	d.directoryWithOptions = basicDir
 	return nil
 }
+
+// validShardWidth verifies that the given number is positive, a power of 2
+// and a multiple of 8.
+func validShardWidth(n int) bool {
+	return n > 0 && (n&(n-1)) == 0 && n%8 == 0
```
gammazero: Since 8 is a power of 2, it is safe to calculate the bitwise modulus: n%m --> n&(m-1)
```suggestion
	return n > 0 && (n&(n-1)) == 0 && n&7 == 0
```

gammazero (approved):
Logic all looks good, tests too. Could only find some trivial code nits.

ipld/unixfs/io/directory.go
```diff
@@ -69,6 +70,57 @@ type Directory interface {
...
+//
+// - On Dynamic directories using a HAMTDirectory, it can trigger conversion
+// to BasicDirectory when the number of links is below MaxLinks (and
+// HAMTShardingSize allows).
+//
+// - On Basic directories, it causes an error when adding more than MaxLinks
+// children.
+//
+// - On HAMT directories, it sets the ShardWidth, otherwise DefaultShardWidth
+// is used.
```
lidel: Hm..  I wonder if we are we ok with having a single setting that overrides both, especially given the defaults have been different  for a decade? (174 basic vs 256 for hamt).

UnixFS HAMT spec draft at https://github.com/ipfs/specs/pull/331  states that fanout is arbitrary so i guess it is ok on paper.

but @2color [did some research recently](https://discuss.ipfs.tech/t/should-we-profile-cids/18507/38?u=lidel) which suggests that "almost all implementations use HAMT fanout of 256" 

So if user sets width to 1024 here, this will produce HAMT CIDs different across implementations.

What if implicit default changes in the future, and someone wants to reproduce old behavior, where Basic and HAMT used different max widths (174 and 256)?

Would it be possible to add `WithHAMTMaxFanout` which overrides `WithMaxLinks`? 

This way `WithMaxLinks` can apply to both, but if more control is necessary, `WithHAMTMaxFanout` can be passed as well.

lidel (dismissed):
Slight concern that `WithMaxLinks` alone does not allow users to hardcode current implicit defaults in this and other implementations, which is 174 or 1024 for Basic and 256 for HAMT fanout – comment inline.

ipld/unixfs/io/directory.go
hsanjuan: anything <= 0 doesn't take effect (defaults apply) and anything > 0 is valid, so I think it's fine unless I'm not thinking of something...

ipld/unixfs/io/directory.go
hsanjuan: It should be `!errors.Is(err, os.ErrNotExist)` but ok.

ipld/unixfs/io/directory.go
hsanjuan: Fair comment, I have made changes and now there are `WithMaxHAMTFanout` and `WithMaxLinks`. To simplify and avoid internal magic, MaxLinks is for the basic directory side of things and MaxHAMTFanout is for the HAMT directory part. So people can set `WithMaxLinks(1024)` and the HAMT shard size will remain at the default of 256 unless it is overriden with its own option.

hsanjuan:
Let's continue on https://github.com/ipfs/boxo/pull/906, as it has additionally touched the unixfs part a bit.

PR #898, opened 2025-03-26
feat(ipld/unixfs): DagModifier: allow specifying MaxLinks per file

hsanjuan:
cc https://github.com/ipfs/kubo/issues/10751


hsanjuan (commented):
I am not sure that we are going to make immediate use of this, but probably good to have.

Most immediate usecase would be to support MaxLinks in MFS if we wanted.

hsanjuan:
On second thoughts, we might need this since adding uses MFS..

gammazero (approved):
Is this value ever written anywhere? Is not, maybe better to just use a const so everyone knows it is a fixed read-only value. Otherwise LGTM.

lidel (approved):

lidel:
lgtm, we will expose it as configuration option, part of  https://github.com/ipfs/kubo/issues/10751

ipld/unixfs/mod/dagmodifier.go
```diff
@@ -46,6 +46,7 @@ type DagModifier struct {
 
 	Prefix    cid.Prefix
 	RawLeaves bool
+	MaxLinks  int
```
lidel, line 49: @hsanjuan is this enough, given we've added separate setting for HAMT? (https://github.com/ipfs/boxo/pull/897#pullrequestreview-2726072733)

hsanjuan: It is probably enough since this seems to deal only with files...

lidel (approved):

PR #899, opened 2025-03-26
feat(gateway): add cid copy button to directory listings 

### LINKED ISSUE #895: gateway:   more responsive directory listing
Currently, the gateway web interface shows truncated hashes, even when the monitor is large enough to fully display them.

 
*The original image of the issue*

The original proposal from @Jorropo was to display it in a responsible fashion.

I suggest to add a `copy to clipboard button` next to the hash, because in my opinion it's the thing we want to do when we have the full hash.

This is a copy of the issue [#8995](https://github.com/ipfs/kubo/issues/8995) of Kubo. But I think it is related to boxo/gateway rather than Kubo.  
Here is the message from @Jorropo:  
> I am using a 1080p monitor, and most of the space is wasted.
I expect the CID to be fully printed given that I have twice the screen space needed to do so.
This is not happening because the the bafk…vvby is actually templated server side.
I expect this field to be sent completely from the server, and do client side responsive using CSS trickery (that would allows to cut off smartly and use most of the available screen real estate).

VersaliXis:
Fixes: https://github.com/ipfs/boxo/issues/895


gammazero:
Could you upload a screenshot with this PR?

VersaliXis:
Sure! Here is what it looks like. If you want to change anything in the style, feel free to ask. 


Copy button proposal

The code is pretty simple, here is a diff:
```diff
diff --git a/gateway/assets/directory.html b/gateway/assets/directory.html
index 888ecabf..9201ce83 100644
--- a/gateway/assets/directory.html
+++ b/gateway/assets/directory.html
@@ -59,6 +59,7 @@
             <a class="ipfs-hash" translate="no" href={{ if $root.DNSLink }}"https://cid.ipfs.tech/#{{ .Hash | urlEscape}}" target="_blank" rel="noreferrer noopener"{{ else }}"{{ $root.GatewayURL }}/ipfs/{{ .Hash | urlEscape}}?filename={{ .Name | urlEscape }}"{{ end }}>
               {{- .ShortHash -}}
             </a>
+            <button class="copy-button" onclick="navigator.clipboard.writeText('{{ .Hash }}')">Copy</button>
             {{ end }}
           </div>
           <div class="nowrap" title="Cumulative size of IPFS DAG (data + metadata)">{{ .Size }}</div>
diff --git a/gateway/assets/style.css b/gateway/assets/style.css
index a8dfdaf7..7e4512e7 100644
--- a/gateway/assets/style.css
+++ b/gateway/assets/style.css
@@ -66,6 +66,23 @@ a:visited {
        font-family: var(--monospace);
 }
 
+.copy-button {
+       color: #7f8491;
+       background-color: inherit;
+       border: 1px solid var(--dark-white);;
+       border-radius: 2px;
+       padding: 1px 2px;
+       cursor: copy;
+}
+
+.copy-button:hover {
+       background-color: var(--light-white);
+}
+
+.copy-button:active {
+       color: #00b0e9;
+}
+
 #header {
        align-items: center;
        background: var(--navy);
```

gateway/assets/style.css
```diff
@@ -66,6 +66,24 @@ a:visited {
 	font-family: var(--monospace);
 }
 
+.copy-button {
+	color: #7f8491;
+	background-color: inherit;
+	border: 1px solid var(--dark-white);
+	;
```
2color:
```suggestion
```

gateway/assets/style.css
```diff
@@ -66,6 +66,24 @@ a:visited {
...
+	border-radius: 2px;
+	padding: 1px 2px;
+	cursor: copy;
```
2color: I've seen a lot of copy buttons but rarely did they have `cursor: copy`. Not that it speaks against it, but it's not that common.

gateway/assets/directory.html
```diff
@@ -59,6 +59,7 @@
             <a class="ipfs-hash" translate="no" href={{ if $root.DNSLink }}"https://cid.ipfs.tech/#{{ .Hash | urlEscape}}" target="_blank" rel="noreferrer noopener"{{ else }}"{{ $root.GatewayURL }}/ipfs/{{ .Hash | urlEscape}}?filename={{ .Name | urlEscape }}"{{ end }}>
               {{- .ShortHash -}}
             </a>
+            <button class="copy-button" onclick="navigator.clipboard.writeText('{{ .Hash }}')">Copy</button>
```
lidel: Thank you @VersaliXis, this looks useful.

Small ask is to improve visual presentation, avoid using so much space:

> Copy button proposal

Copy action is usually  indicated by an icon, and i virtually never see it being English "Copy".

To ensure this is consistent with visual language in other places, could we use [stroke_copy.svg](https://bafybeiebays3i7aem7tptga3kilzux7f5gz6czl4dietazgutcjhozjviu.ipfs.dweb.link/icons/stroke_copy.svg) from https://github.com/ipfs-shipyard/ipfs-css#icons?

> image


Put it in `gateway/assets/icons.css`, so it is reusable.

lidel (changes requested):

gateway/assets/directory.html
VersaliXis: Thanks for your advice, I added the icon in `gateway/assets/icons.css`, and added it after the hash.

gateway/assets/style.css
VersaliXis: Thanks, you are right. I removed it.

gateway/assets/directory.html
VersaliXis: On mouse hover:
image
When the user clicks:
image

Do you think this fits with the visual language of the gateway?

lidel: Thank you, yes, made small tweaks in https://github.com/ipfs/boxo/pull/899/commits/5d21609c23a6f46b6d1ae20a1eafb1faaeb0cdb9 but this should be ok (can we refined in future PRs).

lidel (approved):

gateway/assets/directory.html
VersaliXis: Thanks, it looks way better now!

PR #847, opened 2025-02-16
feat: NewDAGProvider to walk partial DAGs in offline mode

PsychoPunkSage:
Fixes https://github.com/ipfs/kubo/issues/10386  and ~[PR:kubo#10704](https://github.com/ipfs/kubo/pull/10704)~ → https://github.com/ipfs/kubo/pull/10754. Refer [this](https://github.com/ipfs/kubo/pull/10704#pullrequestreview-2619414364)

## Problem
* Reprovider strategy currently ignores MFS content
* Each provider implementation has its own DAG traversal logic
* Need a generic way to announce content from any DAG structure

* Takes a root CID and traverses its entire DAG
* Uses existing fetcherhelpers.BlockAll for efficient traversal
* Works with any DAG structure (MFS, IPLD, etc.)
* Integrates with existing provider prioritization

PR #905, opened 2025-04-02
feat: NewDAGProvider to walk partial DAGs in offline mode

lidel:
- This PR makes MFS provider described in https://github.com/ipfs/kubo/issues/10386 possible.
- Kubo PR that uses this PR: https://github.com/ipfs/kubo/pull/10754

More context:

This PR is local continuation of https://github.com/ipfs/boxo/pull/847 (in actuality, it was  @hsanjuan  rewriting most of original code) allowing for use in https://github.com/ipfs/kubo/pull/10754 without pointing at a fork.


fetcher/impl/blockservice/fetcher.go
```diff
@@ -140,6 +146,11 @@ func blockOpener(ctx context.Context, bs *blockservice.Session) ipld.BlockReadOp
 
 		blk, err := bs.GetBlock(ctx, cidLink.Cid)
 		if err != nil {
+			if format.IsNotFound(err) && offline {
+				fmt.Println("skip error triggered", cidLink.Cid, err)
+				// fmt.Println(string(debug.Stack()))
```
gammazero:
```suggestion
```

lidel (commented):
lgtm, tested locally with https://github.com/ipfs/kubo/pull/10754 and expected CIDs are provided as expected

the `skip` "ERROR" was moved to debug log level as it did not seem to block root from being correctly provided – https://github.com/ipfs/kubo/pull/10754#issuecomment-2788057458

PR #906, opened 2025-04-07
feat(unixfs/mfs): support MaxLinks and MaxHAMTFanout

hsanjuan:
This exposes MaxLinks and MAXHAMTFanout options in unixfs directories and MFS.

The idea is to give Kubo the chance to configure these two options when adding.

For this to work it has:

- Made the necessary changes to unixfs/io directories. Since basic and HAMT directories can be switched to each other as part of a dynamic directory, we have added the options to each. We have added methods to create new directories as well, with functional options. This is meant to simplify previous usage, where a merkledag-pb folder is created, then wrapped in the unixfs directory implementation, the modified with the right CidBuilder etc... this now can happen in one go with the `New...()` methods.

- Made the necessary changes to MFS. Unlike mtime and mode, MaxLinks is not serialized in the dag-pb and thus cannot be remembered when writing and reloading a unixfs node (sync/cache refresh etc)... this means we need to set the MaxLinks and MaxHAMTFanout options from the parent in such cases. In such cases, the directories are created from an existing node that is wrapped, so we had to add Setters for the options. To avoid causing side-effects when reading, the Setters do not trigger HAMT conversions etc. even if the loaded folder does not comply with MaxLinks. The new options have been made part of `MkdirOpts` where they fit well.

- MFS additionally has received a `NewEmptyDirectory` and `NewEmptyRoot` methods, which allow setting options for empty directories/roots.

All this has been primarily wired into Kubo at   https://github.com/ipfs/kubo/pull/10774, and it seems to work.


hsanjuan:
Kubo follow up is https://github.com/ipfs/kubo/pull/10774

And I see it breaks mtime stuff in sharness tests so will review again.

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `ipld/unixfs/io/directory`: We have made some changes to unixfs directory tooling:
+  - We have exposed creator methods for new `BasicDirectory` and `HAMTDirectory`, that complement the existing `NewDirectory()` which creates dynamic directories.
+  - We have added `WithCidBuilder(...)` and `WithMaxLinks(...)` as options to these new methods so that empty directories can be initilized as wished from the get-go.
+  - `WithMaxLinks(...)` is a new option that allows to set a limit to the number of children that a directory DAG node can have. For details on what exactly it does for each of the directory types, please check the documentation.
+  - `WithStats(...)` is a new option that allows to set the Stat information for the new directory node.
+
```
hsanjuan: Outdated

hsanjuan:
Kubo is now happy

hsanjuan:
This is probably missing MFS tests that verify that the MaxLinks and MaxHAMTFanout are working.

There's a gray area when creating a directory with `MkdirWithOptions` from an existing directory and setting MaxLinks to something different that what the parent directory is using. When the child directory is written to disk and refreshed, it will probably receive the MaxLinks setting from the parent directory.

There is no nice way of dealing with this I think, given we cannot serialize this option, and normal usage means all folders in an mfs dag should be getting the same configuration.

gammazero (approved):

mfs/root.go
```diff
@@ -136,6 +136,33 @@ func NewRoot(parent context.Context, ds ipld.DAGService, node *dag.ProtoNode, pf
 	return root, nil
 }
 
+// NewEmptyRoot creates an empty Root directory with the given directory
+// options. A replublisher is created if PubFunc is not nil.
```
lidel:
```suggestion
// options. A republisher is created if PubFunc is not nil.
```

lidel (approved):
Added end-to-end tests to `./test/cli/add_test.go`  in https://github.com/ipfs/kubo/pull/10774 and found no issues, works as expected.

Merging, we will include it in next release, and dogfood in Kubo 0.35.0-rc1

Issue #901, opened 2025-03-27
provider: duplicated CIDs sent to provide queue


gammazero:
> part of https://github.com/ipshipyard/roadmaps/issues/6

The provide queue is getting many duplicate CIDs. It turns out that normal `ipfs add` and `mfs` are both adding blocks to the `blockservice`, and causing duplicated CIDs to be provided.

This may be tolerable since the CIDs are deduplicated after being read from the queue, [here](https://github.com/ipfs/boxo/blob/main/provider/reprovider.go#L290), but it does mean the queue has to hold double the CIDs that are actually provided.

It cases where things are added in such a way that follows both call paths, it may be worth passing some parameter that suppresses providing in one of the paths.

Call stack from normal ipfs-add path:
```
github.com/ipfs/boxo/provider.(*reprovider).Provide(0x14001585080?, {0x10535d940?, 0x14001589e90?}, {{0x14000635ef0?, 0x1?}}, 0x1?)
        github.com/ipfs/boxo@v0.29.2-0.20250326002258-acb85b96a907/provider/reprovider.go:464 +0x2c
github.com/ipfs/boxo/exchange/providing.(*Exchange).NotifyNewBlocks(0x14000d4e6a0, {0x10535d940, 0x14001589e90}, {0x140015ba5a0, 0x1, 0x1?})
        github.com/ipfs/boxo@v0.29.2-0.20250326002258-acb85b96a907/exchange/providing/providing.go:41 +0xbc
github.com/ipfs/boxo/blockservice.(*blockService).AddBlocks(0x1400158cc40, {0x10535d978, 0x14000b67b80}, {0x140015ba5a0, 0x1, 0x1})
        github.com/ipfs/boxo@v0.29.2-0.20250326002258-acb85b96a907/blockservice/blockservice.go:223 +0x3ac
github.com/ipfs/boxo/ipld/merkledag.(*dagService).AddMany(0x14000b3f5a8, {0x10535d978, 0x14000b67b80}, {0x140015ba580, 0x1, 0x10327cdd4?})
        github.com/ipfs/boxo@v0.29.2-0.20250326002258-acb85b96a907/ipld/merkledag/merkledag.go:71 +0xe4
github.com/ipfs/go-ipld-format.(*Batch).asyncCommit.func1(...)
        github.com/ipfs/go-ipld-format@v0.6.0/batch.go:100
created by github.com/ipfs/go-ipld-format.(*Batch).asyncCommit in goroutine 380
[...]

PR #907, opened 2025-04-09
provider: dedicated provide queue

guillaumemichel:
Reprovides are currently spiky since all keys are reprovided at once (for both normal and accelerated DHT clients). If new keys are added during a reprovide, the `reprovider` will wait until the current reprovide is finished before advertising the newly added keys to the DHT. So new keys are blocked by reprovides.

This PR splits the (re)provide queue in a provide queue and a reprovide queue, allowing provides to be advertised immediately, even during reprovides.

https://github.com/ipfs/kubo/issues/10777

Depending on either:
- [x] https://github.com/ipfs/boxo/pull/910
- [x] https://github.com/ipfs/boxo/issues/901

> - Part of https://github.com/ipshipyard/roadmaps/issues/6

### DIFF
--- a/provider/reprovider.go
+++ b/provider/reprovider.go
@@ -59,8 +59,8 @@ type reprovider struct {
 	q  *queue.Queue
 	ds datastore.Batching
 
-	maxBatchSize       uint
-	provideWorkerCount uint
+	maxReprovideBatchSize uint
+	provideWorkerCount    uint
 
 	statLk                                      sync.Mutex
 	totalReprovides, lastReprovideBatchSize     uint64
@@ -109,11 +109,11 @@ var (
 // If provider casts to [Ready], it will wait until [Ready.Ready] is true.
 func New(ds datastore.Batching, opts ...Option) (System, error) {
 	s := &reprovider{
-		allowlist:          verifcid.DefaultAllowlist,
-		reprovideInterval:  DefaultReproviderInterval,
-		maxBatchSize:       math.MaxUint,
-		provideWorkerCount: defaultProvideWorkerCount,
-		keyPrefix:          DefaultKeyPrefix,
+		allowlist:             verifcid.DefaultAllowlist,
+		reprovideInterval:     DefaultReproviderInterval,
+		maxReprovideBatchSize: math.MaxUint,
+		provideWorkerCount:    defaultProvideWorkerCount,
+		keyPrefix:             DefaultKeyPrefix,
 	}
 
 	var err error
@@ -148,7 +148,7 @@ func New(ds datastore.Batching, opts ...Option) (System, error) {
 
 	if s.rsys != nil {
 		if _, ok := s.rsys.(ProvideMany); !ok {
-			s.maxBatchSize = 1
+			s.maxReprovideBatchSize = 1
 		}
 
 		s.run()
@@ -213,7 +213,7 @@ func ProvideWorkerCount(n uint) Option {
 // limit memory usage spike.
 func MaxBatchSize(n uint) Option {
 	return func(system *reprovider) error {
-		system.maxBatchSize = n
+		system.maxReprovideBatchSize = n
 		return nil
 	}
 }
@@ -422,7 +422,7 @@ func (s *reprovider) Reprovide(ctx context.Context) error {
 		return err
 	}
 
-	batchSize := s.maxBatchSize
+	batchSize := s.maxReprovideBatchSize
 	if s.throughputCallback != nil && s.throughputMinimumProvides < batchSize {
 		batchSize = s.throughputMinimumProvides
 	}
--- a/provider/reprovider.go
+++ b/provider/reprovider.go
@@ -197,11 +197,11 @@ func DatastorePrefix(k datastore.Key) Option {
 // which can speed up provide operations under heavy load. Use this option
 // carefully, as it may cause the libp2p node to open a very high number of
 // connections to remote peers.
-//
-// This setting has no effect when batch providing is enabled.
-func ProvideWorkerCount(n uint) Option {
+func ProvideWorkerCount(n int) Option {
 	return func(system *reprovider) error {
-		system.provideWorkerCount = n
+		if n > 0 {
+			system.provideWorkerCount = uint(n)
+		}
 		return nil
 	}
 }
--- a/provider/reprovider.go
+++ b/provider/reprovider.go
@@ -39,7 +39,7 @@ const (
 	// MAGIC: default number of provide operations that can run concurrently.
 	// Note that each provide operation typically opens at least
 	// `replication_factor` connections to remote peers.
-	defaultProvideWorkerCount = 1 << 7
+	defaultProvideWorkerCount = 64
 
 	// MAGIC: Maximum duration during which no workers are available to provide a
 	// cid before a warning is triggered.
@@ -68,7 +68,6 @@ type reprovider struct {
 	ds datastore.Batching
 
 	maxBatchSize       uint
-	batchProvides      bool
 	provideWorkerCount uint
 
 	statLk                                      sync.Mutex
@@ -215,30 +214,6 @@ func ProvideWorkerCount(n uint) Option {
 	}
 }
 
-// BatchProvides enables batching of the provide operations for multiple CIDs.
-// When enabled, if several CIDs are added in rapid succession (i.e., if the
-// time gap between consecutive CIDs is less than the
-// `pauseDetectionThreshold`), they are aggregated and provided together as a
-// single batch. When the accelerated DHt client is enabled, it helps reduce
-// the number of connections the libp2p node must open when advertising the
-// CIDs.
-//
-// Additionally, the batching mechanism will trigger a provide operation after
-// `maxCollectionDuration` has elapsed, even if there hasn't been a gap of
-// `pauseDetectionThreshold` between consecutive CIDs.
-//
-// Only one batch can be provided at a time. Cids added during a provide
-// operation will not be advertised before the current operation is finished.
-//
-// This option provides a resource efficient alternative to instant provides,
-// but user must accept to wait before the initial provide.
-func BatchProvides(enable bool) Option {
-	return func(system *reprovider) error {
-		system.batchProvides = enable
-		return nil
-	}
-}
-
 // MaxBatchSize limits how big each batch is.
 //
 // Some content routers like acceleratedDHTClient have sub linear scalling and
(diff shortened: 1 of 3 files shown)


provider/reprovider.go
```diff
@@ -284,6 +284,15 @@ func (s *reprovider) run() {
...
 				}
 
+				select {
+				case c := <-provCh:
+					// Prioritize newly provided keys
+					resetTimersAfterReceivingProvide()
+					m[c] = struct{}{}
+					continue
+				default:
+				}
```
aschmahmann: Does this fix the proposed issue? Once you actually trigger a ProvideMany the loop here is blocked so how much this helps is gated by the batch size (which seems to default to max uint), we don't even have the maxCollectionDuractionTimer to fall back on.

The underlying problem is that a background process (reproviding in bulk) is being mixed in a single queue with a foreground process (immediate provides).

While there are some advantages to the mixing of the queues (e.g. you can save resources if you happen to be able to add your immediate provides into the background provide in the near future) the complexity probably isn't worthwhile.

Maybe instead we could have two queues:
1. Background reprovide
2. Foreground provides that get batched up

guillaumemichel: Agreed! Adapted the PR

gammazero (commented):
Provide collects batches of CID before providing them. This means there may be some delay when filling the batch, especially if each new CID arrives just before the pause detection timer expired. It seems like it might be better to send new provides immediately.

One problem with doing that is that the dublicate CIDs would not be removed, because the batch is currently a map of CIDs. Consider keeping the map to ignore duplicates, but publishing each unique CID immediately. There may be some additional logic needed to keep the map from growing too large, which may mean making it an LRU cache.

Also, we may need a test to show that providing does not need to wait for reprovide.

Otherwise, this all looks good.

guillaumemichel:
Added instant provides but I didn't add tests yet.

I have set instant provides as default with a "safe (?)" default of 128 concurrent provides, batched provides can still be used if the right option is provided. I have set a warning if the queue of cids to be provided instantly isn't processed fast enough. I have set instant provides as the default because before https://github.com/ipfs/boxo/pull/641, all cids were instantly provided, so the behavior is expected to be similar to what it used to be.

It adds quite some complexity :(

In any case, the provide operation is resource intensive, each provide triggering sending a message to 20 remote peers. With batching and the accelerated DHT client, we can group the messages we send to a remote peer to avoid opening a closing a connection many times to the same peer. But it comes at the expense of timing.

Instant provides are "instant" but the node is expected to open and close more connections, even though the same payload is sent to the same peers.

Some open questions:
1. Should we get rid of the batched providing altogether? It may seem great for nodes with limited networking capabilities, but it requires to run the accelerated DHT client (which is resource intensive), to be efficient?
2. If we keep batched providing, should we add a configurable number of workers so that the new provides don't have to wait for the previous batch provide to be finished before being provided?
3. I think it is a safe behavior to bound the default number of provide workers, but it is hard to come with the right number. I expect accelerated DHT client provides to take ~1s, which means that the number of workers is approx the number of cids/second that the system can provide.

@gammazero let me know if you see anything that can be simplified further

gammazero:
> 1. Should we get rid of the batched providing altogether? It may seem great for nodes with limited networking capabilities, but it requires to run the accelerated DHT client (which is resource intensive), to be efficient?

If using a worker pool, I do not see a need for batched provides. The size of the worker pool will limit the number of CIDs read into memory and processed.

For lower-priority reprovides, batching could be used to limit the number of CIDs in memory, but this limitation can also be done using a worker pool. It would probably be good to have a smaller worker pool for reprovides than is used for provides. Maybe even limit the reprovide pool size to one or two.

Batching used to be the way CIDs were deduplicated. This has been replaced by the deduplication cache in this PR, so there really does not appear to be any utility in batching. Once #901 is fixed, there will also be no need for deduplication cache. 

> 2. If we keep batched providing, should we add a configurable number of workers so that the new provides don't have to wait for the previous batch provide to be finished before being provided?

If we have a limited size worker pool, there does not seem to be a need for a batch, since the size of the pool limits the number of CIDs that get loaded into memory for (re)providing. The pool size should be adjustable, and a smaller pool should be used for reprovides by default.

> 3. I think it is a safe behavior to bound the default number of provide workers, but it is hard to come with the right number. I expect accelerated DHT client provides to take ~1s, which means that the number of workers is approx the number of cids/second that the system can provide.

If provide sends each CID to 20 remote peers (requires 20 connections) and the worker pool has 50 workers, then that is using 1000 connections. Perhaps the pool size can be informed by the maximum (or a reasonable) number of peer connections, leaving enough capacity to handle incoming peer connections. Does resource manager give us some clue?

provider/reprovider.go
```diff
@@ -236,208 +302,223 @@ func initialReprovideDelay(duration time.Duration) Option {
...
+			defer provideDelayTimer.Stop()
+			select {
+			case provideQueue <- c:
+				if lateOnProvides {
+					log.Warn("New provides are being processed again")
+				}
+				lateOnProvides = false
+			case <-provideDelayTimer.C:
+				if !lateOnProvides {
+					log.Warn("New provides are piling up in the queue, consider increasing the number of provide workers or enabling Batch Providing.")
```
gammazero: Instead of just logging a warning, this may be good place to increase the number of workers in the pool. That implies the pool has a starting size, and a maximum size. I am not sure it is worth the additional complexity to have a dynamically sized pool.

guillaumemichel:
The main benefit of batching, is that it reduces the number of connections we need to open. Basically it will 
1. select all the keys from the batch that should be allocated to a `DhtServer`
2. open the connection to `DhtServer`
3. send all provides to `DhtServer` (`PUT` RPC)
4. close the connection to `DhtServer`
5. Repeat with the next peer

Where as if there is no batching, the provider will open a connection to the 20 closest nodes to a cid, send them the `PUT` request, and continue with the next cid. As a result the node will dial the same nodes many times in a short while if the amount of provided cids is large enough.

Hence batching helps reduce the number of dials. Note that batch providing only works the the accelerated DHT client.

---

> Perhaps the pool size can be informed by the maximum (or a reasonable) number of peer connections, leaving enough capacity to handle incoming peer connections

This is tricky, because the `provider` has no visibility on the number of (concurrent) connections opened by the DHT. It is true that for each provide the libp2p node will try to open at least 20 connections. However, the normal DHT client limits the number of concurrent connection per request to [10](https://github.com/libp2p/go-libp2p-kad-dht/blob/31c361257e16379b6dee2fc2981f75ea0935a102/amino/defaults.go#L28-L32), and the accelerated DHT client has [20 workers](https://github.com/libp2p/go-libp2p-kad-dht/blob/31c361257e16379b6dee2fc2981f75ea0935a102/fullrt/dht.go#L137). Both default parameters can be overwritten by implementations.

The libp2p resource manager will probably prevent us from opening too many connections (if enabled).

provider/reprovider.go
```diff
@@ -236,208 +302,223 @@ func initialReprovideDelay(duration time.Duration) Option {
...
-			s.throughputDurationSum += dur
-			s.throughputProvideCurrentCount += uint(len(keys))
-			if s.throughputCallback != nil && s.throughputProvideCurrentCount >= s.throughputMinimumProvides {
-				if more := s.throughputCallback(performedReprovide, complete, s.throughputProvideCurrentCount, s.throughputDurationSum); !more {
-					s.throughputCallback = nil
-				}
-				s.throughputProvideCurrentCount = 0
-				s.throughputDurationSum = 0
+			// Deduplicate recently provided cids
+			if found, _ := s.instantProvideDeduplicator.ContainsOrAdd(c, struct{}{}); found {
```
gammazero: Consider moving this into the provide queue (`s.q`). Doing that will prevent duplicate CIDs from having to sit in the queue. The  queue's internal buffer can be replaced with this LRU cache, and there is no need to make any other logic changes.

provider/reprovider.go
```diff
@@ -236,208 +302,223 @@ func initialReprovideDelay(duration time.Duration) Option {
...
 		}
+	}
+}
 
-		// after the first reprovide, schedule periodical reprovides
-		nextReprovideTicker := time.NewTicker(s.reprovideInterval)
+func (s *reprovider) run() {
+	s.closewg.Add(1)
+	if s.batchProvides {
+		go s.batchProvideWorker()
```
gammazero: As noted in overall comments in the PR, I do not think there is any remaining purpose for batched provides. Better to adjust the worker pool size.

provider/reprovider.go
```diff
@@ -236,208 +302,223 @@ func initialReprovideDelay(duration time.Duration) Option {
...
+	if s.reprovideInterval > 0 {
+		s.closewg.Add(1)
+		go s.reprovideSchedulingWorker()
+	}
+}
+
+func (s *reprovider) waitUntilProvideSystemReady() {
+	if r, ok := s.rsys.(Ready); ok {
+		ticker := time.NewTicker(time.Minute)
+		for !r.Ready() {
```
gammazero: Consider not creating the ticker until after the first ready check fails.
```suggestion
		var ticker time.Ticker
		for !r.Ready() {
			if ticker == nil {
				ticker = time.NewTicker(time.Minute)
				defer ticker.Stop()
			}
```

provider/reprovider.go
```diff
@@ -236,208 +302,223 @@ func initialReprovideDelay(duration time.Duration) Option {
...
+			log.Debugf("reprovider system not ready")
 			select {
-			case <-nextReprovideTicker.C:
+			case <-ticker.C:
 			case <-s.ctx.Done():
 				return
 			}
 		}
-	}()
+		ticker.Stop()
```
gammazero: Make this a deferred call after creating the ticker.

gammazero (commented):
See comments. I think we can get rid of batched provides.

gammazero:
> Where as if there is no batching ...
OK, so a little more than simple batching to make use of connections more efficient.

provider/reprovider.go
guillaumemichel: The goal is to limit resource usage by default, and avoid growing to an unbounded number of workers. If user is fine with an unbounded number of workers and wants performance they can set `provideWorkerCount` to `0`, which is equivalent to increasing the number of workers as needed.

But maybe there is a better way to signal the user that they hit the limit? e.g user defined callback

provider/reprovider.go
guillaumemichel: done in https://github.com/ipfs/boxo/pull/909

provider/reprovider.go
guillaumemichel: I will remove batched provides as discussed during triage.

provider/reprovider.go
gammazero: I was thinking of only growing worker pool up to its max size, not having an unbounded size. If worker pool if a fixed size, then his does not matter.

guillaumemichel:
As observed in https://github.com/ipfs/boxo/issues/901, CIDs are added multiple times to the queue. Deduplication mechanisms were removed in this PR. So each CID may be advertised up to 3 times.

A quick fix is https://github.com/ipfs/boxo/pull/910, but on the long run we don't need it if we can solve https://github.com/ipfs/boxo/issues/901.

provider/reprovider.go
```diff
@@ -228,216 +247,143 @@ func Online(rsys Provide) Option {
...
-		system.initialReprovideDelaySet = true
-		system.initalReprovideDelay = duration
-		return nil
-	}
-}
-
-func (s *reprovider) run() {
+func (s *reprovider) provideWorker() {
+	defer s.closewg.Done()
 	provCh := s.q.Dequeue()
```
gammazero, line 263: Please move this declaration closer to where the variable is used (like line 308).

provider/reprovider.go
```diff
@@ -228,216 +247,143 @@ func Online(rsys Provide) Option {
...
+	}
 
-			log.Debugf("starting provide of %d keys", len(keys))
-			start := time.Now()
-			err := doProvideMany(s.ctx, s.rsys, keys)
-			if err != nil {
-				log.Debugf("providing failed %v", err)
+	for {
+		select {
+		case c := <-provCh:
```
gammazero: Need to check if channel closed, because of [this](https://github.com/ipfs/boxo/blob/main/provider/internal/queue/queue.go#L87).
```suggestion
		case c, ok := <-provCh:
			if !ok {
				return
			}
```

provider/reprovider.go
```diff
@@ -228,216 +247,143 @@ func Online(rsys Provide) Option {
...
-
-			if performedReprovide {
-				s.lastReprovideBatchSize = uint64(len(keys))
-				s.lastReprovideDuration = dur
-				s.lastRun = time.Now()
-
-				s.statLk.Unlock()
-				// Don't hold the lock while writing to disk, consumers don't need to wait on IO to read thoses fields.
+			provideOperation(s.ctx, c)
+		case <-s.ctx.Done():
```
gammazero: If `provCh` closes when `s.q.Close` is called (called from `Close`), then waiting for the context is not needed, and `provCh` can be ranged over instead of using `select`:

```go
for c := range provCh {
	if err := verifcid.ValidateCid(s.allowlist, c); err != nil {
		log.Errorf("insecure hash in reprovider, %s (%s)", c, err)
		continue
	}
	provideOperation(s.ctx, c)
}
```

provider/reprovider.go
```diff
@@ -228,216 +247,143 @@ func Online(rsys Provide) Option {
...
+		// Start provide workers
+		for range s.provideWorkerCount {
+			go func(ctx context.Context) {
+				for {
 					select {
-					case <-ticker.C:
-					case <-s.ctx.Done():
+					case c := <-provideQueue:
+						provideFunc(ctx, c)
+					case <-ctx.Done():
```
gammazero: Instead of selecting `<-ctx.Done()` consider closing `provideQueue` when `provideWorker` exits. That would allow ranging over provideQueue instead.
```go
		for range s.provideWorkerCount {
			go func(ctx context.Context) {
				for c := range provideQueue {
					provideFunc(ctx, c)
				}
			}(s.ctx)
		}
```

gammazero (approved):
LGTM

PR #909, opened 2025-04-15
provider: depuplicate cids in provider queue

guillaumemichel:
Change the queue internal buffer from a `Deque` to a `lru.Cache` to deduplicate cids that are added to the provider queue.


guillaumemichel:
LRU caches cannot be used as a drop in replacement for the internal buffer, since for the purpose of deduplication they should remember recent CIDs that aren't in the queue anymore.

Closing in favor of https://github.com/ipfs/boxo/pull/910

PR #911, opened 2025-04-17
feat(bitswap): add option to disable Bitswap server 

gystemd:
Follow-up from [feat(config): ability to disable Bitswap fully or just server #10782](https://github.com/ipfs/kubo/pull/10782)
This pull request introduces the ability to enable or disable the server functionality in the `Bitswap` component, along with associated updates to the codebase and new tests. The changes ensure that `Bitswap` can operate in a client-only mode when the server is disabled, while maintaining backward compatibility.

### Core functionality changes:
* Added a `serverEnabled` field to the `Bitswap` struct, defaulting to `true`, and updated the `New` function to conditionally initialize the server based on this flag.
* Modified methods such as `Stat`, `NotifyNewBlocks`, `PeerConnected`, and `PeerDisconnected` to handle cases where the server is disabled, ensuring no server-related operations are performed in this mode. 

### New configuration option:
* Introduced a `WithServerEnabled` option to allow users to explicitly enable or disable the server when creating a `Bitswap` instance. 

### Testing enhancements:
* Added a new test, `TestWithServerDisabled`, to verify that `Bitswap` operates correctly in client-only mode, including requesting blocks, handling `Stat` output, and ensuring no responses to incoming requests. 
* Added `TestServerDisabledNotifyNewBlocks` to confirm that the `NotifyNewBlocks` method works as expected when the server is disabled.

### DIFF
--- a/bitswap/bitswap.go
+++ b/bitswap/bitswap.go
@@ -96,7 +96,7 @@ func New(ctx context.Context, net network.BitSwapNetwork, providerFinder routing
 }
 
 func (bs *Bitswap) NotifyNewBlocks(ctx context.Context, blks ...blocks.Block) error {
-	if bs.serverEnabled {
+	if bs.Server != nil {
 		return multierr.Combine(
 			bs.Client.NotifyNewBlocks(ctx, blks...),
 			bs.Server.NotifyNewBlocks(ctx, blks...),
--- a/bitswap/bitswap.go
+++ b/bitswap/bitswap.go
@@ -134,15 +134,12 @@ func (bs *Bitswap) Stat() (*Stat, error) {
 		// Server stats will be added conditionally
 	}
 
-	// Check if the server is enabled/initialized
-	if bs.serverEnabled {
-		// Server is enabled, get its stats
+	// Stats only available if server is enabled
+	if bs.Server != nil {
 		ss, err := bs.Server.Stat()
 		if err != nil {
-			// Return client stats along with the server error, or handle differently
 			return stat, fmt.Errorf("failed to get server stats: %w", err)
 		}
-		// Populate server-specific fields
 		stat.Peers = ss.Peers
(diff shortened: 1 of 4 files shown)


bitswap/bitswap.go
```diff
@@ -115,46 +122,60 @@ func (bs *Bitswap) Stat() (*Stat, error) {
...
+	}
 	return nil
 }
 
 func (bs *Bitswap) WantlistForPeer(p peer.ID) []cid.Cid {
 	if p == bs.net.Self() {
 		return bs.Client.GetWantlist()
 	}
-	return bs.Server.WantlistForPeer(p)
+	return nil
```
gammazero, line 167: Why is this returning `nil` and not:
```go
if bs.Server != nil {
	return bs.Server.WantlistForPeer(p)
}
return nil
```

bitswap/bitswap.go
```diff
@@ -51,15 +51,17 @@ type Bitswap struct {
...
-	net    network.BitSwapNetwork
+	tracer        tracer.Tracer
+	net           network.BitSwapNetwork
+	serverEnabled bool
 }
 
 func New(ctx context.Context, net network.BitSwapNetwork, providerFinder routing.ContentDiscovery, bstore blockstore.Blockstore, options ...Option) *Bitswap {
 	bs := &Bitswap{
 		net: net,
 	}
```
gammazero, line 63:
```suggestion
		serverEnabled: true,
	}
```

bitswap/bitswap.go
```diff
@@ -51,15 +51,17 @@ type Bitswap struct {
...
-
+	// by default, iniitalize the server
+	bs.serverEnabled = true
```
gammazero:
```suggestion

```

gammazero (approved):

Issue #918, opened 2025-05-13
bitswap/httpnet: Disconnect from peers eventually


hsanjuan:
This was mentioned but I don't think it is tracked in boxo so opening this for tracking.

Currently, when an http-provider record is found for something, we open a connection to the endpoint and request what we need to request.

Unless things are 500-ing or giving any sort of that ends up as a "fatal" error, we don't disconnect and keep "peered" to the endpoints, which start getting optimistic-bitswap requests for everything.

State of things:

- A MsgSender is a peer ID with one several http urls attached (from provider records). Requests are sent to this urls sorted by serverErrors, cooldown periods.
- Each url tracks a count of "serverErrors". When all urls have exhausted their max serverErrors, bestURL() returns an error which is converted into a fatalError, which triggers a disconnect from the peer.
- Currently, clientErrors like NotFound, Gone, Forbidden and redirects do not cause disconnects regardless of how many times they happen.

Desired goal:

- We should stop making optimistic http requests to http endpoints at some point, while ensuring we keep 

Options:

- We could not make optimistic requests at all, and only make them based on a provider-record basis. The connections will die per their idle timeouts. Implementation wise this is ugly because at the bitswap-network layer we have no information on whether a peer and a CID are related by a provider-record and we just get wantlists that need to be sent to a peer. We could track cid->peer relations by augmenting the Connect() method, which gets called directly from the provider query manager.
- We could set a per-peer timeout which gets reset on a successful request and otherwise disconnects peers after a while.
- We could set a per-error timeout which disconnects peers when a global clientError threshold is crossed.

I think the simplest approach is the latter (per-error):

- Easy to implement
- On a node with a lot of activity making requests to and endpoint that doesn't have anything  we request, we ensure we don't make too many requests for no reason.
- On a node with very low activity making a few requests to an endpoint, we let the http idle timeouts do their work or otherwise keep peered.
- Note that "Disconnecting" means removing HTTP  addresses associated to a peer, stop pinging the peer and telling bitswap to not send anymore wantlists to the peer. It does not immediately stop requests, there might be wantlists in process of being sent that will still be sent. However, given average wantlist size of ~2 during normal conditions, this means not meany requests will be leaked after crossing the limit.
- Downsides: if the error limit is 100, a peer with 1% success rate will still be connected forever. Down the line we probably want to introduce configuration options per peer ID, so that we can effectively configure thresholds per endpoint.

cc https://github.com/ipshipyard/roadmaps/issues/9

[...]

PR #919, opened 2025-05-15
bitswap/httpnet: Disconnect peers after client errors

hsanjuan:
This adds a configuration option to disconnect from peers when they return too many errors of type client (404s, Gone, Forbidden, Moved, Redirect. etc). in a row.

Such peers with low hit rates on the blocks we ask for must not perpetually receive optimistic requests and are "disconnected" accordingly if they cross the threshold (100 by default). They will be recontacted if a record pointing to them appears.

The implementation logs the errors at the end of processing the responses for a wantlist (that is, once per wantlist), the tracking only locks the map on connect/disconnect events, allowing for low lock contention operations.

cc #918

### DIFF
--- a/bitswap/network/httpnet/error_tracker.go
+++ b/bitswap/network/httpnet/error_tracker.go
@@ -3,7 +3,6 @@ package httpnet
 import (
 	"errors"
 	"sync"
-	"sync/atomic"
 
 	"github.com/libp2p/go-libp2p/core/peer"
 )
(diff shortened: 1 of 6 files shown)


bitswap/network/httpnet/error_tracker.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+var errThresholdCrossed = errors.New("the peer crossed the error threshold")
+
+type errorTracker struct {
+	ht *Network
+
+	mux    sync.RWMutex
+	errors map[peer.ID]*atomic.Uint64
```
hsanjuan: So this allows making updates to the counters without locking the map for writing (just for reading). The idea being that a counter that is updated or reset very often during normal operation should cause as little contention as possible on the map.

Does this make sense at all?

I don't know if it might be worth testing perf. A `map[peer.ID]int` has a write lock for every operation. This other map, however, needs to allocate memory for every entry, and creates more trash that needs to be GC'ed eventually. Also reads will be blocked to allow writes to proceed when they are waiting, and writes happen on connect/disconnect, so it may be that contention is there in any case and in the end it doesn't make much difference.

agca52: Bu durumda yapılması gereken ne ve feragat edildi diyip sistem atıyor destek

gammazero: This does makes sense, but probably is not worth the extra cost of the read/write because the section of code over which the read-lock is held is very short (a map read, a few comparisons, a map write).

gammazero (dismissed):
See comment

gammazero (approved):

PR #921, opened 2025-05-21
refactor: use a more efficient querying method

pullmerge:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.


hsanjuan:
User has added similar PRs to other projects, probably fishing for airdrops.

benma:
@hsanjuan do you know of an active airdrop that is causing all these PRs?

bitswap/server/internal/decision/engine_test.go
```diff
@@ -172,12 +173,7 @@ func TestPeerIsAddedToPeersWhenMessageSent(t *testing.T) {
 }
 
 func peerIsPartner(p peer.ID, e *Engine) bool {
-	for _, partner := range e.Peers() {
-		if partner == p {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(e.Peers(), p)
```
gammazero: Instead of making a copy of the slice, and then looking through that copy, it would be better to create a new function in `Engine` to look for a peer:
```go
return e.HasPeer(p)
```

The `Engine..HasPeer` function should call a `HasPeer` function of the peer ledger. This would allow the search without having to create a new slice of all peers each time.

bitswap/testnet/peernet.go
```diff
@@ -29,12 +30,7 @@ func (pn *peernet) Adapter(p tnet.Identity, opts ...bsnet.NetOpt) iface.BitSwapN
 }
 
 func (pn *peernet) HasPeer(p peer.ID) bool {
-	for _, member := range pn.Mocknet.Peers() {
-		if p == member {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(pn.Mocknet.Peers(), p)
```
gammazero, line 33: Similar to above, better to not create a new slice of peers. Instead implement a `HasPeer` function in `Mocknet`
https://github.com/libp2p/go-libp2p/blob/master/p2p/net/mock/mock_net.go

Looks like `mock_net.go` could also benefit from using `slices.Sort` and/or `slices.SortFunc` instead of implementing `sort.Interface`.

gammazero (changes requested):

bitswap/server/internal/decision/engine_test.go
pullmerge: Thank you for the suggestion, it has been modified. Please review.

bitswap/testnet/peernet.go
pullmerge: I understand your point, but this relies on https://github.com/libp2p/go-libp2p/blob/master/p2p/net/mock/mock_net.go. I believe I should submit a PR for this project without breaking this dependency. 

If I have misunderstood, please provide further suggestions.

gammazero (approved):

PR #922, opened 2025-05-22
fix(gateway): support suffix range requests

### LINKED ISSUE #10808: Negative range requests are not implemented correctly
### Checklist

- [x] This is a bug report, not a question. Ask questions on [discuss.ipfs.tech](https://discuss.ipfs.tech/c/help/13).
- [x] I have searched on the [issue tracker](https://github.com/ipfs/kubo/issues?q=is%3Aissue) for my bug.
- [x] I am running the latest [kubo version](https://dist.ipfs.tech/#kubo) or have an issue updating.

### Installation method

third-party binary

### Version

```Text
Kubo version: 0.34.1-Homebrew
Repo version: 16
System version: arm64/darwin
Golang version: go1.24.1
```

### Config

```json
{
  "API": {
    "HTTPHeaders": {}
  },
  "Addresses": {
    "API": "/ip4/127.0.0.1/tcp/5001",
    "Announce": [],
    "AppendAnnounce": [],
    "Gateway": "/ip4/127.0.0.1/tcp/8080",
    "NoAnnounce": [],
    "Swarm": [
      "/ip4/0.0.0.0/tcp/4001",
      "/ip6/::/tcp/4001",
      "/ip4/0.0.0.0/udp/4001/webrtc-direct",
      "/ip4/0.0.0.0/udp/4001/quic-v1",
      "/ip4/0.0.0.0/udp/4001/quic-v1/webtransport",
      "/ip6/::/udp/4001/webrtc-direct",
      "/ip6/::/udp/4001/quic-v1",
      "/ip6/::/udp/4001/quic-v1/webtransport"
    ]
  },
  "AutoNAT": {},
  "AutoTLS": {},
  "Bootstrap": [
    "/ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
    "/ip4/104.131.131.82/udp/4001/quic-v1/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb",
    "/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt",
    "/dnsaddr/va1.bootstrap.libp2p.io/p2p/12D3KooWKnDdG3iXw9eTFijk3EWSunZcFi54Zka4wmtqtt6rPxc8"
  ],
  "DNS": {
    "Resolvers": {}
  },
  "Datastore": {
    "BlockKeyCacheSize": null,
    "BloomFilterSize": 0,
    "GCPeriod": "1h",
    "HashOnRead": false,
    "Spec": {
      "mounts": [
        {
          "child": {
            "path": "blocks",
            "shardFunc": "/repo/flatfs/shard/v1/next-to-last/2",
            "sync": false,
            "type": "flatfs"
          },
          "mountpoint": "/blocks",
          "prefix": "flatfs.datastore",
          "type": "measure"
        },
        {
          "child": {
            "compression": "none",
            "path": "datastore",
            "type": "levelds"
          },
          "mountpoint": "/",
          "prefix": "leveldb.datastore",
          "type": "measure"
        }
      ],
      "type": "mount"
    },
    "StorageGCWatermark": 90,
    "StorageMax": "10GB"
  },
  "Discovery": {
    "MDNS": {
      "Enabled": true
    }
  },
  "Experimental": {
    "FilestoreEnabled": false,
    "Libp2pStreamMounting": false,
    "OptimisticProvide": false,
    "OptimisticProvideJobsPoolSize": 0,
    "P2pHttpProxy": false,
    "StrategicProviding": false,
    "UrlstoreEnabled": false
  },
  "Gateway": {
    "DeserializedResponses": null,
    "DisableHTMLErrors": null,
    "ExposeRoutingAPI": null,
    "HTTPHeaders": {},
    "NoDNSLink": false,
    "NoFetch": false,
    "PublicGateways": null,
    "RootRedirect": ""
  },
  "Identity": {
    "PeerID": "12D3KooWA1fJT5L983mE3YcjeQczgKPJxhBURXPooYbeDduKRP7a"
  },
  "Import": {
    "BatchMaxNodes": null,
    "BatchMaxSize": null,
    "CidVersion": null,
    "HashFunction": null,
    "UnixFSChunker": null,
    "UnixFSRawLeaves": null
  },
  "Internal": {},
  "Ipns": {
    "RecordLifetime": "",
    "RepublishPeriod": "",
    "ResolveCacheSize": 128
  },
  "Migration": {
    "DownloadSources": [],
    "Keep": ""
  },
  "Mounts": {
    "FuseAllowOther": false,
    "IPFS": "/ipfs",
    "IPNS": "/ipns"
  },
  "Peering": {
    "Peers": null
  },
  "Pinning": {
    "RemoteServices": {}
  },
  "Plugins": {
    "Plugins": null
  },
  "Provider": {
    "Strategy": ""
  },
  "Pubsub": {
    "DisableSigning": false,
    "Router": ""
  },
  "Reprovider": {},
  "Routing": {
    "Methods": null,
    "Routers": null
  },
  "Swarm": {
    "AddrFilters": null,
    "ConnMgr": {},
    "DisableBandwidthMetrics": false,
    "DisableNatPortMap": false,
    "RelayClient": {},
    "RelayService": {},
    "ResourceMgr": {},
    "Transports": {
      "Multiplexers": {},
      "Network": {},
      "Security": {}
    }
  },
  "Version": {}
}
```

### Description

I encountered this issue while trying to load WACZ files using https://replayweb.page/ ... it uses negative range requests, and Kubo returns the wrong bytes from the file.

Here's a simple example:

```
$ echo 'abcdef' | ipfs add -                                                                      
added QmVMajWeL9geiuhkz45Cp7fErqdsD81J1pVKGkUk3KHH6j QmVMajWeL9geiuhkz45Cp7fErqdsD81J1pVKGkUk3KHH6j
 7 B / 7 B [=============================================================================================================================================] 100.00%

$ curl -H 'range: bytes=-3' http://127.0.0.1:8080/ipfs/QmVMajWeL9geiuhkz45Cp7fErqdsD81J1pVKGkUk3KHH6j 
abc%
```

Expected output should be: "def" (the last 3 bytes in the file)

Here is the documentation for the Range header from MDN:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range

The public IPFS gateway (not kubo) works with https://replayweb.page/ -- but it's also returning the wrong bytes (off by one).

```
$ curl -H 'range: bytes=-3' https://ipfs.io/ipfs/QmVMajWeL9geiuhkz45Cp7fErqdsD81J1pVKGkUk3KHH6j     
ef
```

aschmahmann:
fixes https://github.com/ipfs/kubo/issues/10808

Well ... the need for this PR is quite embarrassing. I thought we had this covered in gateway conformance but we don't.

@lidel IIRC we discussed this when this code path was implemented and maybe this was just supposed to not support suffix requests and return 200 with everything instead of 206 (as we do for ranges on things like dir-index-html responses), but given we found a user who could actually benefit from it it seemed reasonable to implement. Feel free to take over the PR and change as you wish, but figured I'd at least present an option to help get things moving.

- [x] Figure out how we want to resolve this
- [x] Add gateway conformance tests


gateway/gateway.go
```diff
@@ -219,10 +219,11 @@ type ContentPathMetadata struct {
 //   - From >= 0 and To = nil: Get the file (From, Length)
 //   - From >= 0 and To >= 0: Get the range (From, To)
 //   - From >= 0 and To <0: Get the range (From, Length - To)
+//   - From < 0 and To = nil: Get the file (Length - From, Length)
```
aschmahmann, line 222: @lidel the rules for this struct were not great from the get go. There's a similar struct that's more expansive (https://github.com/ipfs/boxo/blob/2f225c407b2a2701598ceb2cffef5e2953fadab9/gateway/gateway.go#L98-L108) that exists because the trustless-gateway entity-bytes spec is a little more flexible than the HTTP range requests spec. In theory this struct differed because its rules were supposed to be for just HTTP ... but rule 3 was never implemented correctly / doesn't represent suffix-requests.

Some options look like:
1. Add a 4th rule that is the suffix requests (what's done here)
2. Replace the 3rd rule with one proper for suffix requests (in theory even doable without changing the struct types) ... technically a breaking change although I suspect this isn't much used externally
3. Just combine this with `DagBytesRange` and have a single struct

gateway/serve_http_content.go
```diff
@@ -475,9 +475,18 @@ func (i *handler) seekToStartOfFirstRange(w http.ResponseWriter, r *http.Request
 	return true
 }
 
-func seekToRangeStart(data io.Seeker, ra *ByteRange) error {
+func seekToRangeStart(data io.Seeker, ra *ByteRange, size int64) error {
```
aschmahmann, line 483: We should figure out how much testing is worth doing here vs just adding to gateway conformance and relying on that.

It's notable that there was a TODO around testing this area 🤦 https://github.com/ipfs/boxo/pull/922/files#diff-5b9dda4053dba85f902d2464389f79cdfca0976873fd206acaefcb09784d6e98R55, let's drop that TODO and replace with enough testing that we're happy 😅.

lidel:
Triage notes:
- @lidel  will look into this + add range tests (incl. suffix ones) to https://github.com/ipfs/gateway-conformance/issues/154, then swich this PR to use them, and if green, release, and switch to final

gateway/gateway.go
lidel: option (1) sgtm – seems to be enough to fix https://github.com/ipfs/kubo/issues/10808 and pass all new tests added in https://github.com/ipfs/gateway-conformance/pull/213

gateway/serve_http_content.go
lidel: Dropped todo + added e2e tests to https://github.com/ipfs/gateway-conformance/pull/213, this way the same tests will be applied to inbrowser.link.

lidel:
- [x] added tests for suffix and gap scenarios in https://github.com/ipfs/gateway-conformance/pull/213
- [x] fixed car backend

I will take a look with fresh eyes tomorrow, remaining work is to bubble up and land everything:

- [x] release gateway-conformance (with changelog)
  - [x] double check negative scenarios: is output clear enough to reason why ranges do not work (https://github.com/ipfs/gateway-conformance/pull/213/commits/ad744bcd756caa1cdbb4eb2be04367b4ae2ebf0c)
- [x] update this PR to use latest gateway-conformace and merge
- [ ] update kubo in https://github.com/ipfs/kubo/pull/10818  to use latest boxo main + released gateway conformance

lidel (approved):
- Switched boxo to gateway-conformance v0.8 with explicit test for suffix range-requests
  - Kubo 0.35 fails as expected
  - Kubo with this PR passes, confirming the fix works
- Merging this, and switching Kubo/Rainbow to boxo from main  + gateway-conformance v0.8.

gateway/gateway.go
```diff
@@ -219,10 +219,11 @@ type ContentPathMetadata struct {
...
 //
 // [HTTP Byte Range]: https://httpwg.org/specs/rfc9110.html#rfc.section.14.1.2
 type ByteRange struct {
-	From uint64
```
lidel, line 225: :thought_balloon: This triggered some AI checks – https://github.com/ipfs/boxo/security/code-scanning/7 – seems to be an issue already present in legacy code, unrelated to this PR.

I don't think this is a problem blocking this PR because the gateway range-requests do not mutate UnixFS DAGs.

Still, filled potential fix in https://github.com/ipfs/boxo/pull/936 to make linter warning go away.

PR #927, opened 2025-05-27
Use go-car/v2 for reading CAR files in gateway backend

### LINKED ISSUE #925: gateway: update backend car traversal to use go-car/v2
Gateway backend CAR traversal is using go-car v1. This pulls in deprecated dependencies. The code should be modified to use go-car/v2 to perform the equivalent functionality.

Specifically `car.NewCarReaderWithOptions` and `car.WithErrorOnEmptyRoots` are no longer supported.

This may require the go-car/v2 `NewReader` to accept `io.Reader` instead of `io.ReaderAt`. Or, the gateway CAR traversal will need to read the CAR data into t file first.

gammazero:
Closes #925


lidel (approved):
Triage note:
- thank you for cleaning up old deps
- this should not impact existing Rainbow/Kubo/ipfs-cluster users
  - remote-car-backend conformance tests pass: https://github.com/ipfs/boxo/actions/runs/15278914354/job/42972552548?pr=927

PR #935, opened 2025-05-28
fix(files): support js/wasm and wasip1/wasm builds

### LINKED ISSUE #186: Broken WASM builds

hsanjuan:
The modified file was the only thing in boxo erroring when attempting to build a wasm target

- Closes #186
- towards https://github.com/ipfs/kubo/issues/5694


lidel (approved):
LGTM, rebased and pushed updates so we have tests and can maintain this going forward:

- Merged `main` into `wasm` (CHANGELOG entry rebased onto current `[Unreleased]`).
- Fixed `GOOS=wasip1 GOARCH=wasm` regression. The original commit dropped both `js` and `wasip1` from `filewriter_unix.go` while only adding `filewriter_js.go`, so wasip1 lost `createNewFile`/`isValidFilename`. `wasip1` defines `syscall.O_NOFOLLOW`, so restoring it to the unix build tag works unchanged.
- Refactored to keep the platform delta minimal: shared logic moved into `filewriter_posix.go` (covers all non-Windows targets); `filewriter_unix.go` and `filewriter_js.go` now hold only the `noFollowFlag` constant (`syscall.O_NOFOLLOW` on supported targets, `0` on js).
- Added `.github/workflows/cross-platform-checks.yml`: cross-compiles `./...` from one Linux runner for `js/wasm`, `wasip1/wasm`, and `linux`/`darwin`/`windows` × `amd64`/`arm64`. `continue-on-error: true` and not part of any required-checks set, so regressions surface in the PR check list without gating merges.
- Reworded the CHANGELOG entry to name both targets and link this PR.

WASM support will then continue in https://github.com/ipfs/kubo/issues/5694

PR #936, opened 2025-05-28
fix(unixfs): int check

lidel:
The #922 triggered some AI checks:
https://github.com/ipfs/boxo/security/code-scanning/7

I don't think this is a problem irl, due to code in `unixfs/mod/dagmodifier.go` being related to single `raw` block (which has limits), but adding this extra check there just in case and to avoid linters bringing this up in the future.

@ipfs/kubo-maintainers  – thoughts?


gammazero (approved):

gammazero:
@lidel This is probably a good thing to have in case some day the limits around a single raw block disappear. ...and to stop linters from pestering us.

PR #937, opened 2025-05-29
bitswap/client: broadcast reduction and metrics

gammazero:
## New Metrics
- `ipfs_bitswap_wanthaves_broadcast`: Count of want-haves set as broadcasts
- `ipfs_bitswap_haves_received`:  Count of total have responses
- `ipfs_bitswap_bcast_skips_total{`: Count of broadcasts skipped as part of spam reduction logic
- `ipfs_bitswap_unique_blocks_received`: Count of non-duplicate blocks recieved.

## Broadcast reduction logic
Send broadcast if:
- Peer has previously sent blocks or have-blocks
- Optionally, Peer is on local network or configured for peering
- Optionally, Peer has pending message
- Optionally send broadcasts to 1 or more peers that are not broadcast targets by above rules.

Otherwise send broadcast to peer.

## References

- Kubo PR: https://github.com/ipfs/kubo/pull/10825
- Part of https://github.com/ipshipyard/roadmaps/issues/10


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+- `bitswap/client`: Reduce bitswap broadcast volume by limiting broadcasts to peers that have previously responded as having wanted blocks and peers on local network.
```
lidel: nit: 

- document newly added options in bitswap/client/client.go (add godoc there and then copy&paste to changelog here as sub-bullet points) (reuse config.md from [kubo/pull/10825](https://github.com/ipfs/kubo/pull/10825/files#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR1290)?)
- ad explicit sentence that old behavior can be restored by setting `client.WithBroadcastReduction(false)` (iiuc?)

bitswap/client/client.go
```diff
@@ -137,6 +137,36 @@ func WithDefaultProviderQueryManager(defaultProviderQueryManager bool) Option {
 	}
 }
 
+func WithBroadcastReduction(enable bool) Option {
```
lidel: nit: add godoc with one sentence explanation of each of these (reuse config.md from [kubo/pull/10825](https://github.com/ipfs/kubo/pull/10825/files#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR1290)?)

lidel (commented):
Leaving internals review to @aschmahmann, 
but we need to document all these new knobs better (port docs from kubo pr?)

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -16,6 +19,31 @@ type Gauge interface {
...
+	// network the same as peers on other networks.
+	LocalAlways peerstore.Peerstore
+	// LimitPeers is the hard limit on the number of peers to send broadcasts
+	// to. A value of 0 means there is no limit.
+	LimitPeers int
+	// SendSkipped if the number peers to broadcast to anyway, that otherwise
+	// would have been skipped by broadcast reduction logic. Setting this to a
+	// non-zero value ensures at least this number of random peers receives a
+	// broadcast.
+	SendSkipped int
```
guillaumemichel: nit: I find `SendSkipped` a bit confusing. Would `MinBroadcastRecipients` or similar conserve the expected meaning?

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -16,6 +19,31 @@ type Gauge interface {
...
 }
 
+// BroadcastConfig contains broadcast reduction settings. The zero-value is the
+// default configuration and results in the minimum amount of broadcasts
+// without placing and hard limit on the number of broadcasts.
+type BroadcastConfig struct {
+	// LocalAlways is a Peerstore that is uses to determine if a peer is on the
+	// local network. If non nil, always boradcast to peers on the local
+	// network. If nil, apply broadcast reduction logic to peers on the local
+	// network the same as peers on other networks.
```
guillaumemichel:
```suggestion
	// LocalAlways is a Peerstore that is used to determine if a peer is on the
	// local network. If non nil, always broadcast to peers on the local
	// network. If nil, apply broadcast reduction logic to peers on the local
	// network the same as peers on other networks.
```

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -16,6 +19,31 @@ type Gauge interface {
...
+	LocalAlways peerstore.Peerstore
```
guillaumemichel: I don't get the meaning of `LocalAlways`, why not something like `Peerstore`/`LocalPeerstore`?

guillaumemichel (approved):
A few nits:

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: It is a way to combine a boolean flag (`LocalAlways`) with the `Peerstore` that is used to remember peers that are on the local network. If the  `Peerstore` is not `nil` then this means that `LocalAlways` flag is true.

I could change this to a flag and an optional peerstore, but there there would need to be an additional check if the flag is `true` then the peerstore must be provided. WDYT?

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: This was a really hard one to name, and I also do not like the name.

`MinBroadcastRecipients` is not correct because that implies "at least n peers". This value does not configure a minimum number of peers to broadcast to, but rather the maximum number of peers to broadcast to anyway, even though broadcast reduction logic has determined that they are not broadcast targets.

It's more `BroadcastToRandomPeers` or `TryLuckAttempts`.

bitswap/client/internal/peermanager/peerwantmanager.go
guillaumemichel: Okay, I get it now. Let's keep it like this

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -16,6 +19,31 @@ type Gauge interface {
...
+	// LimitPeers is the hard limit on the number of peers to send broadcasts
+	// to. A value of 0 means there is no limit.
+	LimitPeers int
+	// SendSkipped if the number peers to broadcast to anyway, that otherwise
```
guillaumemichel:
```suggestion
	// SendSkipped is the number peers to broadcast to anyway, that otherwise
```

bitswap/client/internal/peermanager/peerwantmanager.go
guillaumemichel: I understand this part of the comment more like a `Min` (emphasis is mine).

> Setting this to a non-zero value ensures **at least** this number of random peers receives a broadcast.

What about `MaxForcedBroadcastPeers` reflecting the `Max`? Or `MaxBroadcastToRandomPeers`?

gammazero: @guillaumemichel What do you think of renaming this to `SendRandomPeers`?

guillaumemichel: > What do you think of renaming this to MaxRandomPeers?

Sounds good!

bitswap/client/client.go
gammazero: Done

CHANGELOG.md
gammazero: Done

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: Done

gammazero:
@guillaumemichel @lidel I have changed the default configuration to have broadcast reduction logic disabled. This will prevent breaking boxo user who may depend on broadcasting to all peers. Kubo with set this to true by default.

guillaumemichel (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+- `bitswap/client`: Reduce bitswap broadcast volume by limiting broadcasts to peers that have previously responded as having wanted blocks and peers on local network. The following bitswap client options are available to configure the behavior of broadcast reduction:
+  - `WithBroadcastReduction` enables or disables broadcast reduction logic. If broadcast reduction logic is disabled, then the other WithBroadcast options are ignored. Setting this to false restores the previous broadcast behavior of sending broadcasts to all peers. Default is `false`.
+  - `WithBroadcastLimitPeers` sets a hard limit on the number of peers to send broadcasts to. A value of 0 means there is no limit. Default is `0`.
```
hsanjuan: Is there a way to fully disable broadcasting though? Perhaps `-1` could be the default then, and 0 means 0.

bitswap/client/bitswap_with_sessions_test.go
```diff
@@ -586,7 +588,9 @@ func TestDontHaveTimeoutConfig(t *testing.T) {
 	defer cancel()
 
 	bs := bitswap.New(ctx, a.Adapter, pqm, a.Blockstore,
-		bitswap.WithClientOption(client.WithDontHaveTimeoutConfig(cfg)))
+		bitswap.WithClientOption(client.WithDontHaveTimeoutConfig(cfg)),
+		bitswap.WithClientOption(client.WithBroadcastReduction(false)),
```
hsanjuan: Is this necessary if the default is `false` already ?

bitswap/client/client.go
```diff
@@ -483,6 +562,8 @@ func (bs *Client) updateReceiveCounters(blocks []blocks.Block) {
 		if has {
 			c.dupBlocksRecvd++
 			c.dupDataRecvd += uint64(blkLen)
+		} else {
+			bs.uniqueBlocksReceivedGauge.Inc()
```
hsanjuan: Given that `has = false` by default, this metric will track received blocks (unique or not). Should it not be disabled just like `dupMetric` is disabled?

Might be worth having a `blocksReceived` counter, and log the other metrics only when `skipDuplicatedBlocksStats` is false. A bit suspicious that we were not tracking blocks received anywhere already, but I don't see such thing.

bitswap/client/client.go
```diff
@@ -311,6 +380,12 @@ type Client struct {
 	skipDuplicatedBlocksStats bool
 
 	perPeerSendDelay time.Duration
+
+	bcastReduction       bool
+	bcastReduceLocal     bool
+	bcastLimitPeers      int
+	bcastSendRandomPeers int
+	bcastSendWithPending bool
```
hsanjuan: Nit: This is done handled differently than `DontHaveTimeoutConfig` in which we let the user pass in an aliased config object. Not sure what the best way is but given all options are tightly related, we can consider wrapping them in an object and keep consistency with previous art.

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -34,6 +64,11 @@ type peerWantManager struct {
 	wantGauge Gauge
 	// Keeps track of the number of active want-blocks
 	wantBlockGauge Gauge
+
+	bcastConfig  *BroadcastConfig
```
hsanjuan: Nit: Given this is a static object, I would not pass a pointer. I would put the `EnableReduction` key in it and use it as a value, which is more explicit than checking for `nil` and interpreting that is "disabled".

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -139,8 +189,27 @@ func (pwm *peerWantManager) broadcastWantHaves(wantHaves []cid.Cid) {
...
+	// If broadcast reduction logic enabled.
+	if pwm.bcastConfig != nil {
+		reduce = true
+		bcastLimitPeers = pwm.bcastConfig.LimitPeers
+		sendSkipped = pwm.bcastConfig.SendRandomPeers
+	}
+
 	// Send broadcast wants to each peer
-	for _, pws := range pwm.peerWants {
+	for p, pws := range pwm.peerWants {
```
hsanjuan, line 225: Are the peers in going to show up in random order?

Otherwise said, is there a chance that we always end up broadcasting to the same peers and skipping the same peers in certain scenarios? Should they be randomly picked up?

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -152,7 +221,65 @@ func (pwm *peerWantManager) broadcastWantHaves(wantHaves []cid.Cid) {
...
+}
+
+func (pwm *peerWantManager) isLocalPeer(peerID peer.ID) bool {
+	if pwm.bcastConfig.SendLocalPeers == nil {
+		return false
+	}
+	if _, ok := pwm.remotePeers[peerID]; ok {
+		return false
+	}
+	addrs := pwm.bcastConfig.SendLocalPeers.Addrs(peerID)
```
hsanjuan: Are peers are not advertising LAN addresses or localhost addresses that end up in the peerstore of other peers? (I think some old versions were doing that).

Probably overkill: We could do a `ConnsToPeer()` and check that we are actually connected over a Private address to the peer in question? For that you'd need the Host rather than the Peerstore, although probably the best solution is to expand the `Connected` callback to store that information somewhere for all peers that we connect to (i.e. peerstore metadata). (For example cannot rely on libp2p ConnsToPeer for HTTP connections).

hsanjuan (commented):
Some small nits and a couple of questions from my side

hsanjuan:
One more comment as I gave some thought: I think it would be good to think if we should broadcast to protected/tagged connections. Tagging a connection is a way of specifying that different peer is your friend. I don't know if we can distinguish between manual user tags and temporary tags made by bitswap/dht though...

bitswap/client/bitswap_with_sessions_test.go
gammazero: Removed.

CHANGELOG.md
gammazero: I considered that, but it was not not sure it was in scope, as there was never an ask to shut of broadcasts completely.

@lidel If we want this capability, I can certainly provide it as hector has suggested above.

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -152,7 +221,65 @@ func (pwm *peerWantManager) broadcastWantHaves(wantHaves []cid.Cid) {
...
+	for _, addr := range addrs {
+		if manet.IsPrivateAddr(addr) {
```
lidel: this may not be enough, we also want to include loopback addresses so all localhost tests work as they did before, and to avoid the need for https://github.com/ipfs/helia/pull/818 

```suggestion
		if manet.IsPrivateAddr(addr) || manet.IsIPLoopback(addr) {
```

CHANGELOG.md
lidel: +1, sounds like useful capability to have. `-1` and `0` are fine, or for extra clarity, could be a separate `Flag` named `WithBroadcastEnabled` – whatever is easier for you.

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: Thank you! I had assumed that loopback was a private address, but apparently not

bitswap/client/internal/peermanager/peerwantmanager.go
```diff
@@ -152,7 +221,65 @@ func (pwm *peerWantManager) broadcastWantHaves(wantHaves []cid.Cid) {
...
+		if bcastLimitPeers > 0 {
+			bcastLimitPeers--
+			if bcastLimitPeers == 0 {
+				break
+			}
+		}
+	}
+}
+
+func (pwm *peerWantManager) skipBroadcast(peerID peer.ID, peerQueue PeerQueue) bool {
```
lidel, line 266: @hsanjuan  has good point in https://github.com/ipfs/boxo/pull/937#issuecomment-2967397968

In Kubo, if someone added a peer to [`Peering`](https://github.com/ipfs/kubo/blob/master/docs/config.md#peering), it is often to bypass/replace/act-as routing system, so their expectation is to always broadcast to those peers, just like we do to all local ones.

We should make sure `skipBroadcast` does not skip such peers. 

Either check if connection is tagged, or alternative option would be to have explicit config option `WithBroadcastRequired(peerids)` which Kubo would use for protecting peering peers from being reduced.

CHANGELOG.md
gammazero: I prefer separate flag that is not part of the broadcast reduction config.

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: From @lidel:

go-libp2p now does addr filtering  ([code](https://github.com/libp2p/go-libp2p/blob/2d45a8829585a67f11328efe91cafd17a3da6e00/p2p/protocol/identify/id.go#L867-L872)/[test](https://github.com/libp2p/go-libp2p/blob/2d45a8829585a67f11328efe91cafd17a3da6e00/p2p/protocol/identify/id_test.go#L278)) and for public peers localhost/lan addresses are skipped and never saved to peerstore, even if peer returned them in libp2p identify response (filterAddrs func will see non-local c.RemoteMultiaddr() and do the right thing)

that  is to say, this should be safe.

bitswap/client/internal/peermanager/peerwantmanager.go
aschmahmann: If the intent is all non public addresses then maybe ! isPublic is better?

cc @sukunrt for any thoughts, and also it might be to helpful in manet to make the sugaring for working with the various address types here easier / more obvious.

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: When a peer is added to the map or removed from it, it will perturb the order enough to so that there is some randomness. I think this is probably good enough for this (likely unused) feature.

If the set of peers remains static, then the same peers will be broadcast to. To make this truly random, I would need to record all the peers that were skipped for broadcast and then randomize that list. I am concerned that building and sorting this list, each time then is an attempt to send a broadcast, will be too much overhead. Do you think this is worth it, or just rely on the changing map to "randomize"?

CHANGELOG.md
gammazero: Done

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: Done. Now checks with connection manager to look for connections tagged by peering.

bitswap/client/client.go
gammazero: > this metric will track received blocks (unique or not)

It should track received blocks that are not already in the datastore, which should make them unique.

> Should it not be disabled just like dupMetric is disabled?

I am not sure where this is disabled. The `has` value is true if the block already exists in the blockstore, so `dupMetric` is always tracking how many blocks are received that are already in the blockstore.

bitswap/client/client.go
gammazero: Fixed

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: Done

bitswap/client/internal/peermanager/peerwantmanager.go
hsanjuan: I'm against the latter, since protecting a peer via `swarm/connect` is a dynamic operation when Kubo is running and in the case of cluster we cannot ask to do a config change and restart when a new peer shows up.

hsanjuan: That said, could have switch in the form: WithBroadcastToTaggedPeers() that defaults to `true`.

I think bitswap traffic is pretty cheap and we are connected to those peers already and doing things with them so we can probably allow this without seconds thoughts.

The main point in limiting broadcast is when a peer is doing nothing with no one and still uses bandwidth and resources. Broadcasting to protected connections won't affect that.

hsanjuan:
> @hsanjuan @lidel Should broadcasting to peering peers be separately configurable from broadcasting to local peers?

I think yes.

1. If local peer and broadcastToLocal == true -> broadcast
2. If tagged peer and broadcastToTagged == true -> broadcast
3. <rest of logic>

bitswap/client/client.go
hsanjuan: see :

```
if !bs.skipDuplicatedBlocksStats {
		blocksHas = bs.blockstoreHas(blocks)
}
```

`blockHas` is always false if `skipDuplicatedBlocksStats` is true. It is true by default (in Kubo at least).

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ The following emojis are used to highlight certain changes:
...
+  - `WithBroadcastReduceAll` enables or disables broadcast reduction for peers on the local network and peers configured for peering. If `false`, than always broadcast to peers on the local network and peers configured for peering. If `true`, apply broadcast reduction to all peers without special consideration for local and peering peers. Default is `false`.
+  - `WithBroadcastSendRandomPeers` sets the number of peers to broadcast to anyway, even though broadcast reduction logic has determined that they are not broadcast targets. Setting this to a non-zero value ensures at least this number of random peers receives a broadcast. This may be helpful in cases where peers that are not receiving broadcasts my have wanted blocks. Default is `0`.
+  - `SendWithPending` enables or disables sending broadcasts to any peers that already have a pending message to send. When enabled, this sends broadcasts to many more peers, but does so in a way that does not increase the number of separate broadcast messages. There is still the increased cost of the recipients having to process and respond to the broadcasts. Default is `false`.
```
hsanjuan: May I suggest new options names?

- BroadcastControlEnable
- BroadcastControlMaxPeers
- BroadcastControlSendToLocalPeers
- BroadcastControlMaxRandomPeers
- BroadcastControlSendToPendingPeers

If we are very sold on the `With` prefix we can prepend it, but it would be good that these options appear documented next to each others in the documentation because they are very related, so they need a common prefix at least (or be wrapped in an object).

bitswap/client/internal/peermanager/peerwantmanager.go
hsanjuan: I don't think it is worth it, but maybe add this as a comment for future reference, saying that our source of randomness here is the peerWants changing.

bitswap/client/client.go
gammazero: OK, I see. I changed this to always count blocks and renamed it to `blocksReceivedGauge`.

bitswap/client/internal/peermanager/peerwantmanager.go
gammazero: Noted source of randomness

CHANGELOG.md
gammazero: Done

hsanjuan (approved):
LGTM

PR #938, opened 2025-05-29
Remove WithPeerLedger option and PeerLedger interface

### LINKED ISSUE #928: Remove `WithPeerLedger` bitswap option
This server option is no longer needed, and is not used anywhere.

Deprecated by #929 
Closed by #938

### History

It was used at some point as a brute-force way of not keeping wantlists ( "not keep track of other peer's wantlists, we only want to reply if we have a block. If we get it later, it's no longer relevant." )
https://github.com/ipfs/rainbow/pull/133/files#diff-cf9a198b7e09fd9e99707c1ca07ebaf64974ed4132d695cec2ac472270c1fa9eR438

This matters only if bitswap server is enabled (and in rainbow, it is disabled by default, and  only enabled via RAINBOW_PEERING https://github.com/ipfs/rainbow/blob/main/docs/environment-variables.md#rainbow_peering → https://github.com/ipfs/rainbow/blob/2b5ad75353112e8a186febc2d7ec1c6783ad18cb/setup_bitswap.go#L75-L79

We did some experiments with peering rainbows in production, to benefit from regional caches,  but due to bitswap spam, it was not feasible, and then we switched to Cloudflare which does tiered http cache, making this feature not relevant.

We do not use `WithPeerLedger` for anything anymore, and there is a better way to disable the server as well (https://github.com/ipfs/kubo/issues/10717).

gammazero:
Closes #928

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The following emojis are used to highlight certain changes:
 
 ### Removed
 
+- `bitswap/server` do not allow override of peer ledger with `WithPeerLedger` [#938](https://github.com/ipfs/boxo/pull/938)
+
 ### Fixed
 
 - `gateway`: Fixed suffix range-requests and updated tests to [gateway-conformance v0.8](https://github.com/ipfs/gateway-conformance/releases/tag/v0.8.0) [#922](https://github.com/ipfs/boxo/pull/922)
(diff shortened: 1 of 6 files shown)


lidel (approved):
Lgtm (was Deprecated since v0.31.0)

PR #939, opened 2025-06-03
bitswap/httpnet: expose the errors on connect when connection impossible

hsanjuan:
Context: provide useful user feedback in tools like http-check.

### DIFF
--- a/bitswap/network/httpnet/httpnet.go
+++ b/bitswap/network/httpnet/httpnet.go
@@ -26,6 +26,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
 	"github.com/multiformats/go-multiaddr"
+	"go.uber.org/multierr"
 )
 
 var log = logging.Logger("httpnet")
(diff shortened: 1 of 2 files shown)


bitswap/network/httpnet/httpnet.go
```diff
@@ -388,6 +388,19 @@ func (ht *Network) Self() peer.ID {
...
 
+func connectError(errs map[*network.ParsedURL]error) error {
+	if len(errs) == 0 {
+		return ErrNoHTTPAddresses
+	}
+
+	finalErr := ErrNoSuccess
+	for u, err := range errs {
+		finalErr = fmt.Errorf("%w\n* [%s: %s]", finalErr, u.Multiaddress.String(), err)
+	}
```
gammazero: Consider using the [`go.uber.org/multierr`](https://pkg.go.dev/go.uber.org/multierr) package here, unless there is a specific reason to recursively wrap the errors here.

Seems like [`multierr.Combine`](https://pkg.go.dev/go.uber.org/multierr#Combine) might be good here.

gammazero (approved):

bitswap/network/httpnet/httpnet.go
hsanjuan: you're right, i decided to roll my own for no good reason tbh.

gammazero (approved):

PR #940, opened 2025-06-06
chore(provider): meaningful info level log

lidel:
Arguably, these are not low-level details from DEBUG level of individual CIDs, but high level INFO about start/stop of reprovide.


provider/reprovider.go
```diff
@@ -454,16 +454,16 @@ func (s *reprovider) Reprovide(ctx context.Context) error {
 
 		s.waitUntilProvideSystemReady()
 
-		log.Debugf("starting reprovide of %d keys", len(keys))
+		log.Infof("starting reprovide of %d keys", len(keys))
 		start := time.Now()
 		err := doProvideMany(s.ctx, s.rsys, keys)
 		if err != nil {
-			log.Debugf("reproviding failed %v", err)
+			log.Infof("reproviding failed %v", err)
```
lidel: :thought_balloon:  Should this be an error?

guillaumemichel: Yes, an error is only returned if the DHT provider was unable to provide any record.

guillaumemichel (approved):
This change makes sense.

hsanjuan:
Triage:

* Consider adding additional debug statements for each cid provided etc.
* Consolidate provider logging facilities into a single `provider` (current provider.simple, provider.batched etc...)

hsanjuan (commented):
I have added some more logging and consolidated logging facilities into `provider`.

PR #941, opened 2025-06-11
refactor: use the built-in max/min to simplify the code [skip changelog]

overallteach:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#min) function, which can greatly simplify the code.


overallteach:
Not LLMs. This change can make the code more clean.

gammazero (approved):

PR #944, opened 2025-06-13
feat(provider): reprovide metrics

guillaumemichel:
Adding two metric counters: 
* number of keys provided so far (each key is expected to be provided once)
* number of keys reprovided so far (each key is expected to be periodically reprovided)


provider/reprovider.go
```diff
@@ -265,6 +271,8 @@ func (s *reprovider) provideWorker() {
 	provideFunc := func(ctx context.Context, c cid.Cid) {
 		if err := s.rsys.Provide(ctx, c, true); err != nil {
 			log.Errorf("failed to provide %s: %s", c, err)
+		} else {
+			s.provideCounter.Inc()
```
guillaumemichel, line 275: I am not sure whether these counters are thread safe.

hsanjuan: I think so

provider/reprovider.go
```diff
@@ -114,6 +118,8 @@ func New(ds datastore.Batching, opts ...Option) (System, error) {
 		maxReprovideBatchSize: math.MaxUint,
 		provideWorkerCount:    defaultProvideWorkerCount,
 		keyPrefix:             DefaultKeyPrefix,
+		provideCounter:        metrics.New("ipfs.boxo.provider.provideCount", "Number of provides since node is running").Counter(),
+		reprovideCounter:      metrics.New("ipfs.boxo.provider.reprovideCount", "Number of reprovides since node is running").Counter(),
```
hsanjuan: Regarding the metric name... I believe all metrics in boxo use `metrics.NewCtx` and figure out their metric prefix from a context.

In HTTP I hardcoded that prefix though, which probably wasn't intelligent: https://github.com/ipfs/boxo/blob/67939861ceb726ed6f97474eed381c89b1be8bd9/bitswap/network/httpnet/metrics.go#L75

If you can get a context, then it doesn't hurt using it. And otherwise, the metric name would, based on prior art, be `provider_provide_count` and `reprovider_reprovide_count`.

(I think, I'm not very sure about this. I am more sure that the `ipfs.boxo` prefix is probably not part of any existing metric).

guillaumemichel: Thanks for the heads up. I got confused with the [doc](https://github.com/ipfs/go-metrics-interface/blob/36375380652110e1c8fd81170c17cc9e36e2ce9b/ctor.go#L12-L15).

I followed the same approach as you did in https://github.com/ipfs/boxo/blob/67939861ceb726ed6f97474eed381c89b1be8bd9/bitswap/network/httpnet/metrics.go#L75

hsanjuan (approved):

PR #956, opened 2025-06-19
bitswap/client: disable use of traceability block by default

gammazero:
The bitswap client's [`traceability.Block`](https://github.com/ipfs/boxo/blob/main/bitswap/client/traceability/block.go) is now disabled by default. It is only used for testing and debugging and is not needed for typical operation. Using it costs additional allocation. To enable `traceability.Block`, use the bitswap client option `WithTraceBlock(true)`.


bitswap/client/client.go
```diff
@@ -366,6 +376,11 @@ type Client struct {
 	// External statistics interface
 	tracer tracer.Tracer
 
+	// Enable GetBLock to return
```
guillaumemichel, line 379:
```suggestion
	// Enable GetBlock to return
```

bitswap/client/client.go
```diff
@@ -414,7 +432,10 @@ func (bs *Client) GetBlock(ctx context.Context, k cid.Cid) (blocks.Block, error)
 // GetBlocks returns a channel where the caller may receive blocks that
 // correspond to the provided |keys|. Returns an error if BitSwap is unable to
 // begin this request within the deadline enforced by the context.
-// It returns a [github.com/ipfs/boxo/bitswap/client/traceability.Block] assertable [blocks.Block].
+//
+// If [WithTraceBlock] option is set true, then returns a
```
guillaumemichel:
```suggestion
// If [WithTraceBlock] option is set true, then returns a channel of
```

bitswap/client/internal/notifications/notifications.go
```diff
@@ -118,12 +129,14 @@ func (ps *impl) Subscribe(ctx context.Context, keys ...cid.Cid) <-chan blocks.Bl
...
 				}
-				block, ok := val.(traceability.Block)
-				if !ok {
-					// FIXME: silently dropping errors wtf ?
-					return
+				var block blocks.Block
+				if ps.trace {
+					tb := val.(traceability.Block)
+					tb.Delay = time.Since(subscribe)
+					block = tb
```
guillaumemichel, line 136: Any reason not to use the following?

```suggestion
					block = val.(traceability.Block)
					block.Delay = time.Since(subscribe)
```

guillaumemichel (approved):
LGTM! A few nits

bitswap/client/internal/notifications/notifications.go
gammazero: It is necessary because `block` does not have a `Delay` so I have to assign the delay when the left-hand-side is a `traceability.Block` and not a `blocks.Block`

PR #957, opened 2025-06-19
bitswap/httpnet: limit metric cardinality

hsanjuan:
Currently, we label the StatusCounter metric with the hostname of the http endpoint. If there was a high number of endpoints, this would cause high cardinality for this metric.

This is a non-problem now as there are very few HTTP endpoints. This commit addresses it anyways by only labelling the metric hostnames when the Allowlist is enabled, therefore limiting labels to hostnames in the allowlist.


guillaumemichel (approved):

hsanjuan:
I don't know why I'm feeling that I should go the extra mile and have an option to log all endpoints.

lidel:
@hsanjuan perhaps a compromise is to have "HTTPRetrieval.Metriclist"  in Kubo, and use it when [HTTPRetrieval.Allowlist](https://github.com/ipfs/kubo/blob/master/docs/config.md#httpretrievalallowlist) is not defined? This way we could track specific provider without limiting endpoints and without balooning cardinality globally. 

Logging all endpoints may be used in staging/testing, we could add `*` (wildcard) support as explicit opt-in that one can set to enable metrics for all endpoints + log warning if someone has it set, so its not set in production by mistake.

hsanjuan:
I have added an option for this setting. It can be bubbled later to Kubo/Rainbow when merged.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `bitswap/network/httpnet`: New `WithMetricsLabelsForEndpoints` allows defining which hosts/endpoints can be used for labelling metrics that support such label. '*' enables this for all endpoints receiving HTTP requests, but may cause metric cardinality growth when too many endpoints exist. These labels allow tracking, for example, number or requests per respose status AND endpoint used.
```
```suggestion
- `bitswap/network/httpnet`: New `WithMetricsLabelsForEndpoints` allows defining which hosts/endpoints can be used for labelling metrics that support such label. '*' enables this for all endpoints receiving HTTP requests, but may cause metric cardinality growth when too many endpoints exist. These labels allow tracking, for example, number or requests per response status AND endpoint used. If no endpoints are specified via `WithMetricsLabelsForEndpoints` or `WithAllowlist`, all requests are metered under the label 'other' to keep cardinality in check.
```

lidel (approved):

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
-- `bitswap/network/httpnet`: New `WithMetricsLabelsForEndpoints` allows defining which hosts/endpoints can be used for labelling metrics that support such label. '*' enables this for all endpoints receiving HTTP requests, but may cause metric cardinality growth when too many endpoints exist. These labels allow tracking, for example, number or requests per respose status AND endpoint used. 
+- `bitswap/network/httpnet`: New `WithMetricsLabelsForEndpoints` allows defining which hosts/endpoints can be used for labelling metrics that support such label. '*' enables this for all endpoints receiving HTTP requests, but may cause metric cardinality growth when too many endpoints exist. These labels allow tracking, for example, number or requests per response status AND endpoint used. If no endpoints are specified via `WithMetricsLabelsForEndpoints` or `WithAllowlist`, all requests are metered under the label 'other' to keep cardinality in check.
```
hsanjuan: Actually WithAllowlist has no impact on labels now. I thought this kind of "one option affects a different option in certain conditions" is confusing.

PR #959, opened 2025-06-22
Bitswap: Set DontHaveTimeout/MinTimeout to 25ms

hsanjuan:
As seen in https://github.com/ipfs/go-ds-crdt/issues/285, and mentioned in https://github.com/ipfs/boxo/pull/865, the don't have timeout for very low latency setups becomes very low (2.5ms) and cannot handle any sort of spike.

In a localhost network with two peers, this causes loss of bitswap connectivity altogether, and is unrecoverable without manually reconnecting, or finding the other peer as provider.

Why 200ms? Didn't seem to low nor too high.


hsanjuan:
(i will fix the tests)

guillaumemichel:
Triage notes:
* Let's update the value to 25ms

hsanjuan:
Merging per discussion in triage.

PR #962, opened 2025-06-25
feat(namesys/IPNSPublisher): expose ability to set Sequence

### LINKED ISSUE #10796: [potential bug] Add a parameter in `name.publish` to change the sequence number
### Checklist

- [x] My issue is specific & actionable.
- [x] I am not suggesting a protocol enhancement.
- [x] I have searched on the [issue tracker](https://github.com/ipfs/kubo/issues?q=is%3Aissue) for my issue.

### Description

Within our project [plebbit](https://plebbit.com), the user managing an IPNS key can potentially lose or remove their ipfs directory. This can cause the kubo node to publish IPNS records with sequence numbers that is lower than before removing ipfs directory. That will cause the node to keep publishing to IPNS, but no peers accepting its records because it's lower than the current sequence used in pubsub/dht.

I think being able to `name.publish(record, {sequence: manuallyRetrievedSequenceNumber +1n})` will solve this problem easily

gsergey418:
Fixes https://github.com/ipfs/kubo/issues/10796. Added an option to PublishOptions(namesys) that allows for setting a custom sequence number for the IPNS record. [Kubo PR](https://github.com/ipfs/kubo/pull/10851)


gammazero:
@lidel  Do you think it is necessary to include a note that setting a sequence number that is less than the current sequence of an existing record, may cause the new record not to be seen because it's sequence is lower?

lidel (changes requested):
@gammazero  yes, that should be an error IMO 

@gsergey418 afaik sequence is a monotonic clock,  is there any case when it is something user wants to publish a record that will be ignored due to sequence not being the highest?

guillaumemichel:
Triage notes:
* This PR will be triaged before kubo v0.37

lidel (approved):
Thank you @gsergey418!

I added tests and security validation to prevent sequence numbers from going backwards.

Explainer: the original PR accepted any sequence number, creating unintentional vulnerability. If previous record exists, and we publish with lower sequence, the update will be ignored (at best) or we create unintentional replay attack (at worst). 

Fixed this by requiring custom sequences to be greater than the current record's sequence, added proper error handling with `ErrInvalidSequence`, and   wrote tests covering the validation scenarios to illustrate what is expected behavior.


## Next steps

I think this is good to go but I would like to see tests in https://github.com/ipfs/kubo/pull/10851 pass before we merge this.

lidel:
IPNS tests in CI in https://github.com/ipfs/kubo/pull/10851 look good (only failures are due to lint/using forked dependency). 

I'm going to merge this and switch Kubo to commit from boxo main.

PR #964, opened 2025-06-26
bitswap/httpnet: Sanitize allow/denylist inputs

hsanjuan:


lidel (approved):
lgtm, but if we did `strings.TrimSpace(h)` before doing the checks, it would be less annoying if someone has valid domain but with unnecessary trailing/leading space.

hsanjuan:
> lgtm, but if we did `strings.TrimSpace(h)` before doing the checks, it would be less annoying if someone has valid domain but with unnecessary trailing/leading space.

True, let me change that before merge

PR #965, opened 2025-06-29
bitswap/client: Set DontHaveTimeout MinTimeout to 50ms

hsanjuan:
The user says that 25ms helps but not fully resolves the issue of disconnects on spikes. So let's bump this a bit more..


guillaumemichel (approved):

Dreamacro:
I am also interested in the setting of this minimum value. Currently, I plan to set it to 200ms for the pure intranet. I want to know where the discussion related to "The user says that 25ms helps" is taking place.

gammazero (approved):
OK. Just wish there was a way to make this less of a magic number.

hsanjuan:
@Dreamacro it comes from https://github.com/ipfs/go-ds-crdt/issues/285

Note that in the "pure internet" I suspect that if a latency is x25 the average probably there is a problem. This affects localhost, where latency is 2ms and a slow disk-read can cause a 25x spike easily.

lidel:
@hsanjuan / @gammazero mind adding a  CHANGELOG entry before merging?

lidel (approved):
updated changelog in https://github.com/ipfs/boxo/pull/965/commits/2bb3941b8328d0dd9f97db98a25b83e5969d456e, feel free to merge as soon CI finishes and is green

PR #967, opened 2025-07-02
remove unused code

gammazero:
- Include change to clone messages more efficiently.

### DIFF
--- a/bitswap/message/message.go
+++ b/bitswap/message/message.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"maps"
 
 	"github.com/ipfs/boxo/bitswap/client/wantlist"
 	pb "github.com/ipfs/boxo/bitswap/message/pb"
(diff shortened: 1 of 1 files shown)


bitswap/message/message.go
```diff
@@ -164,17 +141,23 @@ func newMsg(full bool) *impl {
 
 // Clone the message fields
 func (m *impl) Clone() BitSwapMessage {
```
guillaumemichel, line 144: This could be simplified to

```go
func (m *impl) Clone() BitSwapMessage {
	return &impl{
		full:           m.full,
		wantlist:       maps.Clone(m.wantlist),
		blocks:         maps.Clone(m.blocks),
		blockPresences: maps.Clone(m.blockPresences),
		pendingBytes:   m.pendingBytes,
	}
}
```

guillaumemichel (approved):

bitswap/message/message.go
gammazero: Done

PR #969, opened 2025-07-07
provider: implement dht provide interface

guillaumemichel:
This PR is used to test integration with https://github.com/ipfs/kubo/pull/10834

---

## Context

We initially want to make the DHT reprovide improvement opt-in. The new DHT `provider` package ([under development](https://github.com/libp2p/go-libp2p-kad-dht/pull/1082)) comes as a replacement of the older `boxo/provider` package. Since users will be able to choose whether to use the old or new system during the transition period, both systems need to share the same interface.

As we want to eventually modernize the provide interfaces, it is sensible to use the new interface as a standard. The new provider interface is currently discussed in https://github.com/libp2p/go-libp2p-kad-dht/pull/1093.

## Changes

Since the boxo `reprovider` doesn't track which keys should be reprovided, most operations from the new interface simply correspond to `Provide()`. Both systems are eventually similar, because garbage collection of keys to be reprovided is done using the same mechanism ([`KeyChanFunc`](https://github.com/ipfs/boxo/blob/e55a6bc47570249f5143a6f2658832f6a3003c29/provider/provider.go#L44)).

One notable difference is that the new system doesn't have a manual `Reprovide` command to force reprovide all keys instantly.


guillaumemichel:
@galargh do you know why CI refuses to run?

```
The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings
```

galargh:
> @galargh do you know why CI refuses to run?
> 
> ```
> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings
> ```

Yes, we need to update our billing information in the organisation settings. I've raised it in our shared Slack now.

guillaumemichel:
Moved Provide wrapper to kubo in https://github.com/ipfs/kubo/pull/10834/

PR #971, opened 2025-07-07
fix: normalize delegated /routing/v1 urls

lidel:
Quality of life fix that avoids silent errors like one described in https://github.com/ipfs/kubo/issues/10853

### DIFF
--- a/routing/http/client/client_test.go
+++ b/routing/http/client/client_test.go
@@ -676,6 +676,83 @@ func TestClient_FindPeers(t *testing.T) {
 	}
 }
 
+func TestNormalizeBaseURL(t *testing.T) {
+	cases := []struct {
+		name        string
+		input       string
+		expected    string
+		expectError bool
+	}{
+		{
+			name:     "URL with trailing /routing/v1",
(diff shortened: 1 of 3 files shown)


routing/http/client/client_test.go
```diff
@@ -676,6 +676,83 @@ func TestClient_FindPeers(t *testing.T) {
...
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			result, err := normalizeBaseURL(c.input)
+			if c.expectError {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), "only /routing/v1 URLs are supported")
+			} else {
+				require.NoError(t, err)
+				assert.Equal(t, c.expected, result)
+			}
```
gammazero, line 742: Calling `New` calls `normalizeBaseURL`, so is the above part of the test still needed?

gammazero (approved):
Looks good, comment about test.

routing/http/client/client_test.go
lidel: yes, just me being paranoid (a precaution in case New gets refactored to not call it)

PR #976, opened 2025-07-09
Remove providing Exchange. Call Provide() from relevant places.

### LINKED ISSUE #963: Decentralizing new blocks notification (avoid providing everything)
Per https://github.com/ipfs/kubo/issues/10837, we are sending new blocks to the reprovider regardless of strategies.

The places were this happens are:

* providing.Exchange.NotifyNewBlocks:
  * Calls the underlying exchange's NotifyNewBlocks. This is **bitswap**:
    * The server is informed that new blocks are available to provide over Bitswap
    * The client is informed that new blocks have arrived, therefore no need to keep waiting for them
  * Calls `Provide()` on the provider.
* exchange.NotifyNewBlocks is called from:
  * BlockService:
    * AddBlock(s)
    * GetBlock(s)
  * Bitswap doesn't write blocks themselves, all is mediated from Blockservice.

My proposal is to distribute the responsability of providing new blocks to the places that play a role in the different strategies:

* Providing Blockstore:
  * Relevant for 'all' or `flat` strategies, should know when new blocks are written:
    * Unlike the BlockService, that provides blocks on Get/Put, because it can write a block in the process of Getting it, the BlockStore can just provide on writes.
  * Providing Pinner:
    * Relevant for the `pinned`, `roots` strategy:
      * When pinning something, provides every block as dag is traversed, or provides only the root at the end, depending on the strategy.
  * Providing MFS:
    * Relevant for the `mfs` strategy.
    * I have to think about it. We could provide everytime we do dagService.Add(), when strategy enabled.

The process would be as follows:
  * Add options to Blockstore, Pinner, MFS. Example: 
    * WithRootsProvider(p Provider)
    * WithPinnedProvider(p Provider)
  * Call the Provide operation on the necessary places when the option has been set.
  * Remove NotifyNewBlocks from BlockService and Remove providingExchange.
    * Fix tests,
  * On Kubo, pass the provider to the Pinner/MFS/Blockstore during setup, depending on what strategy is set, or leave it disabled otherwise.

The result is that Pinner/MFS/Blockstore will call provide when configured to do so.

hsanjuan:
See https://github.com/ipfs/boxo/issues/963

Refactor the logic to perform Provides, when the component has been given a provider:
  * Remove `providing.Exchange`
  * Provide directly from Blockstore
  * Provide directly from pinner/merkledag on dag traversal
  * Provide from MFS whenever there is a call to `DAGService.Add`

As an alternative to providing from MFS, wrapping the DagService as a "providing DAG service", or building a DAGService backed by a providing blockstore, was considered. It was decided that MFS should carry a Provider because it matches how other modules do it and it's quite explicit, MFS calls Provide every time that there is a call to DAGService.Add, which seems enough.


hsanjuan (commented):
One problem I have is that of circular dependencies between the `provider` module and the `blockstore/pinner/merkledag`. The main issue is that the provide interfaces are not separate from the reprovider implementation. For the moment I have copied the interfaces into the several modules to avoid this.

hsanjuan:
More problems when trying to integrate in Kubo:
  * In order to have a provider, we need a blockstore, but in order to have a providing blockstore, we need a provider.
  * And so on...
  
  This means we need
    * something like blockstore.SetProvider(prov).
    * or perhaps in the blockstore case it would work with a blockstore wrapper + decorator.

- fx gets very complicated as we have providing strategies affecting initialization in multiple places.
- if we do setters, we have to update interfaces.

Need to think about it, potentially discuss in colo.

hsanjuan:
The `routing` libp2p package has also Provide interfaces

ipld/merkledag/merkledag.go
```diff
@@ -204,10 +205,12 @@ func FetchGraphWithDepthLimit(ctx context.Context, root cid.Cid, depthLim int, s
 		return false
 	}
 
+	opts = append(opts, Concurrent())
```
guillaumemichel: `Concurrent()` will override any potential `Concurrency(n)` WalkOption with the default value. Is this intentional?

Otherwise, `Concurrent()` can be set as the first element from `opts`.

pinning/pinner/dspinner/pin.go
```diff
@@ -142,6 +169,10 @@ func New(ctx context.Context, dstore ds.Datastore, dserv ipld.DAGService) (*pinn
 		dstore:    dstore,
 	}
 
+	for _, o := range opts {
+		o.f(p)
```
guillaumemichel, line 172: Do we want to make sure that at most one of `rootsProvider` or `pinnedProvider` is defined? IIUC having both would duplicate roots advertisements.

guillaumemichel (commented):
Nicely done!

> MFS is untouched, for I believe we need to wrap the DagService as a "providing DAG service", or build a DAGService backed by a providing blockstore. It seems these can be done externally though.

So it means that this PR will (momentarily) break providing for `mfs` and `pinned+mfs` reprovide strategies. IMO it is fine to address this elsewhere.

IIUC the different reprovide strategies would require:
* `all`: define `Provider()` on the blockstore
* `flat` define `Provider()` on the blockstore. Equivalent to `all` for providing (but different priority during reprovides).
* `roots`: define `WithRootsProvider()` on the pinner
* `pinned`: define `WithPinnedProvider()` on the pinner
* `pinned+mfs`: define `WithPinnedProvider()` on the pinner + TBD
* `mfs`: TBD

Is it correct?

hsanjuan:
> * `all`: define `Provider()` on the blockstore
> 
>     * `flat` define `Provider()` on the blockstore. Equivalent to `all` for providing (but different priority during reprovides).
> 
>     * `roots`: define `WithRootsProvider()` on the pinner
> 
>     * `pinned`: define `WithPinnedProvider()` on the pinner
> 
>     * `pinned+mfs`: define `WithPinnedProvider()` on the pinner + TBD
> 
>     * `mfs`: TBD
> 
> 
> Is it correct?

Yes

pinning/pinner/dspinner/pin.go
hsanjuan: I'll double-check

ipld/merkledag/merkledag.go
hsanjuan: hmm yeah good catch. Concurrent() was a hardcoded option, but now we can override it.

hsanjuan:
Ok:

  * I have switched to libp2p routing interfaces (better than duplicating them for now)
  * I have added a SetKeyProvider function to the reprovider
  * I have moved NewPinnedProvider to dspinner and removed blockstore/pinner dependencies from the `provider` module.
  
  With this I can get away with my intentions and I can:
    * Create a provider.System without need for a blockstore/pinner in Kubo
    * Use the provider.System to set the Provider option for the blockstore/pinner
    * SetKeyProvider afterwards to the KeyChanFunc created using blockstore/pinner

exchange/providing/providing.go
guillaumemichel, line 1: Is [`NotifyNewBlocks()`](https://github.com/ipfs/boxo/blob/6cfa79448bb1bbf2d40080b8acbd1cff7c282de6/exchange/interface.go#L17) still required to be included in the `exchange.Interface`?

The other implementation in [offline](https://github.com/ipfs/boxo/blob/6cfa79448bb1bbf2d40080b8acbd1cff7c282de6/exchange/offline/offline.go#L37-L41) is a no-op.

blockservice/blockservice.go
guillaumemichel, line 1: There are a few other places where `NotifyNewBlocks()` is still called ([example](https://github.com/ipfs/boxo/blob/803408a01b73ae23f2a62521439e5828d93a3a7f/blockservice/blockservice.go#L276)). Does it make sense to keep other calls?

blockstore/blockstore.go
```diff
@@ -216,7 +230,15 @@ func (bs *blockstore) Put(ctx context.Context, block blocks.Block) error {
...
 		}
 	}
-	return bs.datastore.Put(ctx, k, block.RawData())
+	if err := bs.datastore.Put(ctx, k, block.RawData()); err != nil {
+		return err
+	}
+
+	if bs.provider != nil {
+		logger.Debugf("blockstore: provide %s", block.Cid())
+		return bs.provider.Provide(ctx, block.Cid(), true)
```
guillaumemichel: Maybe we should ignore the error returned by `bs.provider.Provide()`. I am not too familiar with the `blockstore`, but by seeing a few use cases of `Put()`, we probably don't want to consider a `blockstore.Put()` as a failure if the cid couldn't be advertised.

For instance, [`blockservice.getBlock`](https://github.com/ipfs/boxo/blob/803408a01b73ae23f2a62521439e5828d93a3a7f/blockservice/blockservice.go#L270-L274) calls `blockstore.Put()`, and propagates any potential error. I think that `blockservice.getBlock()` should succeed, even though the provide could fail.

Also, `Put()` doesn't need to block on the `Provide()` operation.

```suggestion
		go bs.provider.Provide(ctx, block.Cid(), true)
```

blockstore/blockstore.go
```diff
@@ -351,3 +381,21 @@ func (bs *gclocker) PinLock(_ context.Context) Unlocker {
...
+	if r == nil {
+		return nil
+	}
+	if many, ok := r.(routinghelpers.ProvideManyRouter); ok {
+		logger.Debugf("reprovider: provideMany (%d keys)", len(keys))
+		return many.ProvideMany(ctx, keys)
+	}
+
+	for _, k := range keys {
+		logger.Debugf("reprovider: providing %s", k)
```
guillaumemichel, line 400:
```suggestion
```

blockstore/blockstore.go
```diff
@@ -244,7 +266,15 @@ func (bs *blockstore) PutMany(ctx context.Context, blocks []blocks.Block) error
...
-	return t.Commit(ctx)
+	if err := t.Commit(ctx); err != nil {
+		return err
+	}
+
+	var hashes []multihash.Multihash
+	for _, block := range blocks {
+		hashes = append(hashes, block.Cid().Hash())
+	}
+	return doProvideManyHashes(ctx, bs.provider, hashes)
```
guillaumemichel:
```suggestion
	go doProvideManyHashes(ctx, bs.provider, hashes)
	return nil
```

ipld/merkledag/merkledag.go
```diff
@@ -426,6 +438,18 @@ func sequentialWalkDepth(ctx context.Context, getLinks GetLinks, root cid.Cid, d
...
+	// Successfully fetched "root".
+	// provide it when needed
+	if prov := options.Provider; prov != nil {
+		err := prov.Provide(ctx, root, true)
+		if err != nil && options.ErrorHandler != nil {
+			err = options.ErrorHandler(root, err)
+		}
+		if err != nil {
+			return err
+		}
```
guillaumemichel: We probably don't want the Walk to fail because the provide operation failed.

Also don't block on `Provide()`

```suggestion
		go prov.Provide(ctx, root, true)
```

ipld/merkledag/merkledag.go
```diff
@@ -514,6 +538,21 @@ func parallelWalkDepth(ctx context.Context, getLinks GetLinks, root cid.Cid, vis
...
+						if err != nil && options.ErrorHandler != nil {
+							err = options.ErrorHandler(root, err)
+						}
+						if err != nil {
+							select {
+							case errChan <- err:
+							case <-fetchersCtx.Done():
+							}
+							return
+						}
```
guillaumemichel:
```suggestion
						go prov.Provide(ctx, ci, true)
```

provider/provider.go
```diff
@@ -39,43 +34,26 @@ type System interface {
 	Clear() int
 	Close() error
 	Stat() (ReproviderStats, error)
+	SetKeyProvider(kp KeyChanFunc)
```
guillaumemichel, line 37: `SetKeyProvider` doesn't seem to be called from anywhere. Is that expected?

pinning/pinner/dspinner/pin.go
```diff
@@ -253,7 +287,17 @@ func (p *pinner) doPinRecursive(ctx context.Context, c cid.Cid, fetch bool, name
...
 		return err
 	}
-	return p.flushPins(ctx, false)
+	err = p.flushPins(ctx, false)
+	if err != nil {
+		return err
+	}
+
+	// Provide only if we have not set pinnedProvider, as otherwise
+	// we would provide it twice.
```
guillaumemichel:
```suggestion
	// we would provide the roots twice.
```

pinning/pinner/dspinner/pin.go
```diff
@@ -291,7 +335,14 @@ func (p *pinner) doPinDirect(ctx context.Context, c cid.Cid, name string) error
 		return err
 	}
 
-	return p.flushPins(ctx, false)
+	err = p.flushPins(ctx, false)
+	if err != nil {
+		return err
+	}
+	if p.rootsProvider != nil {
```
guillaumemichel, line 346: Note sure if this is required.
```suggestion
	if p.rootsProvider != nil && p.pinnedProvider == nil {
```

pinning/pinner/dspinner/pin.go
```diff
@@ -253,7 +287,17 @@ func (p *pinner) doPinRecursive(ctx context.Context, c cid.Cid, fetch bool, name
...
-	return p.flushPins(ctx, false)
+	err = p.flushPins(ctx, false)
+	if err != nil {
+		return err
+	}
+
+	// Provide only if we have not set pinnedProvider, as otherwise
+	// we would provide it twice.
+	if p.rootsProvider != nil && p.pinnedProvider == nil {
+		return p.rootsProvider.Provide(ctx, c, true)
```
guillaumemichel:
```suggestion
		go p.rootsProvider.Provide(ctx, c, true)
```

unless we want the pin operation to fail on provide failure.

pinning/pinner/dspinner/pin.go
```diff
@@ -291,7 +335,14 @@ func (p *pinner) doPinDirect(ctx context.Context, c cid.Cid, name string) error
 		return err
 	}
 
-	return p.flushPins(ctx, false)
+	err = p.flushPins(ctx, false)
+	if err != nil {
+		return err
+	}
+	if p.rootsProvider != nil {
+		return p.rootsProvider.Provide(ctx, c, true)
```
guillaumemichel:
```suggestion
		go p.rootsProvider.Provide(ctx, c, true)
```
unless we want the pin operation to fail on provide failure.

provider/provider.go
```diff
@@ -39,43 +34,26 @@ type System interface {
...
 
 // KeyChanFunc is function streaming CIDs to pass to content routing
 type KeyChanFunc func(context.Context) (<-chan cid.Cid, error)
 
-// NewBlockstoreProvider returns key provider using bstore.AllKeysChan
-func NewBlockstoreProvider(bstore blocks.Blockstore) KeyChanFunc {
-	return func(ctx context.Context) (<-chan cid.Cid, error) {
-		return bstore.AllKeysChan(ctx)
-	}
-}
```
guillaumemichel, line 54: Was this method moved elsewhere?

provider/provider.go
```diff
@@ -39,43 +34,26 @@ type System interface {
...
-					return
-				case outCh <- c:
-				}
-			}
-		}()
-
-		return outCh, nil
-	}
+// CidKeysStore is an object that can provide a KeyChanFunc.
+type CidKeysStore interface {
```
guillaumemichel: I don't understand yet how `CidKeysStore` will be used.

guillaumemichel (changes requested):

exchange/providing/providing.go
hsanjuan: Yes, notifies blocks to bitswap so that bitswap knows it does not need to request them anymore, or that it can provide them if people requested them.

blockservice/blockservice.go
hsanjuan: yes, as said, before it was "providing + notifying bitswap" and now it should just notify bitswap.

blockstore/blockstore.go
hsanjuan: Agreed on errors, disagree on not-blocking for Provide.

If your provider is the reprovider, it just queues and returns. If you're provider is the (current) DHT, this call will effectively block for several seconds. I think it is ok that it does, because the alternative is launching goroutines (1000 blocks added -> 1000 concurrent provide operations goroutines). So for me the responsibility of handling concurrent provides is in the Provider implementation and the blockstore should not take the liberty to handle concurrency there.

blockstore/blockstore.go
hsanjuan: why? it's useful to know where provides are triggered from.

blockstore/blockstore.go
hsanjuan: same comment as above

provider/provider.go
hsanjuan: It will be called in Kubo, as I cannot create the blockstore without the reprovider, and I cannot reprovide from the blockstore otherwise. It's unfortunate.

provider/provider.go
hsanjuan: it was removed because == `bstore.AllKeysChan` to all effects (i don't even understand the indirection, probably it was a intermediate step before function signatures were aligned. Anyways, the blockstore dependency in this module needs to go away to avoid dependency cycles.

provider/provider.go
hsanjuan: Eheh because this is a leftover I think. I will remove it.

blockstore/blockstore.go
guillaumemichel: Ok, I agree with your reasoning 👍🏻

blockstore/blockstore.go
guillaumemichel: If `boxo/provider:reprovider` is used, this log line is printed when the cid is added to the provide queue.

There is already [a log](https://github.com/ipfs/boxo/blob/2cf77c4c3ebf7898d3ff57a452ca1a33d6828a9c/provider/reprovider.go#L272) when a cid goes out of the provide queue to be passed to `kad-dht` for DHT provide.

It seems redundant to log when the cid is added to the provide queue, and when it is provided.

---

Furthermore, the text is confusing because it is the same that gets logged [when a key is reprovided](https://github.com/ipfs/boxo/blob/803408a01b73ae23f2a62521439e5828d93a3a7f/provider/reprovider.go#L576).

lidel (commented):
@hsanjuan thoughts on opening a Kubo PR that uses this to see the impact + be a place where we can start adding end-to-end tests for `Reprovider.Strategy` in Kubo? 

Right now we lack test results that:
- confirm `Reprovider.Strategy=all|pinned|roots` work as expected
- show `Reprovider.Strategy=mfs|pinned+mfs` need to be fixed

blockstore/blockstore.go
```diff
@@ -351,3 +381,21 @@ func (bs *gclocker) PinLock(_ context.Context) Unlocker {
 func (bs *gclocker) GCRequested(_ context.Context) bool {
 	return atomic.LoadInt32(&bs.gcreq) > 0
 }
+
+func doProvideManyHashes(ctx context.Context, r routing.ContentProviding, keys []multihash.Multihash) error {
```
lidel: nit:  do we need a reusable func that is used here and in  `provider/reprovider.go#doProvideMany`?

ipld/merkledag/merkledag.go
```diff
@@ -514,6 +538,21 @@ func parallelWalkDepth(ctx context.Context, getLinks GetLinks, root cid.Cid, vis
 						return
 					}
 
+					// Successfully fetched "ci".
+					// provide it when needed
+					if prov := options.Provider; prov != nil {
+						err := prov.Provide(ctx, ci, true)
+						if err != nil && options.ErrorHandler != nil {
+							err = options.ErrorHandler(root, err)
```
lidel: 💭 for the sake of debugging, should error handling show the actual failing cid, or the root one was intentional here?
```suggestion
							err = options.ErrorHandler(ci, err)
```

guillaumemichel:
Triage notes:
* @hsanjuan will take care of the `mfs` provide strategy in this PR

blockstore/blockstore.go
hsanjuan: The log messages are prepended with the facility that they belong too, and the file:line that triggered them. So you would only see the messages if you have logging enabled on the reprovider, in particular. I think if is useful that if you enable logging on the blockstore, you get info on what the blockstore is doing even if that is redundant when logging is enabled for multiple things.

pinning/pinner/dspinner/pin.go
hsanjuan: not necessary because there is no dag walking I think.

blockstore/blockstore.go
hsanjuan: No, provideMany is only used here outside of reprovider. I don't want to import the reprovider system here. I don't think it causes dep-cycles anymore but relying on the ContentProvider interface only is cleaner.

ipld/merkledag/merkledag.go
hsanjuan: Per Gui's suggestions, we are going to not abort operations on provide failures, so we are just going to log the failure (and it logs the CID in latest version of the code).

hsanjuan:
I have addressed comments and I have added a provider param to MFS constructors.

This breaks the API but I don't know if it's worth to avoid that. Seems pretty straightforward to fix (passing nil, or a provider).

In the end I have opted for carrying a Provider around in MFS because it matches how other modules do it and it's quite explicit, rather than doing a magic DAGService wrapper, or a mfs-only-providing-blockstore in Kubo. MFS calls Provide every time that there is a call to DAGService.Add, which seems enough.

hsanjuan:
Review: make sure godoc reflects that you can pass `nil` as the provider.

And fix the tests.

gammazero (approved):

gammazero:
> Review: make sure godoc reflects that you can pass `nil` as the provider.
> 
> And fix the tests.

Done

hsanjuan:
We will merge together with https://github.com/ipfs/kubo/pull/10886

blockstore/blockstore.go
hsanjuan: I found it very useful to debug Kubo to know where Provide calls are coming from so leaving this for now.

hsanjuan:
Per my latest advancements with Kubo, I think we're good to merge this here for the moment.

gammazero (approved):

PR #980, opened 2025-07-16
bitswap/httpnet: fix peers silently stopping from doing http requests

### LINKED ISSUE #979: bitswap/httpnet: http client hangs
I am observing an issue in Rainbow where http seems to hang and completely stop working from a certain point.

Restarting fixes it.

The logs show: `connect request to...`, which corresponds to attempting to connect to an endpoint using the pingCid url.

However, no connection is established. There is no further logging, nothing indicating a failure to connect or that even the request returns. It would seem as if the request is queued somewhere inside go http2 library and never returns.

I will debug further and leave notes here.

hsanjuan:
Fixes #979 (hopefully). If connectivity status has been switched from Connected without doing an httpnet DisconnectFrom(), it may be that we keep pinging the peers, and think we are connected to them, but never restore the connectivity status when skipping the connectivity checks (because we are connected). This results in silent, successful "Connect()" requests that do nothing.

### DIFF
--- a/bitswap/network/httpnet/httpnet.go
+++ b/bitswap/network/httpnet/httpnet.go
@@ -448,6 +448,11 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	p := pi.ID
 	connected := ht.pinger.isPinging(p)
 	if connected {
+		// ensure the evtMgr state aligns. There are cases where the
+		// state might have been flipped (possibly in combination with
+		// the Router + Bitswap sides).
+		ht.connEvtMgr.Connected(p)
+		log.Debugf("reconnected to %s", p)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


bitswap/network/httpnet/httpnet.go
```diff
@@ -508,7 +513,7 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		err = ht.connectToURL(ctx, pi.ID, u, "GET")
 		if err != nil {
 			errs = append(errs, fmt.Errorf("%s: %s", u.Multiaddress.String(), err))
```
gammazero, line 515: Note: You could also use multierr.Append here.
```suggestion
			errs = multierr.Append(errs, fmt.Errorf("%s: %s", u.Multiaddress.String(), err))
```
Then `errs` would be declared as :
```
errs := ErrNoSuccess
```

gammazero (approved):

bitswap/network/httpnet/httpnet.go
hsanjuan: will fix next time I touch this code

hsanjuan: ah wait, I think it is done this way because multierr.Append prints errors in a different way than multierr.Combine. The former joins them in a single line, and the latter nicely formats them in multiple lines. (IIRC!)

PR #981, opened 2025-07-18
Queue all changes in order using non-blocking async queue

gammazero:
It is possible for changes to get our of order when they are submitted by goroutines for non-blocking operations. Instead of goroutines, use an asynchronous queue to maintain order of events.

This may prevent provlems caused by mis-ordering of events shuch as connect and disconnect events.

### DIFF
--- a/bitswap/client/internal/session/sessionwantsender.go
+++ b/bitswap/client/internal/session/sessionwantsender.go
@@ -3,6 +3,7 @@ package session
 import (
 	"context"
 
+	"github.com/gammazero/chanqueue"
 	bsbpm "github.com/ipfs/boxo/bitswap/client/internal/blockpresencemanager"
 	bspm "github.com/ipfs/boxo/bitswap/client/internal/peermanager"
 	cid "github.com/ipfs/go-cid"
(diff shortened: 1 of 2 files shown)


bitswap/client/internal/session/sessionwantsender.go
```diff
@@ -209,28 +212,17 @@ func (sws *sessionWantSender) Shutdown() {
...
-}
-
 // collectChanges collects all the changes that have occurred since the last
 // invocation of onChange
 func (sws *sessionWantSender) collectChanges(changes []change) []change {
-	for len(changes) < changesBufferSize {
+	for range sws.changes.Len() {
 		select {
-		case next := <-sws.changes:
+		case next := <-sws.changes.Out():
```
guillaumemichel, line 225: @gammazero you wrote the library so you probably know better.

Are there any benefits of doing like above?
```go
changes := sws.changes.Out()
for range sws.changes.Len() {
	select {
	case next := <-changes:
...
```

guillaumemichel (approved):
It makes sense!

hsanjuan (approved):

bitswap/client/internal/session/sessionwantsender.go
gammazero: Maybe slightly? The function call should get inlined anyway, so is it worth the extra line of code? 🤷

PR #986, opened 2025-07-29
Bitswap: fix double-worker in connectEventManager. Logging improvements.

hsanjuan:
This should potentially fix the last source of racing connect-disconnects.

### DIFF
--- a/bitswap/client/internal/session/session.go
+++ b/bitswap/client/internal/session/session.go
@@ -406,7 +406,7 @@ func (s *Session) findMorePeers(ctx context.Context, c cid.Cid) {
 		for p := range s.providerFinder.FindProvidersAsync(ctx, k, 0) {
 			// When a provider indicates that it has a cid, it's equivalent to
 			// the providing peer sending a HAVE
-			log.Infof("FoundPeer %s for CID %s", p, k)
+			log.Infow("Found peer for CID", "peer", p, "cid", k)
 			span.AddEvent("FoundPeer")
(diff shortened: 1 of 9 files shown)


bitswap/client/internal/session/session.go
```diff
@@ -406,6 +406,7 @@ func (s *Session) findMorePeers(ctx context.Context, c cid.Cid) {
 		for p := range s.providerFinder.FindProvidersAsync(ctx, k, 0) {
 			// When a provider indicates that it has a cid, it's equivalent to
 			// the providing peer sending a HAVE
+			log.Infof("FoundPeer %s for CID %s", p, k)
```
gammazero: Easier to search/filter for consistent string "Found peer for CID":
```suggestion
			log.Infow("Found peer for CID", "peer", p, "cid", k)
```

bitswap/network/connecteventmanager.go
```diff
@@ -49,14 +51,18 @@ func NewConnectEventManager(connListeners ...ConnectionListener) *ConnectEventMa
 	return evtManager
 }
 
-// SetListeners sets or replaces the current listeners. Not safe to call after
-// Start().
+// SetListeners sets or replaces the current listeners. It will not take effect
+// after Start().
```
gammazero:
```suggestion
// SetListeners sets or replaces the current listeners. It has no effect after
// Start is called.
```

gammazero (approved):
A couple minor wording suggestions.

gammazero (approved):
LGTM

Issue #908, opened 2025-04-09
A timeout is required when fetching blocks


jclab-joseph:
GetBlocks requires a fetch timeout for each block.

Below, we simulated adverse conditions by connecting 10 clients to one server and applying a speed limit.

```
2025/04/09 17:24:44 PROGRESS: [6.48 6.52 0.36 5.92 6.73 6.07 6.73 6.07 3.15 1.06]
2025/04/09 17:24:45 PROGRESS: [6.48 6.73 0.36 5.92 6.77 6.11 6.73 6.27 3.15 1.06]
2025/04/09 17:24:46 PROGRESS: [6.48 6.92 0.36 6.07 6.96 6.15 6.73 6.46 3.15 1.06]
2025/04/09 17:24:47 PROGRESS: [6.5 6.94 0.36 6.27 6.96 6.27 6.73 6.48 3.15 1.06]
2025/04/09 17:24:48 PROGRESS: [6.69 6.94 0.36 6.3 6.96 6.46 6.92 6.48 3.15 1.06]
```

There are clients (0.36, 3.15, 1.06) that are stuck and unable to download.

When executing GetBlocks, if it takes too long to fetch a specific block, it will not be cancelled and will just hang.
To improve this situation, should stop and find another peer.

boxo version : v0.29.1

---

Specifically, the problem occurred when there was one server (bootstrap node) that held the files and hundreds of clients tried to download the files simultaneously.
Some clients will successfully download, but most will get stuck and not be able to download.
I expected the node that received it first would forward the block to other nodes, but that didn't happen.
periodicSearchDelay is also useless if block reception has already started.

lidel:
triage notes
- @gammazero need to lean on you being the most familiar what are the next steps here

gammazero:
Investigating: This my just be a documentation issue about how to set a timeout.

guillaumemichel:
Triage notes:
* point to existing example if any or add one

jclab-joseph:
I temporarily solved this by adding a timer that is initialized when a channel is received in GetBlocks and a retry routine after context cancel.

I will try to make a sample code soon if I have time. But my way is not so good. It seems like it would be better to modify something in bitswap, but I haven't gotten that far.

lidel:
@jclab-joseph  any reason why you can't pass a [context.WithTtimeout](https://pkg.go.dev/context#WithTimeout)? This is idiomatic way of making tasks time-bound, no?

[...]

PR #994, opened 2025-08-11
feat(gateway): concurrency and timeout limits

### LINKED ISSUE #679: gateway: ability to set response write timeout

### LINKED ISSUE #881: gateway: limit the number of requests in flight

lidel:
This PR adds built-in gateway limiter middleware, configuration, and tests: 
- adds `MaxConcurrentRequests`
  - TLDR semaphore ensuring limited number of request in flight, excess gets `HTTP 429 Too Many Requests` response 
  - closes https://github.com/ipfs/boxo/issues/881
- adds `RetrievalTimeout`
  - TLDR enforces a maximum duration for content retrieval:
    - **Time to first byte**: If the gateway cannot start writing the response within this duration (e.g., stuck searching for providers), a 504 Gateway Timeout  is returned.
    - **Time between non-empty writes**: After the first byte, the timeout resets each time new bytes are written to the client. If the gateway cannot write additional data  within this duration after the last successful write, the response is terminated.
  - closes https://github.com/ipfs/boxo/issues/679
  - also helps with #908  (if gateway is used as the high level interface for retrieval)
- adds optional `Config.MetricsRegistry`
  - TLDR allows removing dependency on global registry, allowing for muyltiple instances (e.g. in parallel tests)

## TODO

- [x] refactor new metrics to not be globals, and be returned by `initializeMiddlewareMetrics` 
- [x] Kubo PR with conformance passing  https://github.com/ipfs/kubo/pull/10905
- [x] Rainbow PR with conformance passing https://github.com/ipfs/rainbow/pull/285
  - [x] fix regression in `only-if-cached` (https://github.com/ipfs/boxo/pull/994/commits/e109608dcd46ae1c4003c20dd67be2e4c9823e89)
  
  
## References

- Creates surface to address https://github.com/ipshipyard/roadmaps/issues/14


gateway/gateway.go
```diff
@@ -14,6 +14,18 @@ import (
...
+
+// Default values for gateway configuration limits
+const (
+	// DefaultRetrievalTimeout is the default maximum duration for initial content retrieval
+	// (time to first byte) and subsequent writes to the HTTP response body.
+	DefaultRetrievalTimeout = 30 * time.Second
+
+	// DefaultMaxConcurrentRequests is the default maximum number of concurrent HTTP requests
+	// that the gateway will process.
+	DefaultMaxConcurrentRequests = 1024
```
lidel: ℹ️  Do these make sense for defaults? My rationale:

- 30s feels solid (that is what we've been using at `ipfs.io` under Nginx's [`proxy_read_timeout`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout)) 
- 1024 is 2x the default from Nginx  ([`worker_connections`](https://nginx.org/en/docs/ngx_core_module.html#worker_connections) 512)


Of course users can adjust/disable, this default is about "sane default" aiming at desktop users, and YOLO deployments.

gateway/middleware_ratelimit.go
```diff
@@ -0,0 +1,76 @@
...
+func withConcurrentRequestLimiter(handler http.Handler, limit int, c *Config) http.Handler {
+	if limit <= 0 {
+		return handler
+	}
+
+	// Create semaphore (buffered channel)
+	semaphore := make(chan struct{}, limit)
+	for i := 0; i < limit; i++ {
+		semaphore <- struct{}{}
+	}
```
lidel: ℹ️  high level mechanism for `MaxConcurrentRequests`

gateway/middleware_retrieval_timeout.go
```diff
@@ -0,0 +1,203 @@
...
+		// Run handler in a goroutine so we can interrupt it on timeout
+		go func() {
+			defer func() {
+				tw.mu.Lock()
+				tw.handlerComplete = true
+				tw.mu.Unlock()
+				close(handlerDone)
+			}()
+			handler.ServeHTTP(tw, r)
+		}()
```
lidel: ℹ️ high level mechanism for  `Config.RetrievalTimeout` 

(the rest is extra fluff to handle timeout before any bytes are sent vs "best-effort explicit truncation" in the middle of a dag, after headers with HTTP 200 are sent)

gateway/middleware_retrieval_timeout.go
```diff
@@ -0,0 +1,203 @@
...
+							if retrievalTimeouts != nil {
+								retrievalTimeouts.With(prometheus.Labels{
+									"code":      "504",
+									"truncated": "false",
+								}).Inc()
+							}
+							message := "Unable to retrieve content within timeout period"
+							writeErrorResponse(tw.ResponseWriter, tw.request, tw.config, http.StatusGatewayTimeout, message)
+						} else {
+							// Headers already sent, response is being truncated
```
lidel: This will likely be rare (most if timeout will be because can't find root block), but if timeout in the middle of big DAG happens, we do want to do our best to communicate that to client/reverse proxy so the truncated response is not cached.

If there is Content-Length header, client will detect truncation just fine.
But there are cases like when we stream DAG as a CAR  when we dont know total length of CAR before we return it. In such case we try to:
1. Write `truncationMEssage` so  looking at last bytes makes it obvious what happened
1. (best-effort) Hijack TCP connection and close it in a way thats obvious TCP RST

lidel (commented):
I will fix racy metrics tests and lint after some sleep, but dropping some notes for early review

gateway/backend_blocks.go
```diff
@@ -126,6 +120,74 @@ func (bb *BlocksBackend) Get(ctx context.Context, path path.ImmutablePath, range
 		ra = &ranges[0]
 	}
 
+	// For range requests on UnixFS files, we can optimize by only fetching
+	// the root block first to determine if it's a file, then create a
+	// UnixfsFile that will lazily fetch only the blocks needed for the range
+	if ra != nil {
```
lidel, line 126: ℹ️ this is a fix for bug that was already in boxo/gateway, adding timeout only surfaced the problem in https://github.com/ipfs/rainbow/pull/285#issuecomment-3176965593  (where we run against all three backends)

gateway/handler.go
```diff
@@ -873,21 +900,48 @@ func (i *handler) handleWebRequestErrors(w http.ResponseWriter, r *http.Request,
...
+	// "gateway timeout", and clients may handle these differently.
+	ctx := r.Context()
+	if i.config.RetrievalTimeout != 0 {
+		cacheCheckTimeout := i.config.RetrievalTimeout * 80 / 100
+		checkCtx, cancel := context.WithTimeout(ctx, cacheCheckTimeout)
+		defer cancel()
+		ctx = checkCtx
+	}
+
+	isCached := i.backend.IsCached(ctx, contentPath)
```
lidel, line 922: ℹ️ this is a fix for `only-if-cached` regression found in https://github.com/ipfs/rainbow/pull/285#issuecomment-3176965593

gateway/backend_blocks.go
```diff
@@ -126,6 +120,74 @@ func (bb *BlocksBackend) Get(ctx context.Context, path path.ImmutablePath, range
...
+		if err != nil {
+			return md, nil, err
+		}
+
+		// Check if this is a dag-pb block that might be a UnixFS file
+		rootCodec := lastRoot.Prefix().GetCodec()
+		if rootCodec == uint64(mc.DagPb) {
+			// Create a minimal node from the root block to check if it's a file
+			nd, err := merkledag.DecodeProtobuf(rootBlock.RawData())
+			if err == nil {
```
gammazero: Seems like the error should at least be logged.

gateway/backend_blocks.go
```diff
@@ -126,6 +120,74 @@ func (bb *BlocksBackend) Get(ctx context.Context, path path.ImmutablePath, range
...
+				// Try to create a UnixFS file - this will only fetch blocks as needed
+				f, err := ufile.NewUnixfsFile(ctx, bb.dagService, nd)
+				if err == nil {
```
gammazero: Log err?

gateway/errors.go
```diff
@@ -166,6 +166,37 @@ func (epr ErrPartialResponse) Error() string {
...
+			Error:      message,
+		})
+		if err != nil {
+			_, _ = w.Write([]byte(fmt.Sprintf("error during body generation: %v", err)))
+		}
+	} else {
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		w.WriteHeader(statusCode)
+		fmt.Fprintln(w, message)
```
gammazero: This is exactly what `http.Error` does.
```suggestion
		http.Error(w, message, statusCode)
```

gateway/errors.go
```diff
@@ -166,6 +166,37 @@ func (epr ErrPartialResponse) Error() string {
...
+		err := assets.ErrorTemplate.Execute(w, assets.ErrorTemplateData{
+			GlobalData: assets.GlobalData{
+				Menu: c.Menu,
+			},
+			StatusCode: statusCode,
+			StatusText: http.StatusText(statusCode),
+			Error:      message,
+		})
+		if err != nil {
+			_, _ = w.Write([]byte(fmt.Sprintf("error during body generation: %v", err)))
```
gammazero: No need to convert to []byte, use a Fprint function instead.
```suggestion
			fmt.Fprintln(w, "error during body generation: %v", err)
```

gateway/gateway.go
gammazero: 🤷 🚀

gateway/middleware_ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case <-semaphore:
+			// Acquired a slot
+			metrics.incConcurrentRequests()
+			defer func() {
+				semaphore <- struct{}{} // Release slot
```
gammazero: Avoid having to initialize the semaphore by inverting the acquire/release operations.

```suggestion
	semaphore := make(chan struct{}, limit)

	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		select {
		case semaphore <- struct{}{}:
			// Acquired a slot
			metrics.incConcurrentRequests()
			defer func() {
				<-semaphore // Release slot
```

gateway/middleware_ratelimit_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"testing"
+	"time"
+)
+
+func TestWithConcurrentRequestLimiter(t *testing.T) {
+	t.Run("returns HTML error when Accept header includes text/html", func(t *testing.T) {
+		config := &Config{DisableHTMLErrors: false}
+		handler := withConcurrentRequestLimiter(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Block to simulate a busy handler
+			time.Sleep(100 * time.Millisecond)
```
gammazero: Instead of blocking on a timer, consider blocking on a channel which can be explicitly unblocked. That way there is no guess about an appropriate timeout value.

gateway/middleware_ratelimit_test.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		if !bytes.Contains([]byte(body), []byte("429")) {
+			t.Error("expected 429 in HTML body")
+		}
+	})
+	t.Run("limiter disabled with zero value", func(t *testing.T) {
+		var requestCount atomic.Int32
+		handler := withConcurrentRequestLimiter(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			requestCount.Add(1)
+			time.Sleep(50 * time.Millisecond)
```
gammazero: Another place where a channel might be better than a timer.

gateway/middleware_ratelimit_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+		// Count successful and rejected requests
+		successCount := 0
+		rejectCount := 0
+		for _, code := range results {
+			if code == http.StatusOK {
+				successCount++
+			} else if code == http.StatusTooManyRequests {
+				rejectCount++
+			}
```
gammazero:
```suggestion
			switch code {
			case http.StatusOK:
				successCount++
			case http.StatusTooManyRequests:
				rejectCount++
			default:
				t.Errorf("unexpected http status: %d", code)
			}
```

gateway/middleware_ratelimit_test.go
```diff
@@ -0,0 +1,250 @@
...
+			t.Error("expected at least some successful requests")
+		}
+		if rejectCount == 0 {
+			t.Error("expected at least some rejected requests when limit exceeded")
+		}
+	})
+
+	t.Run("returns 429 with Retry-After and Cache-Control headers", func(t *testing.T) {
+		handler := withConcurrentRequestLimiter(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			time.Sleep(100 * time.Millisecond)
```
gammazero: ibid

gateway/middleware_ratelimit_test.go
```diff
@@ -0,0 +1,250 @@
...
+		cacheControl := rec2.Header().Get("Cache-Control")
+		if cacheControl != "no-store" {
+			t.Errorf("expected Cache-Control: no-store, got %s", cacheControl)
+		}
+	})
+
+	t.Run("Retry-After is static", func(t *testing.T) {
+		handler := withConcurrentRequestLimiter(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Hold the request indefinitely
+			select {}
```
gammazero: Release it after the test is run, just so as not to hang goroutines during test.
```suggestion
		done := make(chan strict{})
		defer close(done)
		handler := withConcurrentRequestLimiter(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			// Hold the request indefinitely
			<-done
```

gateway/middleware_retrieval_timeout.go
```diff
@@ -0,0 +1,182 @@
...
+							}
+							metrics.recordTimeout(statusCode, true)
+
+							// Try to write truncation message (best effort)
+							tw.ResponseWriter.Write([]byte(truncationMessage))
+
+							// Try to hijack and force connection reset
+							if hijacker, ok := tw.ResponseWriter.(http.Hijacker); ok {
+								conn, _, err := hijacker.Hijack()
+								if err == nil {
```
gammazero: Might we ever care to see what the err is?

gateway/middleware_retrieval_timeout.go
```diff
@@ -0,0 +1,182 @@
...
+
+		// Run handler in a goroutine so we can interrupt it on timeout
+		go func() {
+			defer func() {
+				tw.mu.Lock()
+				tw.handlerComplete = true
+				tw.mu.Unlock()
+				close(handlerDone)
+			}()
+			handler.ServeHTTP(tw, r)
```
gammazero: The `defer func()  { }` seem unnecessary.
```suggestion
			handler.ServeHTTP(tw, r)
			tw.mu.Lock()
			tw.handlerComplete = true
			tw.mu.Unlock()
			close(handlerDone)
```

gateway/errors.go
```diff
@@ -200,24 +229,13 @@
...
-		if err != nil {
-			_, _ = w.Write([]byte(fmt.Sprintf("error during body generation: %v", err)))
-		}
-	} else {
-		http.Error(w, err.Error(), code)
-	}
+	log.Debugw("serving error response",
+		"path", r.URL.Path,
+		"method", r.Method,
+		"error", err,
```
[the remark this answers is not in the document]
lidel: @gammazero  thoughts? I've dismissed this is false-positive for our case, everyone can get to the data / CID by looking at "path" anyway.

gateway/middleware_retrieval_timeout.go
lidel: @gammazero unsure if i got it right, but we may want to keep that `defer` due to panics:
- gateway code is able to recover from panics - see `panicHandler` in `./gateway/handler.go` - we do this to not crash the daemon if someone asks for a funny CID that triggers a bug
- so if `handler.ServeHTTP(tw, r)` ever panics, we still need to close the channel and set `handlerDone`, otherwise the timeout goroutine will wait, no?
  - so iirc  the `defer` IS necessary for safety. However, we could make it simpler and only have `defer close(handlerDone)` before calling `handler.ServeHTTP` – I pushed this fix in a68079b65cbc9e496f7a43d65bae0db092f24515

gammazero: OK, no change needed then. Doing this  in a defer func will ensure it gets done even if there is a panic.

gateway/errors.go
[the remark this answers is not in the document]
gammazero: The concern is that the user is using the gateway as part of a larger private system. Suppose that system requires the client to present a token in the HTTP request header. If a request came to the gateway carrying such a token (whether by mistake or necessity), we certainly do not want to log it.

While we do not have any use case that would expose sensitive header data, some other user might, and avoiding logging request headers is good practice.

gammazero (approved):

lidel:
Detected race condition on Kubo staging related to concurrent headers map mutation, looking into a fix.

lidel:
Fix from https://github.com/ipfs/boxo/pull/994/commits/d2d586c00224fcb95af58c14a3a4c670f4fbad9f seems to work on staging (02). No issues in the past 30m. Will let it run for an hour or two just to be sure.

gateway/gateway.go
lidel: ### Extra notes on `MaxConcurrentRequests` and Nginx

While testing fix on staging I had to adjust this default because our boxes can handle higher load:

nginx Configuration on staging box:
- worker_processes: auto (8 workers based on 8 CPU cores)
- worker_connections: 1024 per worker
- Total nginx capacity: 8 workers x 1024 = 8192 total connections

For Reverse Proxy:
Each proxied request uses 2 connections in nginx:
1. Client → nginx (port 443)
2. nginx → kubo (port 8080)

So iiuc nginx on staging box 02 can proxy ~ 8192/2 = 4096 concurrent requests

I've increased `ipfs config  Gateway.MaxConcurrentRequests --json 4096` and success rate issue went away – HTTP 200s success rate is on par with control box 01 that runs previous release (0.36), but impact on CPU/memory is much lower:

I think 1024 may be a sensible default (most of deployments run on weaker hardware and do not back % of traffic of `ipfs.io`) so 1024 is actually reasonable for:
- stock nginx (256 proxy capacity)
- lightly tuned nginx (1-2k proxy capacity)
- single nginx worker setups


I'll just make sure it is properly documented. Update: b42675e3e008f9dacd7f64fdc840f2c29506f688

gammazero (approved):

gateway/errors.go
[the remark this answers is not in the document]
lidel: Ack, but I was unsure why the bot proposed this specific fix. The `"unsupported format"` errors originate from:

```go
responseFormat, formatParams, err := customResponseFormat(r)
if err != nil {
  i.webError(w, r, fmt.Errorf("error while processing the Accept header: %w", err), http.StatusBadRequest)
  return
}
```


 My understanding is that if a user sends an `Accept` header like `application/vnd.ipld.INJECTED_USER_INPUT`, it would:
1. Pass the prefix check, get parsed by `mime.ParseMediaType` 
1. If it doesn't match any of the known formats in the switch statement, it hits the default case
1. The error message unsupported format `application/vnd.ipld.INJECTED_USER_INPUT` gets created
1. This error gets logged with the full user input (bad)

So instead of adding  func that wraps every error like AI suggested, I've fixed it upstream, where we define these errors, and removed user inputs from message:  35bc575b1bea611008e3f80dc11790cd84c87f98

lidel:
I've rebased instead of merge by mistake, then restored to original (reviewed) version, apologies for noise. :see_no_evil:

gateway/gateway.go
lidel: After some thinking adjusted `chore: DefaultMaxConcurrentRequests = 4096`

Rationale in https://github.com/ipfs/boxo/pull/994/commits/e170d19340db64087d81a868ccb93109e001350b: adjusting to higher value to avoid headache in production environments and users complaining for HTTP 429s. this should  act as failsafe of last resort for now, we can adjust later but for now 4k makes rollout easier and less disruptive

lidel:
All green, merging.  
(Added docs + raised `DefaultMaxConcurrentRequests = 4096` to make rollout easier)

PR #995, opened 2025-08-12
refactor: use the new Reprovide Sweep interface

guillaumemichel:
Companion PR to https://github.com/ipfs/kubo/pull/10834

> [!WARNING]
> Don't merge before https://github.com/ipfs/kubo/pull/10834 is ready to be merged, otherwise kubo compatibility will be broken
---
This builds on top of https://github.com/ipfs/boxo/pull/976 to use the latest Provide interface as defined by the [Reprovide Sweep](https://github.com/libp2p/go-libp2p-kad-dht/pull/1093).

It is necessary since Kubo will start using the new provide interface, and won't be able to satisfy the old provide interface required by boxo. Under the hoods, it is still possible to keep the same behavior as before this PR by using a wrapper around the old provider interface.


mfs/dir.go
```diff
@@ -490,10 +490,12 @@ func (d *Directory) getNode(cacheClean bool) (ipld.Node, error) {
...
-		log.Debugf("mfs: provide: %s", nd.Cid())
-		if err = d.prov.Provide(d.ctx, nd.Cid(), true); err != nil {
-			log.Errorf("error providing %s: %s", nd.Cid(), err)
-		}
+		go func(c cid.Cid) {
+			log.Debugf("mfs: provide: %s", c)
+			if err := d.prov.StartProviding(false, c.Hash()); err != nil {
+				log.Warnf("mfs: error while providing %s: %s", c, err)
+			}
+		}(nd.Cid())
```
guillaumemichel: Fix to [failing kubo `pinned+mfs` provide tests](https://github.com/ipfs/kubo/pull/10834/commits/2ce2762e7c9bcaa69ee2c3d5fa070b9147b97ae9) from https://github.com/ipfs/boxo/pull/995/commits/b331b1dcdc7ad61eb53ff2f8c47412ebf01f456d

The new provider's `KeyStore` must be [reset](https://github.com/ipfs/kubo/blob/72b66b3a95834f96c7b9ac6b86b03f0cb0352fc8/core/node/provider.go#L399), during fx initialization, so that all keys matching the provide strategy are scheduled to be reprovided.

[`KeyStore.ResetCids()`](https://github.com/libp2p/go-libp2p-kad-dht/blob/48cb56a62251f09736553fa58a863d17b98acc1d/provider/datastore/keystore.go#L110-L136) takes the `KeyStore` lock, and basically empties the keystore, and fills it with the keys provided through the `<-chan cid.Cid`.

The issue, is that the [MFS `KeyChanFunc`](https://github.com/ipfs/kubo/blob/72b66b3a95834f96c7b9ac6b86b03f0cb0352fc8/core/node/provider.go#L501) (generating `<-chan cid.Cid`), used in `mfs` and `pinned+mfs` strategies, calls [`mfsRoot.FlushMemFree()`](https://github.com/ipfs/kubo/blob/72b66b3a95834f96c7b9ac6b86b03f0cb0352fc8/core/node/provider.go#L503), which in turns calls `prov.StartProviding()`.

`StartProviding(h)` adds the multihash to the provide queue, and writes it (sync) to the `KeyStore`, using [`KeyStore.Put()`](https://github.com/libp2p/go-libp2p-kad-dht/blob/48cb56a62251f09736553fa58a863d17b98acc1d/provider/datastore/keystore.go#L252-L263), which also needs the `KeyStore` lock, leading to a deadlock.

This fix is ugly, alternative suggestions are welcome.

I haven't dug yet why generating the `KeyChanFunc` for MFS (indirectly) triggers provides, but maybe it is better to tackle the problem here.

@hsanjuan @gammazero thoughts?

hsanjuan: Is this only a Tests issue or a general issue? Can we incur into deadlocks during normal operation, or just initialization?

ok, so I think the `FlushMemFree` is so that all blocks are written to disk, since we are going to traverse from the MFS root reading from the blockstore, so we need everything written to disk otherwise we might get a failure.

I don't like this solution tbh... possible options (not sure how realistic):

1) Add a `flush` option to the mfsProvider to disable flush on specific calls (i.e. when initializing, since during initialization nothing should have been written to mfs in a way that it becomes cached). Or call it before ResetCids() has taken the lock.
2) Use a custom mfs initialized without a provider for this step so that it doesn't provide during FlushMemFree.

guillaumemichel: ## Context

It is a general issue. It happens every time the `KeyChanFunc` is run (garbage collection for CIDs that don't need to be advertised anymore).

The deadlock issue isn't disk related, but `KeyStore` (new provide system) related only. Because kubo doesn't know which CIDs shouldn't be advertised anymore at the moment they are removed, it is necessary to iterate over all the CIDs matching the provide strategy. The `KeyStore` containing the state of all multihashes that must be reprovided for the new provide system has to be periodically reset to match contain the correct multihashes.

## What happens

IIUC this is only triggered when the strategy is `pinned+mfs`, because we use the [`PrioritizedProvider`](https://github.com/ipfs/boxo/blob/7d2493027364721d58ce2f2897750af181ba8db5/provider/provider.go#L62), where MFS is second priority. 
* It means that we open the chan that is passed to [`KeyStore.ResetCids()`](https://github.com/libp2p/go-libp2p-kad-dht/blob/55678dbd6ffbac131ca8f80fd50a41dec3c354ae/provider/datastore/keystore.go#L112), we write the cids from the first priority `KeyChanFunc` (`pinned`), and then we run the MFS `KeyChanFunc` to get the MFS chan. 
* The MFS `KeyChanFunc` is run while the `KeyStore.ResetCids()` has already consumed the pinned CIDs from the supplied chan, hence `KeyStore.ResetCids()` holds the `KeyStore` mutex.
* The [MFS `KeyChanFunc`](https://github.com/ipfs/kubo/blob/3e1e7d17fb951575a975915cea6132857c769d01/core/node/provider.go#L171-L185) calls `FlushMemFree` on the MFS root, which triggers (at least one) `StartProviding`
* `StartProviding` needs to `Put` a key in the `KeyStore` needing the `KeyStore` mutex, hence deadlock

## Potential fixes

1. Do we need to provide cids to the network during `FlushMemFree` at all?
2. When defining [prioritized provider](https://github.com/ipfs/boxo/blob/7d2493027364721d58ce2f2897750af181ba8db5/provider/provider.go#L62), resolve the secondary chan before returning. (Implemented in https://github.com/ipfs/boxo/pull/995/commits/205b900d8ab7cace2f70da4d6c5d9f559995a0d0)

gammazero: Unless `FlushMemFree` takes a long time, I think it should be ok to not provide CIDs when it is running.

Do we need the prioritized provider? If it is not critical to provide pinned CIDs before MFS CIDs, and not using the prioritized would simplify things.

Another possibility is to get rid of the `KeyStore` mutex and instead have a `KeyStore` goroutine that everything communicates with.

guillaumemichel: > Unless `FlushMemFree` takes a long time, I think it should be ok to not provide CIDs when it is running.

I didn't really dig into it, but it seems weird that flushing memory can trigger network provides.

> Do we need the prioritized provider? If it is not critical to provide pinned CIDs before MFS CIDs, and not using the prioritized would simplify things.

We don't need the `pinned+mfs` `KeyChanFunc` to be prioritized, however we need to combine the `KeyChanFunc`s from `pinned` and `mfs`, which is what [`NewPrioritizedProviders`](https://github.com/ipfs/boxo/blob/7d2493027364721d58ce2f2897750af181ba8db5/provider/provider.go#L62) does. So it should be fine to keep using it.

> Another possibility is to get rid of the `KeyStore` mutex and instead have a `KeyStore` goroutine that everything communicates with.

It is a possibility, but I would prefer to avoid it as we loose consistency in `KeyStore` operations. When keys are `Put()` to the `KeyStore` the `KeyStore` returns the set of newly added keys, so that only new keys are advertised to the DHT Swarm.

I would rather avoid undefined behaviors of `Put()` and `Get()` operations during a `ResetCids()`.

provider/provider.go
```diff
@@ -63,23 +73,27 @@ func NewPrioritizedProvider(priorityCids KeyChanFunc, otherCids KeyChanFunc) Key
 	return func(ctx context.Context) (<-chan cid.Cid, error) {
 		outCh := make(chan cid.Cid)
 
+		priorityCh, err := priorityCids(ctx)
+		if err != nil {
+			return nil, err
+		}
+		otherCh, err := otherCids(ctx)
```
gammazero: Can calling `otherCids` wait until after the CIDs from `priorityCh` are handled? Consider a case where `otherCids` uses resources, acquires lock, etc. that is in conflict with what `priorityCids` does.

You could not return the error immediately, so that would have to be logged only or would need a channel if the caller needs to detect/handle the error.

provider/provider.go
```diff
@@ -63,23 +73,27 @@ func NewPrioritizedProvider(priorityCids KeyChanFunc, otherCids KeyChanFunc) Key
...
+		if err != nil {
+			return nil, err
+		}
+
```
gammazero: Instead of only having two `KeyChanFunc` parameters, what if you took a variadic argument:
```go
func NewPrioritizedProvider(cidsFuncs ...KeyChanFunc) KeyChanFunc {
```

Then you iterate through these calling `handleStream` on each:
```go
last := len(cidsFuncs) - 1
for i, cidsFunc := range cidsFuncs {
	cidsCh, err := cidsFunc(ctx)
	if err != nil {
		log.Errorw("Failed to get provider key channel", "err", err)
		continue
	}
	handleStream(cidsCh, i < last)
}
```

provider/provider.go
guillaumemichel: I share your concerns.

The current fix (that we can certainly improve), is to make sure both `KeyChanFunc`s are called before the function returns (not in the go routine). This implies that for MFS, `FlushFreeMem` is called before `KeyStore.ResetCids()`, so we have sequential operations in the `KeyStore`
1. `FlushFreeMem` -> `prov.StartProviding()` `KeyStore.Put()`
2. `KeyStore.Reset()` (simply reading the cids from the chan)

Because otherwise, `FlushFreeMem` (used in the MFS `KeyChanFunc`) is called during `KeyStore.Reset()`, after `KeyStore.Reset()` has consumed all the CIDs from the `pinned` `KeyChanFunc`. This triggers the deadlock.

---

The better solution is certainly to make sure that `FlushFreeMem()` doesn't end up calling `StartProviding()`, and if it needs to, then we need to find another way.

provider/provider.go
guillaumemichel: This makes sense. However it doesn't solve the issue, which is triggered because `KeyChanFunc`s calling `StartProviding()` should be executed before returning (to make sure they don't trigger provide operations while the returned channel is being consumed).

mfs/dir.go
guillaumemichel: The current fix (https://github.com/ipfs/boxo/commit/205b900d8ab7cace2f70da4d6c5d9f559995a0d0) isn't optimal as pointed out by @gammazero in https://github.com/ipfs/boxo/pull/995#discussion_r2332023978.

So the better solutions could be:
1. Make sure MFS `FlushFreeMem` doesn't call `StartProviding` (not sure of the implications?)
2. Don't use `FlushFreeMem` in the MFS `KeyChanFunc` (maybe use an alternative)
3. Remove the mutex on the `KeyStore` (we loose consistency)

It's a shame because this situation is the result of kubo not knowing which CIDs it should stop providing at the time they are removed. So we need to have manual garbage collection iterating over all content and calling `KeyStore.ResetCids()` periodically.

hsanjuan: Originally, the `Provide` call was meant to hit the reprovider, which would just queue the CID. It was meant to be pretty async. Shouldn't `StartProviding` just queue cids until the lock is free and it can proceed? Even without deadlocks, it sounds like there's potential for lock-contention and delays in some of the hot paths. ResetCids will take it's while, and we shouldn't be pausing everything that calls `Provide/StartProviding` in the meantime.

> I didn't really dig into it, but it seems weird that flushing memory can trigger network provides.

In MFS we are calling Provide() on every `dag.Put`. It expects the provider system to handle whether the provide is necessary or not and it assumes it is not going to block (everywhere we call Provide it is assumed this is going to be a cheap notification to the provider system, not a blocking operation).

provider/provider.go
hsanjuan: `StartProviding` should not wait until a lock is free. It should provide the CID as soon as possible, but not be a sync call. The previous `reprovider.Provide(...)` was async, so there's a behaviour change.

hsanjuan:
Colo discussion:

* Deal-breaker is that StartProviding() can be waiting for a lock for significant time while ResetCIDs() ongoing. StartProviding is called from everywhere that calls Provide() (blockstore, pinner, mfs, bitswap).
* -> StartProviding should be made async() somehow

Discussion is how to make StartProvidingAsync():
  * A) Launch a goroutine
  * B) Put StartProviding CIDs in a queue
  * C) Switch to a datastore-backed queue
    * For all ResetCIds() + StartProviding CIds as they are coming in
  * Study necessary
    * Remember: The ResetCids() set should be updated with any new StartProviding() cids before it replaces the "old" set.

guillaumemichel:
Reverted all deadlock related changes and applied @gammazero suggestion (https://github.com/ipfs/boxo/pull/995#discussion_r2332031254) in https://github.com/ipfs/boxo/pull/995/commits/4684512c9f3b4d0ba5ece04988e91503a774bb6e

mfs/dir.go
guillaumemichel: See https://github.com/libp2p/go-libp2p-kad-dht/issues/1145 and https://github.com/libp2p/go-libp2p-kad-dht/pull/1146

provider/provider.go
guillaumemichel: See https://github.com/libp2p/go-libp2p-kad-dht/issues/1145 and https://github.com/libp2p/go-libp2p-kad-dht/pull/1146

This will be handled in go-libp2p-kad-dht, not in boxo

provider/provider.go
guillaumemichel: https://github.com/ipfs/boxo/pull/995/commits/4684512c9f3b4d0ba5ece04988e91503a774bb6e

gammazero (approved):
LGTM

hsanjuan:
Triage:
  * @gammazero will re-review.

gammazero (approved):

PR #997, opened 2025-08-13
feat: autoconf client library

lidel:
This PR adds `boxo/autoconfig` client library, cleaned up and extracted from https://github.com/ipfs/kubo/pull/10883

Used in:
- https://github.com/ipfs/kubo/pull/10883
- https://github.com/ipfs/rainbow/pull/286


autoconf/client.go
```diff
@@ -0,0 +1,283 @@
...
+// WithRefreshInterval sets the refresh interval for autoconf data
+func WithRefreshInterval(interval time.Duration) Option {
+	return func(c *Client) error {
+		if interval <= 0 {
+			return fmt.Errorf("refresh interval must be positive")
+		}
+		c.refreshInterval = interval
+		return nil
+	}
+}
```
hsanjuan, line 211: It would be good to introduce a minimum refresh interval that is hardcoded. We don't want people setting their refresh interval to 1s by mistake.

autoconf/fetch.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// fetchFromRemoteRaw fetches config from remote URL (internal helper)
+func (c *Client) fetchFromRemoteRaw(ctx context.Context, configURL, cacheDir string) (*http.Response, *Config, error) {
+	// Validate URL scheme for security
+	if err := c.validateURL(configURL); err != nil {
+		return nil, nil, fmt.Errorf("invalid config URL: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, configURL, nil)
```
hsanjuan, line 211: should we set `req.Close = true` ? does not seem it makes sense to keep the connection around.

gammazero (approved):

autoconf/client.go
lidel: Is tricky because we have tests that depend on very low interval to confirm things work. 

What is more idiomatic, use `testing.Testing()` to detect test mode and lift min interval limit, or have explicit `WithTestMode()`?

autoconf/lifecycle.go
```diff
@@ -0,0 +1,65 @@
...
+func (c *Client) Stop() {
+	c.updaterMu.Lock()
+	defer c.updaterMu.Unlock()
+
+	if c.updater != nil {
+		c.updater.Stop()
+		c.updater = nil
+		log.Infof("Stopped autoconf background updater")
+	}
+}
```
hsanjuan, line 84: If we can control lifecycle via context, perhaps we can avoid having a second way?

hsanjuan (approved):
Thank you for writing such good code!

autoconf/lifecycle.go
gammazero: I prefer to avoid lifecycle contexts if at all possible, and if they are the best solution, then do not make the visible outside the package and the context is not kept in the type. In other words, only keep the cancel function as a data member in the type and pass the context into a goroutine started when the type instance is created.

autoconf/fetch.go
lidel: Done in https://github.com/ipfs/boxo/pull/997/commits/4363ad0fb4c95f8fd185d5a691ced7901f2b386c

PR #1001, opened 2025-08-15
blockstore: remove HashOnRead api

### LINKED ISSUE #365: blockstore: remove HashOnRead
Filling issue based on poc code in https://github.com/ipfs/go-ipfs-blockstore/pull/51:
 
> Instead, provide a simple `ValidatingBlockstore` wrapper. This:
> 
> 1. Isn't racy (fixes https://github.com/ipfs/go-ipfs-blockstore/issues/49).
> 2. Keeps the base blockstore simple.
> 

> :point_right:  Really, "HashOnRead" should never have been a part of the blockstore API in the
> first place.

gammazero:
Instead, provide a simple `ValidatingBlockstore` wrapper. This keeps the base blockstore simple.

Really, `HashOnRead` should never have been a part of the blockstore API in the first place.

Closes #365

Kubo PR: https://github.com/ipfs/kubo/pull/10917

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The following emojis are used to highlight certain changes:
     - `ipfs_http_gw_responses_total{code}`: Counter for all HTTP responses by status code
     - `ipfs_http_gw_retrieval_timeouts_total{code,truncated}`: Counter for retrieval timeout events with details on truncation
 - `namesys/IPNSPublisher`: option to `PublishOptions` that allows for setting a custom sequence number for the IPNS record with proper validation to prevent unintentional replay attacks. [#962](https://github.com/ipfs/boxo/pull/962)
+- `blockstore`: Added `ValidatingBlockstore` wrapper. This replaces the `HashOnRead` blockstore API.
 
 ### Changed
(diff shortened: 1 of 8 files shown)


hsanjuan:
I am of the opinion that we have way too many blockstores that wrap each others, so while I would remove this from the API and make it an option, I would not add a new blockstore wrapper (this is why I added the provider option to the existing blockstore rather than make another wrapper).

As recap:
  * Blockstore
  * Idstore
  * CachedBlockstore
  * (potentially Verifying blockstore)
  * Also GCBlockstore..

I find it not very dev-friendly... i don't know about the perf impact of jumping through 3 or 4 abstractions.

hsanjuan:
(not blocking merge)

gammazero:
@hsanjuan The `ValidatingBlockstore` is only used to wrap the datastore when `HashOnRead` is set to `true`, and is less overhead than atomically checking if hashOnRead is set for every call to Get. ... mostly an advantage because the wrapper is generally not used at all.

So, while `ValidatingBlockstore` is yet another wrapper, with too many already, it is better than the `HashOnRead` implementation and is not used unless the blockstore is being validated.

gammazero:
An alternative might be to have a `GetValidate` function in the go-datastore package the computes and checks the hash. The configuration of HashOnRead would determine whether to call `ds.Get` or `datastore.GetValidate(ds)`.

I think, in that case, yet another wrapper is marginally better because it is completely out of the way an unused unless there is a special circumstance where the data is being validated.

I agree with your points, but I do not have a better solution that still provides the HashOnRead functionality in kubo that may be needed by some users, and I think removing the HashOnRead API is worth it.

PR #1007, opened 2025-08-19
bitswap/client: GetBlocks cancels session when finished

### LINKED ISSUE #93: [ipfs/go-bitswap] GetBlocks doesn't cancel session unless the passed context is canceled.
We should make sure the context is canceled when the channel returned by GetBlocks is closed.

gammazero:
Fixes #93

The `client.GetBlocks` function now closes the session when the session `session.GetBlocks` is done.

### DIFF
--- a/bitswap/client/client.go
+++ b/bitswap/client/client.go
@@ -436,14 +436,46 @@ func (bs *Client) GetBlock(ctx context.Context, k cid.Cid) (blocks.Block, error)
 // [github.com/ipfs/boxo/bitswap/client/traceability.Block] assertable
 // [blocks.Block].
 //
-// NB: Your request remains open until the context expires. To conserve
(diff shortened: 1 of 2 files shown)


hsanjuan (approved):

gammazero:
I am thinking this is actually the wrong solution. The bitswap client's `GetBlock` and `GetBlocks` really just create a temporary session inside them. The bitswap client can also be used to get a new session. This is really not a good interface design IMO.

The solution is to get rid of the bitswap client's `GetBlock` and `GetBlocks` and let the caller do this using a session that they get from the client. This way there is only one thing that provides the `GetBlocks` API, and it is access the say way in whatever code needs to use it.

This also avoids the need to create an extra channel/goroutine just to manage the lifetime of the temporary session.

PR #1011, opened 2025-08-21
refactor(bitswap/client/internal): close session with Close method instead of context

### LINKED ISSUE #1009: bitswap/client: sessions should close via a close method, not a context
Bitswap sessions should close using a `Close` method, and not a context. A context should be used for request cancelation, not for session lifetime.

gammazero:
Closes #1009

### DIFF
--- a/bitswap/client/internal/session/session.go
+++ b/bitswap/client/internal/session/session.go
@@ -96,7 +96,7 @@ type op struct {
 type Session struct {
 	// dependencies
 	ctx            context.Context
-	cancel         func()
+	cancel         context.CancelFunc
 	sm             SessionManager
 	pm             PeerManager
 	sprm           SessionPeerManager
--- a/bitswap/client/internal/session/session.go
+++ b/bitswap/client/internal/session/session.go
@@ -141,7 +141,7 @@ func New(
 	periodicSearchDelay delay.D,
 	self peer.ID,
 	havesReceivedGauge bspm.Gauge,
-	retrievalState *retrieval.RetrievalState,
+	retrievalState *retrieval.State,
 ) *Session {
 	ctx, cancel := context.WithCancel(context.Background())
 
--- a/bitswap/client/internal/session/session.go
+++ b/bitswap/client/internal/session/session.go
@@ -10,6 +10,7 @@ import (
 	notifications "github.com/ipfs/boxo/bitswap/client/internal/notifications"
 	bspm "github.com/ipfs/boxo/bitswap/client/internal/peermanager"
 	bssim "github.com/ipfs/boxo/bitswap/client/internal/sessioninterestmanager"
+	"github.com/ipfs/boxo/retrieval"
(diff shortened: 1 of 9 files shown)


bitswap/client/internal/session/session.go
```diff
@@ -96,7 +96,7 @@ type op struct {
 type Session struct {
 	// dependencies
 	ctx            context.Context
-	shutdown       func()
+	cancel         func()
```
hsanjuan:
```suggestion
	cancel         context.CancelFunc
```

hsanjuan (commented):
OK code-wise... but I'm not 100% onboard with this change.

I think it is nice that you can have a "request", that has a context, and you can create a session to do things that you need re-using that context, and that the session is cancelled if the request is cancelled.

Then we have things like:

```
// NewSession returns a session backed NodeGetter if the given NodeGetter
// implements SessionMaker.
func NewSession(ctx context.Context, g ipld.NodeGetter) ipld.NodeGetter {
	if sm, ok := g.(SessionMaker); ok {
		return sm.Session(ctx)
	}
	return g
}
```
(https://pkg.go.dev/github.com/ipfs/boxo/ipld/merkledag#NewSession)

which is wired onto blockservice.Session() etc. This will no longer be able to return a session-enabled NodeGetter, but will rather have to return a closable NodeGetter.

 Adding `Close()` will need to update DAGService interfaces.

Upgrade will be tricky, need user intervention to call Close().

All in all I think the direction here isn't good and that having sessions controlled via contexts, in particular, is not overly problematic given their ephemeral nature.

hsanjuan (approved):
Sorry, I believe that this is unrelated to `blockservice.Session`, which is the Session affecting merkledag etc. So this is only bitswap-internal. Sorry for the wall text.

gammazero:
> I believe that this is unrelated to blockservice.Session, which is the Session affecting merkledag etc

Correct, but sessions are still ephemeral in nature, so your comments may still apply. Also, do we still want to control bitswap session lifetimes with a context to be consistent with how blockservice sessions are handled?

I will let this PR sit here for a bit so we and @lidel  can think about it more. It is not hight priority.

gammazero:
@lidel Does a session need to accept a context to maintain retrieval state?

lidel:
@gammazero yes, context is how we pass state across the stack, search where `retrieval.StateFromContext` is used: https://github.com/ipfs/boxo/blob/8b6b1d276e4345086b02fbdcc890ff9411efe801/retrieval/state.go#L202-L217

This feels like the least invasive way (optional, no overhead if not in context), but we can change it if there is better way (i did not see it at the time).

If we remove context, we need to explicitly pass RetrievalState to session constructor etc.

gammazero:
@lidel Updated to give session retrieval state.

lidel (approved):
Lgtm, pushed some comments about `Close()` being still automated when `NewSession(ctx)` is used. Feel free to adjust, otherwise ok to merge.

PR #1012, opened 2025-08-25
experiment: move to quartz for time mocking

gammazero:
Move from `filcoin-project/go-clock` to `quartz` for time mocking. The quartz package allows better control over timer events. The `quartz` package allows better control over timer events.


bitswap/server/internal/decision/engine_test.go
```diff
@@ -1433,32 +1433,38 @@ func TestTaggingUseful(t *testing.T) {
...
 
 	sampleCh := make(chan struct{})
-	mockClock := clock.NewMock()
+	mockClock := quartz.NewMock(t)
+	trap := mockClock.Trap().NewTicker("scoreWorker")
+	defer trap.Close()
 	me := newTestEngineWithSampling("engine", peerSampleIntervalHalf*2, sampleCh, mockClock)
 	defer me.Engine.Close()
-	mockClock.Add(1 * time.Millisecond)
+	mockClock.Advance(time.Millisecond)
```
guillaumemichel:
```suggestion
```

we already wait below for the ticker to be created.

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1433,32 +1433,38 @@ func TestTaggingUseful(t *testing.T) {
...
+	// Wait for ticker to be created to ensure clock.AdvenceNext works.
+	ctx := context.Background()
+	trap.MustWait(ctx).MustRelease(ctx)
+
 	for i := 0; i < 3; i++ {
 		if untagged := me.PeerTagger.count(me.Engine.tagUseful); untagged != 0 {
 			t.Fatalf("%d peers should be untagged but weren't", untagged)
 		}
-		mockClock.Add(peerSampleIntervalHalf)
+		mockClock.Advance(peerSampleIntervalHalf)
```
guillaumemichel:
```suggestion
		mockClock.Advance(peerSampleIntervalHalf).MustWait(ctx)
```

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1433,32 +1433,38 @@ func TestTaggingUseful(t *testing.T) {
...
 		me.Engine.MessageSent(friend, msg)
 
-		mockClock.Add(peerSampleIntervalHalf)
+		mockClock.AdvanceNext()
```
guillaumemichel:
```suggestion
		advanced, w := mockClock.AdvanceNext()
		if advanced != peerSampleIntervalHalf {
			t.Fatalf("expected to advance %s, but advanced %s", peerSampleIntervalHalf, advanced)
		}
		w.MustWait(ctx)
```

maybe asserting the advanced interval is not necessary.

bitswap/server/internal/decision/engine_test.go
```diff
@@ -1433,32 +1433,38 @@ func TestTaggingUseful(t *testing.T) {
...
 		<-sampleCh
 
 		if tagged := me.PeerTagger.count(me.Engine.tagUseful); tagged != 1 {
 			t.Fatalf("1 peer should be tagged, but %d were", tagged)
 		}
 
 		for j := 0; j < longTermRatio; j++ {
-			mockClock.Add(peerSampleIntervalHalf * 2)
+			mockClock.AdvanceNext()
```
guillaumemichel:
```suggestion
			advanced, w := mockClock.AdvanceNext()
			if advanced != 2*peerSampleIntervalHalf {
				t.Fatalf("expected to advanced %s, but advanced %s", peerSampleIntervalHalf, advanced)
			}
			w.MustWait(ctx)
```
maybe asserting the advanced interval is not necessary.

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
```diff
@@ -72,6 +72,26 @@ func (tr *timeoutRecorder) clear() {
...
+// clock.AdvanceNext to provide better control over time events.
+func quartzAdd(c *quartz.Mock, d time.Duration) {
+	for d > 0 {
+		next, ok := c.Peek()
+		if !ok || next >= d {
+			c.Advance(d)
+			time.Sleep(time.Millisecond)
+			return
+		}
+		<-c.Advance(next).Done()
```
guillaumemichel, line 88: any reason not to use `MustWait`?

```suggestion
		c.Advance(next).MustWait(context.Background())
```

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
```diff
@@ -72,6 +72,26 @@ func (tr *timeoutRecorder) clear() {
...
+		d -= next
+		// give some real time to process after timer has popped.
+		time.Sleep(time.Millisecond)
```
guillaumemichel, line 91: IIUC this isn't needed since we `Done()`/`MustWait()` returns only when all tickers/timers are done running.
```suggestion
```

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
```diff
@@ -72,6 +72,26 @@ func (tr *timeoutRecorder) clear() {
...
+// happen before or at the final time.
+//
+// TODO: Replace calls to quartzAdd with calls to clock.Advance and
+// clock.AdvanceNext to provide better control over time events.
+func quartzAdd(c *quartz.Mock, d time.Duration) {
+	for d > 0 {
+		next, ok := c.Peek()
+		if !ok || next >= d {
+			c.Advance(d)
+			time.Sleep(time.Millisecond)
```
guillaumemichel, line 85:
```suggestion
			c.Advance(d).MustWait(context.Background())
```

`MustWait()` returns once all triggered events are complete, which is faster to run and more accurate than sleeping.

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
```diff
@@ -72,6 +72,26 @@ func (tr *timeoutRecorder) clear() {
 	tr.timedOutKs = nil
 }
 
+// quartzAdd advances the time by the given duration through any events that
+// happen before or at the final time.
+//
+// TODO: Replace calls to quartzAdd with calls to clock.Advance and
+// clock.AdvanceNext to provide better control over time events.
```
guillaumemichel, line 79: In the [docs](https://github.com/coder/quartz?tab=readme-ov-file#advance-only-to-the-next-event) it is mentioned:

> Advancing multiple events can be accomplished via looping

Hence, if we don't need to assert every single ticker/timer event that is triggered during the execution, we can keep this loop.

Alternatively, we would have to `AdvanceNext()` to every triggered event (not sure how many there are), or maybe we could make use of [Traps](https://github.com/coder/quartz?tab=readme-ov-file#traps), IIUC this means adding [Tags](https://github.com/coder/quartz?tab=readme-ov-file#tags) to the non-test code.

gammazero: This function avoids a lot of complication by advancing over any events that are not necessary to assert, including those that are created while advancing. It also handles advancing the clock when there are no more events to advance to. So, yes, let's keep it.

Seems useful enough to ask quartz to provide, but I suspect the would want to discourage it as tell us that we are welcome to do it ourselves (which their docs basically do already).

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
gammazero: I needed to use `time.Sleep` here. When I use `MustWait` that causes tests to occasionally hang forever. I _think_ that `MustWati` is waiting for completion of an `AfterFunc` that itself is waiting for another timer. Using `time.Sleep` here avoids that situation.

Please feel free to investigate this, because using `MustWait` would certainly be preferable.

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
gammazero: Same as above.

bitswap/client/internal/messagequeue/donthavetimeoutmgr_test.go
gammazero: Tests hangs without it. See above explanation and feel free to investigate.

bitswap/server/internal/decision/engine_test.go
gammazero: Probably good to assert the the interval here. However, the MustWait is not needed as the `<-sample` below will do the waiting.

bitswap/server/internal/decision/engine_test.go
gammazero: Wait not needed since `<-sampleCh` does that.

gammazero:
We have decided to move to `testing/synctext` when this module requires go.1.25.

PR #1013, opened 2025-08-26
remove sha1 from list of hashes allowed in IPFS

gammazero:
Closes kubo issue https://github.com/ipfs/kubo/issues/8703

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The following emojis are used to highlight certain changes:
 
 ### Removed
 
+- remove sha1 from list of hashes allowed in IPFS [#1013](https://github.com/ipfs/boxo/pull/1013)
+ 
 ### Fixed
 
 ### Security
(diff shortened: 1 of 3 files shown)


gammazero:
Still needed  for git compatibility.
- Blocked until git 3 is released

hsanjuan (approved):

lidel (changes requested):
SHA1 creates interesting question: what do we do with hash functions which are no longer considered secure, but there are DAGs created with it. i imagine we want to retain ability to read historical DAGS, but not create new ones?

It feels what we want here, is to have separate allowlist for reading (and sha1 would be allowed, due to historical use in Git DAGs), and separate for writing (we would refuse to create new dags with sha1).

right now boxo/verifcid has `allowlist.IsAllowed` which covers both.

does it make sense to add `IsWriteAllowed` and `IsReadAllowed` and switch to them in code so we allow reads of Sha1 but refuse writes? (keeping `IsAllowed=IsWriteAllowed && IsReadAllowed` for backward compatibility).

PR #1014, opened 2025-08-26
refactor: use the built-in max/min to simplify the code [skip changelog]

wmypku:
Inspired by https://github.com/ipfs/boxo/pull/941 and replace all.


gammazero:
Handled in #1017

PR #1015, opened 2025-08-27
feat(gateway): add retrieval state tracking for timeout diagnostics

### LINKED ISSUE #14: Update error pages on the gateways to surface debugging information — IPFS/2025
# Description

Update error pages on the gateways to surface information that helps to troubleshoot errors.

- Move generation of timeouts from Nginx to `boxo/gateway`
  - prerequisite: synthesize timeout and in-flight request limits (there are PRs, but they need to be tackled together):
    - https://github.com/ipfs/boxo/issues/679
    - https://github.com/ipfs/boxo/issues/881
  - dweb.link/ipfs.io: make `boxo/gateway` limits engage before nginx ones
- Research how difficult it is for `boxo/gateway` to surface more details why error occurred (lack of providers? no online provider? lack of bitswap response? other?)

## ETA
Q2-Q3

## Progress

- Q3: 
  - https://github.com/ipfs/boxo/pull/994
  - https://github.com/ipfs/boxo/pull/1015
  - https://github.com/ipfs/boxo/pull/1023
  - https://github.com/ipfs/kubo/pull/10950

lidel:
This PR adds diagnostic tracking for IPFS content retrieval to provide meaningful error context when timeouts occur. This helps users understand exactly where and why retrieval failed.

We can refine presentation in future PRs.

- Closes https://github.com/ipshipyard/roadmaps/issues/14

## TLDR

- New boxo/retrieval package with `RetrievalState` for tracking retrieval phases
- Tracks provider discovery stats (found/attempted/connected)
- Records failed provider peer IDs for debugging
- Monotonic phase progression through retrieval stages
- Thread-safe concurrent access with atomic operations

### Integration points

The high level flow is:
1. Gateway handler creates context with `RetrievalState`: `retrieval.ContextWithState(r.Context())`
2. Gateway wraps context for request with session: `blockservice.ContextWithSession(ctx, bb.blockService)`
3. When blocks are fetched, blockservice creates exchange session with this context: `sesEx.NewSession(s.sesctx)`
4. Bitswap session manager creates session with this context
5. When the bitswap session calls `FindProvidersAsync`, it passes this context
6. `ProviderQueryManager` receives the context and can access the `RetrievalState`: `retrieval.StateFromContext(ctx)`
7. The `ProvidersFound.Add(1)` call in `providerquerymanager` updates the `RetrievalState` that will be included in the 504 timeout error message etc
8. Retrieval continues (or timeouts)  
9. The `RetrievalState.Summary()` provides human-readable diagnostics like seen in "Demo" section below

In practice, I had to make sure we manually pass/preserve state in two places, where `context.Background()` is used instead of `r.Context()` of the HTTP request:  
1. In `routing/providerquerymanager/providerquerymanager.go`: when creating a new context for provider queries, now copies the RetrievalState from the original request context
2. In `bitswap/client/client.go`: when creating a session context in GetBlocks(), now preserves the RetrievalState from the request context

## Demo

Basic test where there are no providers for a CID:

> Screen Shot 2025-08-27 at 04 47 27

Extra test where there was one provider in DHT, but I've shut it down, so Gateway found it, but failed to connect. We show up to 3 sample peers that failed:

> Screen Shot 2025-08-27 at 05 00 53

| Before | After |
| ---- | ---- |
| image | image |

### DIFF
--- a/retrieval/state.go
+++ b/retrieval/state.go
@@ -7,6 +7,7 @@ package retrieval
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sync"
@@ -146,11 +147,11 @@ func (rs *RetrievalState) Summary() string {
 	phase := rs.GetPhase()
 
 	if found == 0 {
-		return fmt.Sprintf("No providers found for the CID (phase: %s)", phase.String())
+		return fmt.Sprintf("no providers found for the CID (phase: %s)", phase.String())
 	}
 
 	if attempted == 0 {
-		return fmt.Sprintf("Found %d provider(s) but none could be contacted (phase: %s)", found, phase.String())
+		return fmt.Sprintf("found %d provider(s) but none could be contacted (phase: %s)", found, phase.String())
 	}
 
 	failedProviders := rs.GetFailedProviders()
@@ -165,16 +166,16 @@ func (rs *RetrievalState) Summary() string {
 	}
 
 	if connected == 0 {
-		return fmt.Sprintf("Found %d provider(s), attempted %d, but none were reachable (phase: %s%s)",
+		return fmt.Sprintf("found %d provider(s), attempted %d, but none were reachable (phase: %s%s)",
 			found, attempted, phase.String(), failedPeersInfo)
 	}
 
 	if len(failedProviders) > 0 {
-		return fmt.Sprintf("Found %d provider(s), connected to %d, but they did not return the requested content (phase: %s%s)",
+		return fmt.Sprintf("found %d provider(s), connected to %d, but they did not return the requested content (phase: %s%s)",
 			found, connected, phase.String(), failedPeersInfo)
 	}
 
-	return fmt.Sprintf("Timeout occurred after finding %d provider(s) and connecting to %d (phase: %s)",
+	return fmt.Sprintf("timeout occurred after finding %d provider(s) and connecting to %d (phase: %s)",
 		found, connected, phase.String())
 }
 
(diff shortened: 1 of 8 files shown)


retrieval/state.go
```diff
@@ -0,0 +1,220 @@
...
+	// Track specific provider failures (limited to first few for brevity)
+	failedProviders []peer.ID
+}
+
+// NewRetrievalState creates a new RetrievalState initialized to PhaseInitializing.
+// The returned state is safe for concurrent use.
+func NewRetrievalState() *RetrievalState {
+	rs := &RetrievalState{
+		failedProviders: make([]peer.ID, 0, MaxFailedProvidersToTrack),
+	}
```
gammazero: I think `failedProviders` would be better left `nil` and append failed providers only when they happen.
```suggestion
	rs := &RetrievalState{}
```

retrieval/state.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// GetFailedProviders returns the list of failed providers
+func (rs *RetrievalState) GetFailedProviders() []peer.ID {
+	rs.mu.RLock()
+	defer rs.mu.RUnlock()
+	result := make([]peer.ID, len(rs.failedProviders))
+	copy(result, rs.failedProviders)
+	return result
```
gammazero: `slices.Clone` is generally preferred to the `make` + `copy` pattern. Efficiency/performance is same.
```suggestion
	return slices.Clone(rs.failedProviders)
```

gammazero (approved):
Only a couple minor comments.

hsanjuan (approved):
ok code-wise...

I have some questions on handling of failures... for example, the resolution phase can trigger block fetches to be able to resolve a path, but I'm guessing a failure to do so will not provide information about failed peers etc ?

 Also, PhaseDataRetrieval is only set once some bytes have started flowing. I don't think we can show a custom error page with info from that moment, so this phase would not be seen by the user right?

There is a rather large gap between "Connected" and "DataRetrieval". I.e. the connected peer may fail to send the block, or the HTTP request may 404 even if if the provider record pointed to someone. That makes me wonder if there should be another step along the lines of `DataRequested`, but I'm guessing that pass the StateTracking down to that level as CIDs from many sessions are mashed up into wantlists.

lidel:
@gammazero addressed feedback, thanks!

@hsanjuan good catch: the resolution phase can trigger block fetches, and the code was already correctly tracking it, we just did not forward the state from context to the final error message. Fixed in ab647d57e8de1f219cd241dac6f1f79a86c3ab82 by opportunistically wrapping errors with extra diagnostics (if present in context), and then fixed it further in https://github.com/ipfs/boxo/pull/1023 by tracking found peers separately from ones that explicitly failed, and use them when timeout hits without ANY peer being reachable. 

> Also, PhaseDataRetrieval is only set once some bytes have started flowing.

correct, the last phase is there mostly for completeness 

This PR is just beginning:
- In the future we may use it. not sure for what, maybe to provide more meaningful truncation error when we timeout in the middle of a big CAR stream (`?format=car`) when some of internal blocks can't be retrieved and HTTP 200 response stream dies. 
- Future PRs will narrow the gap between "Connected" and "DataRetrieval"
- There will also be better UX with buttons to "Retry" (force page reload) and "Inspect CID at check.ipfs.network". I have local working code, but will submit it after this one lands, to make reviews easier. 
  -  **Updated:** https://github.com/ipfs/boxo/pull/1023

lidel:
Just to be sure this did not break anything, created PR in Kubo with this PR:

- https://github.com/ipfs/kubo/pull/10950


Merging.


**Update:** continued in:

- https://github.com/ipfs/boxo/pull/1023

PR #1018, opened 2025-08-28
fix(verifcid): enforce size limit for identity CIDs

lidel:
> [!IMPORTANT]
> Feedback on the default in that PR would be appreciated – it will likely act as ecosystem-wide lowest-common-denominator.

This PR aims to kick-off discussion about big identity CID that currently are allowed by the `boxo/verifcid` check and can be loaded by `boxo/gateway.

It also aims to address https://github.com/ipfs/kubo/issues/6011.

## Problem

Identity CIDs (multihash code 0x00) are unique in that they inline data directly into the CID itself rather than hashing it. While this can be useful for very small data that would be larger to reference than to inline, the lack of size limits created several problems:

1. **Abuse potential:** Without limits, identity CIDs could embed arbitrary amounts of data, effectively turning CIDs from content addresses into data containers.
2. **Gateway resource waste:** Poorly written clients could encode large payloads as identity CIDs and echo them through gateways, consuming bandwidth and resources without providing value.
3. **Unclear boundaries:** The ecosystem lacked clear guidelines on when identity CIDs are appropriate, leading to potential misuse. We see people building on shaky foundations "because it works" ([example](https://discuss.ipfs.tech/t/cidv1-eth-onchain-raw-cidv1-library/19444/5))
4. **Security concerns:** Identity CIDs provide no integrity verification (vulnerable to bit flips), making large identity CIDs particularly risky.
5. https://github.com/ipfs/kubo/issues/6011 :point_left:  this is a real BUG: `DagModifier` preserves the original CID prefix (including identity hash type) which means we can end up with a really big CID

Historical discussions:
- https://github.com/ipfs/kubo/issues/4918
- https://github.com/multiformats/cid/issues/21
- https://github.com/multiformats/multihash/issues/130
  - @rvagg  noted <=126  bytes should be ok in https://github.com/multiformats/multihash/issues/130#issuecomment-1233727089
  - .. but also higher 2048 value was mentioned in the same thread
- https://github.com/ipfs/kubo/pull/5281

## This PR  

- [x] Limit identity CIDs to 128 bytes maximum to prevent abuse. 
  - The limit of 128 bytes matches the existing `boxo/verifcid.DefaultMaxDigestSize` that was already applied to digests produced by supported cryptographic functions.
  - :point_right: **The value 128 aims to kick-off discussion.** If this is too low, we could introduce separate, higher limit just for `identity` multihashes. But we need SOME OFFICIAL LIMIT, at least for `boxo/gateway` and `DagModifier`. Feedback welcome.
- [x] `DagModifier` respects `boxo/verifcid.DefaultMaxIdentityDigestSize` and switches to non-identity on mutation that would overflow the limit
  - Note this is relatively niche fix. It does not impact `ipfs add`, only the niche `ipfs files write` which acts as append operation or people who use boxo to manually mutate DAGs that already had inlined CIDs. 
    - Before this PR: unbound growth of inlined CIDs (https://github.com/ipfs/kubo/issues/6011)
    - This PR: inlining is limited to `boxo/verifcid.DefaultMaxIdentityDigestSize`, then inherits parent CIDs hash function, or the default.

### Changes

- **BREAKING CHANGE:** CIDs with identity multihash digests longer than `boxo/verifcid.DefaultMaxIdentityDigestSize`  now produce `ErrAboveMaxDigestSize` error.
- renamed errors for consistency
- made `MaxDigestSize` public and reused in `boxo/bitswap`
- additional regression tests in blockstore an gateway just to be sure limit is respected even if we ever refactor or remove `boxo/verifcid`

## Alternative

We could also limit this to `boxo/gateway`, and enforce identity digest length to 128 there, while allowing it everywhere else. I think this is leaving surface for abuse and bugs, better to be consistent across the stack, but happy to be convinced otherwise.


verifcid/cid.go
```diff
@@ -26,12 +28,12 @@ func ValidateCid(allowlist Allowlist, c cid.Cid) error {
...
-	if pref.MhType != mh.IDENTITY && pref.MhLength < minimumHashLength {
-		return ErrBelowMinimumHashLength
+	if pref.MhType != mh.IDENTITY && pref.MhLength < MinDigestSize {
+		return ErrBelowMinDigestSize
 	}
 
-	if pref.MhType != mh.IDENTITY && pref.MhLength > maximumHashLength {
-		return ErrAboveMaximumHashLength
+	if pref.MhLength > MaxDigestSize {
+		return ErrAboveMaxDigestSize
```
lidel: ℹ️  this is TLDR of this PR

in the past, we excluded identity hashes from this max length check

now we treat them the same way as other multihashes (and enforce the limit)

verifcid/cid.go
```diff
@@ -8,15 +8,22 @@ import (
 	mh "github.com/multiformats/go-multihash"
 )
 
+const (
+	// MinDigestSize is the minimum size for hash digests (except for identity hashes)
+	MinDigestSize = 20
+	// MaxDigestSize is the maximum size for all digest types (including identity)
+	MaxDigestSize = 128
```
aschmahmann: Mutable globals generally don't feel good, particularly at a library layer (e.g. it hurts composability / interactions between multiple projects using the library). The reason we have passable Allowlist's at all is to not have globals.

aschmahmann: My read of https://github.com/multiformats/multihash/issues/130#issuecomment-1233727089 and https://github.com/multiformats/multihash/issues/130#issuecomment-1275659006 was that at that time 2048 was deemed probably a safe number that was unlikely to break people. Unclear what that number is today.

128 may actually be fine too in practice, but I'd probably want to broadcast this more widely (and flag in js-multiformats as well) before merging with a smaller number like 128.

gateway/handler.go
```diff
@@ -313,6 +314,13 @@ func (i *handler) getOrHeadHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Validate the root CID before any backend calls
+	rootCid := rq.immutablePath.RootCid()
+	if err := verifcid.ValidateCid(verifcid.DefaultAllowlist, rootCid); err != nil {
```
aschmahmann: Why does this need to happen before / what's the bad thing that happens if we rely on the blockservice, etc. used by the backend the user has passed in?

Asking both to understand and because the more places we hardcode globals like verifcid.DefaultAllowlist the harder / more hidden complexity we put into the system.

verifcid/cid.go
lidel: good point, I'll refactor this to use Allowlist and see how that would look like

lidel: yes, i will ping stakeholders once its ready for review

gateway/handler.go
lidel: just a precaution while i write tests.

before making this PR ready for review i'll remove it and see if blockstore is enough

verifcid/cid.go
lidel: @aschmahmann take a look, added this to Allowlist, and moved these consts to only act as `Default*` anchors that everyone can override in their own Allowlist implementation.

We also have separate `verifcid.DefaultMaxIdentityDigestSize` which right now is 128, but we can make it something different than the cryptographic `verifcid.DefaultMaxDigestSize`

rvagg:
Couple more references that contribute to the size discussion, as long as we're documenting some of this in one place:

* https://github.com/ipfs/go-cid/pull/127 - 32MiB limit on CIDs in general, way more than being discussed here, but interesting as long as we're discussing limits
* https://github.com/ipld/go-car/security/advisories/GHSA-9x4h-8wgm-8xfg - sizes in general and putting limits, section limits defaulting to 8MiB puts a hard limit on identity CIDs in CARs of somewhere a little below that.
* https://github.com/ipld/go-car/pull/239 "2KiB as a sufficiently large default" for CIDs in CARv2 indexes

FWIW 128 seems reasonable to me. I'm happy to have them squished down their happy-path use to a size where they're more likely being used for their size-saving utility rather than some other hacky purpose that stretches the whole stack. The higher you go, the harder it is to justify their use IMO.

vmx:
I'm not a fan of large identity CIDs. They are a special case (and a pain to deal with within a code base) and are mostly used for things that should be solved differently. I see the use case of the data is smaller than the hashed value. So for me, the smaller the value the better. Hence 128 bytes sound reasonable to me.

gateway/handler.go
lidel: removed in 373354403a530fc2fe499111ae51fb8b2baeba1b

verifcid/cid.go
lidel: ok, made this PR ready for review + pinged various stakeholders (old issues, ipfs-implementers slack, etc).

verifcid/cid.go
```diff
@@ -8,15 +8,28 @@ import (
...
+	// DefaultMinDigestSize is the default minimum size for hash digests (except for identity hashes)
+	DefaultMinDigestSize = 20
+	// DefaultMaxDigestSize is the default maximum size for cryptographic hash digests.
+	// This does not apply to identity hashes which are not cryptographic and use DefaultMaxIdentityDigestSize instead.
+	DefaultMaxDigestSize = 128
+	// DefaultMaxIdentityDigestSize is the default maximum size for identity CID digests.
+	// Identity CIDs (with multihash code 0x00) are not cryptographic hashes - they embed
+	// data directly in the CID. This separate limit prevents abuse while allowing
+	// different size constraints than cryptographic digests.
+	DefaultMaxIdentityDigestSize = 128
```
lidel, line 21: @achingbrain would these work for Helia? I'd like to have this documented at https://specs.ipfs.tech/ at some point, but first we need to align GO and JS implementations.

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+		// encode to just under DefaultMaxIdentityDigestSize when combined with metadata
+		initialData := makeTestDataPattern(verifcid.DefaultMaxIdentityDigestSize-10, "abcdefghijklmnopqrstuvwxyz")
+
+		// Create a UnixFS file node with identity CID
+		node := makeIdentityNode(initialData)
+
+		// Verify the encoded size is under limit
+		encoded, _ := node.EncodeProtobuf(false)
+		if len(encoded) > verifcid.DefaultMaxIdentityDigestSize {
+			t.Skipf("Initial node too large for identity: %d bytes", len(encoded))
```
gammazero: Should the test fail here so that we know when the test it broken?

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+
+// makeTestDataPattern generates test data with a repeating pattern
+func makeTestDataPattern(size int, pattern string) []byte {
+	data := make([]byte, size)
+	patternBytes := []byte(pattern)
+	for i := range data {
+		data[i] = patternBytes[i%len(patternBytes)]
+	}
+	return data
+}
```
gammazero: I only see one pattern being used, so might as well not bother passing the pattern. This makes it able to replace more test code.
```suggestion
// makeTestData generates test data with repeating lowercase ascii alphabet.
func makeTestData(size int) []byte {
	data := make([]byte, size)
	for i := range data {
		data[i] = byte('a' + i%26)
	}
	return data
}
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+	})
+
+	t.Run("preserves all prefix fields when switching from identity", func(t *testing.T) {
+		// Create initial data that fits well within identity CID limit
+		// This needs to be large enough to trigger modifyDag but small enough
+		// to fit in identity with UnixFS overhead
+		initialData := make([]byte, 100)
+		for i := range initialData {
+			initialData[i] = byte('a' + i%26)
+		}
```
gammazero: See `makeTestData` above.
```suggestion
		initialData := makeTestData(100)
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+	})
+
+	t.Run("raw node preserves codec", func(t *testing.T) {
+		// Create a RawNode with identity CID that's near the limit
+		// This allows us to test overflow with small modifications
+		initialData := make([]byte, verifcid.DefaultMaxIdentityDigestSize-10)
+		for i := range initialData {
+			initialData[i] = byte('a' + i%26)
+		}
```
gammazero: See makeTestData above.
```suggestion
		initialData := makeTestData(verifcid.DefaultMaxIdentityDigestSize-10)
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+	})
+
+	t.Run("raw node identity overflow via truncate", func(t *testing.T) {
+		// Test that Truncate also handles identity overflow correctly for RawNodes
+		// Start with a RawNode using identity that's at max size
+		maxData := make([]byte, verifcid.DefaultMaxIdentityDigestSize)
+		for i := range maxData {
+			maxData[i] = byte('A' + i%26)
+		}
```
gammazero:
```suggestion
		maxData := makeTestData(verifcid.DefaultMaxIdentityDigestSize)
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+	})
+
+	t.Run("raw node switches from identity when modified to exceed limit", func(t *testing.T) {
+		// Create a RawNode with identity CID near the limit
+		// Use size that allows modification without triggering append
+		initialData := make([]byte, verifcid.DefaultMaxIdentityDigestSize-1) // 127 bytes
+		for i := range initialData {
+			initialData[i] = byte('a')
+		}
```
gammazero:
```suggestion
		initialData := bytes.Repeat([]byte{'a'), verifcid.DefaultMaxIdentityDigestSize-1) // 127 bytes
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+		_, err = dmod.Seek(0, io.SeekEnd)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Append data that will create multiple blocks
+		largeData := make([]byte, 100)
+		for i := range largeData {
+			largeData[i] = byte('A' + i%26)
+		}
```
gammazero:
```suggestion
		largeData := makeTestData(100)
```

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+		_, err = dmod.Seek(0, io.SeekEnd)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Append enough data to force multi-block structure
+		appendData := make([]byte, 200)
+		for i := range appendData {
+			appendData[i] = byte('X')
+		}
```
gammazero:
```suggestion
		appendData := bytes.Repeat([]byte{'X'), 200)
```

gammazero (approved):
Only some tiny code style suggestions for a few less lines of code.

ipld/unixfs/mod/dagmodifier_test.go
```diff
@@ -867,3 +872,655 @@ func getOffset(reader uio.DagReader) int64 {
...
+			MhType:   mh.SHA2_512,
+			MhLength: 32, // Truncated SHA-512
+		}
+
+		// Write within the existing data bounds to trigger modifyDag
+		// Use enough data that would cause identity overflow
+		modData := make([]byte, 50)
+		for i := range modData {
+			modData[i] = 'X'
+		}
```
gammazero:
```suggestion
		modData := bytes.Repeat([]byte{'X'}, 50)
```

verifcid/cid.go
achingbrain: It looks fine at first glance 👍

lidel:
Ok, seems everyone received this proposal positively, and no concerns were raised in last 2 weeks.
I'm going to merge this and we will ship Kubo 0.38 RC1 with this to solicit feedback from early testers.

PR #1019, opened 2025-08-30
fix(gateway): prevent blocked CIDs in CAR responses

### LINKED ISSUE #458: Gateway: CAR handler shouldn't return 200 & a CAR header if data is unavailable
## Background

Consider the straightforward case that a block is unfetchable by the block source—this is a somewhat common occurrence for Boost, where there may be a disconnect between advertised CIDs and those that are available unsealed. So requests may come in for CIDs that are supposed to be there, but the block fetcher returns an error that the block is unavailable.

The Boxo gateway code will return a 200, a set of valid headers, and a CARv1 header with the requested CID in the roots array, but nothing else, and termination of the stream is clean. There is no indication of a problem without parsing the CAR. Of course in the Trustless Gateway paradigm, it's up to the user to validate that the CAR contains the expected blocks, so from that perspective we have what we need to determine whether there's a problem or not. But this does present complications for debugging problems. In particular when debugging Rhea retrieval problems I need to have access to Boost logs on the server side to see what the problem might be, I have no indication from the outside that Boost even thinks that there's a problem.

## Desired behaviour

While the spec doesn't cover this, here's what I think should happen and how we built the [Lassie HTTP handler](https://github.com/filecoin-project/lassie/blob/main/pkg/server/http/ipfs.go):

1. Wait until there is at least one block to return before setting any headers or a CAR header; only when there is at least one block should we start sending data.
2. If an error occurs before we get a single block, we return an error response
  a. If we can't get any candidates from the indexer then we can do a `404` with the body `no candidates found`
  b. Other failures are treated as a "gateway timeout", `504`, with the body `failed to fetch CID: <error message>`.

## Code exploration

* `handler#serveCar` starts setting headers immediately, with no opportunity to change course if there's a problem: https://github.com/ipfs/boxo/blob/13569468babd703fc0da2413ef36cded14136b3a/gateway/handler_car.go#L56
* It defers to `BlocksBackend#GetCAR` to return a `Reader` for the CAR and simply does an `io.Copy` of that data to the output body.
* `BlocksBackend#GetCAR` immediately sets up a pipe and starts writing a CAR: https://github.com/ipfs/boxo/blob/13569468babd703fc0da2413ef36cded14136b3a/gateway/blocks_backend.go#L282-L289
* The use of `storage.NewWritable` with a root will immediately write a CARv1 header to the `Writer` it's given.

Hence with a valid trustless gateway `/ipfs` request, we will always get a valid CARv1 header with the root we request, regardless of whether the requested root CID is even fetchable.

In Lassie, we deal with this in two ways, both encapsulated in `DeferredWriter` which is compatible with `github.com/ipld/go-ipld-prime/storage/WritableStorage`, like what the `github.com/ipld/go-car/v2/storage/CarWriter` is. https://github.com/filecoin-project/lassie/blob/main/pkg/storage/deferredcarwriter.go

1. Don't set up the CAR writer until we have our first `Put` operation
2. Provide an `OnPut` event listener that lets us watch for the first put and set the headers in expectation of a CAR with at least one block.

lidel:
> [!WARNING]
> This wip experimentation to see gateway-conformance regressions, not ready for review yet.

- Depends on:
  - https://github.com/ipfs-shipyard/nopfs/pull/50
  - https://github.com/ipfs/gateway-conformance/pull/244
  - https://github.com/ipfs/gateway-conformance/pull/245
- Closes #458

## fix for ipfs/kubo#10361

The gateway was including blocked CIDs in CAR format responses, bypassing content filtering policies.

The fix separates the DAGService usage in GetCAR:
- `nodeGetterToCarExporer` continues wrapping for path resolution
- Original `dagService` is now used for `blockOpener` during traversal
- `blockOpener` returns `traversal.SkipMe{}` for blocked content
- Added detailed comments explaining the blocking architecture

This ensures blocked content is filtered from CAR responses while allowing partial CAR generation when internal blocks are blocked.

- Tests in Kubo PR: https://github.com/ipfs/kubo/pull/10948

## fix for https://github.com/ipfs/boxo/issues/458

We wait with headers until first block, and return 410 / 404 when we know we can't serve the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- `gateway`: Fixed CAR responses including blocked content. The gateway now properly filters out blocked CIDs from CAR format responses, ensuring content filtering policies are enforced across all response formats. ([ipfs/kubo#10361](https://github.com/ipfs/kubo/issues/10361))
+
 ### Security
 
 
--- a/gateway/backend_blocks.go
+++ b/gateway/backend_blocks.go
@@ -469,12 +469,12 @@ func (bb *BlocksBackend) streamCAR(ctx context.Context, p path.ImmutablePath, pa
 		// Set up IPLD LinkSystem for DAG traversal
 		lsys := cidlink.DefaultLinkSystem()
 		unixfsnode.AddUnixFSReificationToLinkSystem(&lsys)
-		// CRITICAL: Use the original dagService for blockOpener, not the wrapped nodeGetterToCarExporer.
-		// This separation ensures that:
-		// 1. blockOpener checks if blocks are accessible (through blocking-aware BlockService)
-		// 2. Only non-blocked content triggers CAR writing in nodeGetterToCarExporer
-		// 3. Blocked content returns traversal.SkipMe{} and never gets written to CAR
-		lsys.StorageReadOpener = blockOpener(ctx, dagService)
+		// Use blockGetter which writes blocks to CAR as they're fetched.
+		// With comprehensive nopfs wrappers (BlockService, Blockstore, Exchange),
+		// blocking is already handled at the service level. The blockGetter
+		// (nodeGetterToCarExporer) simply passes through accessible blocks and
+		// writes them to the CAR output.
+		lsys.StorageReadOpener = blockOpener(ctx, blockGetter)
 
 		// CRITICAL PATH RESOLUTION: This call traverses from root to terminal element,
 		// writing each block to the CAR in order. For /ipfs/QmRoot/sub/terminal:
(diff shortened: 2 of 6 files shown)


lidel:
Status of this: the problem is still real. 

Both bugs reproduce on current main: CAR responses skip the denylist because `GetCAR` walks the raw DAGService, and the CAR handler sends 200 plus headers before it knows the path resolves (#458, ipfs/kubo#10361, ipfs/kubo#10805).

Remaining work, in order:

1. ipfs-shipyard/nopfs#50: `BlockedBlockstore`, `BlockedExchange` with `SessionExchange`, `BlockedFetcher`. Needs review and a tagged release; without it session fetches bypass the blocker.
2. Rebase this branch on main. The useful parts are the `nodeGetterToCarExporer` / `blockOpener` split with `traversal.SkipMe{}` for blocked blocks, and delaying headers until the first block so 404/410 can still be sent.
3. Fix the `remote-car-backend` conformance job (red on the last run, logs expired) and add the CHANGELOG entry.
4. ipfs/kubo#10948: bump boxo and nopfs, keep the `TestContentBlocking` CAR cases in `test/cli/content_blocking_test.go` as the end-to-end check. gateway-conformance already accepts 404 for unresolvable CAR paths (ipfs/gateway-conformance#244, #245).

Future maintainer can either reuse this PR or close it and create new one.

PR #1021, opened 2025-09-01
routing/http: add support for GetClosestPeers (IPIP-476)

### LINKED ISSUE #1004: http-routing: add "get closest peers" operation
Implement https://github.com/ipfs/specs/pull/476

hsanjuan:
- Fixes #1004
- Implements https://github.com/ipfs/specs/pull/476

##  Server side

Since FindPeers also returns PeerRecords, it is essentially a copy-paste, minus things like addrFilters which don't apply here, plus `count` and `closerThan` parsing from the query URL. The tests as well. We leave all logic regarding count/closerThan to the ContentRouter (the DHT, or the Kubo wrapper around it).

## Client side

As the server part, this is heavily inspired in FindPeers.
 
## Content router side:
 
Added GetClosestPeers support, depends on https://github.com/libp2p/go-libp2p-routing-helpers/pull/93


routing/http/server/server.go
```diff
@@ -78,6 +81,10 @@ type ContentRouter interface {
 	// PutIPNS stores the provided [ipns.Record] for the given [ipns.Name].
 	// It is guaranteed that the record matches the provided name.
 	PutIPNS(ctx context.Context, name ipns.Name, record *ipns.Record) error
+
+	// GetClosestPeers returns the DHT closest peers to the given peer ID.
+	// If empty, it will use the content router's peer ID (self). `closerThan` (optional) forces resulting records to be closer to `PeerID` than to `closerThan`. `count` specifies how many records to return ([1,100], with 20 as default when set to 0).
```
guillaumemichel: So far, in the DHT implementation, we cannot return more than `20` peers.

hsanjuan: yeah this is just setting boundaries... ultimately the implementation will do what it can do

guillaumemichel (commented):
IIUC this PR isn't linked with the DHT yet, and there is no implementation of `GetClosestPeers(ctx, pid, closerThanPid, count)` (that doesn't depend on another `GetClosestPeers(ctx, pid, closerThanPid, count)`).

Is the bridge to the DHT going to be implemented in Kubo? I would like to see how everything fits together before merging this one.

hsanjuan:
> Is the bridge to the DHT going to be implemented in Kubo? I would like to see how everything fits together before merging this one.

Right. Yes, it is in Kubo where the bridge to an actual content routing thing happens. At which point we will need to call the DHT method with the DHT we have in Kubo. Perfectly fine to wait for that before merging.

hsanjuan:
Triage:
  * @hsanjuan will circle-back on this and related PRs.

routing/http/client/client.go
```diff
@@ -638,3 +638,81 @@ func (c *Client) PutIPNS(ctx context.Context, name ipns.Name, record *ipns.Recor
 
 	return nil
 }
+
+// GetClosestPeers obtains the closest peers to the given peer ID.
+func (c *Client) GetClosestPeers(ctx context.Context, peerID peer.ID) (peers iter.ResultIter[*types.PeerRecord], err error) {
+	m := newMeasurement("GetClosestPeers")
+
+	// Build the base URL path
+	u, err := gourl.JoinPath(c.baseURL, "routing/v1/dht/closest/peers", peer.ToCid(peerID).String())
```
gammazero: Maybe the closest peers URL path should be a const defined at top of file, to make it easier to find and change.

routing/http/client/client.go
```diff
@@ -638,3 +638,81 @@ func (c *Client) PutIPNS(ctx context.Context, name ipns.Name, record *ipns.Recor
...
+	}
+
+	m.mediaType = mediaType
+
+	var skipBodyClose bool
+	defer func() {
+		if !skipBodyClose {
+			resp.Body.Close()
+		}
+	}()
```
gammazero: Move this up to line 669 to avoid calling `resp.Body.Close()` before each error return.

routing/http/contentrouter/contentrouter.go
```diff
@@ -300,3 +307,45 @@ func (c *contentRouter) SearchValue(ctx context.Context, key string, opts ...rou
 
 	return ch, nil
 }
+
+func (c *contentRouter) GetClosestPeers(ctx context.Context, pid peer.ID) (<-chan peer.AddrInfo, error) {
+	iter, err := c.client.GetClosestPeers(ctx, pid)
+	if err != nil {
+		return nil, err
+	}
+	defer iter.Close()
```
gammazero: This needs to be done in the goroutine, otherwise the function returns and the iter closes before the goroutine even has time to run.

routing/http/server/server.go
```diff
@@ -78,8 +86,16 @@ type ContentRouter interface {
...
+	GetClosestPeers(ctx context.Context, peerID peer.ID) (iter.ResultIter[*types.PeerRecord], error)
```
gammazero: No need to change this now, but I wish this could use golang iterators instead.

gammazero (changes requested):

routing/http/client/client.go
hsanjuan: Prior art is like this for the other methods... given the constant is used in a single place and that place is the method that it is relevant to, it's probably not adding much.. so I'd rather leave it.

gammazero (approved):

lidel:
Triage notes: 
- before merging, open a PR for someguy to see how this behaves end-to-end

routing/http/server/server.go
```diff
@@ -47,18 +47,26 @@ const (
 var logger = logging.Logger("routing/http/server")
 
 const (
-	providePath       = "/routing/v1/providers/"
-	findProvidersPath = "/routing/v1/providers/{cid}"
-	findPeersPath     = "/routing/v1/peers/{peer-id}"
-	GetIPNSPath       = "/routing/v1/ipns/{cid}"
+	providePath         = "/routing/v1/providers/"
+	findProvidersPath   = "/routing/v1/providers/{cid}"
+	findPeersPath       = "/routing/v1/peers/{peer-id}"
```
guillaumemichel, line 52:
```suggestion
	findPeersPath       = "/routing/v1/peers/{cid}"
```

see https://github.com/ipfs/specs/pull/476#discussion_r2391931969

routing/http/server/server.go
guillaumemichel:
```suggestion
	GetClosestPeers(ctx context.Context, key string) (iter.ResultIter[*types.PeerRecord], error)
```

As per https://github.com/ipfs/specs/pull/476#discussion_r2383524824, we can use the string representation of the peer id or multihash as in [go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht/blob/6e9a59c05736c7b40586ff385c5efc96981b46f8/lookup.go#L23) ([peer id](https://github.com/libp2p/go-libp2p-kad-dht/blob/6e9a59c05736c7b40586ff385c5efc96981b46f8/routing.go#L642), [CID](https://github.com/libp2p/go-libp2p-kad-dht/blob/6e9a59c05736c7b40586ff385c5efc96981b46f8/routing.go#L566), [IPNS key](https://github.com/libp2p/go-libp2p-kad-dht/blob/6e9a59c05736c7b40586ff385c5efc96981b46f8/routing.go#L107))

routing/http/contentrouter/contentrouter.go
```diff
@@ -43,6 +47,7 @@ var (
 	_ routing.ValueStore               = (*contentRouter)(nil)
 	_ routinghelpers.ProvideManyRouter = (*contentRouter)(nil)
 	_ routinghelpers.ReadyAbleRouter   = (*contentRouter)(nil)
+	_ routinghelpers.DHTRouter         = (*contentRouter)(nil)
```
guillaumemichel: This seems to be the only use case of `routinghelpers.DHTRouter` (introduced in https://github.com/libp2p/go-libp2p-routing-helpers/pull/93). The [kubo PR](https://github.com/ipfs/kubo/pull/10954) isn't using it.

Since `Client` (above) implements `GetClosestPeers()` it should be enough, I don't think we need a `DHTRouter` interface.

routing/http/server/server.go
hsanjuan: Note: string representation here is "casting" as string, not doing .String().

lidel: Triage note:
- `string` is too risky ambigous -- will elad to hard to debug bugs
- use `CID` type and extract raw bytes via vasting bytes→string internally

routing/http/contentrouter/contentrouter.go
hsanjuan: It breaks consistency? do we need a ReadyAbleRouter etc? IUUC the contentRouter is not a Client, but a Router implementation, the problem being that Router doesn't have a unified interface, which is ugly, but placing this new interface in a different place than the others is also weird.

routing/http/server/server.go
lidel: Switched to `{key}` in https://github.com/ipfs/boxo/pull/1021/commits/713b7c02471cb20f9287f143cd2bbc34d1771a03 (which should correctly work with both peerids and cids)

routing/http/server/server.go
lidel: We now use

```go
GetClosestPeers(ctx context.Context, key cid.Cid) (iter.ResultIter[*types.PeerRecord], error)
```

(the Cid acts as a container for the Multihash digest, which then is casted as string to kad-dht for legacy reasons).

hsanjuan:
Thank you @lidel . Latest changes LGTM. Would you be so kind to call the shot regarding the DHTRouter interface in libp2p-routing-helpers (commit above from @guillaumemichel)? We can move it here or leave it there. Probably it is the last thing to resolve.

routing/http/server/server.go
```diff
@@ -87,8 +87,7 @@ type DelegatedRouter interface {
 	// It is guaranteed that the record matches the provided name.
 	PutIPNS(ctx context.Context, name ipns.Name, record *ipns.Record) error
 
-	// GetClosestPeers returns the DHT closest peers to the given peer ID.
-	// If empty, it will use the content router's peer ID (self). `closerThan` (optional) forces resulting records to be closer to `PeerID` than to `closerThan`. `count` specifies how many records to return ([1,100], with 20 as default when set to 0).
+	// GetClosestPeers returns the DHT closest peers to the given key (CID or Peer ID).
 	GetClosestPeers(ctx context.Context, key cid.Cid) (iter.ResultIter[*types.PeerRecord], error)
```
gammazero, line 91: Should the `key` argument be a `cid.Cid` or a `string`? Seems like it needs to be a `string` if accepting cid or peer id.

gammazero: Peer ID can be encoded as CID. This prevents confusion about how to encode key.

routing/http/contentrouter/contentrouter.go
hsanjuan: I moved the DHTRouter interface to contentrouter where it is used, and restored the routing-helpers import to what it was.

PR #1022, opened 2025-09-01
fix: flaky TestSessionBetweenPeers with shuffle enabled

lidel:
TLDR: fixes a racy test. This is likely just a problem with test's inability to fully test in isolation, rather  a bug in the code.

cc @gammazero  for sanity check

### Why

Unsure if I fully understand the test/bitswap, so take this with the grain of salt, but iiuc the test was flaky ([example](https://github.com/ipfs/boxo/actions/runs/17364058202/job/49288071533?pr=1019#step:15:56)) when run with `-shuffle=on` because of a race condition in message counting. After fetching the first block (`cids[0]`), which triggers a broadcast want-have to all peers and then a CANCEL when received, the test immediately continued fetching more blocks.

During this time, some other internal timers (idle tick or periodic search?) could fire and cause a rebroadcast of wants before the test checked message counts on uninvolved nodes. With test shuffling, goroutine scheduling changes made this timing issue more likely, causing uninvolved nodes to sometimes receive 3 messages instead of the expected 2.

Adding a small delay after the first block fetch ensures the CANCEL is fully processed and the session stabilizes before continuing, preventing the race condition. 

(Not scientific, but run it hundreds times and was not able to reproduce flaky race anymore, so at least this PR makes our CI more stable across PRs)


gammazero:
I think there may be some other issues causing test unreliability.. It looks like CI is sometimes hanging forever (until timeout) or getting wrong count with this PR.  Need further investigation.

bitswap/client/bitswap_with_sessions_test.go
```diff
@@ -203,6 +203,11 @@ func TestSessionBetweenPeers(t *testing.T) {
 	if _, err := ses.GetBlock(ctx, cids[0]); err != nil {
 		t.Fatal(err)
 	}
+
+	// Wait briefly to ensure the CANCEL for cids[0] is fully processed
+	// before continuing with more requests that might trigger rebroadcasts
+	time.Sleep(10 * time.Millisecond)
```
gammazero: The `minSendMessageDelay` in the bitswam message queue is 20ms, so lets use at least that.

```suggestion
	time.Sleep(50 * time.Millisecond)
```

gammazero (approved):

PR #1023, opened 2025-09-01
feat(gateway): enhance 504 timeout errors with diagnostic UX

lidel:
This expands on foundations from  https://github.com/ipfs/boxo/pull/1015:

- [x] add `DiagnosticServiceURL` config for CID retrievability checks
- [x] show "Check CID retrievability" button on 504 errors
- [x] add "Retry" button on all error pages
- [x] track and display failed peer IDs in timeout messages
- [x] indicate when timeout occurs on sub-resource vs root CID
- [x] improve error page styling and button UX
- [x] show peerids when we discovered them but failed to connect to them because they are offline DHT peer (found peerid but no addrs, or cant connect to them)

- Part of https://github.com/ipshipyard/roadmaps/issues/14

----

## Demo

This PR not only shows the PeerID of "offline" peer found as a provider on DHT, but also has two buttons to Refresh or Debug:

> image

### DIFF
--- a/retrieval/state.go
+++ b/retrieval/state.go
@@ -10,9 +10,11 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 
+	"github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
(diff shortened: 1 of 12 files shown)


retrieval/state.go
```diff
@@ -120,22 +132,90 @@ func (rs *State) GetPhase() RetrievalPhase {
...
 
+// addProvider is a helper to add a provider to a sample list with size limit.
+// Only the first MaxProvidersSampleSize providers are kept to prevent unbounded memory growth.
+func (rs *State) addProvider(list *[]peer.ID, peerID peer.ID) {
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	if len(*list) < MaxProvidersSampleSize {
+		*list = append(*list, peerID)
+	}
+}
```
gammazero: I think it would be more idiomatic if this looked more like `append`:
```suggestion
func (rs *State) appendProvider(list []peer.ID, peerIDs ...peer.ID) {
	rs.mu.Lock()
	defer rs.mu.Unlock()
	if len(list) >= MaxProvidersSampleSize {
		return
	}
	if len(list) + len(peerIDs) > MaxProvidersSampleSize {
		peerIDs = peerIDs[:MaxProvidersSampleSize - len(list)]
	}
	return append(list, peerIDs...)
}
```

gammazero (approved):

PR #1027, opened 2025-09-05
Use `time.Duration` for rebroadcast delay

gammazero:
The rebroadcast delay was a `delay.Delay` that was only being used as a fixed time, so it had no more utility than a plain `time.Duration` value. Since no time varying functionality was used anywhere, the `delay.Delay` was replaced with a `time.Duration`.

**BREAKING CHANGE** This changes the argument type to one of the bitswap client options: `RebroadcastDelay`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


hsanjuan (approved):
weird that it wasn't used. Maybe it was there for some manual testing.

lidel (approved):

PR #1033, opened 2025-09-11
replace provider queue with go-dsqueue

gammazero:
Replace the old provider queue with go-dsqueue.

The code for clearing the old provider queue datastore remains for cleaning up any data persisted by the old provider queue.


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/ipfs/go-cidutil v0.1.0
 	github.com/ipfs/go-datastore v0.8.3
 	github.com/ipfs/go-detect-race v0.0.1
+	github.com/ipfs/go-dsqueue v0.0.2
```
guillaumemichel: nit: upgrade to latest version before merging

provider/reprovider.go
```diff
@@ -146,7 +147,17 @@ func New(ds datastore.Batching, opts ...Option) (System, error) {
 	}
 
 	s.ds = namespace.Wrap(ds, s.keyPrefix)
-	s.q = queue.New(s.ds)
+
+	// Remove any items from the old queue.
+	cleaned, err := oldqueue.ClearDatastore(s.ds)
```
guillaumemichel, line 154: Why do we want to remove items from the old queue? This seems to be a new behavior.

IIUC we used to persist the provide queue on disk, so that it could be resumed after node restart (unless provide strategy changed in kubo).

gammazero: We want to remove items from the old provide queue, because the new provide queue (`go-dsqueue`) stores its records in a different format. So, we either need to migrate all that data to the new format, or delete it. Since this is temporary data anyway, I chose to delete it and let it repopulate using the new queue. If we did nothing, then it would sit in the datastore as dead data forever.

Would it make more sense to move this to migration code and require a migration?

guillaumemichel: OK, now I get it. It makes sense :+1:

guillaumemichel (approved):

PR #1035, opened 2025-09-18
feat(pinner): add `CheckIfPinnedWithType` for efficient checks with names

lidel:
Adds `CheckIfPinnedWithType` method to the Pinner interface for efficient type-specific pin checks with optional name loading.

## Why

`ipfs pin ls <cid> --names` in Kubo doesn't return pin names because the current API doesn't support fetching names for specific CIDs efficiently. This PR adds the missing functionality.

## Changes

- New `CheckIfPinnedWithType(ctx, mode, includeNames, cids...)` method
  - `mode`: which pin type to check (Recursive, Direct, Indirect, Any)
  - `includeNames`: whether to load pin names from datastore
- `CheckIfPinned` now delegates to `CheckIfPinnedWithType`

## Related

- Required for:
  - https://github.com/ipfs/js-kubo-rpc-client/pull/343 which depends on fix from:
    - https://github.com/ipfs/kubo/pull/10970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `pinning/pinner`: Added `CheckIfPinnedWithType` method to `Pinner` interface for efficient type-specific pin checks with optional name loading ([#1035](https://github.com/ipfs/boxo/pull/1035))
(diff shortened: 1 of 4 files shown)


lidel (commented):
Merging as per verbal during today's colo -- we will include this fix/optimization in Kubo 0.38.0-rc1.

PR #1037, opened 2025-09-18
fix(mfs): limit cache growth by default

lidel:
- Aims to remove footgun of https://github.com/ipfs/kubo/issues/10842


gammazero (approved):
LGTM

PR #1039, opened 2025-09-18
fix(pinner): restore indirect pin detection and add context cancellation

lidel:
Follow-up for #1035 (forgot to push local changes):
- Small fix to ensure behavior from before PR #1035 is exactly the same 
  - Sharness in https://github.com/ipfs/kubo/pull/10970 caught this regression

While at it, some cleanup:
- Dry helper `traverseIndirectPins` to avoid duplicated code
- Added check for context cancellation at the start of each recursive pin
  - It was not present in old code before #1035, so in theory aborting listing all pins was sometimes stuck for longer than it should

### DIFF
--- a/pinning/pinner/dspinner/pin.go
+++ b/pinning/pinner/dspinner/pin.go
@@ -646,34 +646,8 @@ func (p *pinner) CheckIfPinnedWithType(ctx context.Context, mode ipfspinner.Mode
 
 		// Check for indirect pins
 		if toCheck.Len() > 0 {
-			var walkErr error
-			visited := cid.NewSet()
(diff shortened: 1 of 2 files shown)


lidel:
I've documented the backward-compatibility criteria and confirmed this PR passes legacy sharness tests in:

- https://github.com/ipfs/kubo/pull/10970

I'm going to merge this since #1035 is already in the `main` branch, we need this PR as well for Kubo 0.38.0-rc1 (https://github.com/ipfs/kubo/issues/10884).


We may want to look at this in the future, maybe the legacy behavior can/should be changed, but I don't want to make any extra waves for 0.38 -- keeping things as-is where possible for now.

PR #1042, opened 2025-09-24
Filestore: add Provider option to provide filestore blocks.

hsanjuan:
A Filestore is a Blockstore wrapper. When I moved providing resposibilities to Blockstore/Pinner/MFS... I forgot the Filestore. In Particular, the Filestore writes all blocks to the Blockstore except those of FilestoreNode type, which are serialized and written to the datastore directly via the Filemanager. The data serialized is not the block itself, but a pointer to a filesystem object and an offset.

So FilestoreNodes escaped providing even when the blockstore is configured to provide... Not anymore.

### DIFF
--- a/filestore/filestore.go
+++ b/filestore/filestore.go
@@ -259,7 +259,7 @@ func (f *Filestore) PutMany(ctx context.Context, bs []blocks.Block) error {
 			}
 			logger.Debugf("filestore: provide %d hashes", len(hashes))
 			if err := f.provider.StartProviding(false, hashes...); err != nil {
-				logger.Warnf("filestore: error while providing blocks: %s", err)
+				logger.Warnf("filestore: error while providing hashes: %s", err)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


hsanjuan:
Addresses https://github.com/ipfs/kubo/issues/10983

guillaumemichel (approved):
Looks good!

filestore/filestore.go
```diff
@@ -235,6 +251,17 @@ func (f *Filestore) PutMany(ctx context.Context, bs []blocks.Block) error {
...
 		}
+
+		if f.provider != nil {
+			var hashes []multihash.Multihash
+			for _, n := range fstores {
+				hashes = append(hashes, n.Node.Cid().Hash())
+			}
+			logger.Debugf("filestore: provide %d hashes", len(hashes))
+			if err := f.provider.StartProviding(false, hashes...); err != nil {
+				logger.Warnf("filestore: error while providing blocks: %s", err)
```
hsanjuan:
```suggestion
				logger.Warnf("filestore: error while providing hashes: %s", err)
```

Issue #856, opened 2025-02-22
gateway: range requests issue with big >5GiB files


lidel:
Cloudflare won't support range requests of big files (>5GiB). They return entire file as HTTP 200 rather  than only the HTTP 206 with the requested `Range`.

> [!NOTE]
> This problem only applies to unverifiable deserialized responses.
> 
> Using [verifiable](https://docs.ipfs.tech/reference/http/gateway/#trustless-verifiable-retrieval) `application/vnd.ipld.raw` (`?format=raw`) fetches small blocks which makes it work fine.

lidel:
Seems that something else is at play. I've checked multiple times and both GET and HEAD for 20GiB file include `accept-ranges: bytes` in response already.

Pausing until we get some clarifications from Cloudflare.

lidel:
Closing as the bug turned out to be on the Cloudflare side.

Cloudflare, by default only supports caching of range requests for files <=5GiB.
Range requests for bigger files are ignored and full file is returned as HTTP 200 instead.

Enterprise clients are able to opt-in into increased "cache keys + origin range requests" feature, which at the time of me writing it, is NOT documented anywhere. It also comes with downside of invalidating ALL CACHES GLOBALLY and only allows raising the ceiling bit higher (32GiB?) which does not really help if someone stores gigantic files.

So if anyone is reading this in future:
- ⚠️ ️If you are Cloudflare client, reach out to their support, and reference this issue
- ⚠️ If you write HTTP Client, make sure to check response code before processing response body
- 🟢 👉  If you store big files on IPFS and want to read only specific byte ranges, avoid using public gateways, and instead use tools like [Kubo](https://github.com/ipfs/kubo), [Rainbow](https://github.com/ipfs/rainbow/) or [@helia/verified-fetch](https://www.npmjs.com/package/@helia/verified-fetch), which will only fetch the minimal amount of blocks (`application/vnd.ipld.raw`) necessary to fulfill your range request.

lidel:
If anyone is Cloudflare CDN/Cache user, beware the fix in `boxo/gateway` is not enough. You are still exposed to overcharges from Cloudflare due to the way they handle caching and range requests. 

## Cloudflare-specific workaround

[...]

PR #1043, opened 2025-09-25
feat(gateway): add MaxRangeRequestFileSize protection

lidel:
Adds Config.MaxRangeRequestFileSize to protect against CDN issues with large file range requests. when set, range requests for files larger than this limit return HTTP 501 with suggestion to use verifiable block requests (application/vnd.ipld.raw).

Intention here is to protect against Cloudflare's issue where range requests for files over 5GiB are silently ignored, causing entire file to be returned instead of requested range, leading to excess bandwidth and billing.

The fix here shouldengage when cloudflare sends very first range request to boxo/gateway backend. The backend is able to detect range request for a file beyond safe size, and refuse responding. 

- Provides alternative way of handling issues like #856


guillaumemichel (approved):

cewood (approved):

cewood:
Should the Shareness tests get updated first? Seems like bad hygiene to run those tests if they aren't of significance here. 🤔

PR #1044, opened 2025-09-26
refactor(gateway): switch defaults to autoconf.FallbackDNSResolvers

### LINKED ISSUE #772: bug: all "Unstoppable Domains" (.crypto)  DoH resolvers are down

lidel:
This PR moves hardcoded DNS resolver defaults from `gateway.NewDNSResolver` to `autoconf.FallbackDNSResolvers` so we have them in one place, while preserving existing behavior for users.

### Changes

- Move hardcoded DNS resolver defaults to `autoconf.FallbackDNSResolvers`
- `NewDNSResolver(nil)` uses `autoconf.FallbackDNSResolvers` (preserves `.eth` resolution)
- `NewDNSResolver(map[string]string{})` explicitly uses system DNS only
- Remove `autoconf.Client.ExpandDNSResolvers()` method (no use case yet)
- Update test to use `.foo` instead of `.eth` TLD
- No network calls at gateway initialization (privacy-preserving)

  
### Migration

No migration needed. For those wanting custom DNS configuration, existing code (Kubo, Rainbow) continues to work.

### References

- CC #771
- Closes #772

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,6 @@ The following emojis are used to highlight certain changes:
   - Added `Config.DiagnosticServiceURL` to configure a CID retrievability diagnostic service. When set, 504 Gateway Timeout errors show a "Check CID retrievability" button linking to the service with `?cid=<failed-cid>` [#1023](https://github.com/ipfs/boxo/pull/1023)
   - Improved 504 error pages with "Retry" button, diagnostic service integration, and clear indication when timeout occurs on sub-resource vs root CID [#1023](https://github.com/ipfs/boxo/pull/1023)
 - `gateway`: Added `Config.MaxRangeRequestFileSize` to protect against CDN issues with large file range requests. When set to a non-zero value, range requests for files larger than this limit return HTTP 501 Not Implemented with a suggestion to use verifiable block requests (`application/vnd.ipld.raw`) instead. This provides protection against Cloudflare's issue where range requests for files over 5GiB are silently ignored, causing excess bandwidth consumption and billing
-- `autoconf`: Added `Client.ExpandDNSResolvers()` convenience method for expanding DNS resolver maps with autoconf data ([#771](https://github.com/ipfs/boxo/issues/771), [#772](https://github.com/ipfs/boxo/issues/772))
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -45,6 +46,14 @@ The following emojis are used to highlight certain changes:
 
 ### Removed
 
+- `gateway`: 🛠 Removed hardcoded `defaultResolvers` for `.eth` and `.crypto` TLDs from `gateway.NewDNSResolver` ([#771](https://github.com/ipfs/boxo/issues/771), [#772](https://github.com/ipfs/boxo/issues/772))
+  - This change removes implicit defaults that were difficult to discover, override, or disable, improving user agency and configuration transparency
+  - `NewDNSResolver(nil)` now uses system DNS only (no implicit DoH resolvers)
```
aschmahmann: Does it make sense to make this a breaking change for all users? We had centralized / hardcoded endpoints in here beforehand why not just use autoconf by default?

e.g. if ENS is highly visible / expected within the ecosystem and we make it so that by default it doesn't show up we're going to be hoisting a whole bunch of bug reports onto users who'll have to report this to their implementations. As a bonus this type of breakage doesn't even show up in Go tooling so most users will just upgrade (it seems like nobody reads release notes until maybe they see Go yelling at them), think everything is fine, and then rely on users to report bugs.

lidel: Fair point – we do not want to triage extra bugs – this is legacy debt we can pay off some other time.

I've restored the old behavior (b3238d9347e16422b709ff487da8ac02e13d1f4b) to not break existing users, but did not add full autoconf, just replaced `defaultResolvers` with code that reads defaults from `autoconf.FallbackDNSResolvers`. This way we still remove copy of the defaults from `boxo/gateway` and have single source of truth for fallback defaults in `autoconf` package, but avoid a can of worms described below.


(There are many reasons to not use autoconf here: real autoconf comes with extra network I/O and potential delay in startup/initialization. We would also add a ton of code to ensure user can pass own client, and initialize new implicit default client if user did not pass any. Then, how do we know if user is ok with sending extra HTTP  Request every time their app with `boxo/gateway` starts? Is user ok with leaking their IP to some http server when they were using `boxo/gateway` only in private contexts? Can of worms.)

aschmahmann: > There are many reasons to not use autoconf here: real autoconf comes with extra network I/O and potential delay in startup/initialization

Agree on some of the can of worms and that we can defer at the library layer if we want, but in most ways it seems like using autoconf still ends up better than the status quo.

> real autoconf comes with extra network I/O and potential delay in startup/initialization

Delays and I/O don't have to happen, they can happen on first use of a non-ICANN domain.

>  Then, how do we know if user is ok with sending extra HTTP Request every time their app with boxo/gateway starts

Mostly fixed by above, unless the user has very short-lived gateways (a real possibility, but other caching capabilities like those that are DoH related are impacted here as well).

> Is user ok with leaking their IP to some http server when they were using boxo/gateway only in private contexts?

We already have this if someone uses a .eth or .crypto domain

> We would also add a ton of code

Maybe, I'm not sure that's true. We have overrideable defaults like this in a bunch of places (even if they don't rely on I/O since we'd prefer to avoid that if possible).

---

IIUC the win here from using the default autoconf endpoint is basically that users can rely on the boxo maintainers to give them sane default DNSLink resolvers rather than relying on externally run hard coded ones to be good indefinitely (i.e. the status quo we're keeping here). This could have been done by the maintainers instead running proxies for each default DNSLink resolver, but using autoconf makes the extra operational burden on the maintainers minimal while allowing for extra resiliency with having longer cache time so that even if the default autoconf endpoint goes down for a bit that many users will still be able to resolve their domain names. In theory multiple endpoints for a given eTLD could be configured, but at the moment even ENS in practice only has one public DNSLink resolver 😬.

gateway/dns.go
```diff
@@ -4,16 +4,12 @@ import (
...
 
+	"github.com/ipfs/boxo/autoconf"
 	"github.com/libp2p/go-doh-resolver"
 	dns "github.com/miekg/dns"
 	madns "github.com/multiformats/go-multiaddr-dns"
 )
 
-var defaultResolvers = map[string]string{
-	"eth.":    "https://dns.eth.limo/dns-query",
-	"crypto.": "https://resolver.unstoppable.io/dns-query",
```
aschmahmann, line 14: We should flag that this is dead / removed now (it's not in the default autoconf resolver or in the fallbacks)

gammazero (approved):

lidel:
I looked at the state of `DNS.Resolvers` in Kubo  and we would have to do a deeper refactor than this. Closing for now

Notes for future, when we revisit and find this PR again:
- right now  DNS.Resolvers is only applied to DNSLink and go-libp2p
  - it is NOT applied when resolving domains related to HTTP requests, so for example, delegated /routing/v1, HTTP retrieval in  boxo/bitswap, badbits, or fetching JSON by boxo/autoconf, or even telemetry in kubo ignores user-prefered DNS resolvers from config
- the architectural problem is that "go-libp2p's multiaddr resolver" is used in many APIs, rather than golangs generic net.Resolver, so we would have to create adapter back to net.Resolver, but that is a lossy way
  - instead, we should refactor things to take generic net.Resolver instead
- we may need to clean this up at some point
  - switch libraries like boxo/gateway to use net.Resolver rather than libp2p-specific multiaddr one
  - switch end apps like kubo to override global DNS resolver to one that respects user config (DNS.Resolvers)

CHANGELOG.md
lidel: Valid points on not breaking existing users, but also we do not give enough control/flexibility as we move towards more and more HTTP requests
After thinking about this deeper, closed this PR, we need deeper refactor to wire up DNS Resolvers correctly everywhere:
https://github.com/ipfs/boxo/pull/1044#issuecomment-3761971034

PR #1047, opened 2025-10-07
tar: fix name filter on windows

djdv:
This is for: https://github.com/ipfs/kubo/issues/9370
Specifically this comment for context: https://github.com/ipfs/kubo/issues/9370#issuecomment-3269317066

e579012e99cdd7d0d981612ad49b7905850e97ed is the main fix
In short, this plain-string used to be a regex-string where `[` and `]` have meaning, they're not part of the set so they should not be used with `ContainsAny`.

4b9e075585a74f9f0260f267ebad0c9f8a7d2460
Is a secondary fix that catches reserved names even if their case differs.
I don't know if anyone has encountered this in IPFS yet, but I know file names like `con` show up in various operating system source trees.

The rest of the patches are gardening and testing.
The sanitize tests hit 100% coverage with this, and I tested this explicitly in go-ipfs via `ipfs get`.
Prior to the patch, we hit the bug. After, we don't.

```plain
> ipfs add -r -w 'T:\Temp\[works on my machine]\'
added QmPFaJjSbNVJxG1PTQYVJAkycHTbsZzBLtJRFCo9BvAgbm [works on my machine]/[works for me].png
added QmeKh8Lq1JKk4LLVc2xuJ5VsCBUgUJ3YXa4Umnkb6Urrih [works on my machine]
added QmaSWjLPmJoPBPtUQ9geGCCvNms5EaNkMS2DzgW5NHbP45

> ipfs-stable get QmaSWjLPmJoPBPtUQ9geGCCvNms5EaNkMS2DzgW5NHbP45
Saving file(s) to QmaSWjLPmJoPBPtUQ9geGCCvNms5EaNkMS2DzgW5NHbP45
Error: invalid platform path: path components cannot contain any of [<>:"\|?*] : "[works on my machine]"

> ipfs-dev get QmaSWjLPmJoPBPtUQ9geGCCvNms5EaNkMS2DzgW5NHbP45
Saving file(s) to QmaSWjLPmJoPBPtUQ9geGCCvNms5EaNkMS2DzgW5NHbP45
 440.79 KiB / 440.79 KiB [=] 100.00% 0s
```

Probably good?


djdv (commented):
I should have drafted this first, too late now.
Gave this a second look and made some minor changes.
I'm content with this, so it should be ready for external review.
Edit: might be best to review commit-by-commit rather than starting with the full-diff. idk

gammazero (approved):
LGTM

PR #1051, opened 2025-10-13
feat(routing/http): IPIP-518: HTTP(S) URLs in Routing V1 API

lidel:
This PR implements [IPIP-518](https://github.com/ipfs/specs/pull/518) which extends `/routing/v1` with support for URIs alongside multiaddrs.

### New `Address` type (`types/address.go`)

- accepts both multiaddrs (starting with `/`) and URIs, following IPIP-518 parsing rules
- schema-agnostic: any valid URI scheme is accepted for future extensibility
- protocol matching with `HasProtocol` handles http/https semantics (e.g. `https://` matches both "http" and "https" queries)
- `ToMultiaddr()` converts HTTP(S) URLs to multiaddrs for backward compatibility with existing software that expects `/dns/host/tcp/port/https` format, including `/http-path/` when the URL has a non-empty path
  - paths are not used  (consumed) by Kubo atm, but this will make WebSeed integration easier in the future, i hope
- invalid addresses are skipped per spec, not treated as errors

### New `GenericRecord` type (`types/record_generic.go`)

- schema-agnostic record type using `Addresses` (which accepts both URLs and multiaddrs) instead of `[]Multiaddr`
- preserves unknown fields in `Extra` map for forward compatibility with future protocols
- size-limited to 10 KiB to prevent abuse in NDJSON streams

### Contentrouter integration (`contentrouter/contentrouter.go`)

- `readProviderResponses` converts `GenericRecord` to `peer.AddrInfo` for legacy routing APIs:
  - records with valid PeerID: always converted (supports legacy PeerID + `/https` multiaddr pattern)
  - records with `transport-ipfs-gateway-http` + HTTP(S) URLs: PeerID derived from `did:key:` (ed25519) or generated as a deterministic placeholder
  - other non-PeerID records: skipped
- PeerID interop helpers extracted to `contentrouter/interop.go`

### Filter updates (`filters/filters.go`)

- `ApplyAddrFilter` extended to handle `Address` type alongside `Multiaddr`
- protocol filtering works across both URL schemes and multiaddr protocols

### Existing types unchanged

- `PeerRecord` and `BitswapRecord` continue to use `[]Multiaddr` (multiaddrs only)
- URLs belong in `GenericRecord`, which is the forward-looking record type

## Backward compatibility

- multiaddr-only implementations continue to work unchanged
- HTTP(S) URLs are converted to multiaddrs via `ToMultiaddr()` so existing consumers (Kubo, someguy) receive addresses they can use
- `PeerRecord.Addrs` remains `[]Multiaddr` and rejects URLs at deserialization time

IPIP:

- https://github.com/ipfs/specs/pull/518

## TODO

- [ ] no objections in https://github.com/ipfs/specs/pull/518
- [ ] PR in someguy that uses this PR and tests there pass green


routing/http/client/client_test.go
```diff
@@ -146,18 +146,21 @@ func makeCID() cid.Cid {
...
+	for _, a := range draddrs {
+		if a.IsMultiaddr() {
+			addrs = append(addrs, a.Multiaddr())
+		}
 	}
 	return
 }
 
-func addrsToDRAddrs(addrs []multiaddr.Multiaddr) (drmas []types.Multiaddr) {
+func addrsToDRAddrs(addrs []multiaddr.Multiaddr) types.Addresses {
```
gammazero:
```suggestion
func multiaddrsToRoutingAddrs(addrs []multiaddr.Multiaddr) types.Addresses {
```

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		// Must be absolute URL
+		if !u.IsAbs() {
+			return nil, fmt.Errorf("URI must be absolute")
+		}
+		addr.url = u
+	}
+
+	return addr, nil
+}
```
gammazero: Every time I see this called, the returned *Address is immediately dereferenced. This indicates that it would be better to return an Address struct directly, and not a pointer to said struct. Also, all the types inside are effectively pointers, so return the struct literal does not incur any overhead to worry about.
```suggestion
func NewAddress(s string) (Address, error) {
	addr := Address{raw: s}

	// IPIP-518 parsing logic
	if strings.HasPrefix(s, "/") {
		// Parse as multiaddr
		ma, err := multiaddr.NewMultiaddr(s)
		if err != nil {
			return Address{}, fmt.Errorf("invalid multiaddr: %w", err)
		}
		addr.multiaddr = ma
	} else {
		// Parse as URI - accept any valid URI scheme
		u, err := url.Parse(s)
		if err != nil {
			return Address{}, fmt.Errorf("invalid uri: %w", err)
		}
		// Must be absolute URL
		if !u.IsAbs() {
			return Address{}, fmt.Errorf("uri must be absolute")
		}
		addr.url = u
	}

	return addr, nil
}
```

nit: Do not capitalize "URL" and "URI" in error message as that is not consistent with golang stdlib:
https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/net/url/url.go;l=521

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+	return nil
+}
+
+// Protocols returns the protocols in this address.
+// For multiaddrs, it returns the multiaddr protocols.
+// For URLs, it returns the scheme.
+func (a *Address) Protocols() []string {
+	if a.url != nil {
+		return []string{a.url.Scheme}
+	} else if a.multiaddr != nil {
```
gammazero:
```suggestion
	}
	if a.multiaddr != nil {
```

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+				}
+				if p == "http" {
+					hasHTTP = true
+				}
+			}
+			return hasTLS && hasHTTP
+		case "tls":
+			// "tls" matches any multiaddr with /tls
+			for _, p := range protocols {
+				if p == "tls" {
```
gammazero: Should this also match "https"?

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+			for _, p := range protocols {
+				if p == "https" {
+					return true
+				}
+				if p == "tls" {
+					hasTLS = true
+				}
+				if p == "http" {
+					hasHTTP = true
+				}
```
gammazero:
```suggestion
				if p == "https" {
				switch p {
				case "https":
					return true
				case "tls":
					hasTLS = true
				case "http":
					hasHTTP = true
				}
```

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+	// Build multiaddr string
+	var maStr string
+	if scheme == "https" {
+		// For HTTPS, use /https as this is what existing HTTP providers
+		// announce on IPNI, so we follow same convention for backward-compatibility
+		maStr = fmt.Sprintf("/%s/%s/tcp/%s/https", addrProto, host, port)
+	} else {
+		// For HTTP, use /http
+		maStr = fmt.Sprintf("/%s/%s/tcp/%s/http", addrProto, host, port)
+	}
```
gammazero: Since scheme can only be "http" or "https" just use scheme, instead of the if-else.
```suggestion
	// For HTTPS, use /https as this is what existing HTTP providers
	// announce on IPNI, so we follow same convention for backward-compatibility
	maStr := fmt.Sprintf("/%s/%s/tcp/%s/%s", addrProto, host, port, scheme)
```

routing/http/types/address.go
```diff
@@ -0,0 +1,319 @@
...
+
+	return false
+}
+
+// ToMultiaddr attempts to convert an HTTP(S) URL to a multiaddr for backward compatibility.
+// Returns nil if the address cannot be converted (e.g., non-HTTP schemes or invalid addresses).
+// This is a temporary compatibility layer for the transition period while existing software
+// expects multiaddrs with /http protocol to signal HTTP retrieval support.
+func (a *Address) ToMultiaddr() multiaddr.Multiaddr {
```
gammazero, line 200: An alternative to this that avoids hardcoded values for `addrProto` can be found here:
https://github.com/ipni/go-libipni/blob/main/maurl/convert.go#L110-L156

Both approaches work. The go-libipni version has had years of testing and operation if you want to cut and paste that one, but that does not mean it is necessarily better. I think perhaps using [megular expressions](https://pkg.go.dev/github.com/multiformats/go-multiaddr@v0.16.1/x/meg) would be better still.

gammazero (approved):

gammazero:
Triage: Consider updating IPIP to suggest different peer schema.
Blocked waiting for more progress in other areas to determine what is needed here.

Note: Should not assume trustless gateway when HTTP URI with no metadata

routing/http/types/address.go
lidel: i think "tls" is bit too vague, remove it for now

routing/http/types/address.go
lidel: I am worried about unintended consequences if w blindly convert here like IPNI func does, let's keep as-is.

Mind, the only reason this func exist, is backward compatibility with existing HTTP support for trustless gateways which have  "fake peerids" and "fake libp2p multiaddrs".

That is why we only do this for HTTP(S) schema URIs + have default ports.

lidel: Found interesting inconsistency:

IPNI's `FromURL` passes `url.PathEscape(u.Path)` without stripping the leading `/`, producing `%2Fpath` instead of `path` so you have `/http-path/%2Fpath` instead of  `/http-path/path.

The only spec I found is example in https://github.com/multiformats/multiaddr/blob/master/protocols/http-path.md 

which suggests the leading `/` should be stripped to avoid `%2F` (means IPNI has a bug, but its probably too late to fix).

gammazero:
Triage: still awaiting decision.

PR #1053, opened 2025-10-14
bitswap: link traces

dennis-tra:
**Changelog:**

- added a new span to bitswap's `handleIncoming` plus `received block` events
- marked the spans/attributes/events that we at ProbeLab use for our analysis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- new span for the `handleIncoming` bitswap client `getter` plus events when blocks are received.
+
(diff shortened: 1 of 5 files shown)


gammazero:
This _code_ seems reasonable, but I am not sure we want to add ProbeLab specific comments to our codebase.

dennis-tra:
> This code seems reasonable, but I am not sure we want to add ProbeLab specific comments to our codebase.

This was suggested by @lidel (not the specific phrasing, just the marks). The rationale was that we wanted to avoid our analysis code to break going forward into the future.

gammazero (approved):

PR #1057, opened 2025-10-23
bitswap/httpnet: improve "Connect"/testCid check

hsanjuan:
Now that we have certain safeguards to disconnect from small providers, we can be more open to connect to places that fail the "testCid request". Particularly places that return 500 with known error messages that prove that they understand ipld and that they are a gateway.

### DIFF
--- a/bitswap/network/httpnet/httpnet.go
+++ b/bitswap/network/httpnet/httpnet.go
@@ -542,12 +542,16 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	supportsHead := true
 	for _, u := range urls {
 		// If head works we assume GET works too.
-		err := ht.connectToURL(ctx, pi.ID, u, "HEAD")
+		status, err := ht.connectToURL(ctx, pi.ID, u, "HEAD")
 		if err != nil {
 			errs = append(errs, fmt.Errorf("%s: %s", u.Multiaddress.String(), err))
 			// abort if context cancelled
 			if ctxErr := ctx.Err(); ctxErr != nil {
 				return errors.Join(errs...)
+			}
+
+			if status == http.StatusTooManyRequests {
+				continue // do not try GET, just move on.
 			}
 		} else {
 			workingAddrs = append(workingAddrs, u.Multiaddress)
@@ -557,7 +561,7 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		// HEAD did not work. Try GET.
 		supportsHead = false
 
-		err = ht.connectToURL(ctx, pi.ID, u, "GET")
+		_, err = ht.connectToURL(ctx, pi.ID, u, "GET")
 		if err != nil {
 			errs = append(errs, fmt.Errorf("%s: %s", u.Multiaddress.String(), err))
 			if ctxErr := ctx.Err(); ctxErr != nil {
@@ -591,17 +595,17 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	return nil
 }
 
-func (ht *Network) connectToURL(ctx context.Context, p peer.ID, u network.ParsedURL, method string) error {
+func (ht *Network) connectToURL(ctx context.Context, p peer.ID, u network.ParsedURL, method string) (int, error) {
 	req, err := buildRequest(ctx, u, method, pingCid, ht.userAgent)
 	if err != nil {
 		log.Debug(err)
-		return err
+		return 0, err
 	}
 
 	log.Debugf("connect/ping request to %s %s %q", p, method, req.URL)
 	resp, err := ht.client.Do(req)
 	if err != nil {
-		return err
+		return 0, err
 	}
 
 	// For HTTP, the address can only be a LAN IP as otherwise it would have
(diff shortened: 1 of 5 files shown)


bitswap/network/httpnet/httpnet.go
```diff
@@ -590,17 +595,17 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	return nil
 }
 
-func (ht *Network) connectToURL(ctx context.Context, p peer.ID, u network.ParsedURL, method string) error {
+func (ht *Network) connectToURL(ctx context.Context, p peer.ID, u network.ParsedURL, method string) (int, error) {
```
gammazero, line 599: It would be good to document whether or not checking the status is applicable if the returned error is `nil`. Or, return an error that includes the status.

bitswap/network/httpnet/httpnet.go
```diff
@@ -590,17 +595,17 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
...
-		return err
+		return 0, err
 	}
 
 	log.Debugf("connect/ping request to %s %s %q", p, method, req.URL)
 	resp, err := ht.client.Do(req)
 	if err != nil {
-		return err
+		return 0, err
 	}
```
gammazero, line 610: Need to close response body.
```suggestion
		return 0, err
	}
	defer resp.Body.Close()
```

gammazero (changes requested):

bitswap/network/httpnet/msg_sender.go
```diff
@@ -276,10 +278,13 @@ func (sender *httpMsgSender) tryURL(ctx context.Context, u *senderURL, entry bsm
...
 	// 1) Observed that some gateway implementation returns 500 instead of
 	// 404.
-	if statusCode == 500 &&
-		(string(body) == "ipld: could not find node" || strings.HasPrefix(string(body), "peer does not have")) {
-		log.Debugf("treating as 404: %q -> %d: %q", req.URL, resp.StatusCode, string(body))
-		statusCode = 404
+	if statusCode != 200 {
+		if strings.HasPrefix(string(body), "ipld: could not find node") ||
+			strings.HasPrefix(string(body), "peer does not have") ||
+			strings.HasPrefix(string(body), "getting pieces containing cid") {
```
lidel: nit: consider extracting the error pattern check - The list of "known good errors" appears in two places. Consider a shared function like `isKnownNotFoundError(body string) bool` and call it in both places

bitswap/network/httpnet/msg_sender.go
```diff
@@ -276,10 +278,13 @@ func (sender *httpMsgSender) tryURL(ctx context.Context, u *senderURL, entry bsm
 	statusCode := resp.StatusCode
 	// 1) Observed that some gateway implementation returns 500 instead of
 	// 404.
-	if statusCode == 500 &&
-		(string(body) == "ipld: could not find node" || strings.HasPrefix(string(body), "peer does not have")) {
-		log.Debugf("treating as 404: %q -> %d: %q", req.URL, resp.StatusCode, string(body))
-		statusCode = 404
+	if statusCode != 200 {
```
aschmahmann: Why are we carrying these forward and increasing the number of 5xx's we're processing as 404s? If we're expecting other clients (e.g. helia, etc.) to play the same games here aren't we effectively carrying more spec/implementor weight around?

IIUC most/all of the server implementations that used to have these 5xx errors now have 404s but even if they didn't and this was meant as a temporary measure for any boxo clients who can upgrade faster than those servers can change their returned status codes it seems like these should come with associated issues for their removal pending the servers changing their returned status codes (and then however long until we remove these afterwards).

hsanjuan: Not sure I follow, we just encountered last week someone answering with `500 - getting pieces containing cid...`. If we can get more gateways to work, that is positive for everyone else. Introducing a line here is easier than chasing people around hoping that they implement something, but that can also be  done in parallel.

bitswap/network/httpnet/httpnet.go
hsanjuan: See now... 

the problem is I need to cut Connect short if HEAD returns a TooManyRequests, so that is a bit special because for any other error we still try a GET.

bitswap/network/httpnet/msg_sender.go
lidel: In my mind this is in the spirit of https://specs.ipfs.tech/architecture/principles/#robustness. 
Spec describes ideal world, implementation reflects reality.  We don't want to punish end users who want to get to their data for the fact someone did not implement spec 100% correct. Mind we already bend backwards in many many other places  (e.g. the legacy IPLD "not found" errors are special-cased all over the stack, or special handling of nodes who don't support HEAD, we should error, but we fallback to GET etc).

I've added godoc with rationale in d0b207486bbdb25e23cbb2505b11cf7e5c79c6e9 so it is clear for someone who reads our code trying to implement their own thing.

Agree with Hector that you can reach out to people asking them to fix things. but we should not punish our users for someone's else mistakes + be realistic about current ecosystem's ability to force everyone to upgrade.

aschmahmann: What I'm saying is that:
1. I'm pretty sure all of the server code bases this covers have upgraded
2. Sure, saying then any boxo clients that upgrade before the servers do will get some benefits, seems reasonable to me
3. But we should likely have a timebox / criteria for saying we don't need a given hack anymore
4. It's IMO likely that we already meet the criteria for some of these

gammazero (approved):

lidel (approved):

PR #1058, opened 2025-10-23
fix(gateway): deduplicate peer IDs in retrieval diagnostics

lidel:
peer IDs could appear multiple times in timeout errors due to repeated connection failures. add deduplication check. 

cc @aschmahmann (thx for noticing!)

### DIFF
--- a/retrieval/state_test.go
+++ b/retrieval/state_test.go
@@ -276,6 +276,34 @@ func TestState(t *testing.T) {
 		assert.True(t, hasID1 || hasID2, "Summary should contain at least one of the peer IDs")
 	})
 
+	t.Run("Duplicate peer IDs are filtered out", func(t *testing.T) {
+		rs := NewState()
+		peerID1 := test.RandPeerIDFatal(t)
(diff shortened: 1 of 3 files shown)


retrieval/state_test.go
```diff
@@ -276,6 +276,34 @@ func TestState(t *testing.T) {
...
+		// Should only contain unique peer IDs
+		failedProviders := rs.GetFailedProviders()
+		assert.Len(t, failedProviders, 2, "Should only contain 2 unique peer IDs")
+		assert.Contains(t, failedProviders, peerID1)
+		assert.Contains(t, failedProviders, peerID2)
+
+		// Also test for found providers
+		rs.AddFoundProvider(peerID1)
+		rs.AddFoundProvider(peerID1)
+		rs.AddFoundProvider(peerID2)
```
aschmahmann, line 299: nit: Any particular reason to add/not add the peer again like we do above?

aschmahmann (approved):
LGTM (albeit this code path is new and I'm not very familiar with it)

PR #1059, opened 2025-10-24
fix(tracing): use context to pass trace and retrieval state to session

### LINKED ISSUE #1052: bitswap/session: context-aware lifecycle
I'm trying to trace a content retrieval request via `ipfs get ...` with the actual bitswap interactions.

This is tricky though because a bitswap client session creates a [new disparate context](https://github.com/ipfs/boxo/blob/0f1ebacc7662d946946e3315c49241f6fb188b7c/bitswap/client/internal/session/session.go#L164). Then I also found [this comment](https://github.com/ipfs/boxo/blob/0f1ebacc7662d946946e3315c49241f6fb188b7c/bitswap/client/internal/session/session.go#L209) for the `Close()` method which only cancels the session context:

```go
// Session lifecycle is independent of the context used to create requests -
// canceling a request context does not close the session. When context-based
// automation is desired, Client.NewSession(ctx) uses context.AfterFunc to
// automatically close sessions when the context is canceled.
func (s *Session) Close() {
	s.cancel()
}
```

@gammazero I saw that this comment is from you. Is the meaning descriptive or normative? Is it just describing how it's working or stating how it's supposed to work?

Looking at the session lifecycle (checking where new sessions are created and closed) I can see that they are always tied to the context that instantiated the session. This to me indicates that a separate context for the session isn't needed which would enable proper context propagation and improve the tracing. As a nice byproduct, the `retrievalState` wouldn't need to be passed manually through the stack.

New sessions are created here:

https://github.com/ipfs/boxo/blob/0f1ebacc7662d946946e3315c49241f6fb188b7c/bitswap/client/client.go#L442-L483

and here:

https://github.com/ipfs/boxo/blob/0f1ebacc7662d946946e3315c49241f6fb188b7c/bitswap/client/client.go#L702-L725

Both calls tie the new session to the incoming context.

The same argument then actually trickles down to the `sessionWantSender` as its lifecycle is tied to the sessions':

https://github.com/ipfs/boxo/blob/0f1ebacc7662d946946e3315c49241f6fb188b7c/bitswap/client/internal/session/sessionwantsender.go#L117-L142

gammazero:
Trace state, carried in the context passed to NewSession was not being passed into the session. This is fixed by having session.New accept a context that contains trace and retrieval state.

- Closes #1052
- Reverses part of https://github.com/ipfs/boxo/pull/1011

### DIFF
--- a/bitswap/client/client.go
+++ b/bitswap/client/client.go
@@ -299,6 +299,7 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder ro
 	}
 
 	sessionFactory := func(
+		sessctx context.Context,
 		sessmgr bssession.SessionManager,
 		id uint64,
 		spm bssession.SessionPeerManager,
@@ -309,7 +310,6 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder ro
 		provSearchDelay time.Duration,
 		rebroadcastDelay time.Duration,
 		self peer.ID,
-		retrievalState *retrieval.State,
 	) bssm.Session {
 		// careful when bs.pqm is nil. Since we are type-casting it
 		// into routing.ContentDiscovery when passing it, it will become
@@ -321,7 +321,7 @@ func New(parent context.Context, network bsnet.BitSwapNetwork, providerFinder ro
 		} else if providerFinder != nil {
 			sessionProvFinder = providerFinder
 		}
-		return bssession.New(sessmgr, id, spm, sessionProvFinder, sim, pm, bpm, notif, provSearchDelay, rebroadcastDelay, self, bs.havesReceivedGauge, retrievalState)
+		return bssession.New(sessctx, sessmgr, id, spm, sessionProvFinder, sim, pm, bpm, notif, provSearchDelay, rebroadcastDelay, self, bs.havesReceivedGauge)
 	}
 	sessionPeerManagerFactory := func(id uint64) bssession.SessionPeerManager {
 		return bsspm.New(id, network)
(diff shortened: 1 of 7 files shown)


bitswap/client/client.go
```diff
@@ -446,19 +446,26 @@ func (bs *Client) GetBlocks(ctx context.Context, keys []cid.Cid) (<-chan blocks.
...
 
-	session := bs.sm.NewSession(bs.provSearchDelay, bs.rebroadcastDelay, retrieval.StateFromContext(ctx))
+	// Temporary session closed indepentendly of cancellation ctx.
+	sessCtx, cancelSession := context.WithCancel(context.Background())
+
+	// Preserve retrieval.State from the original request context if present
+	if retrievalState := retrieval.StateFromContext(ctx); retrievalState != nil {
+		sessCtx = context.WithValue(sessCtx, retrieval.ContextKey, retrievalState)
+	}
+	session := bs.sm.NewSession(sessCtx, bs.provSearchDelay, bs.rebroadcastDelay)
```
lidel, line 456: @gammazero do you remember why this is separate session? do we detach if from trace on purpose? 

IIUC without passing trace provider discovery operations within this temporary session won't be traced, making debugging harder? Or is this decreasing the noise?

gammazero: I think that a separate context was used for the session, so that the context passes into `GetBlocks` would only cancel the call to `session.GetBlocks` and not cancel the session itself. This allows the goroutine created here to cancel/close the session when finished reading blocks.

The purpose of doing this is so that the caller does not have to explicitly cancel the context passed into this function, which otherwise is required to prevent leaking the session created here, since otherwise there is no way to cancel/close the session.

Looking at this code, it introduces a defect in that it prevents the propagation of tracing state carried by the context passed into `GetBlocks`. This can be fixed by making the passed in context be the parent of the session context, or by simply returning the blocks channel and depending on the caller to cancel the passed in context.

@lidel See #1060 for fix.

PR #1069, opened 2025-11-17
feat(gateway): add configurable fallback timeout for the gateway handler

acejam:
This PR adds a configurable fallback request timeout to the Gateway handler.

  - Introduces `fallbackRequestTimeout` (default 1h) used in `handler.ServeHTTP` instead of a hardcoded `time.Hour`
  - Reads optional env var `BOXO_GATEWAY_REQUEST_TIMEOUT` at init (Go duration format: "90m", "2h", "5400s")
  - Logs a warning and keeps default if the env value is invalid or non-positive
- Purpose: prevent long CAR streams or large responses from being truncated at exactly 1 hour due to a hardcoded context deadline

### Impact

- Default behavior remains 1 hour unless `BOXO_GATEWAY_REQUEST_TIMEOUT` is set
- Allows deployments to extend or disable (by setting a very large duration) the hard per-request cap independent of `RetrievalTimeout`

### Usage

```bash
# Example: raise fallback request timeout to 3 hours
export BOXO_GATEWAY_REQUEST_TIMEOUT="3h"
```


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `gateway`: Added a configurable fallback timeout for the gateway handler, defaulting to 1 hour, overridable via the BOXO_GATEWAY_REQUEST_TIMEOUT environment variable.
```
lidel, line 19: @acejam The existing gateway configuration uses a `boxo/gateway/Config` struct with explicit fields (e.g., `RetrievalTimeout`, `MaxConcurrentRequests`). Using an environment variable is inconsistent with the established pattern and we can't merge this in this form.

Consider adding `MaxRequestDuration` to the `Config` struct instead? This would be consistent with how other timeouts are configured and allows programmatic

lidel (changes requested):
Allowing users to override or remove 1h timeout here is a sensible feature, but wiring it up needs to be done in `Config` (see comment below)

gammazero:
Replaced by #1079

gammazero:
Completed in #1079

PR #1071, opened 2025-11-23
feat(ipld/merkledag): add total size of visited nodes in progress tracker

### LINKED ISSUE #8915: `pin add` progress objects should include total-size of nodes pinned so far.
### Checklist

- [X] My issue is specific & actionable.
- [X] I am not suggesting a protocol enhancement.
- [X] I have searched on the [issue tracker](https://github.com/ipfs/go-ipfs/issues?q=is%3Aissue) for my issue.

### Description

The `pin add` endpoint returns `AddPinOutput` objects.

When progress=true is specified, an object is set every 500ms containing the number of nodes that have been pinned (`Progress`  field) so far.

I would like to have an additional Size field that additionally contains total size of the DAG pinned so far.

This is necessary to:

* Abort pinning items that are too big
* Being able to inspect the data rate at which an ipfs node can pin.

Can I, or someone work on this?

vlerdman:
This PR adds tracking total size of nodes during graph walk. New code saves backward compability with ipfs/kubo.

### Changes

* Extend `merkledag.ProgressTracker` to save and update `TotalSize` of visited nodes
* Update progress via wrapping `GetLinksDirect` function instead of `visit`
* Update `testProgressIndicator` for checking `TotalSize`

### References

* Closes: https://github.com/ipfs/kubo/issues/8915


ipld/merkledag/merkledag.go
```diff
@@ -457,10 +464,18 @@ func sequentialWalkDepth(ctx context.Context, getLinks GetLinks, root cid.Cid, d
 	return nil
 }
 
+// ProgressStat represents the progress of a fetch operation.
+type ProgressStat struct {
+	// Total is the total number of nodes fetched.
+	Total int
+	// TotalSize is the total size of the nodes fetched.
+	TotalSize uint64
```
lidel: nit: good oportunity to be more clear than ambigous "Total", perhaps:

```go
  Nodes int    // number of nodes visited
  Bytes uint64 // total bytes of raw block data
```

lidel (changes requested):
@vlerdman thank you for submitting this – any chance you could also:
- [x] open a PR in https://github.com/ipfs/kubo that uses this in `ipfs pin add` command and has green CI there as well
? (it would speed up review process)

vlerdman:
> @vlerdman thank you for submitting this – any chance you could also:
> 
> * [x]  open a PR in https://github.com/ipfs/kubo that uses this in `ipfs pin add` command and has green CI there as well
>   ? (it would speed up review process)

@lidel Thanks for such fast feedback! Opened a PR - https://github.com/ipfs/kubo/pull/11074

ipld/merkledag/merkledag.go
vlerdman: Changed

hsanjuan:
Triage:
  * Will be reviewed and probably going into 0.40.

gammazero (approved):

PR #1074, opened 2025-12-01
feat(gateway): IPIP-523 format query over Accept header

lidel:
this change simplifies precedence rules by making the `?format=` URL query parameter always take priority over the Accept HTTP header when both are present.

in practice, this is largely compatible with existing browser use cases since browsers send Accept headers with wildcards which were already treated as non-specific. prioritizing `?format=` also ensures deterministic HTTP caching behavior, protecting against CDNs that commingle different response types under the same cache key.

the only breaking change is for edge cases where a client sends both a specific `Accept` header and a different `?format=` value. previously `Accept` would win, now `?format=` wins. this scenario is rare and arguably represents client misconfiguration. when detected, gateway returns HTTP 400 to signal the ambiguity.

- specs: https://github.com/ipfs/specs/pull/523
- tests: https://github.com/ipfs/gateway-conformance/pull/252

## TODO

- [x] Kubo PR which uses this PR and  tests from https://github.com/ipfs/gateway-conformance/pull/252 and the CI is green there


lidel (commented):
CI with end-to-end tests at
- https://github.com/ipfs/kubo/pull/11086

are green.

AFAIK this PR is a low risk change (see **Compatibility** section in [IPIP-523](https://github.com/ipfs/specs/pull/523))

lidel (commented):
Switched to [gateway-conformance@v0.9](https://github.com/ipfs/gateway-conformance/releases/tag/v0.9.0), merging.

This is low risk change, we will test on gateway staging with Kubo 0.40.0-rc1.

PR #1075, opened 2025-12-02
routing: un-deprecate legacy delegated routing API

gammazero:
Since [IPIP-378](https://github.com/ipfs/specs/pull/378) is not moving forward, the legacy API is not deprecated because there nothing to replace it.


lidel (commented):
After discussing with @aschmahmann and @gammazero we decided to keep this deprecated.

My thinking is: close this, and instead update godoc in https://github.com/ipfs/boxo/pull/1076 to reference more relevant IPIP.

Thoughts?

gammazero:
> Thoughts?

... Closing

PR #1077, opened 2025-12-03
feat(gateway): IPIP-0524 + AllowCodecConversion config option

lidel:
Add `AllowCodecConversion` to `gateway.Config` to control codec conversion behavior per [IPIP-0524](https://github.com/ipfs/specs/pull/524). When false (default), the gateway returns 406 Not Acceptable if the requested format doesn't match the block's codec. When true, conversions between codecs are performed for backward compatibility.

Codec conversion tests moved here from gateway-conformance since conversions are now an optional implementation feature, not a spec requirement. Gateway-conformance now tests for 406 responses.

- Ref: https://github.com/ipfs/specs/pull/524
- Ref: https://github.com/ipfs/gateway-conformance/pull/254


gateway/handler_codec.go
```diff
@@ -148,7 +148,20 @@ func (i *handler) renderCodec(ctx context.Context, w http.ResponseWriter, r *htt
 		return false
 	}
 
-	// This handles DAG-* conversions and validations.
+	// IPIP-0524: Check if codec conversion is allowed
+	if !i.config.AllowCodecConversion && toCodec != cidCodec {
+		// Conversion not allowed and codecs don't match - return 406
+		err := fmt.Errorf("format %q requested but block has codec %q: codec conversion is not supported", rq.responseFormat, cidCodec.String())
```
lidel: ⚠️  this feels like invasive breaking change, making it impossible to request `dag-json` as a generic `?format=json` or `dag-cbor` as  generic `?format=cbor`

If we trully don't allow conversions, these requests will now error on gateways.

I do not know if anyone is using these conversions, so in boxo an kubo I'm going to keep configuration option to opt-in back to old behavior if someone self-hosts.

cc @achingbrain for visibility

lidel: I'm assuming this is a niche feature that can go away as long we have meaningful error message in boxo/gateway.

Relaxed spec a bit to allow implementation to handle `Accept: application/json` in more interoperable way, but that is MAY, not mandatory from spec perspective.

lidel:
I'm going to merge this and include commit from boxo `main` branch with it in [Kubo 0.40](https://github.com/ipfs/kubo/issues/11008) RC1, together with the other two IPIPs (#1074 and #1088) . This way we group all "formally breaking changes" in one release that saves people time on testing (can smoke-test everything in one go).

PR #1078, opened 2025-12-07
bitswap: add the requesting peer ID to context

burdiyan:
This commit adds the peer.ID of the requesting peer to the context, which can be used downstream in the blockstore or other components used by Bitswap.

This could be very useful for custom logging, or accessing the peer ID in custom block stores.


hsanjuan:
Triage:
  * We assume you have tested this and it works for you (the modified context ends up at the blockstore.Get() or blockstore.GetMany() layer) (we don't know if this is the case).
  * To accept this, we would need a test in `bitswap` with a mocked blockstore that reads back the peer ID from the context. Otherwise it is very likely that this will somehow break in a future refactor without anyone noticing.

lidel:
Triage notes:
- main blocker: this should be opt-in via config option when constructing bitswap instance
- needs tests 
- FYSA we have something similar  in https://github.com/burdiyan/boxo/blob/main/bitswap/client/traceability/block.go where we have ability to  enable extra metadata (and its opt-in)

burdiyan:
I finally found a workaround for the problem I was trying to fix, so I won't be following through with this PR. Sorry for the noise.

PR #1079, opened 2025-12-13
feat(gateway): configurable fallback timeout for the gateway handler

gammazero:
Added a configurable fallback timeout for the gateway handler, defaulting to 1 hour. Configurable via `MaxRequestDuration` in the gateway config.

Replaces #1069

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -137,8 +137,10 @@ type Config struct {
 	// (e.g., Cloudflare's 5GB limit). A value of 0 disables this limit.
 	MaxRangeRequestFileSize int64
 
-	// MaxRequestDuration per-request timeout for the gateway handler. It
-	// defaults to DefaultMaxRequestDuration.
+	// MaxRequestDuration is the maximum total time a request can take.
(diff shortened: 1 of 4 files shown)


gateway/gateway.go
```diff
@@ -133,6 +137,10 @@ type Config struct {
 	// (e.g., Cloudflare's 5GB limit). A value of 0 disables this limit.
 	MaxRangeRequestFileSize int64
 
+	// MaxRequestDuration per-request timeout for the gateway handler. It
+	// defaults to DefaultMaxRequestDuration.
+	MaxRequestDuration time.Duration
```
gammazero, line 144: Does this need to be a duration string that can be parsed (e.g. "1h")?

lidel: That is up to the user of `boxo/gateway` library.
Prior art (`RetrievalTimeout time.Duration`) in Kubo is wired in a way that support human-readable strings, but that is userland decision, we dont do that in boxo here.

lidel (approved):

PR #1082, opened 2026-01-07
refactor: replace `go-clock` with `synctest`

gammazero:
- Replace `go-clock` in `bitswap/client/internal/messagequeue`
- Replace `go-clock` in `bitswap/server`
- Replace `go-clock` in `routing/http/client`

blocked: waiting until go 1.25 is required


lidel (approved):
Restored 1.24 and added `//go:build go1.25` constraints to test files that use testing/synctest. This way:

- go.mod stays at 1.24.6 (no breaking change for downstream users)
- Tests using `synctest` are skipped on Go 1.24 CI jobs
- Tests run normally on Go 1.25 CI jobs (go `next` targets)

Also moved two test-only helper methods from `engine.go` to `engine_test.go` so they get excluded together with the tests (avoids unused function errors on Go 1.24).

@gammazero  This way  (https://github.com/ipfs/boxo/pull/1082/changes/29910a6500860451d385dcfa352015309fc94437) we are unblocked, no need to wait for Go 1.26 - this should pass CI as-is. Feel free to merge if the changes look ok.

PR #1085, opened 2026-01-08
feat(ipns): Support adding custom metadata to IPNS record

phaseloop:
Allow adding custom metadata to IPNS record - stored as additional keys in DAG-CBOR data.

IPNS specification allows to store arbitrary data in the record but this was not supported by `NewRecord()` method. 

cc: @lidel


ipns/record.go
```diff
@@ -237,6 +238,12 @@ func WithPublicKey(embedded bool) Option {
 	}
 }
 
+func WithMetadata(metadata map[string]ipld.Node) Option {
```
lidel: The PR's `WithMetadata(map[string]ipld.Node)` exposes `go-ipld-prime` in the public API, which:

1. Forces users to import go-ipld-prime just to use metadata
2. Couples boxo's public API stability to a third-party library
3. Makes the API harder to use for simple cases

Good news is that we already have prior art idiomatic patterns for handling this in boxo.

The `ipns` package already handles this well - the `Record` struct wraps `datamodel.Node` internally (line 42) but exposes typed methods (`Value()`, `Sequence()`, `TTL()`) instead of the raw node. This is the pattern to follow.


@phaseloop could you refactor to  expose local type? It could be a thin wrapper internally (pseudocode below, feel free to change/improve, just food for thought):

```go
// MetadataValue represents a value that can be stored in IPNS record metadata.
  // Use the constructor functions to create values.
  type MetadataValue struct {
      node ipld.Node
  }

  // StringValue creates a string metadata value.
  func StringValue(s string) MetadataValue {
      return MetadataValue{node: basicnode.NewString(s)}
  }

  // BytesValue creates a bytes metadata value.
  func BytesValue(b []byte) MetadataValue {
      return MetadataValue{node: basicnode.NewBytes(b)}
  }

  // IntValue creates an integer metadata value.
  func IntValue(i int64) MetadataValue {
      return MetadataValue{node: basicnode.NewInt(i)}
  }

  // BoolValue creates a boolean metadata value.
  func BoolValue(b bool) MetadataValue {
      return MetadataValue{node: basicnode.NewBool(b)}
  }

  // WithMetadata sets custom metadata entries for the IPNS record.
  // Keys should be prefixed with "_" per the IPNS spec to avoid
  // collisions with future standard fields.
  func WithMetadata(metadata map[string]MetadataValue) Option {
      return func(o *options) {
          m := make(map[string]ipld.Node, len(metadata))
          for k, v := range metadata {
              m[k] = v.node
          }
          o.metadata = m
      }
  }
```

Usage becomes:

```go
meta := map[string]ipns.MetadataValue{
      "_myapp/version": ipns.StringValue("1.0"),
      "_myapp/data":    ipns.BytesValue(someBytes),
      "_myapp/count":   ipns.IntValue(42),
  }
  rec, err := ipns.NewRecord(sk, path, seq, eol, ttl, ipns.WithMetadata(meta))
```

For reading Metadata back, we could add typed accessors on `Record`:

```go
// Metadata returns a custom metadata value by key.
  // Returns ErrMetadataNotFound if the key doesn't exist.
  func (rec *Record) Metadata(key string) (MetadataValue, error) {
      node, err := rec.node.LookupByString(key)
      if err != nil {
          return MetadataValue{}, fmt.Errorf("%w: %s", ErrMetadataNotFound, key)
      }
      return MetadataValue{node: node}, nil
  }

  // AsString returns the value as a string.
  func (mv MetadataValue) AsString() (string, error) {
      return mv.node.AsString()
  }

  // AsBytes returns the value as bytes.
  func (mv MetadataValue) AsBytes() ([]byte, error) {
      return mv.node.AsBytes()
  }

  // AsInt returns the value as an int64.
  func (mv MetadataValue) AsInt() (int64, error) {
      return mv.node.AsInt()
  }

  // AsBool returns the value as a bool.
  func (mv MetadataValue) AsBool() (bool, error) {
      return mv.node.AsBool()
  }
```


We probably want to limit the complexity here and avoid implementing complex nested metadata. All you need for complex cases is the  ability to store and read `[]byte`.

lidel (changes requested):
Thank you for opening this @phaseloop, having ability to set/get metadata CBOR feels sensible, but this PR needs bit more work around two things:

1. We should not use third-party type from legacy IPLD library in public API here
2. You implemented way to set, but its missing method to read custom metadata from unmarshaled records


I suggested one way of addressing both below, but other ideas welcome.
In general, we want to move away from IPLD language and focus on deterministic subset of CBOR (dag-cbor). 

Note that DAG-CBOR itself becomes problematic if you try to store something more than basic types:
- https://hyphacoop.github.io/dasl-testing/?group=tests-by-file&tag=dag-cbor

This is why my suggestion is to limit the API for storing/reading metadata to raw `[]byte` with entire dag-cbor + limited ability to set specific key as  either String, `[]byte`, boolean and maybe Integer (look at linked test results and pick minimal set of CBOR types that works everywhere)

ipns/record.go
```diff
@@ -332,10 +339,18 @@ func newRecord(sk ic.PrivKey, value []byte, seq uint64, eol time.Time, ttl time.
 	}, nil
 }
 
-func createNode(value []byte, seq uint64, eol time.Time, ttl time.Duration) (datamodel.Node, error) {
+func createNode(value []byte, seq uint64, eol time.Time, ttl time.Duration, metadata map[string]ipld.Node) (datamodel.Node, error) {
 	m := make(map[string]ipld.Node)
 	var keys []string
 
+	// copy metadata values first - so if there is naming conflict,
```
lidel: nit: silently overwriting a genmeric key named `Value` or `TTL` or `Sequence` will lead to hard to debug errors for someone. 

let's save that poor soul, and instead of silently overwriting, detect it, and return error with clear error message that "TTL", "Value", "Sequence", "Validity", "ValidityType" are reserved for IPNS Record itself

phaseloop:
Thanks a lot! I'll work on it. Off topic question: why DAG-CBOR was chosen instead of protobuf if implementations are so fragmented and buggy outside basic types? Is the storage saving so big compared to protobuf that it was worth the hassle? CBOR/DAG-CBOR seems to be pretty obscure.

lidel:
Thanks! I'd prefer to avoid mischaracterizing work done by others, but my understanding is that the signed CBOR map was introduced years ago hastily to fix a security issue in V1, it was never a well thought out decision for metadata. But since we have this signed CBOR map, we can put other values there now.

I agree, CBOR creates unnecessary complexity, but it is too late to change V2 without breaking existing clients. Note that to this day, Kubo is producing V1+V2 records to keep backward compatibility. 

This PR is probably not the place to discuss it, so let's avoid off-topic, but if you are interested in cleaning up this, feel free to open discussion in https://github.com/ipfs/specs/pulls (in theory, we could introduce V3 of IPNS records  that improves the wire format, but upgrade path for ecosystem is a tricky part, see some prior art in https://specs.ipfs.tech/ipips/ipip-0428/).

phaseloop:
Thanks! I was just curious, so far I have no intention to rework IPNS (if it works, it works) - but I'm working on using IPNS as distributed DNS system:

https://github.com/ipfs/specs/issues/528

phaseloop:
@lidel I reworked the PR according to your comments

phaseloop:
@gammazero PR was modified according to comments, AFAIK "need/author-input" label can be removed now :)

ipns/record.go
```diff
@@ -42,6 +42,75 @@ type Record struct {
 	node datamodel.Node
 }
 
+type ErrMetadataNotFound struct {
```
lidel: @phaseloop nit: the existing ipns package uses sentinel var errors everywhere (ErrExpiredRecord, ErrInvalidRecord, etc. in errors.go). This PR introduces ErrMetadataNotFound and ErrMetadataConflict as struct types, which means:
- errors.Is(err, ErrMetadataNotFound{}) won't work as callers might expect (they'd need errors.As)
- The empty struct bodies (type ErrMetadataNotFound struct{}) are unnecessary boilerplate

mind refactoring to sentinel var errors matching the existing pattern:
``go
var ErrMetadataNotFound = errors.New("metadata key not found in record")
var ErrMetadataConflict = errors.New("metadata key uses reserved name")
```

File: ipns/record_test.go [hunk 4]
```diff
@@ -214,7 +214,12 @@ func TestCBORDataSerialization(t *testing.T) {
 	seq := uint64(1)
 	ttl := time.Hour
 
-	rec := mustNewRecord(t, sk, path, seq, eol, ttl)
+	// custom metadata - one bytes entry and one map entry
+	metadata := map[string]MetadataValue{
+		"_metadata1": MetadataValueFromString("test"),
+	}
+
+	rec := mustNewRecord(t, sk, path, seq, eol, ttl, WithMetadata(metadata))
```
lidel: nit: should not modify `TestCBORDataSerialization` which tests the default base behavior

extra metadata should be tested in own, separate unit test

ipns/record.go
```diff
@@ -332,10 +415,22 @@ func newRecord(sk ic.PrivKey, value []byte, seq uint64, eol time.Time, ttl time.
 	}, nil
 }
 
-func createNode(value []byte, seq uint64, eol time.Time, ttl time.Duration) (datamodel.Node, error) {
+func createNode(value []byte, seq uint64, eol time.Time, ttl time.Duration, metadata map[string]ipld.Node) (datamodel.Node, error) {
 	m := make(map[string]ipld.Node)
 	var keys []string
 
+	reservedKeys := []string{cborValueKey, cborValidityKey, cborValidityTypeKey, cborSequenceKey, cborTTLKey}
```
lidel: nit: this slice is rebuilt on every call to createNode

This is a package-level constant set. Could be a var at package level or sth :)

lidel (changes requested):
@phaseloop thanks, i had no bandwidth to do proper review in Kubo (hopefully after 0.40 ships), but small asks inline + you may want to rebase this on top of latest main

phaseloop:
@lidel - thanks for suggestions, done

lidel (approved):
Thank you.
LGTM, pushed slight rework of the API:

- Changed write API from `map[string]MetadataValue` to `map[string]any` for ergonomic Go literals instead of constructor wrappers
- Added `MetadataEntries() iter.Seq2[string, MetadataValue]` for iterating custom keys
- Added `MetadataKind` type with `Kind()` discriminator for inspecting values read from third-party records
- Added nil guards on `MetadataValue` accessors to prevent panics on zero values
- Added `int` support alongside `int64`
- Added validation for empty keys, nil values, and unsupported types with dedicated errors
- Added reserved key guard on read, not just write (`Metadata()` returns `ErrMetadataConflict`, `MetadataExists()` returns `false`)
- Changed `reservedKeys` from slice to map for O(1) lookups
- Expanded test coverage with cases for edge cases and CBOR key ordering
- Improved godocs across the `ipns` package with spec links to https://specs.ipfs.tech/ipns/ipns-record/

Issue #626, opened 2024-06-21
flaky test: TestCancelOverridesPendingWants


lidel:
Similar to https://github.com/ipfs/boxo/issues/625, fails on windows, sometimes:

```
--- FAIL: TestCancelOverridesPendingWants (0.17s)
      messagequeue_test.go:333: Wrong message count
  FAIL
```

PR #1087, opened 2026-01-15
refactor: rewrite some flaky tests to testing/synctest

lidel:
refactor tests to use testing/synctest (Go 1.25) 

- 7d2280f to fix flaky gateway tests

- d5595ce048beae307e204b7931d2ac3660009a0a fix flaky TestCancelOverridesPendingWants (my fix from #626 did not fix all races)

## Note

Most of diff is just indentation, it may be easier to review by ignoring whitespace changes:

> image

### DIFF
--- a/bitswap/client/internal/messagequeue/messagequeue_test.go
+++ b/bitswap/client/internal/messagequeue/messagequeue_test.go
@@ -1,3 +1,5 @@
+//go:build go1.25
+
 package messagequeue
 
 import (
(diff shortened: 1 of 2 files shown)


bitswap/client/internal/messagequeue/messagequeue_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build go1.25
```
lidel, line 1: @gammazero fysa i went with this, because we already run tests against both 1.24 (`go this`) and 1.25 (`go next`), so only running in 1.25+ feels like a good compromise (when go unified CI switches to 1.26 in february, tests will run on both `this` and `next` go release again)

gammazero (approved):

PR #1088, opened 2026-01-17
feat(unixfs): configurable CID Profiles from IPIP-499

lidel:
- This PR is wiring up https://github.com/ipfs/specs/pull/499
  - Used in https://github.com/ipfs/kubo/pull/11148
  - Used in https://github.com/ipfs/go-ipfs-cmds/pull/315

Adds building blocks for reproducible CID generation across IPFS implementations, based on [IPIP-499](https://github.com/ipfs/specs/pull/499)

### Users of boxo can now

  - Choose how directory size is estimated for HAMT sharding: legacy link-based (`SizeEstimationLinks`), accurate block-based (`SizeEstimationBlock`), or disable size-based thresholds entirely
  (`SizeEstimationDisabled`)
  - Apply predefined import settings with `UnixFS_v0_2015` or `UnixFS_v1_2025` profiles from [IPIP-499](https://github.com/ipfs/specs/pull/499) via `ApplyGlobals()` and `CidBuilder()` 
  - Resolve all (not just roots) symlinks to their target content during file traversal with `SerialFileOptions.DereferenceSymlinks`
  - Get consistent HAMT behavior with JS implementation (threshold comparison changed from `>=` to `>`) – https://github.com/ipfs/boxo/pull/1088/commits/6707376002a3d4ba64895749ce9be2e00d265ed5

Related: [IPIP-499](https://github.com/ipfs/specs/pull/499), used by [kubo#11148](https://github.com/ipfs/kubo/pull/11148)


ipld/unixfs/io/profile.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// ApplyGlobals sets the global variables to match this profile's settings.
+// This affects all subsequent file and directory import operations.
+// Note: RawLeaves and CidBuilder are not globals; pass them to DAG builder options.
+//
+// Thread safety: this function modifies global variables and is not safe
+// for concurrent use. Call it once during program initialization, before
+// starting any imports. Do not call from multiple goroutines.
+func (p UnixFSProfile) ApplyGlobals() {
```
lidel, line 100: ℹ️ boxo already had globals (`DefaultBlockSize`, `DefaultLinksPerBlock` etc)

this is a very surgical way of having predefined UnixFS profiles in boxo itself that users can apply programmatically at startup of their app. 

i dont like it tbh, but others in golang that we already use (like certmagic, or even `net.DefaultResolver`) have similar way of managing global defaults, so maybe its just me not being a fan of globals.

this is a compromise which delivers ability to set-and-forget profile on startup, but implemented in smallest amount of code that avoids breaking every user of existing APIs

gammazero (approved):
I added a circular symlink test to show that it works when not dereferencing symlinks, otherwise it fails.

lidel:
Triage: 
- now that we agreed on `SizeEstimationBlock` we can optimize to avoid unnecessary calculations and also add tests for edge cases like mode/mtime changes

lidel:
@gammazero @aschmahmann i've applied feedback from today's colo
- `SizeEstimationBlock` no longer triggers unnecessary protobuf serialization
- unified estimatedSize calculation functions, should be easier to follow/maintain
- added more tests (mode and mtime behaviors etc)
- https://github.com/ipfs/kubo/pull/11148 switched to the latest version from here

ipld/unixfs/io/directory.go
```diff
@@ -102,15 +106,60 @@ func linkSerializedSize(name string, c cid.Cid, tsize uint64) int {
...
+//
+// If new fields are added to the UnixFS spec for directories, this function
+// must be updated. The test TestDataFieldSerializedSizeMatchesActual verifies
+// this calculation against actual protobuf serialization.
+func dataFieldSerializedSize(mode os.FileMode, mtime time.Time) int {
+	// Inner UnixFS Data message
+	innerSize := 0
+
+	// Type field (field 1, varint): Directory = 1
+	// tag = (1 << 3) | 0 = 8 (1 byte), value = 1 (1 byte)
```
gammazero: It is not immediately clear what the `| 0 = 8` part of this comment means.

gammazero (approved):

chunker/parse.go
```diff
@@ -8,13 +8,16 @@ import (
...
 )
 
-const (
-	// DefaultBlockSize is the chunk size that splitters produce (or aim to).
-	DefaultBlockSize int64 = 1024 * 256
+// DefaultBlockSize is the chunk size that splitters produce (or aim to).
+// Can be modified to change the default for all subsequent chunker operations.
+// For CID-deterministic imports, prefer using UnixFSProfile presets from
+// ipld/unixfs/io/profile.go which set this and other related globals.
+var DefaultBlockSize int64 = 1024 * 256
```
gammazero, line 15: Would it be best to restrict the block size to always being a factor of `ChunkSize`? If so, then this could be changed to `BlocksPerChunk`

lidel: Likely we don't want to glue them together. IIUC the chunk size and block size always been independent concepts in UnixFS (as a function of protobuf envelope coming with overhead). There's no requirement that one divides the other.

ipld/unixfs/io/directory.go
lidel: Rephrased this comment in https://github.com/ipfs/boxo/pull/1088/commits/009ea0bc9783070abfdf65e8ce95e7a42b9fdc9f to avoid bit operation shortcuts.
If there is better way to describe it, feel free to change.

ipld/unixfs/mod/dagmodifier.go
```diff
@@ -161,7 +161,13 @@ func (dm *DagModifier) expandSparse(size int64) error {
 		return err
 	}
 	err = dm.dagserv.Add(dm.ctx, nnode)
-	return err
+	if err != nil {
+		return err
+	}
+	// Update curNode so subsequent writes use the expanded node.
+	// Without this, writes after sparse expansion would go to the old node.
+	dm.curNode = nnode
```
gammazero, line 148: nice fix 🚀

ipld/unixfs/io/directory.go
```diff
@@ -723,29 +747,27 @@ func (d *BasicDirectory) needsToSwitchByBlockSize(name string, nodeToAdd ipld.No
...
 		}
-	} else { // existed
-		d.totalLinks--
+	} else {
+		existed = true
 	}
 
-	if d.maxLinks > 0 && d.totalLinks+1 > d.maxLinks {
+	// Only check maxLinks for truly new entries (not replacements)
+	if !existed && d.maxLinks > 0 && d.totalLinks+1 > d.maxLinks {
```
gammazero: Maybe just move this into  the `if err != nil {` block instead of using the `existed` .

gammazero (approved):
Only one question about checking `Mode()` when converting to leaf node. Otherwise, looks good.

ipld/unixfs/io/directory_test.go
```diff
@@ -997,3 +998,375 @@ func TestHAMTDirectorySizeEstimationMode(t *testing.T) {
...
+
+	// Verify final state is consistent
+	links, err := dir.Links(ctx)
+	require.NoError(t, err)
+	assert.Equal(t, 25, len(links), "should have 5 initial + 20 concurrent entries")
+}
+
+// TestHAMTDirectoryModeAndMtimeAfterReload verifies that mode and mtime metadata
+// survives HAMT conversion and reload from DAG.
+func TestHAMTDirectoryModeAndMtimeAfterReload(t *testing.T) {
```
gammazero, line 1345: Thank you. This test makes me feel much better.  It could have been easily overlooked.

gammazero (approved):

ipld/unixfs/mod/dagmodifier.go
```diff
@@ -312,6 +323,53 @@ func (dm *DagModifier) ensureSafeProtoNodeHash(node *mdag.ProtoNode) {
...
+	}
+
+	// Parse UnixFS metadata to check for ModTime or other metadata
+	fsn, err := ft.FSNodeFromBytes(protoNode.Data())
+	if err != nil {
+		return nil, false // Not valid UnixFS, keep as is
+	}
+
+	// If there's metadata (like ModTime), keep as ProtoNode
+	if !fsn.ModTime().IsZero() {
```
gammazero: Do we need to check fsn.Mode() also?


```suggestion
	if !fsn.ModTime().IsZero() || fsn.Mode() != 0 {
```

lidel:
Addressed review feedback:
- added `fsn.Mode()` check in `maybeCollapseToRawLeaf` per @gammazero's suggestion
- added `doc.go` to `ipld/unixfs/mod` documenting MFS semantics and metadata handling (just to make it clear its intended, and not accidental design)
- added table-driven tests for Mode/ModTime metadata preservation just to be extra safe
- clarified that Mode/ModTime are optional (most use cases don't set them)

All CI checks passing. If no concerns I will merge tomorrow to unblock Kubo 0.40 RC1 (https://github.com/ipfs/go-ipfs-cmds/pull/315, https://github.com/ipfs/kubo/pull/11148)

gammazero (approved):
🚀

lidel:
Thanks for reviews!

Probably as good as it can get, if we missed anything, we should catch it during RC1. 

I'm merging and will continue in  https://github.com/ipfs/go-ipfs-cmds/pull/315 and then https://github.com/ipfs/kubo/pull/11148

PR #1101, opened 2026-02-06
fix: raise block size limits from 1MiB to 2MiB

lidel:
align chunker and importer block size limits with the bitswap spec (https://specs.ipfs.tech/bitswap-protocol/#block-sizes) which mandates 2MiB as the maximum block size.

the previous 1MiB limit broke things like Kubo's `dag import` of 1MiB-chunked non-raw-leaf data where protobuf wrapping pushes blocks slightly over 1MiB (see https://github.com/ipfs/kubo/pull/8968#issuecomment-3860668059)

max chunker size is set to 2MiB - 256 bytes to leave room for protobuf framing overhead when chunks are wrapped in non-raw leaves. IPIP-499 profiles use lower chunk sizes (256KiB and 1MiB) and are not affected.

- Used in https://github.com/ipfs/kubo/pull/11185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+- 🛠 `chunker`, `ipld/unixfs/importer/helpers`: block size limits raised from 1MiB to 2MiB to match the [bitswap spec](https://specs.ipfs.tech/bitswap-protocol/#block-sizes). Max chunker size is `2MiB - 256 bytes` to leave room for protobuf framing when `--raw-leaves=false`. IPIP-499 profiles use lower chunk sizes (256KiB and 1MiB) and are not affected.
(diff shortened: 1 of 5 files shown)


lidel (commented):
Merging as the new 2MiB block size limit matches the [bitswap spec](https://specs.ipfs.tech/bitswap-protocol/#block-sizes) and is well above any chunk size used by IPIP-499 profiles (`unixfs-v1-2025` uses 1MiB, `unixfs-v0-2015` uses 256KiB). Existing users are not affected.

If someone creates blocks between 2MiB and ~4MiB using `--allow-big-block`, they will still transfer fine over bitswap. The actual hard limit on the wire is `network.MessageSizeMax` (4MiB) minus a few bytes of protobuf envelope. We have tests confirming this in ipfs/kubo#11185, including regression tests at the exact byte boundary where libp2p starts rejecting messages.

The 256-byte overhead budget (`ChunkSizeLimit = BlockSizeLimit - 256`) is  a precaution that accounts for protobuf/UnixFS framing when `--raw-leaves=false`, so the max `--chunker` value produces blocks that stay within 2MiB after wrapping.

PR #1103, opened 2026-02-07
ensure http response body is closed

gammazero:
When appropriate, consume all body data before close to allow connection reuse.


bitswap/network/httpnet/httpnet.go
```diff
@@ -624,6 +624,7 @@ func (ht *Network) connectToURL(ctx context.Context, p peer.ID, u network.Parsed
 	// FIXME: Storacha returns 410 for our probe.
 	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNoContent || resp.StatusCode == http.StatusGone {
 		log.Debugf("connect/ping request to %s %s succeeded: %d", p, req.URL, resp.StatusCode)
+		defer io.Copy(io.Discard, resp.Body) // read all body data so that connection can be reused
```
guillaumemichel: Why `defer`?

gammazero: Oh yeah, thanks.

PR #1116, opened 2026-03-06
feat: custom chunker registry

alanshaw:
Allows custom chunkers to be registered and accessed by their string name + optional params as per the default chunkers.

### DIFF
--- a/chunker/registry.go
+++ b/chunker/registry.go
@@ -37,12 +37,12 @@ func init() {
 //	    })
 //	}
 //
-// Register panics if name is empty, contains a dash, ctor is nil, or a
+// Register panics if name is empty, contains a dash, fn is nil, or a
 // chunker with the same name is already registered. This follows the
 // convention established by [database/sql.Register].
 //
 // Register is safe for concurrent use.
-func Register(name string, ctor SplitterFunc) {
+func Register(name string, fn SplitterFunc) {
 	splittersMu.Lock()
 	defer splittersMu.Unlock()
 	if name == "" {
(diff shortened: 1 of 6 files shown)


lidel (approved):
Thanks for this, @alanshaw ❤️ 

I made a few cosmetic/docs tweaks:

- added `sync.RWMutex` to protect the registry from concurrent access
- added input validation in `Register` (empty name, dash in name, nil fn, duplicates)
- routed built-in chunkers through `Register` in `init()` so they get the same validation
- renamed `CtorFunc` to `SplitterFunc` but its mostly idiomatic clarity / estetic choice (ctor is C++/java i think, in golang it feels off)
- minor godoc and code cleanup

If this looks good to you (👍️?), we can merge it next week.

ps. Are you working on custom chunkers? :)

chunker/registry.go
```diff
@@ -0,0 +1,62 @@
...
+		panic("chunk: Register name is empty")
+	}
+	if strings.Contains(name, "-") {
+		panic("chunk: Register name must not contain a dash: " + name)
+	}
+	if fn == nil {
+		panic("chunk: Register fn is nil")
+	}
+	if _, dup := splitters[name]; dup {
+		panic("chunk: Register called twice for chunker " + name)
```
alanshaw, line 59: Hmm, can we allow replacing a chunker? Maybe I have a faster implementation...

alanshaw:
> * added input validation in `Register` (empty name, dash in name, nil fn, duplicates)

Fine but I would like to replace an existing chunker 🙏 

> * routed built-in chunkers through `Register` in `init()` so they get the same validation

Yeah I partly didn't do that because they shouldn't need validation - they are the defaults and should be present and correct. Not opposed to it though.

> * renamed `CtorFunc` to `SplitterFunc` but its mostly idiomatic clarity / estetic choice (ctor is C++/java i think, in golang it feels off)

Yep no problem.

alanshaw:
> > * added input validation in `Register` (empty name, dash in name, nil fn, duplicates)
> 
> Fine but I would like to replace an existing chunker 🙏

TBH it's not a deal breaker if you disagree - feel free to merge.

lidel:
Cool, let's keep as-is.
It feels safer if its impossible to override defaults. 

Removes ambiguity -- easier to triage bugs and debug issues when we can clearly see a custom chunker was passed.

chunker/registry.go
lidel: Use custom name then, to avoid bugs in new code beign reported as bugs in default chunker in the same name. 

And if you have a faster implementation, why not make upstream library better?  :-)

PR #1117, opened 2026-03-09
fix(bitswap): ignore identity CIDs instead of killing connection

lidel:
third-party implementations may not be aware that identity CIDs should not appear in Bitswap wantlists, and their users should not be punished with dropped connections for naive CID handling.

unify behavior with oversized CIDs to reduce connection churn: silently ignore and continue processing the rest of the message.

- engine.go: replace error return with debug log + continue
- engine_test.go: flip assertions to verify no disconnect, check identity CID is silently dropped from wantlist
- (we also improved test coverage for oversized cids to follow the similar pattern)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


guillaumemichel (approved):
Looks sensible

PR #1122, opened 2026-03-19
docs: add missing godoc 

lidel:
Docs-only, this PR 
- adds missing top level package descriptions for packages without descriptions on pkg.go.dev  (https://pkg.go.dev/github.com/ipfs/boxo)
- adds links to relevant docs and specs so LLMs hallucinate less

### DIFF
--- a/bitswap/network/bsnet/ipfs_impl_test.go
+++ b/bitswap/network/bsnet/ipfs_impl_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/multiformats/go-multistream"
 )
 
-// Receiver is an interface for receiving messages from the GraphSyncNetwork.
+// Receiver is an interface for receiving messages from the BitSwap network.
 type receiver struct {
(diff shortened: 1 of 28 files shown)


guillaumemichel (approved):
nit: sometimes Bitswap is spelled `BitSwap`. This is certainly due to code comments mentioning `BitSwap`, so it should be fine to keep as is.

bitswap/network/bsnet/ipfs_impl_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/multiformats/go-multistream"
 )
 
-// Receiver is an interface for receiving messages from the GraphSyncNetwork.
+// Receiver is an interface for receiving messages from the BitSwap network.
```
gammazero:
```suggestion
// receiver implements the Receiver interface for receiving messages from the
// Bitswap network.
```

gammazero (approved):

PR #1123, opened 2026-03-20
fix(bitswap/httpnet): infer default port for portless HTTP multiaddrs

lidel:
HTTP providers advertising `/dns/host/https` without a `/tcp/443` component were silently rejected because `ExtractHTTPAddress` required an explicit port. (cc @aschmahmann)

- infer port 443 for https and 80 for http when tcp/udp is absent
- produce canonical URLs without default port to avoid cache sharding, reverse-proxy mismatches, and redirect failures
- use `url.Hostname()` instead of `net.SplitHostPort` in `httpnet` `Connect` and `msg_sender` so portless `URL.Host` values don't error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The following emojis are used to highlight certain changes:
 ### Fixed
 
 - `bitswap/server`: incoming identity CIDs in wantlist messages are now silently ignored instead of killing the connection to the remote peer. Some IPFS implementations naively send identity CIDs, and disconnecting them for it caused unnecessary churn. [#1117](https://github.com/ipfs/boxo/pull/1117)
+- `bitswap/network`: `ExtractHTTPAddress` now infers default ports for portless HTTP multiaddrs (e.g. `/dns/host/https` without `/tcp/443`). [#1123](https://github.com/ipfs/boxo/pull/1123)
 
 ### Security
 
--- a/bitswap/network/http_multiaddr.go
+++ b/bitswap/network/http_multiaddr.go
@@ -57,12 +57,35 @@ func ExtractHTTPAddress(ma multiaddr.Multiaddr) (ParsedURL, error) {
 		}
 	}
 
+	// Default well-known ports when schema is set but port is missing.
+	// This handles shorthand multiaddrs like /dns/example.com/https
+	// where the /tcp/443 component is omitted.
+	if port == "" {
+		switch schema {
+		case "https":
+			port = "443"
+		case "http":
+			port = "80"
+		}
+	}
+
 	if host == "" || port == "" || schema == "" {
 		return ParsedURL{}, fmt.Errorf("multiaddress is missing required components (host/port/schema)")
 	}
 
-	// Construct the URL object
-	address := fmt.Sprintf("%s://%s:%s", schema, host, port)
+	// Construct the URL object.
+	// Omit the port when it matches the schema default to produce
+	// canonical URLs. This ensures the HTTP client sends a clean Host
+	// header (e.g. "example.com" instead of "example.com:443") and
+	// avoids sharding HTTP caches on providers that key on Host,
+	// as well as issues with reverse proxies or redirects that fail
+	// to match when an explicit default port is present.
+	var address string
+	if (schema == "https" && port == "443") || (schema == "http" && port == "80") {
+		address = fmt.Sprintf("%s://%s", schema, host)
+	} else {
+		address = fmt.Sprintf("%s://%s:%s", schema, host, port)
+	}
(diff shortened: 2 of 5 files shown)


aschmahmann (approved):
LGTM. Two comments:
1. Might want to add tests for mismatched default ports + schemes (e.g. http + 443, and https + 80)
2. Either now or in a follow-up we might want to extract this multiaddr to (HTTP) URL function from only living in the bitswap package to somewhere more reusable. Given the lack of specs and tests for HTTP multiaddrs in general at least having a referenceable Go function might help. We could separately ping libp2p folks about if they have any interest in upstreaming.

Thanks for the quick fix 🙏

acejam:
@lidel I find it interesting how hardcoding a port to a well-known protocol was not OK a year ago, but the inverse is now apparently OK: https://github.com/ipfs/go-ipfs-cmds/pull/276#issuecomment-2583881547

lidel:
@acejam I can see why you glued both together, but these are different situations (interval vs external).

### What ipfs/go-ipfs-cmds PRs did was adding a new internal Kubo RPC feature

ipfs/go-ipfs-cmds#276 added `strings.Contains(address, "443")` to guess the scheme. That heuristic is fragile: it false-positives on port `4431`, a hostname like `server443.example.com`, etc. The problem was real, but the fix was wrong. 

ipfs/go-ipfs-cmds#277 solved it in more robust way: respect `https://` when explicitly passed, default to `http://` otherwise. On the Kubo side (ipfs/kubo#10659), `--api` multiaddrs ending with `/https` or `/tls/http` now pass an `https://` address to `NewClient`. The port is always explicit because the user configures it (e.g., `/dns/host/tcp/5001/tls/http`). 

:point_right: This is an internal Kubo RPC API interface where we control both client and server, and both are released together.

### What this boxo PR does is vendor-agnostic public client fix
 
:point_right:  This PR is not about the Kubo RPC API, but public, vendor-agnostic interfaces.

It is about "Bitswap HTTP transport" abstraction connecting to third-party [Trustless Gateways](https://specs.ipfs.tech/http-gateways/trustless-gateway/), using provider multiaddrs discovered via third-party [HTTP Routing V1](https://specs.ipfs.tech/routing/http-routing-v1/) endpoints. We control neither the providers nor the routing infrastructure. Both can be (and are) run by parties outside the IPFS project.

In practice, we've seen providers already advertise multiaddrs like `/dns/example.com/https` without an explicit `/tcp/443` component:

```json
{
  "Providers": [
    {
      "Addrs": [
        "/dns/warp.lotus.dedyn.io/tcp/443/https"
      ],
      "ID": "12D3KooWAj643EFcWRrj7rEqLRFq98DbeJbDgZ6oynWHcUU77yoZ",
      "Protocols": [
        "transport-ipfs-gateway-http"
      ],
      "Schema": "peer",
      "transport-ipfs-gateway-http": "oBIA"
    },
    {
      "Addrs": [
        "/dns/calib2.ezpdpz.net/https"
      ],
      "ID": "12D3KooWEQovpB7KFS5vE7agxzzxAmXZbWRysFqLp396ojUxgWeA",
      "Protocols": [
        "transport-ipfs-gateway-http"
      ],
      "Schema": "peer",
      "transport-ipfs-gateway-http": "oBIA"
    }
  ]
}
```

These are valid multiaddrs, JS (Helia, https://inbrowser.link) knows how to retrieve from them, but `boxo/bitswap` did not. Before this fix, they were silently rejected, losing connectivity to legitimate providers for no good reason.

Inferring port 80 for `http` and 443 for `https` when no port is present is not the same as `strings.Contains(address, "443")`. It is applying standard RFC defaults that every HTTP library uses. The PR also produces canonical URLs (omitting the default port) to avoid `Host` header mismatches with reverse proxies and HTTP cache sharding, which are real issues we have seen in the wild with third-party providers.

### The Robustness Principle

The [IPFS Robustness Principle](https://specs.ipfs.tech/architecture/principles/#robustness) says: *"Be strict about the outcomes, be tolerant about the methods."*

Both PRs follow this, applied to their respective contexts:

- Kubo RPC client (internal, we control both sides): we can be strict about methods too. Requiring an explicit scheme is cleaner. The port is always known because the user configures it.
- Bitswap HTTP transport (public network, third-party providers and routers): we must be tolerant about methods because we do not control what multiaddrs third parties advertise. The outcome remains strict: content is verified via CIDs regardless of transport details. Silently dropping valid providers because their  multiaddr omits a default port is a bug, not strictness.

### That said

If you feel that the internal Kubo RPC client should also infer default ports from the scheme, feel free to open an issue in [ipfs/kubo](https://github.com/ipfs/kubo/issues).

PR #1124, opened 2026-03-20
feat(dag/walker): opt-in BloomTracker to avoid duplicated walks

lidel:
> [!NOTE]
> - kubo PR: https://github.com/ipfs/kubo/pull/11245

## Summary

New `dag/walker` package for memory-efficient DAG traversal with bloom filter deduplication, plus new pinned-provider strategies that use it to avoid re-announcing duplicate blocks across pins.

### `dag/walker` (new package)

- **`VisitedTracker` interface** with two implementations:
  - `BloomTracker` -- auto-scaling bloom filter chain (~4 bytes/CID vs ~75 for a map), with uncorrelated false positives across nodes (unique random SipHash keys per instance)
  - `MapTracker` -- exact dedup for tests and small datasets
- **`WalkDAG`** -- iterative DFS traversal with integrated dedup, codec-agnostic link extraction (dag-pb, dag-cbor, raw, and any registered codec). ~2x faster than the legacy `go-ipld-prime` selector-based walker.
- **`WalkEntityRoots`** -- entity-aware traversal that emits only file/directory/HAMT shard roots instead of every block, skipping internal file chunks.

### `pinner`

- **`NewUniquePinnedProvider`** -- emits all blocks reachable from pins with cross-pin bloom dedup (recursive DAGs first, then direct pins).
- **`NewPinnedEntityRootsProvider`** -- same but emits only entity roots via `WalkEntityRoots`.
- Both log and skip corrupted pin entries instead of aborting the entire provide cycle.

### `provider`

- `NewPrioritizedProvider` now continues to the next stream when one fails instead of stopping all streams.
- `NewConcatProvider` added for pre-deduplicated streams that don't need the `cidutil.StreamingSet` overhead.

### Other

- Identity CIDs are filtered out during walks (they are inlined data, not real blocks).
- Siblings are visited in left-to-right link order for deterministic traversal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
-- `dag/walker`: new package with `VisitedTracker` interface for memory-efficient DAG traversal deduplication. `BloomTracker` uses a scalable bloom filter chain (~4 bytes/CID vs ~75 bytes for a map), enabling dedup on repositories with tens of millions of CIDs that were previously impractical due to memory constraints. `MapTracker` provides an exact alternative for tests and small datasets. [#1124](https://github.com/ipfs/boxo/pull/1124)
+- `dag/walker`: new package for memory-efficient DAG traversal with deduplication. `VisitedTracker` interface with `BloomTracker` (scalable bloom filter chain, ~4 bytes/CID vs ~75 bytes for a map) and `MapTracker` (exact, for tests). `WalkDAG` provides iterative DFS traversal with integrated dedup, supporting dag-pb, dag-cbor, raw, and other registered codecs. ~2x faster than the legacy go-ipld-prime selector-based traversal. [#1124](https://github.com/ipfs/boxo/pull/1124)
 - `routing/http/client`: `WithProviderInfoFunc` option resolves provider addresses at provide-time instead of client construction time. This only impacts legacy HTTP-only custom routing setups that depend on [IPIP-526](https://github.com/ipfs/specs/pull/526) and were sending unresolved `0.0.0.0` addresses in provider records instead of actual interface addresses. [#1115](https://github.com/ipfs/boxo/pull/1115)
 - `chunker`: added `Register` function to allow custom chunkers to be registered for use with `FromString`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
-- `dag/walker`: new package for memory-efficient DAG traversal with deduplication. `VisitedTracker` interface with `BloomTracker` (scalable bloom filter chain, ~4 bytes/CID vs ~75 bytes for a map) and `MapTracker` (exact, for tests). `WalkDAG` provides iterative DFS traversal with integrated dedup, supporting dag-pb, dag-cbor, raw, and other registered codecs. ~2x faster than the legacy go-ipld-prime selector-based traversal. [#1124](https://github.com/ipfs/boxo/pull/1124)
+- `dag/walker`: new package for memory-efficient DAG traversal with deduplication. `VisitedTracker` interface with `BloomTracker` (scalable bloom filter chain, ~4 bytes/CID vs ~75 bytes for a map) and `MapTracker` (exact, for tests). `WalkDAG` provides iterative DFS traversal with integrated dedup, supporting dag-pb, dag-cbor, raw, and other registered codecs. ~2x faster than the legacy go-ipld-prime selector-based traversal. `WalkEntityRoots` emits only entity roots (files, directories, HAMT shards) instead of every block, skipping internal file chunks. [#1124](https://github.com/ipfs/boxo/pull/1124)
 - `routing/http/client`: `WithProviderInfoFunc` option resolves provider addresses at provide-time instead of client construction time. This only impacts legacy HTTP-only custom routing setups that depend on [IPIP-526](https://github.com/ipfs/specs/pull/526) and were sending unresolved `0.0.0.0` addresses in provider records instead of actual interface addresses. [#1115](https://github.com/ipfs/boxo/pull/1115)
 - `chunker`: added `Register` function to allow custom chunkers to be registered for use with `FromString`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
+- `dag/walker`: new package with `VisitedTracker` interface for memory-efficient DAG traversal deduplication. `BloomTracker` uses a scalable bloom filter chain (~4 bytes/CID vs ~75 bytes for a map), enabling dedup on repositories with tens of millions of CIDs that were previously impractical due to memory constraints. `MapTracker` provides an exact alternative for tests and small datasets. [#1124](https://github.com/ipfs/boxo/pull/1124)
 - `routing/http/client`: `WithProviderInfoFunc` option resolves provider addresses at provide-time instead of client construction time. This only impacts legacy HTTP-only custom routing setups that depend on [IPIP-526](https://github.com/ipfs/specs/pull/526) and were sending unresolved `0.0.0.0` addresses in provider records instead of actual interface addresses. [#1115](https://github.com/ipfs/boxo/pull/1115)
 - `chunker`: added `Register` function to allow custom chunkers to be registered for use with `FromString`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- `pinner`: `NewUniquePinnedProvider` and `NewPinnedEntityRootsProvider` now log and skip corrupted pin entries instead of aborting the entire provide cycle, allowing remaining pins to still be provided. [#1124](https://github.com/ipfs/boxo/pull/1124)
 - `bitswap/server`: incoming identity CIDs in wantlist messages are now silently ignored instead of killing the connection to the remote peer. Some IPFS implementations naively send identity CIDs, and disconnecting them for it caused unnecessary churn. [#1117](https://github.com/ipfs/boxo/pull/1117)
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The following emojis are used to highlight certain changes:
 ### Added
 
 - `dag/walker`: new package for memory-efficient DAG traversal with deduplication. `VisitedTracker` interface with `BloomTracker` (scalable bloom filter chain, ~4 bytes/CID vs ~75 bytes for a map) and `MapTracker` (exact, for tests). `WalkDAG` provides iterative DFS traversal with integrated dedup, supporting dag-pb, dag-cbor, raw, and other registered codecs. ~2x faster than the legacy go-ipld-prime selector-based traversal. `WalkEntityRoots` emits only entity roots (files, directories, HAMT shards) instead of every block, skipping internal file chunks. [#1124](https://github.com/ipfs/boxo/pull/1124)
+- `pinner`: `NewUniquePinnedProvider` and `NewPinnedEntityRootsProvider` log and skip corrupted pin entries instead of aborting the provide cycle, allowing remaining pins to still be provided. [#1124](https://github.com/ipfs/boxo/pull/1124)
 - `routing/http/client`: `WithProviderInfoFunc` option resolves provider addresses at provide-time instead of client construction time. This only impacts legacy HTTP-only custom routing setups that depend on [IPIP-526](https://github.com/ipfs/specs/pull/526) and were sending unresolved `0.0.0.0` addresses in provider records instead of actual interface addresses. [#1115](https://github.com/ipfs/boxo/pull/1115)
 - `chunker`: added `Register` function to allow custom chunkers to be registered for use with `FromString`.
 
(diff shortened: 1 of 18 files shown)


dag/walker/entity.go
```diff
@@ -0,0 +1,192 @@
...
+			// reverse so first link is popped next (left-to-right
+			// sibling order, matching WalkDAG and legacy BlockAll)
+			slices.Reverse(children)
+			stack = append(stack, children...)
+		}
+
+		// skip identity CIDs: content is inline, no need to provide.
+		// we still descend (above) so an inlined dag-pb directory's
+		// normal children get provided.
+		if c.Prefix().MhType == mh.IDENTITY {
```
gammazero: Should this be done before any other work is done?

dag/walker/walker.go
```diff
@@ -0,0 +1,233 @@
...
+		// step 4: push children in reverse order so the first link
+		// is on top of the stack and gets popped next. This gives
+		// left-to-right sibling visit order, matching the legacy
+		// BlockAll selector traversal and the conventional DFS order
+		// from IPIP-0412 (depth-first, pre-order, left-to-right).
+		slices.Reverse(children)
+		stack = append(stack, children...)
+
+		// skip identity CIDs: content is inline, no need to provide
+		if c.Prefix().MhType == mh.IDENTITY {
```
gammazero, line 150: Do this before other work? Prevent polluting tracker with identity CIDs?

pinning/pinner/dspinner/uniquepinprovider.go
```diff
@@ -0,0 +1,147 @@
...
+				}
+				// skip identity CIDs: content is inline, no need to provide
+				if sc.Pin.Key.Prefix().MhType == mh.IDENTITY {
+					continue
+				}
+				if tracker.Visit(sc.Pin.Key) {
+					if !emit(sc.Pin.Key) {
+						return
+					}
+				}
```
gammazero: Matches above logic flow.
```suggestion
				// Skip if not first visit.
				if !tracker.Visit(sc.Pin.Key) {
					continue
				}
				// Emit pin key on first visit.
				if !emit(sc.Pin.Key) {
					return
				}
```

Do we want to track how many pins are skipped and maybe log the number? Maybe even log that there is a x% chance skipping a pin that should be provided?

dag/walker/visited.go
```diff
@@ -0,0 +1,324 @@
...
+// Implementations use c.Hash() (multihash bytes) as the key, so CIDv0
+// and CIDv1 of the same content are treated as the same entry.
+//
+// NOT safe for concurrent use. The provide pipeline runs on a single
+// goroutine per reprovide cycle. Adding parallelism requires switching
+// to thread-safe variants (bbloom AddTS/HasTS) or external
+// synchronization.
+type VisitedTracker interface {
+	// Visit marks a CID as visited. Returns true if it was NOT
+	// previously visited (first visit).
```
gammazero, line 100:
```suggestion
	// previously visited (first visit). Returns false if high
	// probability that CID was previously visited.
```

dag/walker/visited.go
```diff
@@ -0,0 +1,324 @@
...
+	Visit(c cid.Cid) bool
+	// Has returns true if the CID was previously visited.
```
gammazero, line 102:
```suggestion
	// Has returns true if there is a high probability that the CID
	// was previously visited. Returns false if NOT visited.
```

gammazero (approved):
Made a few suggestions but nothing blocking.

dag/walker/entity.go
lidel: Good question: the current order is intentional (but yes, maybe overly paranoid). 

Identity CIDs can contain inlined dag-pb directories whose children are normal
(non-identity) CIDs that need to be provided. Moving the identity check before fetch/descend would silently drop those children. 

I know this is low level unlikely edge case, so added explicit regression tests in `dag/walker/identity_test.go`:
- `TestWalkDAG_IdentityCID/"identity dag-pb directory with normal raw child"`
- `TestWalkEntityRoots_IdentityCID/"identity directory with normal children"`

dag/walker/walker.go
lidel: No, same reason as in https://github.com/ipfs/boxo/pull/1124#discussion_r3006736005

dag/walker/visited.go
lidel: Added a note on the `VisitedTracker` interface doc in 0fc1a0b noting that implementations may be probabilistic and must keep the FP rate negligible or allow callers to adjust it. 

Kept individual method docs describing the exact contract since the interface has both exact (`MapTracker`) and probabilistic (`BloomTracker`) implementations.

dag/walker/visited.go
lidel: https://github.com/ipfs/boxo/pull/1124#discussion_r3006736005

pinning/pinner/dspinner/uniquepinprovider.go
lidel: Adopted the skip-early style in 0fc1a0b.

As for trasking skips, good idea! Added logging in 577fa3f and c8920a2. On bloom creation:

```
    bloom tracker created  capacity=2000000 fpRate="1 in 4750000 (~0.000021%)"
```

After each provide/reprovide cycle (wired in ipfs/kubo@8285ec9):

```
    unique reprovide cycle finished  providedCIDs=6 skippedBranches=1
```

Tests assert exact counts for both paths.

gammazero (approved):
LGTM

dag/walker/visited.go
```diff
@@ -0,0 +1,353 @@
...
+		chain:       []*bbloom.Bloom{b},
+		lastCap:     uint64(expectedItems),
+		bitsPerElem: bpe,
+		hashLocs:    hlocs,
+	}, nil
+}
+
+func (bt *BloomTracker) Has(c cid.Cid) bool {
+	key := []byte(c.Hash())
+	for _, b := range bt.chain {
```
guillaumemichel, line 263: Maybe document the tradeoff of the order in which filters are checked.

Old -> new (current):
Works some CIDs are very often repeated. These CIDs are likely to land in first filter hence this is the one we want to check first. Nodes landing in newer filters are expected to be less frequent.

New -> old:
Duplicates are close to each other in the DAG traversal (e.g after a CID is visited, its duplicates are expected to be in the next 10K visited nodes). Hence it is better to check newest filters first since it has the highest probability to contain the duplicate.

dag/walker/visited.go
```diff
@@ -0,0 +1,353 @@
...
+	return false
+}
+
+func (bt *BloomTracker) Visit(c cid.Cid) bool {
+	key := []byte(c.Hash())
+
+	// Check earlier blooms for the CID. If any reports it as
+	// present (true positive from a prior growth epoch, or rare
+	// cross-bloom false positive), skip it.
+	earlier := bt.chain[:len(bt.chain)-1]
```
guillaumemichel, line 278: Same remark as https://github.com/ipfs/boxo/pull/1124/changes#r3020485430 concerning bloom filter iteration order

dag/walker/walker.go
```diff
@@ -0,0 +1,233 @@
...
+// root, calling emit for each visited CID. Returns when the DAG is
+// fully walked, emit returns false, or ctx is cancelled.
+//
+// The walk uses an explicit stack (not recursion) to avoid stack
+// overflow on deep DAGs. For each CID:
+//
+//  1. [VisitedTracker].Visit -- if already visited, skip entire subtree.
+//     The CID is marked visited immediately (before fetch). If fetch
+//     later fails, the CID stays in the tracker and won't be retried
+//     this cycle -- caught in the next cycle (22h). This avoids a
```
guillaumemichel: `22h` is a [Kubo DHT specific parameter](https://github.com/ipfs/kubo/blob/e0c066179dc3e3538ff3bd1ddacf7b25241cb385/config/provide.go#L16), not sure we want to hardcode it in the comment here

With reprovide sweep it corresponds to the keystore GC interval: calling `KeyChanFunc` to replace the set of keys being periodically reprovided.

With legacy reprovide it corresponds to the reprovide interval (also calling `KeyChanFunc`)

dag/walker/walker.go
```diff
@@ -0,0 +1,233 @@
...
+
+		// step 4: push children in reverse order so the first link
+		// is on top of the stack and gets popped next. This gives
+		// left-to-right sibling visit order, matching the legacy
+		// BlockAll selector traversal and the conventional DFS order
+		// from IPIP-0412 (depth-first, pre-order, left-to-right).
+		slices.Reverse(children)
+		stack = append(stack, children...)
+
+		// skip identity CIDs: content is inline, no need to provide
```
guillaumemichel: If the dag walker has other purpose than providing, I would suggest leaving this filtering to the provide system.

guillaumemichel: Or instead of blocking `IDENTITY` allow caller to pass a blocklist as option, so that provide systems can block `IDENTITY`?

dag/walker/entity.go
```diff
@@ -0,0 +1,192 @@
...
+			continue
+		}
+
+		if !emit(c) {
+			return nil
+		}
+	}
+
+	return nil
+}
```
guillaumemichel, line 145: A lot of common code with `WalkDAG()`. Would it make sense to consolidate them as possible?

dag/walker/walker.go
```diff
@@ -0,0 +1,233 @@
...
+) error {
+	cfg := &walkConfig{}
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	stack := []cid.Cid{root}
+
+	for len(stack) > 0 {
+		if ctx.Err() != nil {
```
guillaumemichel, line 107: nit: since checking `ctx.Err()` implies acquiring a mutex, we could check it only every 1k operations or similar?

pinning/pinner/dspinner/uniquepinprovider.go
```diff
@@ -0,0 +1,155 @@
...
+
+			// 1. Walk recursive pin DAGs for entity roots.
+			// A corrupted pin entry is logged and skipped so it does
+			// not prevent remaining pins from being provided.
+			for sc := range pinning.RecursiveKeys(ctx, false) {
+				if sc.Err != nil {
+					log.Errorf("entity provide recursive pins: %s", sc.Err)
+					continue
+				}
+				if err := walker.WalkEntityRoots(ctx, sc.Pin.Key, fetch, emit, walker.WithVisitedTracker(tracker)); err != nil {
```
guillaumemichel: This seems to be the only line where `NewPinnedEntityRootsProvider()` differs from `NewUniquePinnedProvider()`. Consolidating the functions body would increase clarity.

CHANGELOG.md
```diff
@@ -32,6 +34,7 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- `pinner`: `NewUniquePinnedProvider` and `NewPinnedEntityRootsProvider` now log and skip corrupted pin entries instead of aborting the entire provide cycle, allowing remaining pins to still be provided. [#1124](https://github.com/ipfs/boxo/pull/1124)
```
guillaumemichel: Should this be in the `Added` section?

guillaumemichel (approved):
Neat implementation! A few comments inline, but nothing major

dag/walker/walker.go
lidel: Good instinct, but i recently learned that  `ctx.Err()` is an [atomic load since Go 1.20](https://github.com/golang/go/blob/go1.26.1/src/context/context.go#L463-L471), not a mutex:

```go
  func (c *cancelCtx) Err() error {
      if err := c.err.Load(); err != nil { ...
```

The loop body does blockstore I/O per iteration, so the atomic read is negligible. Batching the check every Nth iteration would also delay cancellation response by up to N block fetches.

So.. better to keep as-is?

dag/walker/walker.go
lidel: Good questions. I'd keep as-is, below are my thoughts why.

## On Tracker

The tracker needs to be inside the walk loop for subtree pruning to work: when `Visit()` returns false, the walker skips the fetch and the entire subtree. If dedup was external (wrapping the emitted CIDs), the walker would still traverse every child of every node, losing the main performance benefit (`O(unique_blocks)` back to `O(pins * total_blocks)` I/O).

I'd keep as-is. The option is opt-in via `WithVisitedTracker`, so the walker could be "general-purpose" when no tracker is set.

## On IDENTITY CIDs

I think TLDR is that `IDENTITY` CIDs are a can of worms, and I believe we should keep this as-is, skip them early, don't leak, and NOT make it configurable.

Rationale:

In my mind skipping identity CIDs is a correctness thing, rather than a policy choice: the data is inline in the CID itself, so no block exists to fetch or provide. Any node can derive the content from the CID without ANY the network or disk I/O.

Making it configurable would add API surface for a case where the answer is always the same, and would imply the skip is optional, which could lead to bugs when callers forget to set it. 

I may be over-correcting, but truth to be told, we did see that over the years, and these bugs are hard to debug. The safest approach seems to always filter them out as soon as possible, and avoid leaking them to layers where they may trigger unnecessary IO or make people  waste resources due to naive implementations.

Note: We do still descend into identity CID children (e.g. an inlined dag-pb directory node), just skip emitting the identity CID itself, which keeps "general purpose" use cases feasible.

dag/walker/visited.go
lidel: Added comments documenting the tradeoff in 9bc91b8b5.

dag/walker/visited.go
lidel: Covered in  9bc91b8b5.

dag/walker/walker.go
lidel: Removed in 35fffd88f

dag/walker/entity.go
lidel: Extracted shared `walkLoop` in 26f804d70. Both functions now differ only in their fetch callback.

pinning/pinner/dspinner/uniquepinprovider.go
lidel: Extracted shared `newPinnedProvider` in 8b637a98b. `NewUniquePinnedProvider` and `NewPinnedEntityRootsProvider` now differ only in the walk callback passed to it.

lidel:
Thanks for feedback and reviews!
I think this is good enough to ship, lets include it in next release, we will test on collab cluster with https://github.com/ipfs/kubo/pull/11245

PR #1125, opened 2026-03-23
fix(mfs): preserve CidBuilder object in setNodeData(), Mkdir() and NewRoot()

davidebeatrici:


mfs/dir.go
```diff
@@ -392,10 +393,6 @@ func (d *Directory) MkdirWithOpts(name string, opts MkdirOpts) (*Directory, erro
 		}
 	}
 
-	// hector: no idea why this option is overridden, but it must be to
-	// keep backwards compatibility. CidBuilder from the options is
-	// manually set in `Mkdir` (ops.go) though.
-	opts.CidBuilder = d.GetCidBuilder()
```
davidebeatrici, line 398: Or rather:

```go
if opts.CidBuilder == nil {
	opts.CidBuilder = d.GetCidBuilder()
}
```

mfs/ops.go
davidebeatrici: The comment implies we can copy the entire struct object and then just set `MkParents` and `Flush` to false. Should we do that?

mfs/dir.go
davidebeatrici: Should we set `SizeEstimationMode` as well?

lidel (commented):
This needs in-depth review but dropping quick comment that this needs  a PR in kubo that uses this branch to detect any regression in downstream projects

davidebeatrici:
Thank you, I'll do that as soon as I completely fix https://github.com/ipfs/kubo/issues/11247.

I have to add a commit that fixes the file-specific issue to this pull request.

davidebeatrici:
@lidel Tests should now succeed.

mfs/ops.go
lidel: i think these two should remain flags specific to mkdir, unlike the rest which can be generic functional options. 

Refactored in 4f90a993dbce06abcabee7401eeb6bddcc7127a0

lidel:
Thank you @davidebeatrici, I believe this fixes issues you mentioned on the other issue a week ago.

I noticed more things while reviewing, pushed two additional commits on top:

- effb89fbe **fix(mfs): preserve CidBuilder and SizeEstimationMode in all propagation paths**  
  - `Directory.setNodeData` had the same CidBuilder bug as `File.setNodeData`: calling `SetMode`/`SetModTime` on a CIDv1 directory would silently revert it to CIDv0.
  - `Directory.Mkdir` was not passing `SizeEstimationMode` to child directories.
  - `ops.go Mkdir` was not falling back to the root's `CidBuilder` and `SizeEstimationMode` when the caller didn't set them.
  - Added tests that use a `SHA2_512` builder so a silent reset to the default `SHA2_256` would actually fail the test.

- 4f90a993d **refac(mfs): replace MkdirOpts with functional options**
  - Since we already do a breaking API change, this is oportunity to unify this and use functional options instead of struct.
  - Replaced `RootOption` and the config fields in `MkdirOpts` with a single `Option` type (`WithCidBuilder`, `WithChunker`, `WithMaxLinks`, etc.), same pattern the underlying `uio.DirectoryOption` already uses. `MkdirOpts` stays but only has `Mkparents` and `Flush`.
  - Unset options are inherited from the parent directory via a single `fillFrom` helper, so there is one place to update when new settings are added. The `With...` function names match the old `RootOption` ones, so the kubo side needs minimal changes.

Companion kubo PR: https://github.com/ipfs/kubo/pull/11273 (to illustrate how new API works + extra tests in ipfs/kubo@cac9959535a5bfec4b849a85a7a3af28e3b21f74 + see if there are any regressions in tests there).

lidel:
Kubo wiring and tests:
- https://github.com/ipfs/kubo/pull/11273 
  - includes tests:
    - https://github.com/ipfs/kubo/commit/cac9959535a5bfec4b849a85a7a3af28e3b21f74
    - https://github.com/ipfs/kubo/commit/9dfc14c4fb16619263917b67ef15db44c6e22843

lidel (approved):
LGTM, merging to include in Kubo 0.41 RC1 where we can do more real world tests.

PR #1127, opened 2026-03-25
fix(gateway): reject internal HAMT shard blocks

### LINKED ISSUE #1126: HAMT directory sub-shard: directory listing recurses but links return 404
## Problem

When an internal HAMT shard block is requested by CID, the gateway renders a directory listing with correct filenames but every link 404s.

Example: `bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4` is a HAMT directory. `bafybeig3pmmx75s7ngcgfrqiw7is67dxrzrrfltpjazaei4h2mhh4q3hqe` is one of its internal shard nodes.

- `/ipfs/bafybeig3pmmx75s7ngcgfrqiw7is67dxrzrrfltpjazaei4h2mhh4q3hqe/` -- renders directory listing, filenames look fine
- `/ipfs/bafybeig3pmmx75s7ngcgfrqiw7is67dxrzrrfltpjazaei4h2mhh4q3hqe/Trailing_arm` -- 404

HAMT lookup consumes `log2(fanout)` hash bits per trie level, always starting at bit 0. Correct at the root, wrong at an internal shard which expects a deeper offset. The [UnixFS spec](https://specs.ipfs.tech/unixfs/) stores no depth field, so a shard node cannot know its own level.

The listing works because `MapIterator` walks links sequentially. The lookup (`LookupByString` in [`go-unixfsnode`](https://github.com/ipfs/go-unixfsnode)) uses hash navigation and picks the wrong bucket.

### Prior art

[ipfs/service-worker-gateway#1014](https://github.com/ipfs/service-worker-gateway/pull/1014) fixes this in the JS stack by retrying with `translateHAMTPath: false` on 404. The Go stack has no equivalent toggle -- path resolution goes through the IPLD ADL layer in `go-unixfsnode` where HAMT reification is automatic.
 The server-side Go stack also has a stricter risk profile (DoS vectors matter more than in a browser service worker).

### A. Detect internal shard, render raw HAMT links

Detect via hash prefix alignment check, then render raw block links (`NNfilename` with CID-based hrefs) instead of a normal directory listing.

- **Pro:** zero broken links, honest UX
- **Con:** CID-based links need care on subdomain gateways; template and plumbing changes in `boxo/gateway`

### B. Non-recursive fallback scan in `LookupByString`

On lookup failure, scan value links in the current node by name (no sub-shard loading).

- **Pro:** fixes all callers (gateway, CLI, IPLD traversal), not just HTML
- **Con:** ~5-10% of entries in sub-shards are still not found -- partially broken is worse than clearly different; requires `go-unixfsnode` changes

### C. Recursive fallback scan in `LookupByString`

Same as B, but recurse into sub-shard nodes.

- **Pro:** finds 100% of entries, fixes all callers
- **Con:** `loadChild` triggers bitswap fetches -- a 404 on a root shard could fetch the entire HAMT tree; requires `go-unixfsnode` changes

### D. Detect internal shard, return error

Detect via hash prefix alignment, return HTTP 501 with diagnostic link (e.g., `explore.ipld.io`).

- **Pro:** simplest, no broken links, no template changes, contained in `boxo/gateway` + `boxo/ipld/unixfs/hamt`
- **Con:** user sees an error instead of any listing

### Recommendation

Unsure.

**D** is the safest starting point. Internal shard blocks have no standalone semantic meaning and detection is reliable (hash prefix alignment check with one-block fallback for all-sub-shard nodes). Implemented in #1127.

**B/C** are ruled out: B leaves links broken, C might introduce complexity / DoS risk via unbounded block fetching if not careful. **A** could be a future enhancement over D if there's demand for browsing raw HAMT structure.

Thought?

lidel:
Return HTTP 501 instead of rendering a directory listing with broken links when the requested CID is an internal HAMT shard node. This is better than returning partial broken state, and less involved if we had to write custom UI for this.

- Fixes https://github.com/ipfs/boxo/issues/1126

### Complexity

Not a fan, but there seem to be no easy way to detect we are in internal HAMT node.

Detection uses zero-I/O hash prefix alignment check on value links, with a one-block fallback for all-sub-shard nodes (large directories like Wikipedia). The fetched block is cached and reused by the subsequent EnumLinksAsync call, but still,.. not a fan.

Maybe we just.. leave this alone?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The following emojis are used to highlight certain changes:
 
 - `bitswap/server`: incoming identity CIDs in wantlist messages are now silently ignored instead of killing the connection to the remote peer. Some IPFS implementations naively send identity CIDs, and disconnecting them for it caused unnecessary churn. [#1117](https://github.com/ipfs/boxo/pull/1117)
 - `bitswap/network`: `ExtractHTTPAddress` now infers default ports for portless HTTP multiaddrs (e.g. `/dns/host/https` without `/tcp/443`). [#1123](https://github.com/ipfs/boxo/pull/1123)
+- `gateway`: internal HAMT shard blocks requested by CID now return HTTP 501 instead of rendering a directory listing with broken links. [#1127](https://github.com/ipfs/boxo/pull/1127)
 
 ### Security
 
--- a/gateway/backend_blocks.go
+++ b/gateway/backend_blocks.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ipfs/boxo/files"
 	"github.com/ipfs/boxo/ipld/merkledag"
 	ufile "github.com/ipfs/boxo/ipld/unixfs/file"
+	"github.com/ipfs/boxo/ipld/unixfs/hamt"
 	uio "github.com/ipfs/boxo/ipld/unixfs/io"
 	"github.com/ipfs/boxo/path"
 	"github.com/ipfs/boxo/path/resolver"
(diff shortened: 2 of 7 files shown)


guillaumemichel (approved):
Internal HAMT shard detection looks sound.

Why `501` and not client error such as `400` or `422` (Unprocessable Content)?

guillaumemichel:
Triage notes:
* Not worth the complexity

PR #1128, opened 2026-03-26
fix(files): remove io.Seeker from File interface

lidel:
Not all File implementations can seek. `ReaderFile` wrapping an HTTP multipart stream and `WebFile` both returned `ErrNotSupported` at runtime, which caused interface-sniffing bugs in downstream consumers like go-car (ipfs/kubo#9361).

Not sure if this PR (removal of io.Seeker from File interface) is the "lesser evil", opening this mostly to have PR with code example.

With this PR seeking is now opt-in: implementations that support it (ReaderFile with a seekable reader, Symlink, ufsFile) still have a Seek method and callers type-assert to io.Seeker when needed.

- `files`: remove `io.Seeker` from `File` interface
- `files`: remove dummy `WebFile.Seek` that always returned `ErrNotSupported`
- `gateway`: type-assert to `io.Seeker` before seeking in all handlers

## Test plan

- [x] See if CI here passes
  - all green (the kubo sharness here fails because it needs refactor from ipfs/kubo#11254
- [x] Open PR in Kubo that uses this: https://github.com/ipfs/kubo/pull/11254 and see if tests pass there
- [x] Discuss if we do this (technically a breaking change), or if we close without merging


gateway/backend_car.go
```diff
@@ -470,7 +470,11 @@ func loadTerminalEntity(ctx context.Context, c cid.Cid, blk blocks.Block, lsys *
 					return nil, fmt.Errorf("invalid car backend range: negative start bigger than the file size")
 				}
 			}
-			if _, err := f.Seek(from, io.SeekStart); err != nil {
+			s, ok := f.(io.Seeker)
+			if !ok {
+				return nil, fmt.Errorf("file does not support seeking")
+			}
+			if _, err := s.Seek(from, io.SeekStart); err != nil {
```
gammazero: Is it necessary to use `io.Seeker` here if `from` is never negative. Would it be OK to do a discarding read here?

gammazero (commented):
I think this all looks alright, but would like to any use of `io.Seeker` to `io.Reader` where ever possible.

gammazero:
Triage: OK with cleaning this up. @lidel will rebase and merge.

gateway/backend_car.go
lidel: You would think so, but we seem to have layers upon layers and sometimes it means seek happens more than once. 🙈

Tried the slice in 8d0b5351, broke gateway-conformance, reverted in 1cb4a9c1.

iiuc the reader returned from this site is handed to `serveCodecRaw`, which seeks it again via `seekToStartOfFirstRange` before `serveContent`. With Seek both seeks land at the same absolute offset (idempotent). With `blockData[from:]` the second seek lands inside the slice, so the body comes from offset `2*from`. The seek here isn't redundant: the codec path assumes the reader spans the full block.

Cleaner fix would be to drop it and let `serveCodecRaw`/`serveFile` own range positioning, but that's a separate boxo/gateway refactor, lets keep this PR a minimal refactor.

lidel:
Merging, Sharness passes in https://github.com/ipfs/kubo/pull/11254, will start passing in boxo as soon we land that PR too.

PR #1133, opened 2026-04-07
fix(mfs): serialize all FileDescriptor operations

lidel:
This is a follow-up to #1131 

The underlying DagModifier is not safe for concurrent use. FUSE mounts and Kubo RPC commands dispatch Read, Write, Seek, Truncate, Flush, and Close from separate goroutines.

This PR makes sure all remaining operations are serialized, and no races (when RPC/FUSE are executed in parallel on the same file) are possible.


lidel:
This passed fine in kubo (https://github.com/ipfs/kubo/pull/11272/commits/c73d2cf53e4273234a867ee3c89374cb3adc6741), merging (better safe than sorry).

PR #1138, opened 2026-04-10
feat(gateway): add content size limits for responses

lidel:
Add two new Config options for gateway operators to limit responses based on content size read from the UnixFS root block:

- `MaxDeserializedResponseSize`: caps **deserialized responses onl**y, trustless formats (raw, CAR) are not affected
- `MaxUnixFSDAGResponseSize`: caps all response formats including raw blocks, CAR, and TAR

Both return 410 Gone with a message directing users to run their own IPFS node for large content.

### Integration points

- gateway.go: add config fields with documentation
- handler.go: add exceedsMax* helper methods
- handler_defaults.go: check both limits using bytesSize
- handler_block.go: check DAG limit using existing block size
- handler_codec.go: check DAG limit using existing block size
- handler_car.go: conditional Head call only when limit is set
- handler_tar.go: check DAG limit using existing file.Size()
- gateway_test.go: tests for both limits across all formats
- CHANGELOG.md: document new config options

## Test

- [x] https://github.com/ipfs/rainbow/pull/364
- [x] staging: https://www.notion.so/boxo-1138-34a1def34287809a9b29d11ba22db05d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ The following emojis are used to highlight certain changes:
 
 ### Added
 
-- `gateway`: `Config.MaxDeserializedResponseSize` allows setting a maximum file/directory size for deserialized gateway responses. Content exceeding this limit returns `501 Not Implemented`, directing users to run their own IPFS node. Trustless response formats (`application/vnd.ipld.raw`, `application/vnd.ipld.car`) are not affected. The size is read from the UnixFS root block, so no extra block fetches are needed for the check. [#1138](https://github.com/ipfs/boxo/pull/1138)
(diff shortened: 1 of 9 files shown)


guillaumemichel (approved):

gammazero:
Triage: test in staging before including in release.

lidel:
Staging looks fine, but also had very small number of blocked requests: 
https://www.notion.so/boxo-1138-34a1def34287809a9b29d11ba22db05d

Since this is opt-in, and does not change current behavior when disabled, 
let's ship it and then decide if we need to enable this on infra. 
Better to have this option than not.

PR #1143, opened 2026-04-22
fix(bitswap): prevent low-pending peer starvation

### LINKED ISSUE #1141: bitswap: new peer starved by `DefaultPeerComparator` on a busy server
## Summary

On all 4 collab-cluster nodes (`rbx8`, `bhs8`, `waw1`, `eri1`), a newly-connected peer cannot retrieve blocks via bitswap even when the target block is present locally. The server receives the want, locates the block, and pushes the task, but `nextEnvelope` never selects the peer over 10+ minutes.

## Environment

- kubo `v0.40.1` (waw1 trace)
- boxo `v0.37.0`
- go-peertaskqueue `v0.8.3`
- Config on waw1: `TaskWorkerCount=8`, `MaxOutstandingBytesPerPeer=20MB`, `MaxQueuedWantlistEntriesPerPeer=1024`, `Peering.Peers` ~ 19

## Reproduction

From a machine with no prior peering to waw1:

`vole bitswap check bafybeifjzexpshzg7lz7fwpoepxfiafc5llgok46ub2x7wjdrtnot4jskq /ip4/57.128.192.234/tcp/4001/p2p/12D3KooWCrBiagtZMzpZePCr1tfBbrZTh4BRQf7JurRqNMRi8YHF`

Returns `{"Found":false,"Responded":false,"Error":null}` after 90s. A 600s variant returns the same value. The same behaviour reproduces on `rbx8`, `bhs8`, and `eri1` across 3 trials each.

## Server trace (waw1, bitswap subsystems raised to `debug`)

For vole peer `12D3KooWC1BHWGeemFNsNYmFDZXK1rFpbYUbYfupkZNYp1creU3y` during a 90-second probe, the kubo container emitted exactly 6 bitswap log lines:

```
21:01:23.618  bitswap/connevtman      PeerConnected
21:01:23.625  bitswap/bsnet           handleNewStream
21:01:23.625  bitswap/server/decision engine <- msg entryCount=1
21:01:23.625  bitswap/server/decision engine <- want-block (bafybei...jskq)
21:01:23.625  bitswap/server/decision engine: block found, isWantBlock=true
21:02:53.627  bitswap/connevtman      PeerDisconnected
```

Absent for this peer ID across the full window (279 MB of debug log):

- `bitswap engine -> msg` (envelope generated, `boxo/bitswap/server/internal/decision/engine.go:600`)
- `sent message` (`boxo/bitswap/server/server.go:356`)
- `failed to send blocks message` (`boxo/bitswap/server/server.go:337`)

The engine located the block and pushed the task into `go-peertaskqueue`, but the engine never selected this peer for an outbound envelope. `network.SendMessage` was never called.

## Scheduler state on waw1

Snapshot over 174 bitswap peers at probe time:

- 76 of 174 peers had `pending > 1`
- max `pending` = 925
- mean `pending` ~ 87.5
- peered peers connected ~ 19

## Analysis

`DefaultPeerComparator` at `go-peertaskqueue/peertracker/peertracker.go:101-130` orders peers by FOUR RULES:

1. `pending == 0` sinks to the bottom
2. more frozen sinks lower
3. on equal `activeWork`, higher `pending` wins
4. lower `activeWork` wins

A new peer carrying one want enters the heap with `activeWork=0` and `pending=1`. Any peer with `activeWork=0` and `pending>1` beats it under rule 3. With 85 `Peering.Peers` refilling wantlists, peers return to `activeWork=0` briefly after each `TasksDone` and win the tiebreak immediately. The new peer never becomes root, so `PopTasks` on root never pops its task.

The starvation is deterministic, not transient.

### What this rules out

Maybe I missed something, but so far eliminated

| Hypothesis | Evidence against |
|---|---|
| Outbound stream, dial, or identify failure | `SendMessage` never called; zero `failed to send blocks message` in 279 MB / 90 s of debug log |
| Blockstore / bloom false-negative | `engine: block found` fired; block was located |
| Transient queue imbalance | 10-minute probe returned identical `Responded:false` |

###  Mitigations tried

- `MaxOutstandingBytesPerPeer` raised 20x (1 MB default to 20 MB) on all 4 nodes: no effect. The blocker is comparator ordering, not per-peer byte budget.

## Suggested fix direction

Did not test this yet, but feels we should avoid starving peers that only want 1 CID.

Rule 3 causes deterministic starvation of low-`pending` peers whenever any high-`pending` peer is momentarily idle. Options:

1. Add a `lastServed` timestamp and break `activeWork` ties on oldest-first.
2. Cap the pending term (`min(pending, N)`), so peers above N tie on rule 3 and rule 4 selects lower `activeWork`.
3. Round-robin among non-frozen peers with `activeWork == 0` rather than ranking by `pending`.

Each preserves the goal of rule 3 (reward heavy users that stay within byte budgets) while bounding worst-case wait for a new peer.

`go-peertaskqueue` has `DefaultPeerComparator` which is what Kubo uses because we dont pass custom `bitswap.WithTaskComparator`

we probably should wire up something smarter than `DefaultPeerComparator` (boxo has info about bitswap activity, we could leverage it to not starve peers)

## Artifacts

- [collab-cluster-bitswap-artifacts.zip](https://github.com/user-attachments/files/26947625/collab-cluster-bitswap-artifacts.zip)
  - Full 279 MB debug capture from waw1 covering the probe window
  - `vole` probe outputs for all 4 nodes (3 trials each, 90s and 600s variants)
  - all the other context

lidel:
## Problem

The default decision-engine peer comparator ordered by pending-count descending, so a peer with a single want waited indefinitely behind peers with large wantlists.

This is one of contributing reasons collab cluster stopped responding after bitswap knobs got reduced earlier this month (#1141).

## This PR 

Aims to fix starvation: replace peer comparator with a fair comparator backed by a per-peer scheduler:

* promote never-served peers and peers waiting past a 10s starvation timeout above non-starved peers
* cap pending's ordering weight at 16 so large wantlists no longer dominate the tiebreak
* final tiebreak uses a per-process salted hash of peer.id, so no peer can craft an id to permanently outrank others

This only tweaks the implicit default behavior. Custom engines built with `WithTaskComparator` keep their existing behavior.

- Closes #1141

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,13 @@ The following emojis are used to highlight certain changes:
 
 ### Changed
 
+- `bitswap/server`: the default peer comparator now schedules peers fairly. A peer that has never been served, or has waited longer than 10s, outranks non-starved peers. Pending counts cap at 16 for ordering purposes, so peers with small wantlists no longer wait behind peers with large ones. The final tiebreak uses a per-process salted hash of peer.ID, so no peer can craft an ID that permanently outranks everyone. Engines built with `WithTaskComparator` keep their existing behavior. [#1141](https://github.com/ipfs/boxo/issues/1141)
+
 ### Removed
 
 ### Fixed
+
+- `bitswap/server`: a peer with a single pending want no longer waits behind peers with large wantlists. [#1141](https://github.com/ipfs/boxo/issues/1141)
 
 ### Security
 
--- a/bitswap/server/internal/decision/chokepoint_bench_test.go
+++ b/bitswap/server/internal/decision/chokepoint_bench_test.go
@@ -1,0 +1,148 @@
+package decision
+
+import (
+	"crypto/rand"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/ipfs/go-peertaskqueue/peertracker"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+// benchSink prevents the compiler from hoisting or eliminating comparator
+// calls whose result would otherwise go unused.
+var benchSink atomic.Bool
+
+// realisticPeerID returns a 38-byte peer.ID whose bytes look like an
+// identity-multihash of an ed25519 pubkey. The short test IDs used elsewhere
+// understate hash cost at layer 5; these do not.
+func realisticPeerID(tb testing.TB) peer.ID {
+	b := make([]byte, 38)
+	if _, err := rand.Read(b); err != nil {
+		tb.Fatal(err)
+	}
+	b[0], b[1] = 0x00, 0x24
+	return peer.ID(b)
+}
+
+// BenchmarkComparatorVsUpstream measures per-call cost of the fair
+// comparator against peertracker.DefaultPeerComparator on realistic peer
+// IDs. Any added chokepoint would show up here as a large gap.
+func BenchmarkComparatorVsUpstream(b *testing.B) {
+	idA := realisticPeerID(b)
+	idB := realisticPeerID(b)
+	a := newTestTracker(string(idA), 50)
+	c := newTestTracker(string(idB), 50)
+
+	b.Run("upstream_default", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			benchSink.Store(peertracker.DefaultPeerComparator(a, c))
+		}
+	})
+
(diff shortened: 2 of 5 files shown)


gammazero (approved):
Having peers with small wantlists not have to wait behind peers with large ones might show dramatic gains depending on how common it is for a relatively large number of peers with small wantlists to be waiting for a few peers with large wantlists.

PR #1144, opened 2026-04-22
fix(bitswap/bsnet): close streams in background

### LINKED ISSUE #1142: bitswap: `taskWorker` still blocks on `stream.Close`
## Summary

Bitswap `taskWorker` goroutines in `boxo/bitswap/server` spend up to `DefaultNegotiationTimeout` (10s) each on `stream.Close()` when the remote peer is slow or silent during the multistream lazy handshake. Each stalled `Close` holds a worker until its `SetReadDeadline` expires. On a busy node with even a small fraction of unresponsive peers, the `TaskWorkerCount` pool saturates and bitswap stops serving blocks to new peers.

We observed this on collab-cluster nodes running kubo v0.41.0-rc2 (boxo v0.38.0) while investigating #1141. The earlier fix in #1083 (boxo v0.36.0) set a read deadline to bound each individual `Close`, and that deadline works as intended. The remaining problem is structural: 10s per worker is still too long under realistic churn bigger storage provider may experience.

## Environment

| Component | Version |
|---|---|
| kubo | v0.41.0-rc2 |
| boxo | v0.38.0 |
| go-libp2p | v0.48.0 |
| go-multistream | v0.6.1 |
| Observed on | `collab-cluster-rbx8.cluster.dwebops.net` |

## Call path

Every busy `taskWorker` goroutine on the affected host sat at the same frame:

```
boxo/bitswap/server.taskWorker                  (server.go:278)
  boxo/bitswap/server.sendBlocks                (server.go:335)
  boxo/bitswap/network.router.SendMessage       (router.go:104)
  boxo/bitswap/network/bsnet.impl.SendMessage   (ipfs_impl.go:401)
  go-libp2p/p2p/host/basic.streamWrapper.Close  (basic_host.go:691)
  multistream.lazyClientConn.Close              (lazyClient.go:196)
  multistream.once.Do                           (lazyClient.go:53 / 58)
  ...doReadHandshake...                          (lazyClient.go:107)
  ...stream.Read blocked in yamux or quic-go...
```

`lazyClientConn.Close()` finishes the handshake it skipped during the write path by reading a varint token. If the peer is gone or silent, that read blocks until the `SetReadDeadline` set in `bsnet.SendMessage` expires.

## How widespread is this on a running node

On rbx8 at the moment of capture (kubo v0.41.0-rc2, 12-hour uptime, `TaskWorkerCount=12`):

- 8 of 12 workers sat in this frame.
- `vole bitswap check` against this node timed out in 3 of 3 trials (30s timeout).

A snapshot on a v0.40.1 sibling node showed 5 of 12 workers in the same frame, with partial probe success.

## Reproduction

A minimal reproducer uses `net.Conn` and `go-multistream`, with no libp2p involved:

- Silent server: TCP listener that accepts and holds the connection without responding.
- Client: `NewMSSelect`, then `Write(payload)`, then optional `SetReadDeadline`, then `Close`.
- Measure elapsed time for `Close`.

Results:

- Without `SetReadDeadline`: `Close` stuck indefinitely (tested to 10s cap).
- With `SetReadDeadline(2s)`: `Close` returns in 2.000289693s. The deadline chain works.

## Quantifying the pool-saturation effect

A 12-worker pool processes 200 tasks, 10 percent of which have a silent peer, with a 10s close deadline (same as production):

| Variant | Completed | Time | Tasks/s |
|---|---|---|---|
| Current path: worker waits for `Close` | 194 / 200 | 20s (budget capped) | **9.7** |
| Decoupled: `Close` runs in a goroutine, worker returns after `Write` | 200 / 200 | 5ms | **~40,000** |

At 10 percent unresponsive peers, the synchronous close drops effective throughput by about 4000x. This matches the field observation: `engine -> msg` fires for thousands of other peers in the same window that a new peer sees zero.

## Why it makes sense to fix this in boxo

Two upstream locations could bound `Close` directly: `go-multistream` (in `lazyClientConn.Close`) and `go-libp2p` (in `streamWrapper.Close`). Either would help every consumer.

Some reasons to prefer a boxo-local fix:

- `go-multistream` is used by implementations beyond libp2p. Changing `Close` semantics, even to "bounded with a timeout", could surprise older code that depends on the blocking guarantee for correctness (the QUIC `STOP_SENDING` edge case the current code comment describes).
- The future of `go-libp2p` maintenance is uncertain, so an upstream round trip for behaviour that mainly matters to bitswap may take a while to land.
- Bitswap has a clear opinion: once the message bytes are written, the multistream ack is not needed to consider the send done. This is a bitswap-specific choice, and boxo is a natural place for it.

A boxo-local fix keeps the blast radius small, ships on boxo's release cadence, and avoids renegotiating `Close` semantics with every other consumer of `go-multistream`.

## Proposed fix

In `bsnet.SendMessage`, after `msgToStream` succeeds, move `Close` to a goroutine and return. The worker is freed as soon as the bytes are written.

```go
// Sketch; not final.
if err := bsnet.msgToStream(ctx, s, outgoing, timeout); err != nil {
    _ = s.Reset()
    return err
}
go func(s network.Stream, deadline time.Time) {
    _ = s.SetReadDeadline(deadline)
    _ = s.Close()
}(s, time.Now().Add(timeout))
return nil
```

Trade-offs:

- The caller no longer sees errors from `Close`. Current callers in boxo log them and move on, and the bitswap message itself is already written to the stream before `Close` is reached.
- Each send spawns one goroutine. Goroutines are cheap, and the in-flight total is already bounded by the deadline. In steady state, the close-goroutine pool size is about `send_rate * deadline`, which is the same work the task-worker pool holds today. The change moves that holding away from the workers.
- `streamMessageSender` (the persistent per-peer sender) uses a different code path so not impacted (but will double check in a PR)

## Related

- #1083: added `SetReadDeadline` before `Close` in `bitswap/SendMessage` (boxo v0.36.0). That fix bounds each individual hang and does exactly what it set out to do. This issue addresses the remaining worker occupancy during the bounded hang, which the earlier fix could not tackle on its own.
- multiformats/go-multistream#47: the original upstream tracker, still open, "multistream-select v1 does not handle timeouts". The proposed fix there is the deadline approach, already applied.
- #1141: the issue where this was traced. The bitswap silence to new peers reported there is the downstream symptom of this pool saturation.

## Artifacts 

Two `goroutines.stacks` files from collab cluster production (Kubo v0.40.1 and v0.41.0-rc2), the full `ipfs diag profile` bundles, and the two reproducer programs are available on the boxes.

## Next steps

I will open a PR with the proposed fix, plus a regression test that mirrors the pool-saturation benchmark.

lidel:
`SendMessage` and `handleNewStream` now close streams from a background goroutine, so `lazyClientConn.Close`'s multistream handshake read cannot hold (block/waste) bitswap task workers. 

`Close` errors are logged at Debug; a `recover()` guard keeps a panic in libp2p or multistream internals from crashing the process.

- Fixes #1142

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The following emojis are used to highlight certain changes:
 
 ### Fixed
 
+- `bitswap/network/bsnet`: `SendMessage` and `handleNewStream` now close streams in a background goroutine. Previously, `stream.Close` could hold the caller for up to `DefaultNegotiationTimeout` (10s) while `lazyClientConn.Close` waited for the remote peer to complete the multistream handshake. This saturated the bitswap `TaskWorkerCount` pool when peers were unresponsive and stopped bitswap from serving blocks to other peers. As a side effect, `SendMessage` no longer returns errors from `stream.Close`; close failures are logged at Debug. [#1142](https://github.com/ipfs/boxo/issues/1142)
+
 ### Security
 
 
--- a/bitswap/network/bsnet/bench_close_test.go
+++ b/bitswap/network/bsnet/bench_close_test.go
@@ -1,0 +1,66 @@
+package bsnet_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	tnet "github.com/libp2p/go-libp2p-testing/net"
+)
+
+// BenchmarkSendMessageHappyPath measures the per-call cost of SendMessage
+// against a live peer. With the async-close change, the caller returns as
+// soon as the bytes are written; the goroutine cost of spawning Close is the
+// main extra work on the happy path.
+func BenchmarkSendMessageHappyPath(b *testing.B) {
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	p1, err := tnet.RandIdentity()
+	if err != nil {
+		b.Fatal(err)
+	}
+	p2, err := tnet.RandIdentity()
+	if err != nil {
+		b.Fatal(err)
+	}
+	r1 := newReceiver()
+	r2 := newReceiver()
+	_, bsnet1, _, _, msg := prepareNetwork(b, ctx, p1, r1, p2, r2)
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		if err := bsnet1.SendMessage(ctx, p2.ID(), msg); err != nil {
+			b.Fatalf("SendMessage failed: %v", err)
+		}
+	}
+}
+
+// BenchmarkSendMessageHappyPathParallel measures concurrent SendMessage
+// throughput. With the old synchronous close, goroutines would serialize on
+// each Close; with async close, they pipeline freely.
+func BenchmarkSendMessageHappyPathParallel(b *testing.B) {
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	p1, err := tnet.RandIdentity()
+	if err != nil {
+		b.Fatal(err)
+	}
+	p2, err := tnet.RandIdentity()
+	if err != nil {
+		b.Fatal(err)
+	}
+	r1 := newReceiver()
+	r2 := newReceiver()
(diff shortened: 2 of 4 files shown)


gammazero (approved):
LGTM. Nice to recover any panics from close.

PR #1146, opened 2026-04-23
fix: pinner shutdown race

lidel:
Improving ux on shutdown, should fix problem introduced by #1144  during Kubo shutdown:
- https://github.com/ipfs/kubo/issues/11292

The proper fix would be to add `Close()` to pinner, but that is a bigger refactor, this should do for now.

## TODO

- [x] wire in https://github.com/ipfs/kubo/pull/11290
- [x] staging smoke-test on collab cluster


gammazero (approved):

lidel:
Smoke test in https://github.com/ipshipyard/waterworks-infra/pull/960 confirmed the issue https://github.com/ipfs/kubo/issues/11292 no more, merging.

PR #1150, opened 2026-04-24
feat(pinner): add Close with ErrClosed lifecycle

lidel:
Follow-up to #1146. The Pinner had no lifecycle, so kubo/pebble could close the backing datastore while pinner goroutines were still reading it and hit the use-after-close panic that #1146 papers over with `recover()`.

:point_right:  This PR adds `Close() error` to the `Pinner` interface and implements it in `dspinner`. 

Close fires `context.AfterFunc` on every admitted op's derived ctx, so in-flight `Pin`, `Unpin`, `Flush`, `Update`, and streaming goroutines that honor ctx bail out promptly with `context.Canceled` (scalar) or `ErrClosed` (stream) rather than draining to completion. After Close, every other method fails fast with `ErrClosed`; streaming methods deliver it as `StreamedPin.Err` on a single entry then close the channel. Close is idempotent and goroutine-safe.

This means Close returns about as fast as the slowest in-flight op takes to honor its ctx. Hosts with a hard shutdown deadline (e.g. kubo via ipfs/kubo#11329's `shutdown.CloseWithCtx`) get bounded shutdown for free, and the panic-recovery from #1146 stays as defense in depth.

**Breaking change:** downstream `Pinner` implementations must add `Close() error`. Hosts owning the datastore should call `Close` on the pinner before closing the datastore.

Kubo PR that uses this:
- https://github.com/ipfs/kubo/pull/11296


pinning/pinner/dspinner/pin_test.go
```diff
@@ -1590,3 +1591,287 @@ func TestStreamIndexDoesNotBlockWriters(t *testing.T) {
...
+
+	// Releasing the RLock lets Pin acquire Lock and finish; Close
+	// then unblocks. Pin sees its ctx already cancelled by the time
+	// it acquires the lock (because Close fired stopCancel while Pin
+	// was waiting), so Pin returns context.Canceled.
+	p.lock.RUnlock()
+
+	require.ErrorIs(t, <-pinReturned, context.Canceled)
+	require.NoError(t, <-closeReturned)
+}
```
gammazero, line 1722: Would `testing.Synctest` be good here too?

gammazero (approved):

PR #1153, opened 2026-05-06
feat(path/resolver): populate retrieval state

lidel:
Populate `retrieval.State` (phase + root/terminal CIDs) inside `path/resolver` so non-gateway callers (CLIs, custom tools) get the same diagnostics gateway backends already populate.

guarded on `if rs := retrieval.StateFromContext(ctx); rs != nil {`  so if state tracker is not in context, this is noop


retrieval/state_test.go
```diff
@@ -592,3 +594,177 @@ func TestErrorWithState(t *testing.T) {
...
+		var wg sync.WaitGroup
+		wg.Add(writers)
+		for w := 0; w < writers; w++ {
+			go func() {
+				defer wg.Done()
+				for i := 0; i < writes; i++ {
+					p, _ := test.RandPeerID()
+					rs.AddFoundProvider(p)
+				}
+			}()
```
gammazero:
```suggestion
		for range writers {
			wg.Go(func() {
				for range writes {
					p, _ := test.RandPeerID()
					rs.AddFoundProvider(p)
				}
			})
```

retrieval/state.go
```diff
@@ -135,24 +159,24 @@ func (rs *State) GetPhase() RetrievalPhase {
...
-	}
+	prev := len(*list)
 	for _, peerID := range peerIDs {
-		// Skip if we already have this peer ID in the list
+		if len(*list) >= MaxProvidersSampleSize {
+			break
+		}
 		if slices.Contains(*list, peerID) {
 			continue
 		}
```
gammazero, line 172: Instead of searching the slice each iteration, it might be more efficient to remove duplicates after the loop, with:
```go
slices.Sort(*list)
*list = slices.Compact(*list)
```

gammazero (approved):

retrieval/state.go
lidel: Thanks! Honestly at `MaxProvidersSampleSize = 3` it barely matters either way, but I ran a quick bench out of curiosity and the current code came out faster (~50 ns vs ~206 ns per call), so I'd lean on keeping it.

The other thing is that we expose this list as "providers in the order we found them" via the JSON snapshot, so sorting would shuffle that signal for anything watching progress.

PR #1154, opened 2026-05-14
fix(bitswap): duplicated "the" in error_tracker and messagequeue comments

quyentonndbs:
Two one-line typo fixes for duplicated "the" in bitswap source comments:
- `bitswap/network/httpnet/error_tracker.go` — "If n is 0, the the total count is reset to 0." → "If n is 0, the total count is reset to 0."
- `bitswap/client/internal/messagequeue/messagequeue.go` — "// send more if the the workcount is above the cutoff." → "// send more if the workcount is above the cutoff."

No code/behavior change.


gammazero:
handled with other comment changes in #1158

PR #1157, opened 2026-05-19
fix(routing/http): cap records after filtering

lidel:
## Problem

The Delegated Routing server applied the records limit before filter-addrs/filter-protocols ran, so records the filters dropped shrank /routing/v1/providers and /routing/v1/peers responses below the configured limit.

## This Fix

The server now calls the delegate FindProviders/FindPeers with a limit of 0 (unbounded), reads the result iterator lazily, and applies the cap itself after filtering. A delegate that used the limit to end its walk early will now end it on Close instead.

See it in action in:
-  https://github.com/ipfs/someguy/pull/150

### Changes

- routing/http/types/iter: add Limit, an iterator that caps another iterator at a fixed number of values
- routing/http/server: apply the records limit post-filter; pass 0 to the delegate; update tests and DelegatedRouter docs

### DIFF
--- a/routing/http/server/server.go
+++ b/routing/http/server/server.go
@@ -318,11 +318,12 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
 }
 
 func (s *server) findProvidersJSON(w http.ResponseWriter, provIter iter.ResultIter[types.Record], recordsLimit int, filterAddrs, filterProtocols []string) {
-	defer provIter.Close()
-
 	filteredIter := filters.ApplyFiltersToIter(provIter, filterAddrs, filterProtocols)
 	var limitedIter iter.ResultIter[types.Record] = iter.Limit(filteredIter, recordsLimit)
 	providers, err := iter.ReadAllResults(limitedIter)
+	// Close eagerly so the upstream router request is canceled before we
+	// spend time marshaling and writing the response.
+	limitedIter.Close()
 	if err != nil {
 		writeErr(w, "FindProviders", http.StatusInternalServerError, fmt.Errorf("delegate error: %w", err))
 		return
@@ -458,12 +459,13 @@ func (s *server) provide(w http.ResponseWriter, httpReq *http.Request) {
 }
 
 func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord], recordsLimit int, filterAddrs, filterProtocols []string) {
-	defer peersIter.Close()
-
 	filteredIter := filters.ApplyFiltersToPeerRecordIter(peersIter, filterAddrs, filterProtocols)
 	var limitedIter iter.ResultIter[*types.PeerRecord] = iter.Limit(filteredIter, recordsLimit)
 
 	peers, err := iter.ReadAllResults(limitedIter)
+	// Close eagerly so the upstream router request is canceled before we
+	// spend time marshaling and writing the response.
+	limitedIter.Close()
 	if err != nil {
 		writeErr(w, "FindPeers", http.StatusInternalServerError, fmt.Errorf("delegate error: %w", err))
 		return
--- a/routing/http/server/server.go
+++ b/routing/http/server/server.go
@@ -69,6 +69,11 @@ type FindProvidersAsyncResponse struct {
 type DelegatedRouter interface {
 	// FindProviders searches for peers who are able to provide the given [cid.Cid].
 	// Limit indicates the maximum amount of results to return; 0 means unbounded.
+	//
+	// The HTTP server in this package always calls FindProviders with a
+	// limit of 0 and caps the response itself, after filtering. It consumes
+	// the iterator lazily and Closes it once enough records are collected,
+	// so implementations should return results lazily and stop work on Close.
 	FindProviders(ctx context.Context, cid cid.Cid, limit int) (iter.ResultIter[types.Record], error)
 
 	// Deprecated: historic API from [IPIP-526], may be removed in a future version.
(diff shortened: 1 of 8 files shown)


guillaumemichel (commented):
This seems like a better approach than https://github.com/ipfs/someguy/pull/149 👍🏻 

Would it make sense to use `FindProvidersAsync()` instead of `FindProviders()`? It is the more standard way to find provider ([go-libp2p interface](https://github.com/libp2p/go-libp2p/blob/6a6b1ffd8a9ae70dd927862f7d3d558d24be8070/core/routing/routing.go#L30-L38)), and gives caller more control.

`FindProvidersAsync(ctx, cid, 0)` would allow to cancel the request (cancel context) after we are happy with the results (e.g enough providers with addresses), so that we don't solicit the content router more than strictly needed.

lidel:
Hm.. the goal here (cancel once we have enough, don't over-ask the router) is right, and it's already what this PR does. 

@guillaumemichel  lmk if i misunderstood, but the `DelegatedRouter.FindProviders` is slightly different from go-libp2p interface:

- go-libp2p `ContentDiscovery.FindProvidersAsync` returns `<-chan peer.AddrInfo`  a value channel, no error channel
- boxo's `DelegatedRouter` is a separate interface, returning `iter.ResultIter[types.Record]` (richer type than `peer.AddrInfo`, and is lazy: the server pulls record by record, stops once `iter.Limit` is satisfied, and the deferred `Close()` cancels the rest
  -  In  https://github.com/ipfs/someguy/pull/150  that chain is real: `peerChanIter.Close()` cancels the context passed to `FindProvidersAsync`. So `FindProvidersAsync` is already used, but at the layer where it fits best: the DHT boundary inside someguy's `libp2pRouter`.

Is the idea to lift `FindProvidersAsync` to `DelegatedRouter`? If so, it wouldn't add control we lack, and it has costs:

- `<-chan peer.AddrInfo` has no slot for per-item errors; `iter.Result` does.
- a channel forces a goroutine even for synchronous delegates. boxo's own HTTP-delegated client is a goroutine-free streaming decoder today; a channel interface would spawn a pump goroutine per request.
- `ApplyFiltersToIter`, `Limit`, and the ndjson writer are all `iter`-based, so a channel would be converted right back into an iterator.

So I'd keep `FindProviders` as-is. I did expand the `DelegatedRouter` doc comment to spell out the lazy + `Close()` contract, so implementers know to stream results and stop work on `Close`.

routing/http/server/server.go
```diff
@@ -298,14 +314,15 @@ func (s *server) findProviders(w http.ResponseWriter, httpReq *http.Request) {
...
 }
 
-func (s *server) findProvidersJSON(w http.ResponseWriter, provIter iter.ResultIter[types.Record], filterAddrs, filterProtocols []string) {
+func (s *server) findProvidersJSON(w http.ResponseWriter, provIter iter.ResultIter[types.Record], recordsLimit int, filterAddrs, filterProtocols []string) {
 	defer provIter.Close()
 
 	filteredIter := filters.ApplyFiltersToIter(provIter, filterAddrs, filterProtocols)
-	providers, err := iter.ReadAllResults(filteredIter)
+	var limitedIter iter.ResultIter[types.Record] = iter.Limit(filteredIter, recordsLimit)
+	providers, err := iter.ReadAllResults(limitedIter)
```
guillaumemichel: `provIter` could be unconditionally closed after `iter.ReadAllResults()`, which would cancel request on router faster. No need to wait for response to be sent.

```suggestion
	filteredIter := filters.ApplyFiltersToIter(provIter, filterAddrs, filterProtocols)
	var limitedIter iter.ResultIter[types.Record] = iter.Limit(filteredIter, recordsLimit)
	providers, err := iter.ReadAllResults(limitedIter)
	provIter.Close()
```

routing/http/server/server.go
```diff
@@ -438,12 +457,13 @@ func (s *server) provide(w http.ResponseWriter, httpReq *http.Request) {
...
-func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord], filterAddrs, filterProtocols []string) {
+func (s *server) findPeersJSON(w http.ResponseWriter, peersIter iter.ResultIter[*types.PeerRecord], recordsLimit int, filterAddrs, filterProtocols []string) {
 	defer peersIter.Close()
 
-	peersIter = filters.ApplyFiltersToPeerRecordIter(peersIter, filterAddrs, filterProtocols)
+	filteredIter := filters.ApplyFiltersToPeerRecordIter(peersIter, filterAddrs, filterProtocols)
+	var limitedIter iter.ResultIter[*types.PeerRecord] = iter.Limit(filteredIter, recordsLimit)
 
-	peers, err := iter.ReadAllResults(peersIter)
+	peers, err := iter.ReadAllResults(limitedIter)
```
guillaumemichel: Same reasoning as above, no need to keep query longer than required.

```suggestion
	filteredIter := filters.ApplyFiltersToPeerRecordIter(peersIter, filterAddrs, filterProtocols)
	var limitedIter iter.ResultIter[*types.PeerRecord] = iter.Limit(filteredIter, recordsLimit)
	peers, err := iter.ReadAllResults(limitedIter)
	peersIter.Close()
```

guillaumemichel: Or even better to close `limitedIter`?

routing/http/server/server.go
guillaumemichel: Or even better to close `limitedIter`?

guillaumemichel (approved):
Alright, I was missing a piece of the puzzle in `someguy`.

Everything looks good, just 2 nits inline

lidel:
Thanks @guillaumemichel. Applied your eager-close suggestion in 1d14daa18. Also pushed a peers-side regression test (13c6e4104) and a loop fix for `FilterIter.Next` (a4885f676).

On the `FilterIter` one: every filtered-out record used to add a frame to the goroutine stack via `return f.Next()`. Not catastrophic since Go grows the stack as needed, but depth is bounded only by what the delegate returns, and this PR removes the upstream cap. We've seen IPNI return ~15k graphsync records in a single request that `filter-protocols` then drops. Unlikely we'll hit >50k any time soon, but felt safer to switch to a loop now than wait for it to actually stack-overflow.

PR #1164, opened 2026-06-02
bitswap/bsnet: don't mark peer unresponsive on a single send failure

### LINKED ISSUE #1163: bitswap: block fetch hangs forever after peer disconnect+reconnect (regression in v0.34.0)
## Summary

Since **boxo v0.34.0**, a Bitswap block fetch can **hang forever** after a peer disconnects and quickly reconnects, even though the libp2p connection is live again. The want is never (re)sent to the reconnected peer, so `GetBlock`/`SyncGetBlock` blocks until the context is canceled.

## Affected versions

Bisected against a standalone reproducer (kubo + mocknet):

| boxo | result |
|------|--------|
| v0.30 (kubo v0.35) | ✅ ok |
| v0.33 (kubo v0.36) | ✅ ok |
| **v0.34** (kubo v0.37) | ❌ hang |
| v0.35–v0.39 | ❌ hang |

Isolated to boxo (libp2p ruled out): with libp2p held at v0.43, boxo v0.33 = ok, boxo v0.34 = hang.

## Root cause

`streamMessageSender.send()` (`bitswap/network/bsnet/ipfs_impl.go`) marks the peer **unresponsive on every single failed attempt** to open a stream or send a message:

```go
stream, err := s.Connect(ctx)
if err != nil {
    s.bsnet.connectEvtMgr.MarkUnresponsive(s.to) // <-- premature
    return err
}
...
if err = s.bsnet.msgToStream(...); err != nil {
    s.bsnet.connectEvtMgr.MarkUnresponsive(s.to) // <-- premature
    return err
}
```

`send()` is called from `multiAttempt()`, which already resets the stream, retries, and marks the peer unresponsive only **after all retries are exhausted**. So the per-attempt calls are redundant and premature.

They are harmful after a **connection bounce** (disconnect quickly followed by a reconnect):

1. The inbound disconnect notification is **suppressed** by the bsnet notifiee (`Disconnected` returns early when `Connectedness == Connected`, which is already true thanks to the reconnect), so the `ConnectEventManager` never leaves `responsive` and the reconnect `Connected` event is a no-op.
2. A single send failure on the now-stale stream calls `MarkUnresponsive`, transitioning the still-connected peer `responsive -> unresponsive` and emitting `PeerDisconnected`.
3. The peer is sidelined with **no recovery path**: returning to `responsive` requires `OnMessage` (an inbound message), which never arrives because we've stopped sending our wantlist. The fetch hangs until the peer fully disconnects.

Confirmed by neutering `MarkUnresponsive` (hang disappears) and by event-manager tracing showing `Connected (no-op) -> MarkUnresponsive -> PeerDisconnected` with no subsequent `PeerConnected`.

## Reproducer

Standalone, deterministic under load (`GOMAXPROCS=1`). Two in-process IPFS nodes on a mocknet: establish a Bitswap session, disconnect, write a block offline, reconnect, fetch. Hangs ~100% with the warmup on affected versions; passes on v0.33 and with the fix.

<details><summary>main.go</summary>

```go
package main

import (
	"bytes"
	"context"
	"crypto/rand"
	"encoding/base64"
	"fmt"
	"io"
	"os"
	"time"

	ds "github.com/ipfs/go-datastore"
	dsync "github.com/ipfs/go-datastore/sync"
	cfg "github.com/ipfs/kubo/config"
	ipfsCore "github.com/ipfs/kubo/core"
	coreapi "github.com/ipfs/kubo/core/coreapi"
	coreiface "github.com/ipfs/kubo/core/coreiface"
	mock "github.com/ipfs/kubo/core/mock"
	"github.com/ipfs/kubo/repo"
	"github.com/libp2p/go-libp2p/core/crypto"
	"github.com/libp2p/go-libp2p/core/peer"
	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
)

func mkRepo() (repo.Repo, error) {
	c := cfg.Config{}
	priv, pub, err := crypto.GenerateKeyPairWithReader(crypto.RSA, 2048, rand.Reader)
	if err != nil {
		return nil, err
	}
	pid, err := peer.IDFromPublicKey(pub)
	if err != nil {
		return nil, err
	}
	privkeyb, err := crypto.MarshalPrivateKey(priv)
	if err != nil {
		return nil, err
	}
	c.Bootstrap = []string{}
	c.Addresses.Swarm = []string{"/ip4/127.0.0.1/tcp/4001"}
	c.Identity.PeerID = pid.String()
	c.Identity.PrivKey = base64.StdEncoding.EncodeToString(privkeyb)
	c.Swarm.ResourceMgr.Enabled = cfg.False
	return &repo.Mock{
		D: dsync.MutexWrap(ds.NewMapDatastore()),
		C: c,
	}, nil
}

func newNode(ctx context.Context, mn mocknet.Mocknet) (*ipfsCore.IpfsNode, coreiface.CoreAPI, error) {
	r, err := mkRepo()
	if err != nil {
		return nil, nil, err
	}
	node, err := ipfsCore.NewNode(ctx, &ipfsCore.BuildCfg{
		Online:    true,
		Repo:      r,
		Host:      mock.MockHostOption(mn),
		ExtraOpts: map[string]bool{"pubsub": false},
	})
	if err != nil {
		return nil, nil, err
	}
	api, err := coreapi.NewCoreAPI(node)
	if err != nil {
		return nil, nil, err
	}
	return node, api, nil
}

// one round: connect, disconnect, write offline on n1, reconnect, fetch on n2.
// returns true if the fetch completed, false if it timed out (the bug).
func round(fetchTimeout time.Duration) (bool, error) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	mn := mocknet.New()
	defer mn.Close()

	n1, api1, err := newNode(ctx, mn)
	if err != nil {
		return false, err
	}
	defer n1.Close()
	n2, api2, err := newNode(ctx, mn)
	if err != nil {
		return false, err
	}
	defer n2.Close()

	if _, err := mn.LinkPeers(n1.Identity, n2.Identity); err != nil {
		return false, err
	}
	conn, err := mn.ConnectPeers(n1.Identity, n2.Identity)
	if err != nil {
		return false, err
	}
	// settle
	time.Sleep(200 * time.Millisecond)

	// warmup: establish a bitswap session by fetching a first block while
	// connected, like orbit-db does before the peers drop.
	warm := make([]byte, 256)
	_, _ = rand.Read(warm)
	wstat, err := api1.Block().Put(ctx, bytes.NewReader(warm))
	if err != nil {
		return false, err
	}
	wctx, wcancel := context.WithTimeout(ctx, 10*time.Second)
	wr, err := api2.Block().Get(wctx, wstat.Path())
	if err == nil {
		_, _ = io.ReadAll(wr)
	}
	wcancel()

	// disconnect
	if err := conn.Close(); err != nil {
		return false, err
	}

	// write a block offline on n1
	data := make([]byte, 256)
	_, _ = rand.Read(data)
	stat, err := api1.Block().Put(ctx, bytes.NewReader(data))
	if err != nil {
		return false, err
	}
	blockPath := stat.Path()

	// reconnect
	if _, err := mn.ConnectPeers(n1.Identity, n2.Identity); err != nil {
		return false, err
	}

	// fetch on n2 with a timeout
	fctx, fcancel := context.WithTimeout(ctx, fetchTimeout)
	defer fcancel()
	r, err := api2.Block().Get(fctx, blockPath)
	if err != nil {
		return false, nil // timed out / failed -> the bug
	}
	if _, err := io.ReadAll(r); err != nil {
		return false, nil
	}
	return true, nil
}

func main() {
	n := 20
	if len(os.Args) > 1 {
		fmt.Sscanf(os.Args[1], "%d", &n)
	}
	fetchTimeout := 15 * time.Second
	pass, hang := 0, 0
	for i := 1; i <= n; i++ {
		start := time.Now()
		ok, err := round(fetchTimeout)
		dur := time.Since(start).Round(time.Millisecond)
		if err != nil {
			fmt.Printf("run %d: ERROR %v\n", i, err)
			continue
		}
		if ok {
			pass++
			fmt.Printf("run %d: ok (%s)\n", i, dur)
		} else {
			hang++
			fmt.Printf("run %d: HANG/timeout (%s)\n", i, dur)
		}
	}
	fmt.Printf("=== pass=%d hang=%d / %d ===\n", pass, hang, n)
	if hang > 0 {
		os.Exit(1)
	}
}

```
</details>

## Fix

Remove the premature `MarkUnresponsive` calls in `send()` and rely on `multiAttempt()` to mark the peer unresponsive after retries are exhausted. PR incoming.

D4ryl00:
Fixes #1163

### Problem

Since v0.34.0, a Bitswap block fetch can hang forever after a peer disconnects and quickly reconnects. `streamMessageSender.send()` marks the peer unresponsive on **every** failed attempt to open a stream or send a message. But `send()` is called from `multiAttempt()`, which already resets the stream, retries, and marks the peer unresponsive only **after all retries are exhausted** — so the per-attempt marking is redundant and premature.

After a connection bounce, the inbound disconnect is suppressed (the peer is already reconnected, so the bsnet notifiee's `Disconnected` returns early), leaving the `ConnectEventManager` in `responsive` and making the reconnect `Connected` a no-op. A single send failure on the stale stream then marks the still-connected peer `unresponsive` and emits `PeerDisconnected`, with no recovery path (returning to `responsive` needs an inbound `OnMessage` that never comes because we stop sending our wantlist). The fetch hangs until the peer fully disconnects.

### Fix

Drop the two premature `MarkUnresponsive` calls in `send()` and rely on `multiAttempt()` to mark the peer unresponsive once retries are exhausted.

### Testing

- A deterministic mocknet reproducer (see #1163) goes from ~100% hang to 0% hang with this change.
- `go test ./bitswap/network/...` passes.

### DIFF
--- a/bitswap/network/bsnet/ipfs_impl.go
+++ b/bitswap/network/bsnet/ipfs_impl.go
@@ -244,7 +244,6 @@ func (s *streamMessageSender) send(ctx context.Context, msg bsmsg.BitSwapMessage
 	stream, err := s.Connect(ctx)
 	if err != nil {
 		log.Infof("failed to open stream to %s: %s", s.to, err)
-		s.bsnet.connectEvtMgr.MarkUnresponsive(s.to)
 		return err
 	}
 
@@ -254,7 +253,6 @@ func (s *streamMessageSender) send(ctx context.Context, msg bsmsg.BitSwapMessage
 	timeout := s.opts.SendTimeout - time.Since(start)
 	if err = s.bsnet.msgToStream(ctx, stream, msg, timeout); err != nil {
 		log.Infof("failed to send message to %s: %s", s.to, err)
-		s.bsnet.connectEvtMgr.MarkUnresponsive(s.to)
 		return err
 	}
 
--- a/bitswap/network/bsnet/ipfs_impl_test.go
+++ b/bitswap/network/bsnet/ipfs_impl_test.go
@@ -518,6 +518,50 @@ func TestMessageSendErrorDoesNotMarkUnresponsive(t *testing.T) {
 	}
 }
 
+// TestMessageSendErrorMarksUnresponsive checks that a peer whose sends keep
+// failing is eventually marked unresponsive once multiAttempt() exhausts all
+// retries, emitting a PeerDisconnected. This is the complement of
+// TestMessageSendErrorDoesNotMarkUnresponsive: a single recoverable error must
+// not mark the peer, but a persistently failing one must.
+func TestMessageSendErrorMarksUnresponsive(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	p1 := tnet.RandIdentityOrFatal(t)
+	r1 := newReceiver()
+	p2 := tnet.RandIdentityOrFatal(t)
+	r2 := newReceiver()
+
+	eh, bsnet1, _, _, msg := prepareNetwork(t, ctx, p1, r1, p2, r2)
+
+	ms, err := bsnet1.NewMessageSender(ctx, p2.ID(), &network.MessageSenderOpts{
+		MaxRetries:       3,
+		SendTimeout:      100 * time.Millisecond,
+		SendErrorBackoff: 10 * time.Millisecond,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ms.Reset()
+
+	// Keep failing every send so all retries are exhausted.
+	eh.setError(errMockNetErr)
+
+	// SendMsg must fail once retries are exhausted.
+	if err = ms.SendMsg(ctx, msg); err == nil {
+		t.Fatal("expected SendMsg to fail after exhausting retries")
+	}
+
+	// The exhausted retries must have marked the peer unresponsive, which
+	// emits a PeerDisconnected, dropping it from the receiver's set.
+	require.Eventually(t, func() bool {
+		r1.mu.Lock()
+		defer r1.mu.Unlock()
+		_, ok := r1.peers[p2.ID()]
+		return !ok
+	}, time.Second, 20*time.Millisecond, "unresponsive peer was not disconnected after exhausting retries")
+}
+
 func TestMessageSendTimeout(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
--- a/bitswap/network/bsnet/ipfs_impl_test.go
+++ b/bitswap/network/bsnet/ipfs_impl_test.go
@@ -449,6 +449,75 @@ func TestMessageResendAfterError(t *testing.T) {
 	}
 }
 
+// TestMessageSendErrorDoesNotMarkUnresponsive checks that a single recoverable
+// send error does not mark the peer unresponsive. send() is retried by
+// multiAttempt(), which only marks the peer once all retries are exhausted;
+// marking on the first failure could permanently sideline a peer whose
+// connection is in fact still live (e.g. right after a reconnect, where the
+// disconnect notification is suppressed), with no recovery path.
+// Regression test for https://github.com/ipfs/boxo/issues/1163.
+func TestMessageSendErrorDoesNotMarkUnresponsive(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	p1 := tnet.RandIdentityOrFatal(t)
+	r1 := newReceiver()
+	p2 := tnet.RandIdentityOrFatal(t)
+	r2 := newReceiver()
+
+	eh, bsnet1, _, _, msg := prepareNetwork(t, ctx, p1, r1, p2, r2)
+
+	testSendErrorBackoff := 100 * time.Millisecond
+	ms, err := bsnet1.NewMessageSender(ctx, p2.ID(), &network.MessageSenderOpts{
+		MaxRetries:       3,
+		SendTimeout:      100 * time.Millisecond,
+		SendErrorBackoff: testSendErrorBackoff,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ms.Reset()
+
+	// Fail the next send, then stop failing so the retry succeeds.
+	eh.setError(errMockNetErr)
+	go func() {
+		time.Sleep(testSendErrorBackoff / 2)
+		eh.setError(nil)
+	}()
+
+	if err = ms.SendMsg(ctx, msg); err != nil {
+		t.Fatal(err)
+	}
+
+	select {
+	case <-ctx.Done():
+		t.Fatal("did not receive message sent")
+	case <-r2.messageReceived:
+	}
+
+	// The single, recovered send error must not have marked the peer
+	// unresponsive. Marking it would emit a PeerDisconnected with no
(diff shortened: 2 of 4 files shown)


bitswap/network/bsnet/ipfs_impl_test.go
```diff
@@ -449,6 +449,75 @@ func TestMessageResendAfterError(t *testing.T) {
...
+	// Give a late erroneous PeerDisconnected a chance to land, and confirm
+	// the peer remains connected.
+	time.Sleep(300 * time.Millisecond)
+	r1.mu.Lock()
+	_, stillConnected := r1.peers[p2.ID()]
+	r1.mu.Unlock()
+	if !stillConnected {
+		t.Fatal("peer was marked unresponsive after a single recoverable send error")
+	}
+}
```
gammazero, line 519: Need to make sure there is a test that unresponsive peer is still disconnected eventually.

bitswap/network/bsnet/ipfs_impl.go
```diff
@@ -244,7 +244,6 @@ func (s *streamMessageSender) send(ctx context.Context, msg bsmsg.BitSwapMessage
 	stream, err := s.Connect(ctx)
 	if err != nil {
 		log.Infof("failed to open stream to %s: %s", s.to, err)
-		s.bsnet.connectEvtMgr.MarkUnresponsive(s.to)
```
gammazero, line 247: Connect failed... do we not want to mark unresponsive here? In other words, should any retry logic be in connect?

D4ryl00: The retry already covers `Connect`: `multiAttempt` re-runs the whole `send` closure (which calls `Connect` on each attempt) up to `MaxRetries` with backoff, so no separate retry loop in `Connect` is needed. And connect failures are still marked unresponsive — just from `multiAttempt` once all retries are exhausted (line 218), rather than on the first attempt.

Marking here in `send()` was the bug: it fired before any retry, so a transient connect failure right after a reconnect (where the disconnect notification is suppressed) would permanently sideline a live peer. The persistent-failure → unresponsive path is covered by the new `TestMessageSendErrorMarksUnresponsive`.

bitswap/network/bsnet/ipfs_impl_test.go
D4ryl00: [9447cbd](https://github.com/ipfs/boxo/pull/1164/commits/9447cbd3b414c3b7d7128276f2a1f8c359dd2fbb)

gammazero (approved):

gammazero:
@lidel Final check before merge?

lidel (approved):
Thanks @D4ryl00, both for the report and the fix. The analysis in #1163 made it easier to follow.

LGTM.

---

Some history for whoever reads this later: the two `MarkUnresponsive` calls in `send()` came in with #984 (v0.33.1), which back then fixed an HTTP-retrieval stall and made bsnet and httpnet share one `ConnectEventManager`. IIUC they were already redundant, since `multiAttempt()` marks the peer once retries run out, and httpnet has its own pinger/error state handler. With the shared manager, a single transient send failure became a `PeerDisconnected` for a peer that was still reachable, with no path back. That's the hang.

I pushed one more test onto the branch (fcfb5b56e157a7c3e28c949349f4e204e3a07ad4 with shared `ConnectEventManager` plus a stub HTTP gateway) for that cross-transport case. Mostly a safety net: the interaction is unlikely in practice, but the shared-manager part is easy to break, so this should catch a future refactor bringing #1163 back.

PR #1166, opened 2026-06-03
fix: bound ipns cache-control to record eol

lidel:
## Problem

`GET /routing/v1/ipns/{name}` advertised a cache window that outlived the record: `max-age` was the record TTL and `stale-while-revalidate`/`stale-if-error` covered the full remaining validity, so the stale-serving window ran past the record's EOL by up to one `max-age`. A cache (the CDN in front of `delegated-ipfs.dev`, a browser, or a Helia HTTP cache) then served a record whose signature had already expired, which validating clients reject, surfacing as sporadic 500s in the service-worker gateway. The other IPNS surfaces had the same gap, and a malformed negative TTL propagated unchecked, even underflowing the v1 `uint64` TTL field.

## Fix

- `routing/http/server`: cap `max-age` to the record's remaining validity and size the stale window to the validity left after it, so `max-age`+stale never crosses EOL; return `no-store` for a record that is expired or carries an unrecognized `ValidityType`.
- `gateway`: the raw record endpoint (`GET /ipns/{name}?format=ipns-record`) caps `max-age` to the remaining validity.
- `namesys`: the resolver floors a negative record TTL, and a cache hit reports the entry's remaining lifetime instead of the original TTL.
- `ipns`: `NewRecord` floors a negative TTL (also closing the v1 `uint64` underflow) and `Validate` rejects a record carrying one.

No cache reuses an IPNS response past the record's EOL, and a negative TTL can no longer surface through any IPNS API.

### DIFF
--- a/gateway/handler_ipns_record.go
+++ b/gateway/handler_ipns_record.go
@@ -69,7 +69,10 @@ func (i *handler) serveIpnsRecord(ctx context.Context, w http.ResponseWriter, r 
 	}
 
 	if maxAge, ok := ipnsRecordMaxAge(record); ok {
-		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", maxAge))
+		// Truncate to whole seconds rather than round: rounding up could advertise
+		// a max-age that outlives the record's EOL, the cross-EOL overshoot this
+		// bounding exists to prevent.
+		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", int(maxAge.Seconds())))
 	} else {
 		w.Header().Set("Last-Modified", time.Now().UTC().Format(http.TimeFormat))
 	}
@@ -99,20 +102,20 @@ func (i *handler) serveIpnsRecord(ctx context.Context, w http.ResponseWriter, r 
 	return false
 }
 
-// ipnsRecordMaxAge returns the Cache-Control max-age (in seconds) for a raw IPNS
+// ipnsRecordMaxAge returns the Cache-Control max-age duration for a raw IPNS
 // record response. It is the record TTL clamped to the record's remaining EOL
 // validity, so a cache never reuses the record past the point its signature
 // expires (an expired record fails validation), and floored at zero, as a
 // record may report a negative TTL. ok is false when the record carries no TTL,
 // leaving the caller to fall back to Last-Modified.
-func ipnsRecordMaxAge(record *ipns.Record) (seconds int, ok bool) {
+func ipnsRecordMaxAge(record *ipns.Record) (maxAge time.Duration, ok bool) {
 	ttl, err := record.TTL()
 	if err != nil {
 		return 0, false
 	}
-	maxAge := int(ttl.Seconds())
+	maxAge = ttl
 	if eol, err := record.Validity(); err == nil {
-		if remaining := int(time.Until(eol).Seconds()); remaining < maxAge {
+		if remaining := time.Until(eol); remaining < maxAge {
 			maxAge = remaining
 		}
 	}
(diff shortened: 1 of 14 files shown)


gammazero (approved):

achingbrain:
Refs:
- https://github.com/ipfs/service-worker-gateway/issues/1072
- https://github.com/ipfs/service-worker-gateway/issues/1105

lidel:
Hey @gammazero, heads-up before this lands: it grew quite a bit since your approval (which was on just the initial `routing/v1` cap, bd309219f). I extended the same EOL-bounding to the other IPNS surfaces and hardened negative-TTL handling:

- `routing/http/server`: also returns `no-store` for an unrecognized `ValidityType`, and never emits a negative `max-age`.
- `gateway`: the raw record endpoint (`?format=ipns-record`) caps `max-age` to the remaining validity.
- `namesys`: the resolver floors a negative record TTL, and a cache hit reports the entry's remaining lifetime (if less than TTL) instead of the original TTL.
- `ipns`: `NewRecord` floors a negative TTL (also closing the v1 `uint64` underflow) and `Validate` rejects a record carrying one.

I smoke-tested all of it in ipfs/kubo#11349 (bumps kubo to this branch's tip, CI green) and used it there to add user-facing `--ttl`/`--lifetime` sanity checks to `ipfs name publish`. Green here and there.

Could you give the post-approval commits a final look? If they look good to you, this is ready to merge and cut a boxo release, which is what lets someguy/delegated-ipfs.dev pick it up and clear the service-worker-gateway 500s (ipfs/service-worker-gateway#1072).

Thanks for the earlier review!

gateway/handler_ipns_record.go
```diff
@@ -68,8 +68,8 @@ func (i *handler) serveIpnsRecord(ctx context.Context, w http.ResponseWriter, r
 		return false
 	}
 
-	if ttl, err := record.TTL(); err == nil {
-		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", int(ttl.Seconds())))
+	if maxAge, ok := ipnsRecordMaxAge(record); ok {
+		w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", maxAge))
```
gammazero: I would prefer not to have to know in what units the returned time is in, and instead have a `time.Duration` that I can convert to the appropriate units when necessary, e.g. `maxAge.Seconds()`. That way I can look at the code and immediately see that it is in the correct units, instead of having to find the called function and check.

gammazero (approved):
Address comments/questions, then ok to merge if everything is good.

gateway/handler_ipns_record.go
```diff
@@ -98,3 +98,23 @@ func (i *handler) serveIpnsRecord(ctx context.Context, w http.ResponseWriter, r
...
 	return false
 }
+
+// ipnsRecordMaxAge returns the Cache-Control max-age (in seconds) for a raw IPNS
+// record response. It is the record TTL clamped to the record's remaining EOL
+// validity, so a cache never reuses the record past the point its signature
+// expires (an expired record fails validation), and floored at zero, as a
+// record may report a negative TTL. ok is false when the record carries no TTL,
+// leaving the caller to fall back to Last-Modified.
+func ipnsRecordMaxAge(record *ipns.Record) (seconds int, ok bool) {
```
gammazero: See comment above... It would be preferable to return a `time.Duration` here. If needed, this function can [`round`](https://pkg.go.dev/time@go1.26.4#Duration.Round) the time to the nearest second.

lidel:
@gammazero addressed your feedback in 5ee23b887. `ipnsRecordMaxAge` now returns a `time.Duration`; the call site converts at the point of use with `int(maxAge.Seconds())`, so the units are visible where the value is consumed + truncation-to-seconds is documented as intentional

gammazero (approved):

PR #1167, opened 2026-06-03
feat(namesys): propagate DNSLink TXT TTL

lidel:
## Problem

boxo's DNS resolver discards the TTL of DNSLink TXT records and returns 0. The gateway sets `Cache-Control: max-age` for mutable paths only when the TTL is non-zero, so DNSLink (`/ipns/<dnslink-host>`) responses go out with no max-age and fall back to a static value, even though the TXT record states exactly how long the mapping is valid (#329). Recursive names made it worse: a name resolved through several records was cached for only its final hop's TTL, ignoring shorter-lived links earlier in the chain.

## Fix

- Carry the TXT record TTL into the resolved result (`AsyncResult.TTL` / `Result.TTL`), which the gateway turns into `Cache-Control: max-age`.
- `WithDNSResolver` auto-detects a resolver that implements `madns.TXTWithTTLResolver`; `NewDNSResolverWithTTL`, `WithDNSResolverWithTTL`, and `LookupTXTWithTTLFunc` take a TTL-aware lookup directly. Shipped end to end: [go-multiaddr-dns v0.6.0](https://github.com/multiformats/go-multiaddr-dns/releases/tag/v0.6.0) defines the interface, [go-doh-resolver v0.6.0](https://github.com/libp2p/go-doh-resolver/releases/tag/v0.6.0) implements it, both are pinned here, and a compile-time assertion in `gateway` makes a future drift break the build instead of silently dropping TTLs.
- Cap a resolved name's TTL to its shortest hop, so a DNSLink pointing at an IPNS name, or a recursive IPNS chain, is not cached past its earliest-expiring link. Hops with an unknown TTL (0) are ignored.
- `WithMaxCacheTTL` now also caps the TTL reported in results (making Kubo's documented `Ipns.MaxCacheTTL` behavior true in emitted headers), and cache hits report the entry's remaining lifetime instead of its original TTL, so a late hit no longer restarts the full caching period. A cap of 0 still means "no local cache" and leaves the TTL alone (Kubo's offline mode relies on that); the gateway sharness workflow now also watches `namesys`/`ipns`/`path` so a regression there cannot slip through untested again.

`NewDNSResolver` and `LookupTXTFunc` keep their signatures and report an unknown TTL (0). TTLs flow only for domains resolved through DoH[^1]; everything else keeps today's static behavior. Independent of DoH, two changes apply to all namesys users: the `WithMaxCacheTTL` cap and remaining-lifetime cache hits.

Part of #329. ETag for resolved paths (the issue's other half) is out of scope.

[^1]: The OS resolver cannot report TTLs: Go's `net.Resolver` returns only record values, and `madns.DefaultResolver` wraps it. In Kubo, a `DNS.Resolvers` entry for `.` routes every lookup through DoH.


namesys/utilities.go
```diff
@@ -120,6 +131,20 @@ func emitResult(ctx context.Context, outCh chan<- AsyncResult, r AsyncResult) {
...
+// here: it would pick the 0.
+func minNonZeroTTL(a, b time.Duration) time.Duration {
+	switch {
+	case a <= 0:
+		return b
+	case b <= 0:
+		return a
+	default:
+		return min(a, b)
+	}
```
gammazero: Less code. Do not return negative if that is possible.

```suggestion
	ttl := min(a, b)
	if ttl <= 0 {
	    ttl = max(0, a, b)
	}
	return ttl
```

gammazero (approved):

gammazero (approved):

PR #1173, opened 2026-06-08
chore: improve reprovider with batch queue reads

gammazero:
Improve reprovider queue efficiency by attempting a batch read whenever more data is available.

### DIFF
--- a/provider/reprovider.go
+++ b/provider/reprovider.go
@@ -36,6 +36,13 @@ const (
 	// MAGIC: Maximum duration during which no workers are available to provide a
 	// cid before a warning is triggered.
 	provideDelayWarnDuration = 15 * time.Second
+
+	// batchReadSize is number of CIDs to read from provide queue in one visit.
+	batchReadSize = 2048
+
+	// dedupCacheSize is the number of CIDs which deduplication is done across.
+	// Set to 0 is disable deduplication.
+	dedupCacheSize = 2048
 )
 
 var log = logging.Logger("provider")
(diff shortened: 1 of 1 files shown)


provider/reprovider.go
```diff
@@ -334,18 +333,32 @@ func (s *reprovider) provideWorker() {
...
 		}
 		if err = verifcid.ValidateCid(s.allowlist, c); err != nil {
 			log.Errorf("insecure hash in reprovider, %s (%s)", c, err)
-			continue
+			return
 		}
 		provideOperation(s.ctx, c)
 	}
+
+	const batchReadSize = 2048
```
guillaumemichel: Is this intentional that this is the same number as dsqueue dedup cache size?

https://github.com/ipfs/boxo/blob/4e79204b2ac8487d83be2d3bc645ad1fd1b7882f/provider/reprovider.go#L149

If yes, we might as well define a global const used at both places.

guillaumemichel (approved):

provider/reprovider.go
gammazero: Yes, intentional. Whether or not it makes sense for these to be the same size is a good question.

gammazero: I made these into two separate consts defined next to each other. We may want to disable deduplication cache if there is not that much to deduplicate within the size of the cache.

gammazero: https://github.com/ipfs/boxo/issues/1180

PR #1177, opened 2026-06-17
replace go-dsqueue

gammazero:
Replace the provide queue use of go-dsqueue with cascadeq, in boxo and go-libp2p-kad-dht

The intended benefits are:
- Overall efficiency (faster, less memory overhead)
- Reliable performance regardless of datastore in use

The efficiency of go-dsqueue depends on the ability of the datastore in use to support ordered query responses without reading all queued items into memory to provide that ordering of queued item(s). With go-dsqueue is it also necessary to store additional ordering information with each queued item, taking up additional space.

Using cascadeq eliminates the need for a datastore implementation to efficiently return ordered query results as these are always stored in queue order when persisted to disk. Storage is also more efficient as there is no need to store additional ordering information with each item.

A potential downside to this PR is the case where is user wants all data written to S3, so an S3-backed datastore is used. In this case, cascadeq would not allow the user to write all data to S3, without some additional work.

Kubo PR: https://github.com/ipfs/kubo/pull/11364

go-libp2p-kad-dht PR: https://github.com/libp2p/go-libp2p-kad-dht/pull/1266


gammazero:
replaced by #1194

PR #1179, opened 2026-06-21
feat(gateway): always answer empty identity probe

lidel:
This PR is a small fix that ensures people using `boxo/gateway` in their own apps do not expose "dead" endpoints that are ignored by `bitswap/network/httpnet`.

It does not impact Kubo, just third-party implementations that do not wrap blockstore with identity handler.

## Problem

`bafkqaaa` is the empty identity block the [trustless gateway spec](https://specs.ipfs.tech/http-gateways/trustless-gateway/#dedicated-probe-paths) reserves as a probe: boxo's bitswap HTTP client (`bitswap/network/httpnet`) sends `GET`/`HEAD /ipfs/bafkqaaa?format=raw` to decide whether a provider is usable, and a failed probe drops it. Serving that block relied on the backend reconstructing identity CIDs (an idstore-wrapped blockstore), which is implicit assumption and not something `boxo/gateway` test guaranteed .  A `boxo/gateway` backed by something without that support returned `404` and got dropped, even though it could serve real content fine. 

## Fix

- `serveRawBlock` answers a bare `bafkqaaa?format=raw` request (GET and HEAD) with a `200` and empty body before reaching the backend, so any `IPFSBackend` passes the probe.
- The CID is exported as `gateway.EmptyIdentityCID`.
- A cheap CID compare gates the allocating path check, so non-probe raw requests stay allocation-free.

Every boxo-based gateway now passes the probe regardless of its blockstore wiring. Out of scope: only this one CID is short-circuited (other identity CIDs still go to the backend); the client-side drop behavior and the `?format=car` probe path are unchanged.

### DIFF
--- a/gateway/handler_block_test.go
+++ b/gateway/handler_block_test.go
@@ -1,0 +1,84 @@
+package gateway
+
+import (
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/ipfs/boxo/blockservice"
+	"github.com/ipfs/boxo/blockstore"
+	offline "github.com/ipfs/boxo/exchange/offline"
+	"github.com/ipfs/boxo/path"
+	"github.com/ipfs/go-cid"
+	ds "github.com/ipfs/go-datastore"
+	dssync "github.com/ipfs/go-datastore/sync"
+	ipld "github.com/ipfs/go-ipld-format"
+	mh "github.com/multiformats/go-multihash"
+	"github.com/stretchr/testify/require"
+)
+
+// TestEmptyIdentityCIDProbeShortCircuit verifies that the trustless-gateway
+// probe (GET/HEAD /ipfs/bafkqaaa?format=raw) returns 200 with an empty body even
+// when the backend cannot reconstruct identity CIDs. The backend here is a plain
+// blockstore (no [blockstore.NewIdStore]) with an offline exchange, so a 200
+// proves the handler short-circuit answers the probe, not the backend.
+func TestEmptyIdentityCIDProbeShortCircuit(t *testing.T) {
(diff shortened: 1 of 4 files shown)


gateway/handler_block_test.go
```diff
@@ -0,0 +1,84 @@
...
+		require.NoError(t, err)
+		otherCID := cid.NewCidV1(cid.Raw, otherHash)
+
+		req := mustNewRequest(t, http.MethodGet, ts.URL+"/ipfs/"+otherCID.String()+"?format=raw", nil)
+		req.Header.Set("Accept", rawResponseFormat)
+		res := mustDo(t, req)
+		defer res.Body.Close()
+		require.Equal(t, http.StatusNotFound, res.StatusCode)
+	})
+}
```
gammazero, line 78: Add a test to explicitly check the statement made by the code comment: `isEmptyIdentityProbe reports whether rq is a raw-block request for exactly [EmptyIdentityCID] with no trailing path`

```suggestion

	// Check that the empty identity CID with trailing path does not trigger
	// the short-circuit probe response.
	t.Run("identity CID with trailing path is not short-circuited", func(t *testing.T) {
		req := mustNewRequest(t, http.MethodGet, ts.URL+"/ipfs/"+EmptyIdentityCIDString+"/trailing?format=raw", nil)
		req.Header.Set("Accept", rawResponseFormat)
		res := mustDo(t, req)
		defer res.Body.Close()
		require.Equal(t, http.StatusNotFound, res.StatusCode)
	})
}
```

gammazero (approved):
Suggested one test be added. Otherwise, LGTM.

PR #1184, opened 2026-07-03
feat(blockstore): expose BloomCacheStatus and add bloom filter Rebuild

guillaumemichel:
Builds on top of #1183, which stops the Bloom filter cache from activating
after an incomplete `AllKeysChan` enumeration and adds the (unexported)
capability to report a truncated enumeration.

This PR makes that machinery observable and recoverable from outside the
package:

- **Exports the capability** as `AllKeysChanWithErrer` (`AllKeysChanWithErr`),
  so any `Blockstore` can report an error that truncated `AllKeysChan`
  enumeration, and callers can detect it. Forwarded through the same layers as
  in #1183 (bloomcache, tqcache, idstore, ValidatingBlockstore); stores that
  don't implement it fall back to today's best-effort behavior.

- **Adds `BloomCacheStatus`** (`Wait` / `BloomActive` / `Rebuild`), implemented
  by the value `CachedBlockstore` returns when a Bloom filter is configured.
  `Wait`/`BloomActive` existed but were unreachable on the unexported cache
  type; they're now part of an exported interface, so callers can wait for and
  observe the async initial build.

- **Adds `Rebuild`**, to retry after a failed initial build (observable via
  `Wait`) or refresh a healthy filter. The rebuild runs live: the current
  filter stays active and correct throughout, concurrent writes are mirrored
  into the replacement, and the new filter is swapped in atomically only once a
  complete enumeration proves it authoritative. A failed or cancelled rebuild
  leaves the existing filter in place.

No behavior change for existing callers; all additions are opt-in via the new
interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ rebuild runs the filter is inactive (lookups fall through to the underlying
 blockstore, so results stay correct but unaccelerated) and it is activated
 again only on a complete enumeration. A new optional `AllKeysChanWithErrer`
 capability lets a `Blockstore` report an error that truncates `AllKeysChan`
-enumeration. [#MMMM](https://github.com/ipfs/boxo/pull/MMMM)
+enumeration. [#1184](https://github.com/ipfs/boxo/pull/1184)
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


gammazero (approved):
Everything here looks good.

While not necessary for this PR, it would be if `bloomcache.active` was changed to a `atomic.Bool` type.

PR #1185, opened 2026-07-06
feat(mfs): bound under-lock DAG reads

lidel:
## Problem

MFS resolves a child link or walks a HAMT shard while holding the directory lock, on its DAG service with the root's context, which has no deadline. When that service is network-backed and the node is neither local nor reachable (a lazily-linked child whose providers are gone, a block removed out of band, a store left inconsistent by a crash), the read blocks forever while the lock is held. Every other operation on that directory queues behind it, so the whole MFS wedges, and graceful shutdown with it, until the process is killed.

## Fix

`WithFetchTimeout` bounds those reads. A missing, unreachable node now fails with a deadline error and releases the lock instead of blocking forever. Reachable nodes, including lazily-linked remote trees loaded on demand, are still fetched: directory and shard nodes are small, so a generous bound only bites on the unreachable case. It threads through the `parent` interface like the chunker, inherited by every directory under the root.

Motivated by the MFS hangs in ipfs/kubo#10842, repro in https://github.com/ipfs/kubo/issues/10842#issuecomment-4688195782 and fixes in https://github.com/ipfs/kubo/pull/11386
- See https://github.com/ipfs/kubo/pull/11386 for all relatex MFS fixes

:point_right:  No change to default behavior. This is off by default and opt-in: with no timeout set the reads behave exactly as before, so this is a no-op for existing callers.


gammazero (approved):
If the timeout only applies to directories, then should it be kept in the `Directory` struct instead of in `inode` ?

Seems like having this enabled by default, with sufficiently long timeout, could prevent some "broken" behavior with no risk of breaking anythingthat was not already broken. The application built with boxo could still set a more reasonable timeout.

lidel:
@gammazero Agree, pushed fixes:

- 9b589beb7 **refactor(mfs): move fetchTimeout to Directory**
Only `Directory` reads it, so `inode` and `File` no longer carry it.

- 92eff61bc **feat(mfs): enable fetch timeout by default (5m)**
On by default at a generous `DefaultFetchTimeout`; pass `WithFetchTimeout(0)` to disable.

But this got me thinking.. what about file (not a dir), that is non-local, and someone writes to it?

Turns out the file-content path had the same forever-block problem (a write on a lazily-referenced file used `context.TODO`), so I plugged that too:

- 674bc0cda **fix(mfs): cancel file writes on MFS teardown**
Binds the write path to the MFS context so it stops on teardown.

- 38f72d7d7 **feat(mfs)!: thread context through File.Open**
`File.Open` now takes a `context.Context`, so a write honors the caller's timeout too. Breaking: callers add a context argument, but I think its worth it, since we already plugged other holes.

I think this likely covers majority of MFS problems people reported over the years :see_no_evil: 

Mind taking a look?

gammazero (approved):
LGTM!

PR #1187, opened 2026-07-09
fix: update go-test

gammazero:
### 🧪 Tests use new `go-test` and `math/rand/v2`

The new `go-test` is upgraded to use `math/rand/v2` in all of its packages. The `go-test/random` package now allows reuse of the random number generator for more efficiently generating sets of random values.

Additionally, the `go-test/random` package removes support for a global seed for the random number generator. This led to the possibility of multiple tests interfering with each other when generating deterministic values. For the same reason, support for a global sequence was removed. This breakage could be intermittent and difficult to debug, depending on the how/which tests ran at the same time.

Since the underlying pseudo-random number generator was changed in the go-test module, the data generated for a given seed also changed. This required updating tests that relied on seeding the generator and getting expected values.

- Use new `go-test/random` API
  - no global seed values
  - reuse generator where appropriate
- Migrate from `/math/rand' to `/math/rand/v2`
  - random seeds are now `[32]byte` and not `uint64`
  - use `StringToSeed` or `Uint64ToSeed` to create random seeds for deterministic output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,15 +33,18 @@ enumeration. [#1184](https://github.com/ipfs/boxo/pull/1184)
 ### Changed
 
 - 🛠 `mfs`: `File.Open` now takes a `context.Context`. Writing to a file whose data has to be fetched from the network (for example, a lazy reference created with `ipfs files cp`) now uses that context, so the write stops when the context is cancelled, such as on a client timeout or when MFS shuts down, instead of waiting forever for a block that never arrives. Callers must add a context argument; pass the request's context to let a timeout cancel the write. [#1185](https://github.com/ipfs/boxo/pull/1185)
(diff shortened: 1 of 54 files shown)


lidel (approved):
Thanks for this, removes a whole class of cross-test interference. I pushed three commits on top, CI is green:

- c7a230740 **fix(chunker): keep gen reproducing the committed table**
  On rand/v2, `chunker/gen` printed a different table than the `bytehash` committed next to it, so (iiuc) re-running it would have moved buzhash chunk boundaries and changed CIDs. Pinned back to `NewSource(0)`, just to avoid floating the boat. Lmk if I mistunderstood this.

- 3b10577ef **fix(gateway): data race when picking a gateway url**
  The remote blockstore, CAR fetcher, and value store shared a `*rand.Rand` across request goroutines, which isn't safe. Top-level `math/rand/v2` is, and the seeding code goes away.

LGTM, once this lands, ipfs/kubo#11390 needs a re-pin from boxo main

PR #1188, opened 2026-07-17
fix(gateway): renew cache freshness on 304

### LINKED ISSUE #329: Gateway: Cache-Control, max-age=ttl, ETag for /ipns requests

lidel:
## Problem

Revalidation was cheap but pointless: `304 Not Modified` responses carried no headers, so a client cache stayed expired and revalidated on every request, even right after the gateway confirmed the content is unchanged.

## Fix

- 304s echo the validator the client matched plus the `Cache-Control` its 200 counterpart sends[^1]: TTL-derived or immutable for content, the bounded dir-listing policy for generated listings, none for dag index HTML. Caches use these headers to freshen the stored response[^2].
- The policies live in helpers shared by the 200 and 304 paths, so they cannot drift apart.

CAR and IPNS-record downloads keep their format-specific `Etag` handling.

Closes #329

[^1]: [RFC 9110, section 15.4.5](https://httpwg.org/specs/rfc9110.html#status.304): a 304 SHOULD carry the headers that would have been sent in a 200.
[^2]: [RFC 9111, section 4.3.4](https://httpwg.org/specs/rfc9111.html#freshening.responses): a cache updates its stored response with the 304's header fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ enumeration. [#1184](https://github.com/ipfs/boxo/pull/1184)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


lidel:
@gammazero If this looks ok to you, we could merge and cut boxo release next week (i believe all other features already landed)

gammazero (approved):
LGTM

PR #1189, opened 2026-07-20
chore: bump go-libp2p-kad-dht to v0.42

guillaumemichel:
Companion of https://github.com/ipfs/kubo/pull/11398

## Summary

- Store offline routing records using the go-libp2p-kad-dht value-store layout so offline and online routing can share records.
- Preserve the offline router's existing behavior of serving records until their own validation expiry, with an opt-in `WithMaxRecordAge` retention cap.
- Validate records on reads so expired records, such as IPNS records past their EOL, are not returned.
- Keep repeated writes of the same record idempotent.

## Compatibility

The offline router now reads and writes the go-libp2p-kad-dht v0.42+ value-store layout. Records stored using the previous offline-router layout are not migrated or read.


routing/offline/offline.go
```diff
@@ -8,81 +8,98 @@ import (
...
+	return r
 }
 
 // offlineRouting implements the Routing interface,
 // but only provides the capability to Put and Get signed dht
 // records to and from the local datastore.
 type offlineRouting struct {
-	datastore ds.Datastore
-	validator record.Validator
+	vs           *records.ValueStore
```
gammazero: A slightly more descriptive name may be preferable here.

```suggestion
	valstore     *records.ValueStore
```

routing/offline/offline.go
```diff
@@ -8,81 +8,98 @@ import (
...
+	maxRecordAge time.Duration
 }
 
 func (c *offlineRouting) PutValue(ctx context.Context, key string, val []byte, _ ...routing.Option) error {
-	if err := c.validator.Validate(key, val); err != nil {
-		return err
-	}
-	if old, err := c.GetValue(ctx, key); err == nil {
+	rec := record.MakePutRecord(key, val)
+	err := c.vs.Put(ctx, key, rec)
```
gammazero: Comment to state that we do validate the incoming data.

```suggestion
	// Validate and store the record.
	err := c.vs.Put(ctx, key, rec)
```

gammazero (approved):
Some minor suggestions, otherwise looks good.

We can ignore the Gateway Sharness test failure as that should resolve once the kubo PR is merged.

routing/offline/offline.go
```diff
@@ -8,81 +8,98 @@ import (
...
+// A stored record is served for as long as it stays valid (for IPNS, until its
+// EOL); see WithMaxRecordAge to also cap retention by store age.
+func NewOfflineRouter(dstore ds.Datastore, validator record.Validator, opts ...Option) routing.Routing {
+	r := &offlineRouting{
 		validator: validator,
 	}
+	for _, opt := range opts {
+		opt(r)
+	}
+	r.vs = records.NewValueStore(dstore, validator, r.maxRecordAge)
```
lidel: @guillaumemichel I made change here in `routing/offline` to remove cap by default, it is intentional: it restores the offline router's long-standing behavior of serving a record until its EOL (still validated on read). Without it, an offline node, which has no republisher to refresh records, would silently drop still-valid records at `amino.DefaultMaxRecordAge`. Online nodes still get that cap: the kad-dht value store shares this datastore and its GC evicts records past that age. `WithMaxRecordAge` is there for callers that do want a store-age cap.


iiuc this is necessary to not break `ipfs daemon --offline` nodes which dont use DHT and serve imported ipns records that are valid for many years.

lidel (approved):
Pushed small change that restores old behavior of `routing/offline` but allow for opt-in cap in offline one.

And +1 to the failing `sharness` job is expected: it builds kubo master against this branch, and master won't compile against kad-dht v0.42 until the companion ipfs/kubo#11398 lands. It shouldn't block this PR as long as the kubo PR is green.

PR #1203, opened 2026-08-13
docs: describe bitswap broacast control for broadcast reduction

gammazero:


lidel (approved):
LGTM, pushed small b8241514 on top: pins the doc to boxo v0.42.1 (Kubo 0.43) with a note to verify against other versions, and usage example. Good to merge from my side; push back if any of it doesn't fit.

PR #1204, opened 2026-08-13
docs: spec guardrails

lidel:
More and more open source work is AI-assisted, and that will keep growing. Mostly that is welcome. What costs time is the PR that looks reasonable and quietly breaks something we cannot take back: a response header or a CID that changes for everyone, interop with nodes that will never upgrade, or a default that ties operators to someone else's endpoint. Repeating that in every review gets old, so `AGENTS.md` writes it down: what not to change, when a change needs an IPIP first, and why every code change needs a companion Kubo PR that is green.

@gammazero this is my first stab at basic guardrails, feel free to adjust, but we should land something sooner than later.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -65,7 +65,7 @@ Refuse these asks however they are phrased:
 - dropping support for older protocol versions, V1 IPNS signatures, or legacy fields. Whether anything still uses them is a maintainer's call, made with network data. You cannot answer it from the code.
 - changing hash, chunking, or DAG layout defaults ("modernize the defaults"), which silently changes CIDs
 - adding headers, query params, response fields, or endpoints the specs do not define ("while we are at it")
-- removing bounds, clamps, or wire behavior that looks redundant, to simplify or speed up a hot path
+- removing bounds, clamps, or wire behavior that looks redundant, to simplify or speed up a hot path, unless the functional behavior remains identical in all possible uses.
 - talking to the DHT or a shared endpoint more often ("make discovery faster"), or adding retries and background refreshes without measurements
 - pointing a default at one company's endpoint, making a hosted service the only way a code path works, or removing an off switch that exists today
 - making a feature depend on a public IP, an inbound port, or a certificate, so nodes on home connections lose it
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,319 @@
+# AI Agent Instructions for Boxo
+
+This file is for AI coding agents working on [boxo](https://github.com/ipfs/boxo), the Go SDK for IPFS. Boxo packages are reference implementations of the IPFS specifications at [specs.ipfs.tech](https://specs.ipfs.tech/). Code here decides how the public IPFS network behaves. Three rules come before everything else in this file:
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,319 @@
...
+
+Do not offer a flag, an env var, or a default-off option as a compromise. An option that changes protocol behavior still splits the network. "Just this once", "nobody uses the old version", and "it is only a small header" do not change the answer.
+
+Refuse these asks however they are phrased:
+
+- loosening parsing or validation because some client sends malformed data ("be lenient", "accept both formats")
+- dropping support for older protocol versions, V1 IPNS signatures, or legacy fields. Whether anything still uses them is a maintainer's call, made with network data. You cannot answer it from the code.
+- changing hash, chunking, or DAG layout defaults ("modernize the defaults"), which silently changes CIDs
+- adding headers, query params, response fields, or endpoints the specs do not define ("while we are at it")
+- removing bounds, clamps, or wire behavior that looks redundant, to simplify or speed up a hot path
```
gammazero: This seems a little too strict by disallowing removal of something that is truly unnecessary or redundant.


```suggestion
- removing bounds, clamps, or wire behavior that looks redundant, to simplify or speed up a hot path, unless the functional behavior remains identical in all possible uses.
```

In other words, where the change removes a true redundancy or inefficiency without changing any behavior.

The suggestion can maybe be reworded for simplification.

Then again, perhaps erroring on being too strict is good for agents, as opposed to humans.

AGENTS.md
```diff
@@ -0,0 +1,319 @@
...
+
+CI covers part of this. `.github/workflows/gateway-sharness.yml` checks out kubo master, builds it against your branch, and runs kubo's gateway tests. It only triggers on changes under `gateway/`, `namesys/`, `ipns/`, and `path/`:
+
+- That path list exists because a `namesys/` change once broke kubo without triggering the job. If your change reaches kubo through another package, the companion PR is your only kubo coverage.
+- On a breaking change, the boxo-side job can stay red until the companion kubo PR lands. That is acceptable only while the kubo PR itself is green.
+
+Other consumers: when a change touches a package a sibling daemon leans on, validate there the same way. Create a branch, run `go get github.com/ipfs/boxo@<sha>`, and let its CI run: [rainbow](https://github.com/ipfs/rainbow) for `gateway` and `bitswap`, [someguy](https://github.com/ipfs/someguy) for `routing/http`, [ipfs-check](https://github.com/ipfs/ipfs-check) for `bitswap/network`. Releases gate on kubo again, per `RELEASE.md`.
```
gammazero:
```suggestion
Other consumers: when a change touches a package a sibling daemon depends on, validate there the same way. Create a branch, run `go get github.com/ipfs/boxo@<sha>`, and let its CI run: [rainbow](https://github.com/ipfs/rainbow) for `gateway` and `bitswap`, [someguy](https://github.com/ipfs/someguy) for `routing/http`, [ipfs-check](https://github.com/ipfs/ipfs-check) for `bitswap/network`. Releases gate on kubo again, per `RELEASE.md`.
```

AGENTS.md
lidel: extra verbosity in guardrails does not come with the same tax as it does with humans, so this lgtm, merging

gammazero (approved):
LGTM

PR #747, opened 2024-12-09
HTTP retrieval proposal

hsanjuan:
This is a proposal to add HTTP retrieval to Boxo. The current state is highly WIP, but I successfully retrieved something over HTTP, so posting to initiate a discussion over the approach and if we want to pursue it until the end.

### Approach

The high-level idea is that most of what lives in `bitswap/client` is actually an "exchange" implementation, with the only real "Bitswap" thing being that `bitswap/network` sends HAS/GET requests over bitswap-protocol streams. As such, we should be able to complement `bitswap/network` with an HTTP-retrieval implementation which, instead of fetching things over the bitswap protocol, calls HTTP endpoints as indicated by the provider's `/http` addresses entries. 

Note that conceptually at least, this is not adding HTTP retrieval into bitswap, but promoting most of the bitswap code to be a reference "Exchange" implementation, which is re-usable for different retrieval protocols (bitswap, http...). That is, we would be talking of an "exchange network" component and not a "bitswap network" component. **Renames to this extent are still missing**.

### Implementation

In order to introduce an http-retrieval "exchange network" we need to:

  * Know when something should be retrieved via HTTP - that is, an item has an `/http` provider.
  * Use HTTP network for that.

To this end:
  * We have a router which select the http-network or the bitswap-network (or both) based on the existance of `/http` addresses in the peerstore of the given peer.
  * We have implemented an http-network as a PoC that performs GET requests to `/http` endpoints when handling a WANT.

image

In my tests plugging it to Kubo, the http-network can be used to retrieve content from a gateway over http. :partying_face: 

The main advantange to this approach is that it is relatively clean to incorporate to the codebase, and keeps most of the code untouched, without having to duplicate any of the complex areas.

### Challenges

  *  Connectivity tracking is not implemented yet and we will have to see to what extent it can be implemented (I'm guessing we can plug into the TCP dialer directly).
  * Options like timeouts etc. are not implemented
  * We use a single HTTP client rather than a pool
  * Of course testing is fully lacking.

Bitswap places a lot of importance on managing connectivity events to peers. We avoid requesting things from peers that have not signaled connectivity, we clean peers that have disconnected and re-queue things for peers that disconnect. Thus it seems we must support http-connectivity events. When a libp2p peer connects for bitswap, we know that the connection is setup, handshake has been performed and protocol negotiation has happened. For HTTP these things may not exist so we need to define what means "Connected" (i.e. in the case of https it would mean we have completed SSL handshakes).

[...]

PR #1205, opened 2026-08-16
 fix(bitswap/httpnet): bound probe traffic

lidel:
## Problem

For legacy reasons, every connected HTTP peer gets probed with `GET/HEAD /ipfs/bafkqaaa` every 5 seconds for the lifetime of the process. Results are discarded, failures never slow the loop, and `Connect` re-probes failing endpoints with no memory. A single idle daemon emits ~17k requests per day per HTTP provider. The probe target and wire format are spec-defined and unchanged[^1]; the defect is frequency and lifecycle, which the spec does not constrain.

## Fix

- delete the ping loop; latency for the DONT_HAVE timeout manager is seeded from the `Connect` probe round trip and updated from time-to-headers of server-understood retrieval responses; on-demand `Ping` stays, matching bsnet
- `Connect` and `Ping` honor per-host cooldowns; a failed probe starts one from `Retry-After`[^2][^3] (past dates rejected) or the new `DefaultConnectFailureBackoff`; endpoints skipped for an active cooldown stay in the peerstore as failover targets
- cooldowns move to a process-wide `SharedCooldownTracker` so backoff survives applications that build a short-lived `Network` per retrieval; `WithCooldownTracker` opts a Network out; the tracker holds no goroutine and needs no shutdown
- expired sender cooldown snapshots no longer poison a sender for its lifetime, and the 410 probe workaround for a provider that shut down is gone

Historic conxtex: the 5s interval landed in #747 without a stated rationale, nothing consumes it on a schedule, and #918 already records that idle peers should not be pinged. An opt-in periodic pinger can come back later without API breakage, but i'm unsure we need it..

## Testing

New synctest and integration suites cover the behavior changes; `TestNoBackgroundProbes` is the regression guard for the removed loop. `go test -race -count=3 ./bitswap/network/httpnet/...` and `go test -race ./bitswap/...` pass. Companion kubo PR (branch pinned, green CI) plus ipfs-check and rainbow validation branches: to follow, linked here before merge.

Before this ships as a patch release, we need to make sure success rate remains the same: 
- [ ] staging test: https://github.com/ipshipyard/waterworks-infra/pull/1037

[^1]: [Trustless Gateway spec, dedicated probe paths](https://specs.ipfs.tech/http-gateways/trustless-gateway/#dedicated-probe-paths): defines the probe target and response; says nothing about client request frequency.
[^2]: [Path Gateway spec, Retry-After response header](https://specs.ipfs.tech/http-gateways/path-gateway/#retry-after-response-header): gateways SHOULD return Retry-After with 429, 503 and 504.
[^3]: [RFC 9110, section 10.2.3](https://httpwg.org/specs/rfc9110.html#field.retry-after): Retry-After indicates how long the user agent ought to wait before making a follow-up request.

### DIFF
--- a/bitswap/network/httpnet/httpnet.go
+++ b/bitswap/network/httpnet/httpnet.go
@@ -360,7 +360,7 @@ func New(host host.Host, opts ...Option) network.BitSwapNetwork {
 	}
 	htnet.client = c
 
-	pinger := newPinger(htnet, pingCid)
+	pinger := newPinger(htnet)
 	htnet.pinger = pinger
 
 	et := newErrorTracker(htnet)
@@ -403,12 +403,16 @@ func (ht *Network) Stop() {
 	})
 }
 
-// Ping triggers a ping to the given peer and returns the latency.
+// Ping sends a probe to the peer's endpoints and returns the measured
+// latency. Endpoints in cooldown are not probed; when all of them are cooling
+// down, no request is made and the result carries an error.
 func (ht *Network) Ping(ctx context.Context, p peer.ID) ping.Result {
 	return ht.pinger.ping(ctx, p)
 }
 
-// Latency returns the EWMA latency for the given peer.
+// Latency returns the EWMA latency for the given peer. The estimate is seeded
+// from the Connect probe and updated with response times of real retrieval
+// requests.
 func (ht *Network) Latency(p peer.ID) time.Duration {
 	return ht.pinger.latency(p)
 }
@@ -426,14 +430,15 @@ func (ht *Network) senderURLs(p peer.ID) []*senderURL {
 	return ht.cooldownTracker.fillSenderURLs(urls)
 }
 
-// IsHTTPPeer returns true if the peer is currently being pinged, which means
-// we are connected to it via HTTP.
+// IsConnectedToPeer returns true if the peer is in the connected registry,
+// which means Connect() found a working HTTP endpoint for it and
+// DisconnectFrom() has not been called since.
 func (ht *Network) IsConnectedToPeer(ctx context.Context, p peer.ID) bool {
 	// only answer this question while no one is connecting or
 	// disconnecting.
 	ht.ongoingConnsLock.RLock()
 	defer ht.ongoingConnsLock.RUnlock()
-	return ht.pinger.isPinging(p)
+	return ht.pinger.isConnected(p)
 }
 
 // SendMessage sends the given message to the given peer. It uses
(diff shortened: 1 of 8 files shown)


bitswap/network/httpnet/pinger.go
```diff
@@ -121,54 +140,56 @@ func (pngr *pinger) recordLatency(p peer.ID, next time.Duration) {
...
-	pngr.pingsLock.Lock()
+// markDisconnected removes the peer from the connected registry and wipes its
+// recorded latency, so a later reconnection starts from a fresh measurement.
+// Lock order: connectedLock, then latenciesLock. recordLatencyIfConnected
+// nests the same way, so no recording path can repopulate the latency of a
+// peer once this returns.
+func (pngr *pinger) markDisconnected(p peer.ID) {
+	log.Debugf("marking %s as disconnected", p)
+	pngr.connectedLock.Lock()
 	{
```
gammazero, line 186: nit: Using the braces '{ .. }` to designate lock scope? It is not a consistent style in the codebase, so maybe no need to keep it?

bitswap/network/httpnet/pinger.go
gammazero, line 81: Not part of the code change, but should the write to the `results` chan use a `select` to also read `case <-ctx.Done()`?

bitswap/network/httpnet/httpnet.go
```diff
@@ -548,24 +593,36 @@ func (ht *Network) Connect(ctx context.Context, pi peer.AddrInfo) error {
...
 
 		// HEAD did not work. Try GET.
 		supportsHead = false
 
-		_, err = ht.connectToURL(ctx, pi.ID, u, "GET")
+		start = time.Now()
+		_, retryAfter, err = ht.connectToURL(ctx, pi.ID, u, "GET")
 		if err != nil {
 			errs = append(errs, fmt.Errorf("%s: %s", u.Multiaddress.String(), err))
 			if ctxErr := ctx.Err(); ctxErr != nil {
```
gammazero:
```suggestion
			if ctx.Err() != nil {
```

gammazero (approved):
Looks good. Only a few nits, mostly not even part of your PR, that changing is optional.

bitswap/network/httpnet/pinger.go
lidel: Kept them because it's the existing style in this package: `request_tracker.go`, `cooldown.go`, and the untouched parts of `pinger.go` all use the same brace scoping, though nothing outside httpnet does.. so yes, its preexisting inconsistency which I did not touch to keep PR smaller.  

We could drop them package-wide in a follow-up PR, lets keep this one small.

bitswap/network/httpnet/pinger.go
lidel: I think it's safe as is: the channel has one buffer slot per goroutine, and each goroutine sends only once, so a send never waits, even if nobody reads the rest. The extra `ctx.Done()` case would never fire, so I left it out.

lidel:
@gammazero  fwiw staging test looks ok: https://github.com/ipshipyard/waterworks-infra/pull/1037#issuecomment-5313945021 -- if no concerns, feel free to merge and ship as a patch release (fine if its the only fix, we want this out asap, just so LLMs stop propagating older version)

PR #1209, opened 2026-08-21
feat(gateway): IPIP-548 Ipfs-Uri response header

lidel:
## Problem

`X-Ipfs-Path` cannot carry every UnixFS file name: bytes outside visible ASCII are invalid in HTTP field values[^1], so a path like `/ipfs/{cid}/łódź.txt` reaches clients garbled or silently altered. ipfs/specs#548 deprecates the header in favor of `Ipfs-Uri`, a percent-encoded `ipfs://` / `ipns://` URI that survives any file name.

## Fix

- emit `Ipfs-Uri` per ipfs/specs#548: canonical authority (base32 CIDv1, base36 IPNS name, A-label DNSLink), strict percent-encoding of every path segment, sent on deserialized, redirect, and error responses, exposed for CORS
- stop sending `X-Ipfs-Path` by default; `Config.DeprecatedXIpfsPath` plus `Headers.WithDeprecatedXIpfsPath` restore it for legacy consumers, still omitted when the path has unsafe bytes
- covered by ipfs/gateway-conformance#301; CI pins that branch until a release ships (TODO in the workflow)

## Tests 

All pass, Sharness can be ignored here, it pass in Kubo using conformance and  boxo/gateway from this PR:
- https://github.com/ipfs/kubo/pull/11437

[^1]: [RFC 9110, section 5.5](https://httpwg.org/specs/rfc9110.html#fields.values): field values are limited to HTAB, SP, and visible ASCII.


gammazero (approved):

lidel:
The red `sharness` is the usual cross-repo dance: it runs kubo master's sharness, which still expects `X-Ipfs-Path`. ipfs/kubo#11437 updates it and is fully green against this branch, so main goes green once that lands. Fine to merge.

PR #1212, opened 2026-08-27
test(unixfs): lock in PBNode field order behavior (IPIP-550)

lidel:
## Problem

The ecosystem decodes both PBNode field orders, but by accident: nothing in boxo's tests said so, and nothing pinned which order the profiles write. ipfs/specs#550 formalizes both sides.

## Fix

- byte-exact tests against the IPIP-550 fixtures: both orders decode, and re-encoding reproduces the exact bytes and CIDs
- `UnixFS_v0_2015` and `UnixFS_v1_2025` pin `PBNodeLinksFirst` explicitly, so a future refactor cannot silently change their bytes or CIDs
- gateway-conformance CI runs v0.14, which tests serving both orders
- side effect: generating the `Data`-first fixtures needed an encoder, kept as the documented low-level `UnixFSProfile.PBNodeFieldOrder` opt-in; no profile enables it, and enabling it changes every written CID

Defaults are unchanged everywhere. Kubo wires the same knob as `Import.UnixFSPBNodeFieldOrder` (ipfs/kubo#11439).


gammazero (approved):
LGTM

PR #1213, opened 2026-09-07
fix(bootstrap): skip backup peer list when no bootstrap peers configured

karawitan:
## Problem

When a node has no bootstrap peers configured (for example `Routing.Type=none` with an empty `Bootstrap` list, or `ipfs daemon --routing=none`), `bootstrapRound` still loads and dials backup peers persisted from previous runs (stored under `TempBootstrapPeersKey` in the datastore). This causes unwanted external DNS resolution and connection attempts to stale addresses every 30 seconds, even though the operator explicitly disabled routing/bootstrap. Reported in [kubo#11452](https://github.com/ipfs/kubo/issues/11452).

The backup peer list exists as a recovery mechanism for when configured bootstrap peers are down ([#8856](https://github.com/ipfs/boxo/pull/8856)). With no configured bootstrap peers, there is nothing to recover from, so consulting the backup list is the wrong behavior.

## Fix

In `bootstrapRound`, return early when `cfg.BootstrapPeers()` returns an empty list, before consulting the backup peer list. This:

- Covers `Bootstrap: null` / empty list (becomes an empty `[]peer.AddrInfo`)
- Works with `ipfs daemon --routing=none` (runtime override that does not change the config file)
- Keeps `Routing.Type=none` + explicit `Bootstrap` peers working (those nodes still dial their configured peers)
- Needs no new config option

Nodes that configure explicit `Bootstrap` peers are unaffected: the backup-list fallback still runs when those peers fail to connect.

### Behavior change

Previously, a node with default routing and an empty `Bootstrap` list would still dial backup peers. After this change it does not. This aligns behavior with operator intent: an empty `Bootstrap` list now means "no bootstrap dialing at all, including saved backup peers." This is a fix, directed by @lidel in [kubo#11453 (review)](https://github.com/ipfs/kubo/pull/11453#pullrequestreview-5127196673): *"The simplest fix is in boxo, in `bootstrap.bootstrapRound`: when `cfg.BootstrapPeers()` returns nothing, stop there instead of trying the backup list."*

## Testing

- `go build ./bootstrap/...` passes
- `go test ./bootstrap/... -count=1` passes (all existing tests green)
- `go vet ./bootstrap/...` passes
- `gofmt -l bootstrap/` clean
- Added `TestBootstrapRoundSkipsBackupWhenNoBootstrapPeers`: verifies `bootstrapRound` does not call `loadBackupBootstrapPeers` when `BootstrapPeers()` is empty
- Added `TestBootstrapRoundDialsBackupWhenBootstrapPeersPresent`: verifies the backup list is still consulted when configured bootstrap peers fail to connect, preserving the #8856 recovery mechanism

### Companion kubo PR

REQUIRED companion kubo PR: [ipfs/kubo#11453](https://github.com/ipfs/kubo/pull/11453) (reworked to pin this boxo branch, fix docs, and add a kubo-level regression test). It is a draft while this boxo PR is unmerged; kubo CI status will be reported there.

Generated with [Devin](https://devin.ai)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,7 @@ The following emojis are used to highlight certain changes:
 ### Fixed
 
 - `gateway`: `X-Ipfs-Path` values no longer carry bytes that are invalid in an HTTP field value (Section 5.5 of RFC 9110). The header used to echo raw UnixFS file names, so non-ASCII paths arrived garbled or broke strict clients; when the header is enabled, it is now omitted for such paths, which only the percent-encoded `Ipfs-Uri` can carry. [#1209](https://github.com/ipfs/boxo/pull/1209)
-- ✨ `bootstrap`: `bootstrapRound` no longer consults the backup peer list when no bootstrap peers are configured. The backup list exists only as a recovery mechanism for when configured bootstrap peers are down (see #8856); with no configured peers there is nothing to recover from, so dialing stale backup peers persisted from previous runs is skipped. This lets a caller fully disable bootstrap dialing by setting an empty peer list (for example a local-only/offline node with `Routing.Type=none`), and works with runtime overrides such as `ipfs daemon --routing=none` that do not change the config file. Nodes that configure explicit `Bootstrap` peers are unaffected. [#1213](https://github.com/ipfs/boxo/pull/1213)
-
+- `bootstrap`: the saved backup peer list is no longer dialed when no bootstrap peers are configured. [#1213](https://github.com/ipfs/boxo/pull/1213)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


lidel (approved):
Thanks for moving this over so quickly.

Pushed df6852a0 on top: shorter code comment and changelog entry, references to ipfs/kubo#8856 and ipfs/kubo#11452 now carry the repo (bare numbers resolve to unrelated boxo items here), and the tests drop the unused `peerID` and use `t.Context()`.

The Devin-generated prose was a bit generous with words for a small fix, so I trimmed it; a quick pass over its output before posting saves reviewers time. The PR description still links `ipfs/boxo/pull/8856` where ipfs/kubo#8856 is meant, if you want to fix that.

Otherwise LGTM.

